PR #157, opened 2023-07-12
FEAT: support pytorch models

pangyoki:
ATT


xinference/model/llm/__init__.py
```diff
@@ -178,3 +179,25 @@ def install():
             cls=ChatglmCppChatModel,
         )
     )
+
+    MODEL_FAMILIES.append(
```
UranusSeven, line 183: Replace opt with baichuan?

xinference/model/llm/core.py
```diff
@@ -57,6 +61,81 @@ def __setitem__(self, key: str, value: Any):  # type: ignore
         super().__setitem__(key, value)
 
 
+def _to_prompt(
```
UranusSeven, line 64: Better write a mixin class instead of putting the util functions in the global scope. For the definition of mixin, see https://zhuanlan.zhihu.com/p/95857866

xinference/model/llm/core.py
```diff
@@ -216,97 +306,240 @@ def __init__(
         self._user_name: str = user_name
         self._assistant_name: str = assistant_name
 
-    def _to_prompt(
+    def chat(
```
UranusSeven, line 309: Can you move llamacpp and pytorch models into 2 different files respectively?

xinference/core/restful_api.py
```diff
@@ -193,7 +193,7 @@ class CreateChatCompletionRequest(BaseModel):
 
     # llama.cpp specific parameters
     top_k: int = top_k_field
-    repeat_penalty: float = repeat_penalty_field
+    repetition_penalty: float = repetition_penalty_field
```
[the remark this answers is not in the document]
pangyoki: as discussed offline, `repetition_penalty` renamed in generate method.

xinference/model/llm/__init__.py
pangyoki: done

xinference/model/llm/core.py
pangyoki: done

xinference/model/llm/core.py
pangyoki: done

PR #169, opened 2023-07-13
ENH: add user and assistant name to chat

RayJi01:


xinference/model/llm/core.py
```diff
@@ -220,15 +220,17 @@ def _to_prompt(
         self,
         prompt: str,
         system_prompt: str,
+        username,
```
UranusSeven, line 223: Type hint.

xinference/model/llm/core.py
```diff
@@ -182,7 +182,7 @@ def generator_wrapper(
             for _completion_chunk in self._llm(prompt=_prompt, **_generate_config):
                 yield _completion_chunk
 
-        logger.debug(
+        logger.error(
```
UranusSeven, line 185: Restore code for debugging.

xinference/model/llm/core.py
```diff
@@ -140,7 +140,7 @@ def _sanitize_model_config(
         else:
             context_length = 2048
 
-        llamacpp_model_config.setdefault("n_gpu_layers", self._gpu_layers)
+        # llamacpp_model_config.setdefault("n_gpu_layers", self._gpu_layers)
```
UranusSeven, line 143: Restore code for debugging.

xinference/model/llm/core.py
RayJi01: resolved in the latest version

xinference/model/llm/core.py
RayJi01: resolved in the latest version

xinference/model/llm/core.py
RayJi01: Resolved in the latest version

UranusSeven (approved):

UranusSeven:
There's not a OpenAI compatible way to pass the user and assistant name. 

In this case, we need to use the low level API `generate` and piece together a prompt based on the chat history.

PR #192, opened 2023-07-18
FEAT: support vicuna-v1.3 33B

### LINKED ISSUE #178: FEAT: support vicuna-v1.3 33B

Bojun-Feng:
resolve #178

### DIFF
--- a/xinference/model/llm/__init__.py
+++ b/xinference/model/llm/__init__.py
@@ -88,12 +88,14 @@ def install():
         else (
             "https://huggingface.co/TheBloke/vicuna-13b-v1.3.0-GGML/resolve/main/"
             f"vicuna-13b-v1.3.0.ggmlv3.{quantization}.bin"
+            if model_size == 13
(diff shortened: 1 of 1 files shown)


xinference/model/llm/__init__.py
```diff
@@ -88,12 +88,14 @@ def install():
         else (
             "https://huggingface.co/TheBloke/vicuna-13b-v1.3.0-GGML/resolve/main/"
             f"vicuna-13b-v1.3.0.ggmlv3.{quantization}.bin"
+            if model_size == 13
```
UranusSeven, line 91: Would it be more readable if reorgnize the func like this?

```python
if model_size == 7:
    ...
elif model_size == 13:
    ...
else:
    ...
```

Bojun-Feng: It seems that lambda functions do not support elif, so I just used a nested if to add the condition. I can translate the generator into a normal function and make it more readable.

UranusSeven (approved):

PR #205, opened 2023-07-19
ENH: set default device to `mps` on MacOS

pangyoki:
ATT

### DIFF
--- a/xinference/model/llm/core.py
+++ b/xinference/model/llm/core.py
@@ -90,6 +90,14 @@ class Model(abc.ABC):
         self.model_uid = model_uid
         self.model_spec = model_spec
 
+    @staticmethod
+    def _is_darwin():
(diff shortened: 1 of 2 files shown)


xinference/model/llm/core.py
```diff
@@ -90,6 +90,14 @@ def __init__(self, model_uid: str, model_spec: "ModelSpec", *args, **kwargs):
         self.model_uid = model_uid
         self.model_spec = model_spec
 
+    @staticmethod
+    def _is_darwin():
```
UranusSeven, line 94: Does `mps` works with macOS intel?

There's already a util function `_is_darwin_and_apple_silicon`.

pangyoki: done

UranusSeven (approved):

PR #252, opened 2023-07-27
ENH: support XINFERENCE_ENDPOINT

### LINKED ISSUE #249: ENH: client supports env variable XINFERENCE_ENDPOINT
It could be troublesome when starting xinference with a custom port. You have to specify the endpoint every time when you use xinference CLI.

### Describe the solution you'd like
I hope we can optimize user experience by using an env variable XINFERENCE_ENDPOINT.

When you starts an xinference service, an endpoint will be printed. Then in the terminal for xinference CLI, execute `export XINFERENCE_ENDPOINT=${endpoint}`.

Xinference CLI should firstly check if `-e` is specified. If not, check the env var XINFERENCE_ENDPOINT. If neither of them is set, use the default endpoint.

RayJi01:
Resolves #249 

Adding the Environment Variable Support for the user:

Demo for the effects:

We first lanuch a model with a non-default port:
image

Then we manually export the endpoint into the environment:
image

Now we call Xinference launch without port specified. Previously it will go to default port 9997, which will cause connection error in our case. Instead, it will work as shown:
Worker side:
image
The server side:
image

When calling Xinference list, also without port specification, it works as well:
image

This indicate that this PR sucessfully implement the feature enhancement we want.


xinference/deploy/cmdline.py
```diff
@@ -27,6 +28,14 @@
 )
 
 
+def getExistingEndpoint():
+    logger.error("XINFERENCE_ENDPOINT" in os.environ)
```
UranusSeven, line 32: Incorrect logging level.

xinference/deploy/cmdline.py
```diff
@@ -27,6 +28,14 @@
...
+    if "XINFERENCE_ENDPOINT" in os.environ:
```
UranusSeven, line 33: Make XINFERENCE_ENDPOINT a constant. See xinference/constants.py

xinference/deploy/cmdline.py
```diff
@@ -27,6 +28,14 @@
 )
 
 
+def getExistingEndpoint():
```
UranusSeven, line 31: Lacking return value type hint.

xinference/deploy/cmdline.py
```diff
@@ -27,6 +28,14 @@
...
+    logger.error("XINFERENCE_ENDPOINT" in os.environ)
+    if "XINFERENCE_ENDPOINT" in os.environ:
+        return os.environ["XINFERENCE_ENDPOINT"]
+    else:
+        return XINFERENCE_DEFAULT_ENDPOINT_PORT
```
UranusSeven, line 36: This util is supposed to return an endpoint, not a port.

xinference/deploy/cmdline.py
```diff
@@ -73,6 +82,8 @@ def supervisor(
 
     address = f"{host}:{get_next_port()}"
 
+    port = getExistingEndpoint()
```
UranusSeven, line 85: Should not overwrite the argument `port`.

xinference/deploy/cmdline.py
```diff
@@ -92,6 +103,8 @@ def worker(log_level: str, endpoint: str, host: str):
         logging.basicConfig(level=logging.getLevelName(log_level.upper()))
     logging_conf = dict(level=log_level.upper())
 
+    endpoint = getExistingEndpoint()
```
UranusSeven, line 106: Should not overwrite the argument `endpoint`.

xinference/deploy/cmdline.py
```diff
@@ -121,6 +134,8 @@ def model_launch(
     model_format: str,
     quantization: str,
 ):
+    endpoint = getExistingEndpoint()
```
UranusSeven, line 137: Ditto.

xinference/deploy/cmdline.py
```diff
@@ -147,6 +162,8 @@ def model_list(endpoint: str, all: bool):
 
     from ..model import MODEL_FAMILIES
 
+    endpoint = getExistingEndpoint()
```
UranusSeven, line 165: Ditto.

xinference/deploy/cmdline.py
```diff
@@ -205,6 +222,7 @@ def model_terminate(
     endpoint: str,
     model_uid: str,
 ):
+    endpoint = getExistingEndpoint()
```
UranusSeven, line 225: Ditto.

PR #263, opened 2023-07-28
FEAT: add opt-125m pytorch model and add ut

pangyoki:
ATT


xinference/core/gradio.py
```diff
@@ -28,7 +28,7 @@
 MODEL_TO_FAMILIES: Dict[str, LLMFamilyV1] = dict(
     (model_family.model_name, model_family)
     for model_family in LLM_FAMILIES
-    if model_family.model_name not in ["baichuan", "baichuan-base", "llama-2"]
```
UranusSeven, line 31: how about `"chat" not in model_family.model_abilities`?

UranusSeven (approved):

PR #272, opened 2023-07-31
BUG: delete unnecessary path

pangyoki:
ATT


pangyoki:
the same pr: https://github.com/xorbitsai/inference/pull/276

PR #285, opened 2023-08-02
ENH: Support Enviroment Variable

RayJi01:
ATT

This implementation is capable to:

If the user didn't type in the -e or --endpoint when required:
    The program will search if user have exported an endpoint into OS environ. If so, the endpoint variable in the program 
    will automatically be set to that endpoint in OS. If not, we will give it the default endpoint back, which is 
    http://127.0.0.1:9997. 
if the user did have an input at the -e parameters, we keep the users' input to the endpoint variable.


xinference/constants.py
```diff
@@ -22,3 +22,4 @@
 XINFERENCE_DEFAULT_LOCAL_HOST = "127.0.0.1"
 XINFERENCE_DEFAULT_DISTRIBUTED_HOST = "0.0.0.0"
 XINFERENCE_DEFAULT_ENDPOINT_PORT = 9997
+XINFERENCE_ENV_ENDPOINT_VARIABLE = "XINFERENCE_ENDPOINT"
```
UranusSeven, line 25: `XINFERENCE_ENV_ENDPOINT_VARIABLE` -> `XINFERENCE_ENV_ENDPOINT`

xinference/deploy/cmdline.py
```diff
@@ -24,9 +26,22 @@
     XINFERENCE_DEFAULT_DISTRIBUTED_HOST,
     XINFERENCE_DEFAULT_ENDPOINT_PORT,
     XINFERENCE_DEFAULT_LOCAL_HOST,
+    XINFERENCE_ENV_ENDPOINT_VARIABLE,
 )
 
 
+def getExistingEndpointFromEnv(endpoint: Optional[str]) -> str:
```
UranusSeven, line 33: `getExistingEndpointFromEnv` -> `get_endpoint`

PR #290, opened 2023-08-03
FEAT: custom model

UranusSeven:


xinference/model/llm/llm_family.py
```diff
@@ -132,3 +153,125 @@ def cache_from_huggingface(
...
+def unregister_llm(model_name: str, permanent: bool):
+    with UD_LLM_FAMILIES_LOCK:
+        llm_family = None
+        for i, f in enumerate(UD_LLM_FAMILIES):
+            if f.model_name == model_name:
+                llm_family = f
+                break
+        if llm_family:
+            UD_LLM_FAMILIES.remove(llm_family)
+            if permanent:
```
pangyoki, line 260: The word `permanent` is ambiguous. When `permanent=True`, the file path is deleted. The semantics do not match.

xinference/client.py
```diff
@@ -480,6 +480,24 @@ def _gen_model_uid(cls) -> str:
         # generate a time-based uuid.
         return str(uuid.uuid1())
 
+    def register_model(self, model_type: str, model: str, persist: bool):
```
pangyoki, line 483: Can you give a test sample of `register_model`?

xinference/model/llm/llm_family.py
```diff
@@ -132,3 +153,125 @@ def cache_from_huggingface(
...
+
+    if not is_valid_model_name(llm_family.model_name):
+        raise ValueError(
+            f"Invalid model name {llm_family.model_name}. The model name must start with a letter"
+            f" or a digit, and can only contain letters, digits, underscores, or dashes."
+        )
+
+    with UD_LLM_FAMILIES_LOCK:
+        for family in BUILTIN_LLM_FAMILIES + UD_LLM_FAMILIES:
+            if llm_family.model_name == family.model_name:
```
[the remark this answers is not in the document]
UranusSeven: Duplicated `model_id` won't harm since HF will not cache a same model twice.

Also, we are going to support caching a model with an URI in the near future, in which case the `model_id` may not exists.

xinference/client.py
UranusSeven: Done.

xinference/model/llm/llm_family.py
UranusSeven: Removed.

pangyoki (approved):
LGTM

PR #304, opened 2023-08-04
DEMO: LangChain QA System with Xinference LLMs and Milvus Vector DB

jiayini1119:


.gitignore
```diff
@@ -138,6 +138,7 @@ static/
 
 # doc
 doc/source/savefig/
+*.txt
```
UranusSeven, line 141: Is this change relavent to the demo?

examples/xinference_demo.ipynb
UranusSeven, line 1: Better to have a more informative filename.

.gitignore
jiayini1119: I put the query document in the repo initially, but I think you are right we don't need to make this change. I will delete it.

UranusSeven (approved):

PR #309, opened 2023-08-04
BUG: handle ChatGLM ggml specific case for RESTful API

### LINKED ISSUE #306: BUG: RESTful API not compatible with OpenAI API
### Describe the bug
you said your RESTful API(compatible with OpenAI API)，but i post a openai json ，it response a errror.
image

image

my json payload is from openai  and change model name to model id:
```
{
     "model": "c0021658-3289-11ee-89d1-00155d722c2a",
     "messages": [{"role": "user", "content": "Say this is a test!"}],
     "temperature": 0.7
   }
```
image

the error info:
```
getattr(self._model, "chat")(prompt, *args, **kwargs)
TypeError: [address=127.0.0.1:32829, pid=116] ChatglmCppChatModel.chat() takes from 2 to 4 positional arguments but 5 were given
```

jiayini1119:
fix #306.

Chatgml ggml's chat method does not have the optional argument `system_prompt`, which is different from other LLMs.

### DIFF
--- a/xinference/core/restful_api.py
+++ b/xinference/core/restful_api.py
@@ -480,7 +480,7 @@ class RESTfulAPIActor(xo.Actor):
             (msg["content"] for msg in body.messages if msg["role"] == "system"), None
         )
 
-        chat_history = body.messages
+        chat_history = body.messages[:-1]  # exclude the prompt
 
         model_uid = body.model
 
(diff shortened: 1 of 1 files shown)


xinference/core/restful_api.py
```diff
@@ -494,16 +494,39 @@ async def create_chat_completion(
...
+
+        except ValueError as ve:
+            logger.error(str(ve), exc_info=True)
+            raise HTTPException(status_code=400, detail=str(ve))
+
+        except Exception as e:
+            logger.error(e, exc_info=True)
+            raise HTTPException(status_code=500, detail=str(e))
+
+        is_chatglm = desc.get("model_format") == "ggmlv3" and "chatglm" in desc.get(
```
UranusSeven, line 508: `is_chatglm_ggml` sounds more accurate.

xinference/core/restful_api.py
```diff
@@ -494,16 +494,39 @@ async def create_chat_completion(
...
+            "model_name", ""
+        )
+
+        if is_chatglm and system_prompt is not None:
+            raise HTTPException(
```
UranusSeven, line 513: I'd prefer just ignore the system prompt. What do you think?

UranusSeven (approved):

PR #320, opened 2023-08-07
ENH: Include model revision

### LINKED ISSUE #317: ENH: including model revision into model spec
HF models & scripts can be updated, which may result in model re-downloading or introduce incompatible changes with xinference.

### Describe the solution you'd like
Add a model revison to each pytorch model spec. When caching a model, specify the model revision.

RayJi01:
Resolves #317 

For GGML Model, all Model Revision are set to null for placeholder (later changes may appear and thus may change the revision)
For Pytorch Model, all revision are currently set to the latest commit on HuggingFace.

### DIFF
--- a/xinference/model/llm/llm_family.json
+++ b/xinference/model/llm/llm_family.json
@@ -31,8 +31,7 @@
           "q8_0"
         ],
         "model_id": "TheBloke/baichuan-llama-7B-GGML",
-        "model_file_name_template": "baichuan-llama-7b.ggmlv3.{quantization}.bin",
-        "model_revision": null
+        "model_file_name_template": "baichuan-llama-7b.ggmlv3.{quantization}.bin"
       },
       {
         "model_format": "pytorch",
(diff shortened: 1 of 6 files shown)


xinference/model/llm/llm_family.json
```diff
@@ -257,7 +264,8 @@
           "q8_0"
         ],
         "model_id": "TheBloke/vicuna-33B-GGML",
-        "model_file_name_template": "vicuna-33b.ggmlv3.{quantization}.bin"
+        "model_file_name_template": "vicuna-33b.ggmlv3.{quantization}.bin",
+        "model_revision": null
```
UranusSeven, line 268: Explicitly specifying null is unnecessary in this case.

RayJi01: every model_revision: null is deleted in the latest version

xinference/model/llm/pytorch/core.py
```diff
@@ -79,7 +79,7 @@ def _sanitize_model_config(
     ) -> PytorchModelConfig:
         if pytorch_model_config is None:
             pytorch_model_config = PytorchModelConfig()
-        pytorch_model_config.setdefault("revision", "main")
+        pytorch_model_config["revision"] = self.model_spec.model_revision
```
UranusSeven, line 82: Use `setdefault`.

RayJi01: updated in the latest version

UranusSeven (approved):

PR #331, opened 2023-08-08
BUG: fix mps and cuda device detection for pytorch model

pangyoki:
ATT


xinference/model/llm/pytorch/core.py
```diff
@@ -217,6 +197,32 @@ def load(self):
             self._model.to(self._device)
         logger.debug(f"Model Memory: {self._model.get_memory_footprint()}")
 
+    def _select_device(self, device):
```
UranusSeven, line 200: type hint.

UranusSeven (approved):

Issue #164, opened 2023-07-13
BUG: logs from model not correctly printed


UranusSeven:
### Describe the bug
The log from this line is not working even when logging level is set to debug at the beginning.

https://github.com/UranusSeven/inference/blob/8310506590613372cb1b8b28c7d6adbbcb0d27ff/xinference/model/llm/core.py#L185

### To Reproduce
To help us to reproduce this bug, please provide information below:

1. Your Python version.
2. The version of xinference you use.
3. Versions of crucial packages.
4. Full stack of the error.
5. Minimized code to reproduce the error.

PR #335, opened 2023-08-09
BUG: Fix log level on subprocess

RayJi01:
Now if user specified log-level in their command line, like: xinference --log-level debug(any upper lower case of debug), all the processes, including main and subprocess, will have correct log level corresponding to the log log level input.


UranusSeven (approved):

UranusSeven:
Also fixes #164

PR #344, opened 2023-08-11
ENH: Add old ui as example

Bojun-Feng:
Rewrote deprecated Gradio UI with RESTful Client and added it to examples as `gradio_arena.py`.

Instructions should be available by running the file with the `--help` tag.
e.g. `python path/to/inference/examples/gradio_arena.py --help`

Here are the instructions, in case anyone needs it:
```
instructions to run:

    1. Install Xinference, Llama-cpp-python, and other dependencies if necessary
    2. Run command `xinference --host "localhost"` in terminal
    3. You should see something similar to the following output:

    INFO:xinference:Xinference successfully started. Endpoint: http://localhost:9997
    INFO:xinference.core.service:Worker 127.0.0.1:21561 has been added successfully
    INFO:xinference.deploy.worker:Xinference worker successfully started.

    4. In the output, locate the endpoint. In the above case it is `http://localhost:9997`
    5. Run this python file in new terminal window, change the endpoint accordingly

    example run command (feel free to copy):

    python gradio_arena.py \
    --endpoint http://localhost:9997
```


examples/gradio_arena.py
```diff
@@ -0,0 +1,430 @@
+import uuid
```
UranusSeven, line 1: License.

examples/gradio_arena.py
```diff
@@ -0,0 +1,430 @@
...
+
+        def update_message(text_in: str):
+            return "", text_in
+
+        msg.submit(update_message, inputs=[msg], outputs=[msg, model_text])
+        gr.ClearButton(components=[chat, msg, model_text])
+
+    def build(self):
+        with gr.Blocks() as blocks:
+            with gr.Tab(self._locale("Chat")):
```
UranusSeven, line 387: I think we can just keep the Arena tab.

UranusSeven:
Please also add this example to the docs: https://inference.readthedocs.io/en/latest/examples/index.html

You need to create a new rst file in [this folder](https://github.com/xorbitsai/inference/tree/main/doc/source/examples) and add it to the [index's tree of contents](https://github.com/xorbitsai/inference/blob/main/doc/source/examples/index.rst).

doc/source/examples/ai_podcast.rst
```diff
@@ -74,6 +74,4 @@ Chinese (AI_Podcast_ZH.py)
 
 **Source Code** :
 
-    * `AI_Podcast <https://github.com/xorbitsai/inference/blob/main/examples/AI_podcast.py>`_ (English Version)
-
-    * AI_Podcast_ZH (Chinese Version)
\ No newline at end of file
+    * `AI_Podcast <https://github.com/xorbitsai/inference/examples/gradio_arena.py>`
```
UranusSeven, line 77: Seems like a mistake.

Bojun-Feng:
Close as stale

PR #358, opened 2023-08-16
FEAT: Support Launching Model with Uid

### LINKED ISSUE #254: FEAT: Support launch model with specific model uid
Support launch model with specific model uid when using command line to launch model.

Bojun-Feng:
Add `--model-uid -i` in the terminal as an optional argument that overrides the default randomly generated uid.

Resolves #254 

example: Run the following two commands in two separate terminal windows.
1. `xinference`
2. `xinference launch --model-name chatglm --model-uid chatglm`


UranusSeven:
More safety check is required, including the length, encoding, valid chars.

Bojun-Feng:
Close as stale

PR #368, opened 2023-08-20
FEAT: Adding support for stream returning completion results

takatost:
Currently, although i can pass the stream parameter and stream the response, it's actually still blocking and waits for the complete result before streaming the output. To achieve non-blocking streaming, i need to combine it with the stream parameter of requests.

### DIFF
--- a/xinference/client.py
+++ b/xinference/client.py
@@ -392,13 +392,15 @@ class RESTfulChatModelHandle(RESTfulGenerateModelHandle):
             for key, value in generate_config.items():
                 request_body[key] = value
 
-        response = requests.post(url, json=request_body)
+        stream = bool(generate_config and generate_config.get("stream"))
+        response = requests.post(url, json=request_body, stream=stream)
+
         if response.status_code != 200:
             raise RuntimeError(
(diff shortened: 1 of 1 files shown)


jiayini1119:
Hi, thanks for the contribution!

xinference/client.py
```diff
@@ -282,13 +282,15 @@ def generate(
             for key, value in generate_config.items():
                 request_body[key] = value
 
-        response = requests.post(url, json=request_body)
+        stream = bool(generate_config and generate_config.get("stream"))
```
UranusSeven, line 285: Get may result in a KeyError if stream is absent.

xinference/client.py
```diff
@@ -390,13 +392,15 @@ def chat(
...
```
UranusSeven, line 395: Ditto.

xinference/client.py
```diff
@@ -452,13 +456,15 @@ def chat(
...
```
UranusSeven, line 459: Ditto.

UranusSeven (approved):

PR #414, opened 2023-09-01
FEAT: Support CodeLlama-Instruct

### LINKED ISSUE #408: FEAT: Support CodeLlama-Instruct

jiayini1119:
Resolves #408.

### DIFF
--- a/xinference/model/llm/utils.py
+++ b/xinference/model/llm/utils.py
@@ -76,7 +76,10 @@ class ChatModelMixin:
                 else:
                     ret += role
             return ret
-        elif prompt_style.style_name == "LLAMA2":
+        elif (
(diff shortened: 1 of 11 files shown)


xinference/model/llm/utils.py
```diff
@@ -76,7 +76,10 @@ def get_prompt(
                 else:
                     ret += role
             return ret
-        elif prompt_style.style_name == "LLAMA2":
+        elif (
+            prompt_style.style_name == "LLAMA2"
```
UranusSeven, line 80: The prompt style name can be reuse. How about just use "LLAMA2".

UranusSeven (approved):

PR #417, opened 2023-09-01
FEAT: Improve torch inference performance

codingl2k1:


aresnow1:
We need to add tests to check if it works well for GGML models.

codingl2k1:
> Seems it always hang, it may not be caused by new tests, we can try to run all tests locally to find the reasons.

Fixed. Thanks.

UranusSeven (approved):

PR #444, opened 2023-09-12
FEAT: Support OpenBuddy

codingl2k1:


xinference/model/llm/utils.py
```diff
@@ -174,6 +174,9 @@ def get_prompt(
                 else:
                     ret += role + ": Let's think step by step."
             return ret
+        elif prompt_style.style_name == "INSTRUCTION":
```
UranusSeven, line 177: Multi-round conversation is not supported？

UranusSeven:
Please add doc for this model.

UranusSeven:
Also, README needs to be updated.

xinference/model/llm/utils.py
codingl2k1: Yes. Multi-round is not good for translate.

codingl2k1:
> Also, README needs to be updated.

Fixed. Thanks.

UranusSeven (approved):

PR #446, opened 2023-09-13
FEAT: Support gguf

aresnow1:


xinference/model/llm/ggml/llamacpp.py
```diff
@@ -176,15 +210,21 @@ def load(self):
...
+            model_path = self._convert_ggml_to_gguf(model_path)
+
+        try:
+            self._llm = Llama(
+                model_path=model_path,
+                verbose=False,
+                **self._llamacpp_model_config,
+            )
+        except AssertionError:
+            raise RuntimeError(f"Load model {self.model_spec} failed")
```
UranusSeven, line 223: Better add the `self.model_family.model_name` to the error message.

xinference/model/llm/ggml/llamacpp.py
```diff
@@ -176,15 +210,21 @@ def load(self):
...
 
     @classmethod
     def match(cls, llm_family: LLMFamilyV1, llm_spec: LLMSpecV1) -> bool:
-        if llm_spec.model_format != "ggmlv3":
+        if llm_spec.model_format != "ggmlv3" and llm_spec.model_format != "ggufv2":
```
UranusSeven, line 227: How about `llm_spec.model_format not in ["ggmlv3", "ggufv2"]`

xinference/model/llm/ggml/llamacpp.py
```diff
@@ -258,7 +298,7 @@ def __init__(
...
```
UranusSeven, line 301: Ditto.

xinference/model/llm/llm_family.json
```diff
@@ -1322,6 +1322,66 @@
         ],
         "model_id": "TheBloke/CodeLlama-34B-fp16",
         "model_revision": "f91d0cf7fc338cdc726f9c72d5ea15fe51bb16e9"
+      },
+      {
+        "model_format": "ggufv2",
+        "model_size_in_billions": 7,
+        "quantizations": [
+          "Q2_K",
```
UranusSeven, line 1330: A problem here. 

The previous naming of quantizations are like `q_XX`, but recent naming became `Q_XX`. This inconcsistency could confuse users since they can hardly konw the correct quantization name to use.

Maybe we should have our own names for ggml quantizations. What do you think?

xinference/model/llm/llm_family.json
```diff
@@ -1522,5 +1702,31 @@
         "model_revision": "fa88072fee36e36282287410e00897df2f59e09b"
       }
     ]
+  },
+  {
+    "version": 1,
+    "context_length": 2048,
+    "model_name": "TinyLlama",
```
UranusSeven, line 1709: All of our model names are in lowercase. To follow the convention, how about using the name `tiny-llama`?

xinference/model/llm/llm_family.json
aresnow1: I prefer to use what llamacpp use, If there is a new quantization that does not match previous one, compatibility will become a major issue.

UranusSeven (approved):

Issue #448, opened 2023-09-14
BUG: Incorrect traceback if raises an exception when stream iteration 


codingl2k1:
### To Reproduce
To help us to reproduce this bug, please provide information below:

```python
xinference/tests/test_client.py:188 (test_RESTful_client)
self = <urllib3.response.HTTPResponse object at 0x2b3f437c0>

    def _update_chunk_length(self):
        # First, we'll figure out length of a chunk and then
        # we'll try to read it from socket.
        if self.chunk_left is not None:
            return
        line = self._fp.fp.readline()
        line = line.split(b";", 1)[0]
        try:
>           self.chunk_left = int(line, 16)
E           ValueError: invalid literal for int() with base 16: b''

../../.pyenv/versions/3.11.4/lib/python3.11/site-packages/urllib3/response.py:761: ValueError

During handling of the above exception, another exception occurred:

self = <urllib3.response.HTTPResponse object at 0x2b3f437c0>

    @contextmanager
    def _error_catcher(self):
        """
        Catch low-level python exceptions, instead re-raising urllib3
        variants, so that low-level exceptions are not leaked in the
        high-level api.
    
        On exit, release the connection back to the pool.
        """
        clean_exit = False
    
        try:
            try:
>               yield

../../.pyenv/versions/3.11.4/lib/python3.11/site-packages/urllib3/response.py:444: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../.pyenv/versions/3.11.4/lib/python3.11/site-packages/urllib3/response.py:828: in read_chunked
    self._update_chunk_length()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <urllib3.response.HTTPResponse object at 0x2b3f437c0>

    def _update_chunk_length(self):
        # First, we'll figure out length of a chunk and then
        # we'll try to read it from socket.
        if self.chunk_left is not None:
            return
        line = self._fp.fp.readline()
        line = line.split(b";", 1)[0]
        try:
            self.chunk_left = int(line, 16)
        except ValueError:
            # Invalid chunked protocol response, abort.
            self.close()
>           raise InvalidChunkLength(self, line)
[...]

PR #449, opened 2023-09-14
ENH: Safe iterate stream of ggml model

codingl2k1:
- [x] Use FastAPI stream.
- [x] Fix parallel ggml stream undefined behaviour.

Fix: https://github.com/xorbitsai/inference/issues/448

### DIFF
--- a/xinference/client.py
+++ b/xinference/client.py
@@ -226,7 +226,11 @@ def chat_streaming_response_iterator(
     """
 
     for line in response_chunk:
-        yield json.loads(line.decode("utf-8"))
+        content = json.loads(line.decode("utf-8"))
+        error = content.get("error", None)
+        if error is not None:
+            raise Exception(str(error))
+        yield content
 
 
 class RESTfulModelHandle:
--- a/xinference/client.py
+++ b/xinference/client.py
@@ -329,7 +329,7 @@ class RESTfulGenerateModelHandle(RESTfulEmbeddingModelHandle):
             )
 
         if stream:
-            return streaming_response_iterator(response.iter_lines())
+            return streaming_response_iterator(response.iter_content(chunk_size=None))
 
         response_data = response.json()
         return response_data
--- a/xinference/client.py
+++ b/xinference/client.py
@@ -208,15 +208,15 @@ def streaming_response_iterator(
 
 # Duplicate code due to type hint issues
 def chat_streaming_response_iterator(
-    response_lines: Iterator[bytes],
+    response_chunk: Iterator[bytes],
 ) -> Iterator["ChatCompletionChunk"]:
     """
     Create an Iterator to handle the streaming type of generation.
 
     Parameters
     ----------
-    response_lines: Iterator[bytes]
-        Generated lines by the Model Generator.
+    response_chunk: Iterator[bytes]
+        Generated chunk by the Model Generator.
 
     Returns
     -------
@@ -225,11 +225,8 @@ def chat_streaming_response_iterator(
 
     """
 
-    for line in response_lines:
-        line = line.strip()
-        if line.startswith(b"data:"):
-            data = json.loads(line.decode("utf-8").replace("data: ", "", 1))
-            yield data
+    for line in response_chunk:
+        yield json.loads(line.decode("utf-8"))
 
 
 class RESTfulModelHandle:
@@ -405,7 +402,9 @@ class RESTfulChatModelHandle(RESTfulGenerateModelHandle):
             )
 
         if stream:
-            return chat_streaming_response_iterator(response.iter_lines())
+            return chat_streaming_response_iterator(
+                response.iter_content(chunk_size=None)
+            )
 
         response_data = response.json()
         return response_data
@@ -469,7 +468,9 @@ class RESTfulChatglmCppChatModelHandle(RESTfulEmbeddingModelHandle):
             )
 
         if stream:
-            return chat_streaming_response_iterator(response.iter_lines())
+            return chat_streaming_response_iterator(
+                response.iter_content(chunk_size=None)
+            )
 
         response_data = response.json()
         return response_data
--- a/xinference/client.py
+++ b/xinference/client.py
@@ -182,14 +182,14 @@ class ChatglmCppChatModelHandle(EmbeddingModelHandle):
 
 
 def streaming_response_iterator(
-    response_lines: Iterator[bytes],
+    response_chunk: Iterator[bytes],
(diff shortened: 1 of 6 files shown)


xinference/core/model.py
```diff
@@ -168,6 +171,11 @@ async def next(
 
         def _wrapper():
             try:
+                if self._lock is not None and generator_uid not in self._generators:
+                    raise Exception(
```
UranusSeven, line 175: How about raising an RuntimeException since this is totally unexpected？

xinference/core/model.py
```diff
@@ -110,6 +110,9 @@ def load(self):
 
     async def _wrap_generator(self, ret: Any):
         if inspect.isgenerator(ret):
+            if self._lock is not None:
+                # Make sure only one iterator is valid.
+                self._generators.clear()
```
UranusSeven, line 115: How about simply raise an exception when self._lock is not None and self._generators is not empty?

xinference/core/restful_api.py
```diff
@@ -517,32 +513,16 @@ async def create_completion(self, request: Request, body: CreateCompletionReques
...
-                recv_chan, data_sender_callable=partial(event_publisher, send_chan)
-            )
 
+            async def encode_iterator():
+                try:
+                    iterator = await model.generate(body.prompt, kwargs)
+                    async for item in iterator:
+                        yield json.dumps(item)
+                except Exception as ex:
+                    yield json.dumps({"error": str(ex)})
```
UranusSeven, line 523: Better log the error for debugging since `str(ex)` may not contain the full invocation stack.

xinference/core/restful_api.py
```diff
@@ -640,37 +620,21 @@ async def create_chat_completion(
...
+                    if is_chatglm_ggml:
+                        iterator = await model.chat(prompt, chat_history, kwargs)
+                    else:
+                        iterator = await model.chat(
+                            prompt, system_prompt, chat_history, kwargs
+                        )
+                    async for item in iterator:
+                        yield json.dumps(item)
+                except Exception as ex:
+                    yield json.dumps({"error": str(ex)})
```
UranusSeven, line 635: Ditto.

xinference/core/restful_api.py
```diff
@@ -517,32 +513,16 @@ async def create_completion(self, request: Request, body: CreateCompletionReques
...
-                            await inner_send_chan.send(dict(closing=True))
-                            raise e
-                    except Exception as e:
-                        raise HTTPException(status_code=500, detail=str(e))
-
-            return EventSourceResponse(
-                recv_chan, data_sender_callable=partial(event_publisher, send_chan)
-            )
 
+            async def encode_iterator():
```
UranusSeven, line 517: What does `encode_iterator` mean? Since GPTs are decoder-only models, maybe `decode_iterator` sounds more reasonable?

Or simply name it `stream_results`?

xinference/core/restful_api.py
codingl2k1: Good suggestion.

xinference/core/restful_api.py
codingl2k1: OK

xinference/core/model.py
codingl2k1: Current behavior similars to C++ iterator, append data invalidate old iterators.

xinference/core/model.py
codingl2k1: OK

xinference/core/model.py
codingl2k1: Fixed. Thanks.

xinference/core/restful_api.py
codingl2k1: Fixed.

xinference/client.py
```diff
@@ -208,15 +208,15 @@ def streaming_response_iterator(
 
 # Duplicate code due to type hint issues
 def chat_streaming_response_iterator(
-    response_lines: Iterator[bytes],
```
UranusSeven, line 211: Does `streaming_response_iterator` need to be updated?

xinference/core/model.py
```diff
@@ -110,6 +110,8 @@ def load(self):
 
     async def _wrap_generator(self, ret: Any):
         if inspect.isgenerator(ret):
+            if self._lock is not None and self._generators:
+                raise Exception("Parallel iteration is not supported by ggml.")
```
UranusSeven, line 114:
```suggestion
                raise Exception("Parallel generation is not supported by ggml.")
```

xinference/core/restful_api.py
```diff
@@ -517,32 +513,17 @@ async def create_completion(self, request: Request, body: CreateCompletionReques
...
-                recv_chan, data_sender_callable=partial(event_publisher, send_chan)
-            )
 
+            async def stream_results():
+                try:
+                    iterator = await model.generate(body.prompt, kwargs)
+                    async for item in iterator:
+                        yield json.dumps(item)
+                except Exception as ex:
+                    logger.exception("Chat completion stream got an error: %s", ex)
```
UranusSeven, line 523:
```suggestion
                    logger.exception("Completion stream got an error: %s", ex)
```

xinference/client.py
codingl2k1: Yes. The respose stream is iterated by chunk instead of by line.

UranusSeven: I mean the method `streaming_response_iterator`.  You've modified `chat_streaming_response_iterator`, but 
 not `streaming_response_iterator`.

codingl2k1: > I mean the method `streaming_response_iterator`. You've modified `chat_streaming_response_iterator`, but not `streaming_response_iterator`.

Fixed, thanks. But why two same methods?

xinference/tests/test_client.py
```diff
@@ -231,12 +231,33 @@ def test_RESTful_client(setup):
     completion = model.chat("What is the capital of France?")
     assert "content" in completion["choices"][0]["message"]
 
-    streaming_response = model.chat(
-        prompt="What is the capital of France?", generate_config={"stream": True}
-    )
+    def _check_stream():
+        streaming_response = model.chat(
+            prompt="What is the capital of France?", generate_config={"stream": True}
```
aresnow1, line 236: generate_config={"stream": True, "max_tokens": 5}

codingl2k1: Fixed, thanks.

aresnow1 (approved):
LGTM

PR #484, opened 2023-09-21
FEAT: Support stable diffusion

codingl2k1:
- Enable self-host CI for stable diffusion


xinference/model/multimodal/core.py
```diff
@@ -0,0 +1,100 @@
...
+from ...constants import XINFERENCE_CACHE_DIR
+from ..core import ModelDescription
+from .stable_diffusion.core import DiffusionModel
+
+MAX_ATTEMPTS = 3
+
+logger = logging.getLogger(__name__)
+
+
+class MultimodalModelSpec(BaseModel):
```
UranusSeven, line 30: In `xinference`, a "model family" refers to a group of models sharing similar characteristics such as network structure. Each model within this family has its own detailed specification, termed "model spec."

Given this, perhaps a more suitable name for this class would be `MultimodalModelFamilyV1`. The `V1` suffix aids in differentiating potential incompatible versions.

xinference/client.py
```diff
@@ -680,6 +713,8 @@ def get_model(self, model_uid: str) -> "ModelHandle":
                 raise ValueError(f"Unrecognized model ability: {desc['model_ability']}")
         elif desc["model_type"] == "embedding":
             return EmbeddingModelHandle(model_ref, self._isolation)
+        elif desc["model_type"] == "multimodal":
```
UranusSeven, line 716: `multimodal` includes text2img, img2text, text2audio, audio2text, etc. 

`images` could be a better name and is also the way OpenAI names their API.

xinference/model/multimodal/stable_diffusion/core.py
```diff
@@ -0,0 +1,86 @@
...
+                buffered = BytesIO()
+                _img.save(buffered, format="jpeg")
+                return base64.b64encode(buffered.getvalue())
+
+            with ThreadPoolExecutor() as executor:
+                results = list(map(partial(executor.submit, _gen_base64_image), images))
+                image_list = [Image(url=None, b64_json=s.result()) for s in results]
+            return ImageList(data=image_list)
+        else:
+            raise Exception(f"Unsupported response format: {response_format}")
```
UranusSeven, line 86: How about raising a ValueError?

UranusSeven:
A new RESTful API needs be added and it should follow the OpenAI's API spec: https://platform.openai.com/docs/api-reference/images/create

xinference/types.py
```diff
@@ -17,6 +17,15 @@
 from typing_extensions import Literal, NotRequired, TypedDict
 
 
+class Image(TypedDict):
+    url: Optional[str]
+    b64_json: Optional[str]
+
+
+class ImageList(TypedDict):
+    data: List[Image]
```
UranusSeven, line 26: Besides `data`, there should be another field `created`.

Here's an example from OpenAI's doc:
```json
{
  "created": 1589478378,
  "data": [
    {
      "url": "https://..."
    },
    {
      "url": "https://..."
    }
  ]
}
```

xinference/client.py
```diff
@@ -181,6 +182,38 @@ def chat(
         return self._isolation.call(coro)
 
 
+class MultimodalModelHandle(ModelHandle):
```
UranusSeven, line 185: A RESTful impl is also needed.

xinference/types.py
codingl2k1: Fixed. Thanks.

xinference/client.py
codingl2k1: Yes, multimodal includes text2img, img2text, text2audio, audio2text, etc. We can add additional methods to the `MultimodalModelHandle`.

xinference/model/multimodal/stable_diffusion/core.py
codingl2k1: Fixed. Thanks.

xinference/model/multimodal/core.py
codingl2k1: Fixed. Thanks.

xinference/client.py
codingl2k1: Fixed. Thanks.

xinference/client.py
codingl2k1: Fixed. Thanks.

UranusSeven (approved):

PR #507, opened 2023-10-06
ENH: Disable 4-bit and 8-bit quantization on MacOS

### LINKED ISSUE #483: ENH: Disable 4-bit and 8-bit quantization on MacOS

Bojun-Feng:
Resolve #483 at frontend level


UranusSeven:
For most models, 8-bit quant should work on macOS. 

However, if the model uses bf16, it cannot run on macOS, even the non-quant model.

Bojun-Feng:
In this case, maybe we can add a field to the model json specifying whether model is bf16. Am I missing a easier way to distinguish bf16 models?

Bojun-Feng:
Close as stale

PR #520, opened 2023-10-09
REF: Integrate xinference-client package

ChengjieLi28:


xinference/tests/test_client.py
```diff
@@ -189,222 +181,6 @@ def test_client_custom_model(setup):
     assert custom_model_reg is None
 
 
-@pytest.mark.skipif(os.name == "nt", reason="Skip windows")
```
UranusSeven, line 192: I see you've moved the tests to xinference-client. But these tests are not just unittests for the client, they also serve as end to end test for the system. Although some test cases are also covered by `xinference/core/tests/test_restful_api.py`, parallel generation and embedding creation are not.

To ensure the test quality, I think we can keep them for now. And once `test_restful_api.py` covers all the end to end test cases, they could be savely removed.

xinference/model/llm/pytorch/core.py
```diff
@@ -33,20 +35,6 @@
 logger = logging.getLogger(__name__)
 
 
-class PytorchGenerateConfig(TypedDict, total=False):
```
UranusSeven, line 36: Let the server decide what the generate config should be. It is OK to have a duplicated definition on the client side.

xinference/model/llm/ggml/llamacpp.py
```diff
@@ -30,34 +32,9 @@
 from ..utils import ChatModelMixin
 from .ctransformers import CTRANSFORMERS_SUPPORTED_MODEL
 
-if TYPE_CHECKING:
-    from llama_cpp import LogitsProcessorList, StoppingCriteriaList
-
 logger = logging.getLogger(__name__)
 
 
-class LlamaCppGenerateConfig(TypedDict, total=False):
```
UranusSeven, line 39: Let the server decide what the generate config should be. It is OK to have a duplicated definition on the client side.

PR #529, opened 2023-10-12
FEAT: Add --model-uid to launch sub command

codingl2k1:
- In order to use xinference more flexibly, provide `--model-uid` for launch sub command. e.g. specify a fixed model uid for continue integration.

### DIFF
--- a/xinference/client/restful/restful_client.py
+++ b/xinference/client/restful/restful_client.py
@@ -315,6 +315,7 @@ class Client:
         self,
         model_name: str,
         model_type: str = "LLM",
+        model_uid: str = None,
         model_size_in_billions: Optional[int] = None,
         model_format: Optional[str] = None,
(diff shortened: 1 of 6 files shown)


UranusSeven:
There's already another PR for this: https://github.com/xorbitsai/inference/pull/358

The problem is by adding this feature, there's a way to attack the system by passing huge, or invalid model uid to the system.

codingl2k1:
Close this PR due to duplication.

codingl2k1:
> There's already another PR for this: #358
> 
> The problem is by adding this feature, there's a way to attack the system by passing huge, or invalid model uid to the system.

But, the launch cmdline is only used for local user or xinference administrator. Currently, the model uid is generated by restful client which is vulnerable to attack the system.

UranusSeven (approved):

Issue #498, opened 2023-09-27
BUG: 启动chatglm2-6b-32k报错，got multiple values for keyword argument 'trust_remote_code'，模型已缓存


tjudadc:
### Describe the bug
![image](https://github.com/xorbitsai/inference/assets/42208489/8755e0e6-b63e-46bb-ac76-99c9122977f4)

### To Reproduce
To help us to reproduce this bug, please provide information below:

1. Your Python version.
2. The version of xinference you use.
3. Versions of crucial packages.
4. Full stack of the error.
5. Minimized code to reproduce the error.

UranusSeven:
Hi, thanks for reporting this issue! We are aware of the problem and will fix it ASAP.

A workaround for this issue is to downgrade to v0.4.4.

I will let you know once this problem has been resolved.

richzw:
@UranusSeven , I got the same error under version 0.5.3. 

```
File "/home/ec2-user/.local/lib/python3.9/site-packages/xinference/core/restful_api.py", line 419, in launch_model
    model_uid = await self._supervisor_ref.launch_builtin_model(
  File "xoscar/core.pyx", line 288, in __pyx_actor_method_wrapper
  File "xoscar/core.pyx", line 422, in _handle_actor_result
  File "xoscar/core.pyx", line 465, in _run_actor_async_generator
  File "xoscar/core.pyx", line 466, in xoscar.core._BaseActor._run_actor_async_generator
  File "xoscar/core.pyx", line 471, in xoscar.core._BaseActor._run_actor_async_generator
  File "/home/ec2-user/.local/lib/python3.9/site-packages/xinference/core/supervisor.py", line 249, in launch_builtin_model
    yield _launch_one_model(rep_model_uid)
  File "xoscar/core.pyx", line 476, in xoscar.core._BaseActor._run_actor_async_generator
  File "xoscar/core.pyx", line 422, in _handle_actor_result
  File "xoscar/core.pyx", line 465, in _run_actor_async_generator
  File "xoscar/core.pyx", line 466, in xoscar.core._BaseActor._run_actor_async_generator
  File "xoscar/core.pyx", line 471, in xoscar.core._BaseActor._run_actor_async_generator
  File "/home/ec2-user/.local/lib/python3.9/site-packages/xinference/core/supervisor.py", line 223, in _launch_one_model
    yield worker_ref.launch_builtin_model(
[...]

PR #555, opened 2023-10-18
BUG: set trust_remote_code to true by default

richzw:
Try to fix the error ` got multiple values for keyword argument 'trust_remote_code'` when load falcon-instruct model, the related the issue https://github.com/xorbitsai/inference/issues/498

### DIFF
--- a/xinference/model/llm/pytorch/falcon.py
+++ b/xinference/model/llm/pytorch/falcon.py
@@ -57,7 +57,6 @@ class FalconPytorchModel(PytorchModel):
         model = AutoModelForCausalLM.from_pretrained(
             self.model_path,
             low_cpu_mem_usage=True,
-            trust_remote_code=True,
             **kwargs,
         )
         tokenizer.pad_token_id = 9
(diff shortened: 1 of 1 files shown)


richzw:
> Hello, and thank you for your work!
> 
> I'd like to point out something about issue #498. The error you're addressing is located on line 113, but the changes you made are on line 57.
> 
> Here's what I found:
> 
> In this link:
> 
> https://github.com/xorbitsai/inference/blob/9066d5b798726d40eb7501ae2d71f5e253e1fb88/xinference/model/llm/pytorch/core.py#L138-L140
> 
> `trust_remote_code` is given a value inside the parent class.
> 
> But in this link:
> 
> https://github.com/xorbitsai/inference/blob/9066d5b798726d40eb7501ae2d71f5e253e1fb88/xinference/model/llm/pytorch/falcon.py#L109-L120
> 
> on line 113, `trust_remote_code` is still used directly.
> 
> To fix this, please remove `trust_remote_code` on line 116.
> 
> Additionally, on line 109, `trust_remote_code` should be included when using `AutoTokenizer.from_pretrained`, similar to `FalconPytorchModel._load_model`. Could you please make this adjustment as well?
> 
> Thank you again for your help!

@UranusSeven , Thank you very much for your detailed reply, Sorry for the misunderstanding of the codes in the previous comment. Please review the commit again and let me know if there is any other issues.

UranusSeven (approved):
LGTM

PR #573, opened 2023-10-23
BUG: When change model revision, xinference still uses the previous model

ChengjieLi28:


xinference/model/llm/llm_family.py
```diff
@@ -380,20 +380,26 @@ def _get_meta_path(
     cache_dir: str,
     model_format: str,
     model_hub: str,
+    model_revision: Optional[str],
     quantization: Optional[str] = None,
 ):
     if model_format == "pytorch":
         if model_hub == "huggingface":
-            return os.path.join(cache_dir, "__valid_download")
+            return os.path.join(cache_dir, f"__valid_download_{model_revision}")
```
UranusSeven, line 388: The meta file has included the model revision , see`_generate_meta_file`. Is it necessary to add model revision at the end of the filename?

ChengjieLi28: > The meta file has included the model revision , see`_generate_meta_file`. Is it necessary to add model revision at the end of the filename?

But you are deciding whether to download the model or not according to the file `__valid_download`.

xinference/model/utils.py
```diff
@@ -102,3 +103,22 @@ def retry_download(
             raise RuntimeError(
                 f"Failed to download model '{model_name}' after multiple retries"
             )
+
+
+def valid_model_revision(
+    meta_path: str, expected_model_revision: Optional[str]
+) -> bool:
+    if not os.path.exists(meta_path):
```
UranusSeven, line 111: Address legacy meta files to ensure that upgrading xinference does not require users to download the existing model again.

To accomplish this, verify if the meta file content is in JSON format. If it's not in JSON format, this indicates that the meta file is legacy and should return True directly. Otherwise, check the model revision.

Log an entry when legacy meta files are detected and include a unit test.

UranusSeven (approved):

PR #596, opened 2023-10-31
ENH: Supports request limits for the model

ChengjieLi28:
Add a new parameter `request_limits` to `launch_model` interface to control the visit count on a single model instance at the same time.

### DIFF
--- a/xinference/core/model.py
+++ b/xinference/core/model.py
@@ -46,24 +46,28 @@ T = TypeVar("T")
 
 
 def request_limit(fn):
+    """
+    Used by ModelActor.
+    As a decorator, added to a ModelActor method to control
+    how many requests are accessing that method at the same time.
+    """
+
     async def wrapped_func(self, *args, **kwargs):
         logger.debug(
             f"Request {fn.__name__}, current serve request count: {self._serve_count}, request limit: {self._request_limits} for the model {self.model_uid()}"
         )
         if self._request_limits is not None:
-            async with self._limit_lock:
-                if 1 + self._serve_count <= self._request_limits:
-                    self._serve_count += 1
-                else:
-                    raise RuntimeError(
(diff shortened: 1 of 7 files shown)


xinference/core/model.py
```diff
@@ -45,6 +45,33 @@
...
+                    self._serve_count += 1
+                else:
+                    raise RuntimeError(
+                        f"Rate limit reached for the model. Request limit {self._request_limits} for the model: {self.model_uid()}"
+                    )
+        try:
+            ret = await fn(self, *args, **kwargs)
+        finally:
+            if self._request_limits is not None:
+                async with self._limit_lock:
```
codingl2k1, line 65: The `self._limit_lock` is not necessary.

xinference/core/model.py
```diff
@@ -45,6 +45,33 @@
 T = TypeVar("T")
 
 
+def request_limit(fn):
+    async def wrapped_func(self, *args, **kwargs):
```
codingl2k1, line 49: The `request_limit` decorator is only available for the ModelActor, can you add some comments?

xinference/core/model.py
```diff
@@ -109,24 +136,38 @@ async def __pre_destroy__(self):
...
     def load(self):
         self._model.load()
 
+    def model_uid(self):
+        return (
+            self._model.model_uid
+            if hasattr(self._model, "model_uid")
+            else (
+                self._model._model_uid
+                if hasattr(self._model, "_model_uid")
```
codingl2k1, line 166: Why both of `model_uid` and `_model_uid` are checked here?

ChengjieLi28: > Why both of `model_uid` and `_model_uid` are checked here?

Because `self._model` can be `EmbeddingModel` or `LLMModel`, `EmbeddingModel` has the attribute `_model_uid`,  and `LLMModel` has the attribute `model_uid`.

codingl2k1 (approved):
LGTM

PR #606, opened 2023-11-05
BUG: Fixing circular import by renaming types file, issue with emotes in streaming response, embedding tokenizer and system chat messages

AndiMajore:
Major Fix:

- Renamed xinference.types to xinference.xinference_types to fix circular dependency!

Some fixes for minor issues:

- Chat response streaming which formerly broke emotes
- Embedding use where the pad_token for the tokenizer was not initialized fully
- Issue when latest message for chatbot was a system command


aresnow1:
Can you describe what the problem of circular dependency is?

AndiMajore:
The problem is, that after restarting xinference and registering a model the application throws a circular dependency error. This, as far as I found out, that xinference.types.py is overriding the original python types class and as soon as the python types would imported by some nested dependency of the xinferenece.types.py it again tries to import the xinference.types.py which closes the circle. As a result I renamed the xinference.types file to not have any ambiguous imports happening.

Find my log regarding this issue attached.
[xinference_circular_dep.log](https://github.com/xorbitsai/inference/files/13270662/xinference_circular_dep.log)

aresnow1:
According to the logs, it seems like your working directory is "xinference", and Python find the "types.py" under xinference when import types, I guess it would not be a problem if change your working directory.

qinxuye:
I agree with renaming `types`, but maybe we need a better name than `xinference_types`.

PR #611, opened 2023-11-06
BUG: `No slot available` error for embedding and LLM model on one card

ChengjieLi28:
When there are only one GPU available and user launches vLLM-backend models first, then launches embedding models to GPU, `No slot available` error would be raised.
Here' the changes:
- Embedding models use their own slots, no relation with LLM models.
- The embedding model will first avoid the card where the LLM model is located, and if there is no slot, the user will be prompted to launch the embedding model first.
- After finding the cards where the embedding models can be placed, try to equalize the load (number of models) between the cards.

### DIFF
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -15,8 +15,9 @@
 import asyncio
 import os
 import platform
+from collections import defaultdict
 from logging import getLogger
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
 import xoscar as xo
 from xorbits._mars.resource import cuda_count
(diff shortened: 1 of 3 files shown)


xinference/core/worker.py
```diff
@@ -76,6 +78,45 @@ async def __pre_destroy__(self):
...
+                if not is_vllm_model:
+                    candidates.append(_dev)
+
+        if len(candidates) == 0:
+            raise RuntimeError(
+                "No available slot found for the embedding model. "
+                "We recommend to launch the embedding model first, and then launch the LLM models."
+            )
+
+        chosen_one = [-1, -1]  # dev, count
```
UranusSeven, line 107: Consider utilizing two variables for enhanced readability. What do you think?

aresnow1 (approved):

PR #616, opened 2023-11-06
ENH: cache status

UranusSeven:


xinference/core/supervisor.py
```diff
@@ -102,46 +106,91 @@ def get_status(self) -> Dict:
             "workers": self._worker_status,
         }
 
+    @staticmethod
+    def _to_llm_reg(llm_family: "LLMFamilyV1", is_builtin: bool) -> Dict[str, Any]:
```
aresnow1, line 110: Use lru_cache to wrap this function.

xinference/core/supervisor.py
```diff
@@ -102,46 +106,91 @@ def get_status(self) -> Dict:
...
+
+        specs = []
+        for spec in llm_family.model_specs:
+            cache_status = get_cache_status(llm_family, spec)
+            specs.append({**spec.dict(), "cache_status": cache_status})
+
+        return {**llm_family.dict(), "is_builtin": is_builtin, "model_specs": specs}
+
+    @staticmethod
+    def _to_embedding_model_reg(
```
aresnow1, line 121: Ditto.

xinference/core/supervisor.py
```diff
@@ -102,46 +106,91 @@ def get_status(self) -> Dict:
...
+
+        cache_status = get_cache_status(model_spec)
+        return {
+            **model_spec.dict(),
+            "cache_status": cache_status,
+            "is_builtin": is_builtin,
+        }
+
+    @staticmethod
+    def _to_image_model_reg(
```
aresnow1, line 134: Ditto.

xinference/core/supervisor.py
UranusSeven: Using an LRU cache might not be suitable for this scenario because the return value is subject to change.

xinference/core/supervisor.py
UranusSeven: Using an LRU cache might not be suitable for this scenario because the return value is subject to change.

PR #618, opened 2023-11-07
BUG: Rotating log does not create a new one when recreate the xinference cluster

ChengjieLi28:
This PR fixes these issues:
- Too many logs that have no relationship with `xinference` module in the rotate log
- When create a new cluster after destroying the previous one, the rotate log is not be a new one.
- `modelscope` overwrites the streaming handler that leads to log lost on console.

### DIFF
--- a/xinference/deploy/worker.py
+++ b/xinference/deploy/worker.py
@@ -22,8 +22,22 @@ from xorbits._mars.resource import cuda_count
 from xoscar import MainActorPoolType
 
 from ..core.worker import WorkerActor
+from ..deploy.utils import rollover_log_file
 
 logger = logging.getLogger(__name__)
+
+
+def _config_logging(address: str, supervisor_address: str):
+    worker_host = address.split(":")[0]
+    supervisor_host = supervisor_address.split(":")[0]
+    # Local deployment, worker not do rollover
+    if (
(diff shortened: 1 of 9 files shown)


xinference/deploy/worker.py
```diff
@@ -22,10 +22,24 @@
...
+    worker_host = address.split(":")[0]
+    supervisor_host = supervisor_address.split(":")[0]
+    # Local deployment, worker not do rollover
+    if (
+        worker_host == "127.0.0.1"
+        or worker_host == "0.0.0.0"
+        or worker_host == supervisor_host
+    ):
+        return
+    rollover_log_file(logger, "Xinference worker ")
```
UranusSeven, line 40: A more effective approach might be to direct the supervisor and worker logs to separate files. What are your thoughts on this?

ChengjieLi28: > A more effective approach might be to direct the supervisor and worker logs to separate files. What are your thoughts on this?

This approach requires that the filename to be written to the log be passed in dynamically, rather than determined at the outset, and may require relatively larger changes to existing code. I'm not sure it needs to be done in the current PR.

UranusSeven (approved):

PR #619, opened 2023-11-07
BUG: Compatible with openai 1.1

codingl2k1:
- Use `sse_starlette` handle fastapi stream response.

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -615,11 +615,14 @@ class RESTfulAPI:
         if body.stream:
 
             async def stream_results():
+                iterator = None
                 try:
                     iterator = await model.generate(body.prompt, kwargs)
                     async for item in iterator:
                         yield dict(data=json.dumps(item))
                 except Exception as ex:
+                    if iterator is not None:
+                        await iterator.destory()
                     logger.exception("Completion stream got an error: %s", ex)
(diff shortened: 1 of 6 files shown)


aresnow1:
Are these changes compatible with older version of OpenAI? If not, we need to update our requirements.

xinference/api/restful_api.py
```diff
@@ -811,15 +810,13 @@ async def stream_results():
                             prompt, system_prompt, chat_history, kwargs
                         )
                     async for item in iterator:
-                        yield f"data: {json.dumps(item)}\n"
+                        yield dict(data=json.dumps(item))
                 except Exception as ex:
```
UranusSeven, line 814: Client disconnects need to be handled: https://github.com/sysid/sse-starlette/blob/159bbece1a58434d12d67934dc99fab556737247/examples/example.py#L62

codingl2k1: I have handled all the exceptions including `asyncio.CancelledError`.

codingl2k1:
> Are these changes compatible with older version of OpenAI? If not, we need to update our requirements.

It's compatible with old version of OpenAI.

aresnow1:
The issue mentioned by comment is not introduced in this PR, merge this PR first and track this issue in #624

aresnow1 (approved):

PR #626, opened 2023-11-08
ENH: running model location & accelerators

UranusSeven:
The `list models` and `desc model` APIs have been updated to include the model address and the accelerators in use. Corresponding updates have been made to the frontend. The columns and actions now vary depending on the model type.

image

### DIFF
--- a/xinference/client/tests/test_client.py
+++ b/xinference/client/tests/test_client.py
@@ -125,6 +125,8 @@ def test_replica_model(setup):
 
     client2 = RESTfulClient(endpoint)
     info = client2.describe_model(model_uid=model_uid)
+    assert "address" in info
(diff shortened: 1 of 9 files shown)


aresnow1:
Has it tested on GPUs? Additionally, we could detect if the models use Apple's GPU with ggml format, I will open another issue to track this issue.

UranusSeven:
> Has it tested on GPUs? Additionally, we could detect if the models use Apple's GPU with ggml format, I will open another issue to track this issue.

Yes, I've tested it on GPUs. And for better readibility, the column name has been changed from `Accelerator` to `GPU Index`.

aresnow1 (approved):

PR #646, opened 2023-11-14
FEAT: Support TensorRT-LLM backend

aresnow1:
Support TensorRT-LLM backend.

- [ ] Implements TRTModel with generate method.
- [ ] Expose `launch_trt_model` to client.
- [ ] Doc and example


aresnow1:
Python API of in-flight batching is needed for this PR, and TensorRT-LLM team says it will be implemented in next versions.

Issue #647, opened 2023-11-15
BUG: The restful api server performance is poor


codingl2k1:
### Describe the bug

![profile](https://github.com/xorbitsai/inference/assets/138426806/4384fb02-d853-4be9-a9e5-a9d937e26d95)

The code that most affects performance is
1. field.validate in `serialize_response`
2. jsonable_encoder

### To Reproduce
To help us to reproduce this bug, please provide information below:

1. Your Python version.
3. The version of xinference you use.
4. Versions of crucial packages.
5. Full stack of the error.
6. Minimized code to reproduce the error.

PR #648, opened 2023-11-15
BUG: Fix restful api serialization slow

codingl2k1:
Fixes: https://github.com/xorbitsai/inference/issues/647

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -41,7 +41,7 @@ from fastapi.staticfiles import StaticFiles
 from PIL import Image
 from pydantic import BaseModel, Field
 from sse_starlette.sse import EventSourceResponse
-from starlette.responses import RedirectResponse
+from starlette.responses import RedirectResponse, JSONResponse as StarletteJSONResponse
 from typing_extensions import NotRequired, TypedDict
 from uvicorn import Config, Server
 from xoscar.utils import get_next_port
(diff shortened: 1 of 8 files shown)


xinference/api/restful_api.py
```diff
@@ -572,7 +617,7 @@ async def create_embedding(self, request: CreateEmbeddingRequest):
 
         try:
             embedding = await model.create_embedding(request.input)
-            return embedding
+            return Response(embedding, media_type="application/json")
```
ChengjieLi28, line 620: Is this `media_type="application/json"` neccessary? All other interfaces don't specify this option, should it be the default value?

xinference/api/restful_api.py
```diff
@@ -60,11 +63,28 @@
...
     top_p_field,
 )
-from ..types import ChatCompletion, Completion, CreateCompletion, Embedding, ImageList
+from ..types import ChatCompletion, Completion, CreateCompletion, ImageList
 
 logger = logging.getLogger(__name__)
 
 
+class JSONResponse(Response):
+    media_type = "application/json"
```
ChengjieLi28, line 72: Why this default value not set to the `__init__` method?

xinference/api/restful_api.py
codingl2k1: It is required because this is a `Respone` not a `JSONResponse`. The embedding responses are serialized in the ModelActor to offload the pressure of fast api.

xinference/api/restful_api.py
codingl2k1: This class is similar to the `JSONResponse` in `starlette/responses.py`, setting the `media_type` here is more clear.

ChengjieLi28: > This class is similar to the `JSONResponse` in `starlette/responses.py`, setting the `media_type` here is more clear.

Should we disable the `media_type` parameter in the `__init__` function? Because JsonResponse must be `application/json`.

codingl2k1: I changed the JSONResponse to inherit from starlette JSONResponse

ChengjieLi28 (approved):

Issue #653, opened 2023-11-16
BUG: Client http error is not raised


codingl2k1:
### Describe the bug

```python
Traceback (most recent call last):
  File "/home/codingl2k1/anaconda3/lib/python3.11/site-packages/requests/models.py", line 971, in json
    return complexjson.loads(self.text, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/codingl2k1/anaconda3/lib/python3.11/site-packages/simplejson/__init__.py", line 514, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/codingl2k1/anaconda3/lib/python3.11/site-packages/simplejson/decoder.py", line 386, in decode
    obj, end = self.raw_decode(s)
               ^^^^^^^^^^^^^^^^^^
  File "/home/codingl2k1/anaconda3/lib/python3.11/site-packages/simplejson/decoder.py", line 416, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/codingl2k1/test_client.py", line 3, in <module>
    model_uid = client.launch_model(
                ^^^^^^^^^^^^^^^^^^^^
  File "/home/codingl2k1/anaconda3/lib/python3.11/site-packages/xinference/client/restful/restful_client.py", line 538, in launch_model
    f"Failed to launch model, detail: {response.json()['detail']}"
                                       ^^^^^^^^^^^^^^^
  File "/home/codingl2k1/anaconda3/lib/python3.11/site-packages/requests/models.py", line 975, in json
    raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

The actual traceback should be
```python
Traceback (most recent call last):
  File "/home/codingl2k1/test_client.py", line 3, in <module>
    model_uid = client.launch_model(
                ^^^^^^^^^^^^^^^^^^^^
  File "/home/codingl2k1/anaconda3/lib/python3.11/site-packages/xinference/client/restful/restful_client.py", line 537, in launch_model
[...]

PR #657, opened 2023-11-16
BUG: Fix http error is not raised

codingl2k1:
Fixes: https://github.com/xorbitsai/inference/issues/653

### DIFF
--- a/xinference/client/restful/restful_client.py
+++ b/xinference/client/restful/restful_client.py
@@ -38,7 +38,7 @@ if TYPE_CHECKING:
 def _get_error_string(response: requests.Response) -> str:
     try:
         if response.content:
-            return _get_error_string(response)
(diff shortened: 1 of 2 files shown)


xinference/client/restful/restful_client.py
```diff
@@ -35,6 +35,19 @@
...
+    try:
+        if response.content:
+            return response.json()["detail"]
+    except Exception:
+        pass
+    try:
+        response.raise_for_status()
+    except requests.HTTPError as e:
+        return str(e)
+    return "Unknown error"
```
aresnow1, line 48: Could we print all stack if has unknown errors.

codingl2k1: The str(e) may contain the server-side stack.

aresnow1 (approved):

PR #667, opened 2023-11-21
FEAT: Support registering custom embedding model

ChengjieLi28:


aresnow1:
Conflicts should be resolved.

xinference/core/supervisor.py
```diff
@@ -243,25 +255,37 @@ def get_model_registration(
 
     @log_async(logger=logger)
     async def register_model(self, model_type: str, model: str, persist: bool):
-        if model_type == "LLM":
+        if model_type == "LLM" or model_type == "embedding":
```
aresnow1, line 258: Better use different code branches to handle different model types, otherwise it would be complex when types become more and more.

xinference/core/supervisor.py
```diff
@@ -243,25 +255,37 @@ def get_model_registration(
...
+                register_embedding if model_type == "embedding" else register_llm
+            )
+            register_fn(model_spec, persist)
         else:
             raise ValueError(f"Unsupported model type: {model_type}")
 
     @log_async(logger=logger)
     async def unregister_model(self, model_type: str, model_name: str):
-        if model_type == "LLM":
+        if model_type == "LLM" or model_type == "embedding":
```
aresnow1, line 281: Ditto.

xinference/core/worker.py
```diff
@@ -185,21 +185,33 @@ def _check_model_is_valid(self, model_name: str, model_format: Optional[str]):
     @log_sync(logger=logger)
     def register_model(self, model_type: str, model: str, persist: bool):
         # TODO: centralized model registrations
-        if model_type == "LLM":
+        if model_type == "LLM" or model_type == "embedding":
```
aresnow1, line 188: Same as the comments above, if you don't want to write too many conditions, use a dict to handle different types, like {"model_type": [spec_cls, register_func]}.

ChengjieLi28:
> Conflicts should be resolved.

Fix.

xinference/core/worker.py
ChengjieLi28: Fix

aresnow1:
Have you tested launching a custom embedding model and check if it works well?

ChengjieLi28:
> Have you tested launching a custom embedding model and check if it works well?

Yes. I tried to use `bge-base-zh` model dir as a custom embedding model. And it works.

aresnow1 (approved):

PR #675, opened 2023-11-22
FEAT: Supports `qwen.cpp` for `qwen-chat` with `ggml` format

ChengjieLi28:


xinference/api/restful_api.py
```diff
@@ -765,7 +769,7 @@ async def stream_results():
                 iterator = None
                 try:
                     try:
-                        if is_chatglm_ggml:
+                        if is_chatglm_ggml or is_qwen:
```
aresnow1, line 772: Have you checked if qwen.cpp supported system prompt?

xinference/model/llm/ggml/qwen.py
```diff
@@ -0,0 +1,291 @@
...
+
+    @classmethod
+    def match(
+        cls, llm_family: "LLMFamilyV1", llm_spec: "LLMSpecV1", quantization: str
+    ) -> bool:
+        if llm_spec.model_format != "ggmlv3":
+            return False
+        if "qwen" not in llm_family.model_name:
+            return False
+        if "chat" not in llm_family.model_ability:
```
aresnow1, line 107: As we implemented generate method, seems like generate model is also supported.

xinference/types.py
```diff
@@ -155,6 +155,17 @@ class ChatglmCppGenerateConfig(TypedDict, total=False):
     stream: bool
 
 
+class QWenCppModelConfig(TypedDict, total=False):
```
aresnow1, line 158: Does it needed?

aresnow1 (approved):

Issue #639, opened 2023-11-13
BUG: model is launched but can not find any items in UI


waltcow:
### Describe the bug
 model is launched via UI, but can not find any items in UI (Running model)

### To Reproduce

<img width="2180" alt="image" src="https://github.com/xorbitsai/inference/assets/3941174/a699d869-27fc-45e4-9fc0-e5a209a2b230">

ChengjieLi28:
Hi, @waltcow . I have noticed that you have closed the current xinference service via ctrl-C:
<img width="705" alt="image" src="https://github.com/xorbitsai/inference/assets/109656400/8dec6d7d-d25b-4b92-bdce-aa3a09096f4e">
Please check that whether you open the UI of a xinference service that was started elsewhere.

ChengjieLi28:
> Your quote block is just my tail log command.

In this case, please check the port of the web UI you open is the same as the cluster web port where you launch the model.

waltcow:
> > Your quote block is just my tail log command.
> 
> In this case, please check the port of the web UI you open is the same as the cluster web port where you launch the model.

I have double check the UI and model launch in the same host

UranusSeven:
Could you please verify if there are any error logs in the browser?

waltcow:
<img width="1688" alt="image" src="https://github.com/xorbitsai/inference/assets/3941174/6e5707a9-6336-4b29-a5dd-63bf5f36e62c">


it seems like related to domain security

ChengjieLi28:
> <img alt="image" width="1688" src="https://user-images.githubusercontent.com/3941174/282433309-6e5707a9-6336-4b29-a5dd-63bf5f36e62c.png">
> it seems like related to domain security

Currently we only support http services for xinference, the reason for this is that https servers make requests to http resources.

UranusSeven:
@waltcow Hello, are you accessing the endpoint using https? You might want to try the following workaround mentioned here: https://stackoverflow.com/a/76909337/7098025

PR #712, opened 2023-12-04
ENH: upgrade insecure requests when necessary

waltcow:
fix https://github.com/xorbitsai/inference/issues/639

### DIFF
--- a/xinference/web/ui/public/index.html
+++ b/xinference/web/ui/public/index.html
@@ -24,6 +24,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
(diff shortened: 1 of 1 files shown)


waltcow:
I try to use nginx forward port to local dev sever via https schema.
but when i open frontend ui, the console log show that 

image

ChengjieLi28:
> I try to use nginx forward port to local dev sever via https schema. but when i open frontend ui, the console log show that
> 
> image

Great fix. Thank you very much.

ChengjieLi28:
Hi, @waltcow .
We have reverted your PR #712 since if there is only http service, the UI would enconter an error.
Could you please test #722 in your environment to verify whether it is worked? Thanks.

Issue #723, opened 2023-12-05
ENH: Auto set metal env for mac


codingl2k1:
### Is your feature request related to a problem? Please describe

chatglm-cpp require environent var to enable metal support.

PR #724, opened 2023-12-05
ENH: Set metal environment var for mac

codingl2k1:
Close: https://github.com/xorbitsai/inference/issues/723


codingl2k1:
But, chatglm cpp needs to explict set the environment var `GGML_METAL=ON`

codingl2k1:
Set environment var does not affect pip install dependencies. It seems that there is no way to set an environment var for the dependency in pyproject.toml or setup.cfg.

PR #746, opened 2023-12-08
ENH: [UI] Supports `model_uid` input when launching models

ChengjieLi28:
image


qinxuye:
Nice work, my suggestion is putting the text input of model uid at the bottom.

qinxuye (approved):
LGTM

PR #747, opened 2023-12-08
Bugs: Fixing issue with emote encoding in streaming chat, Fixing issue with missing pad_token for pytorch tokenizers, allowing system message as latest message in chat

AndiMajore:
Collection of three small fixes for issues:
- Streaming mode for chats broke emote encoding; fixed by trimming incomplete/non-decodable characters from chunk
- Running llama models would throw missing pad_token errors: fixed by setting to eos_token
- Ran into issue when initializing behavior by sending system message only: allowing system or user messages to be latest message


aresnow1:
Could you provide a minimal reproduction for the first issue?

aresnow1 (approved):

PR #772, opened 2023-12-15
BUG: UI Crash on Search when `model_format` and `model_size` have been selected

### LINKED ISSUE #765: BUG: [UI] when `model_format` and `model_size` have been selected, UI may crash when we search model then.
Note that the issue tracker is NOT the place for general support.
Reproduce:
1. First of all:
select `model_format` and `model_size` in the **first** model card:
image

2. And then input `q` in the text search box (assume I want to search `qwen-chat`)
image

3. The ui crashed.

Bojun-Feng:
Fix #765 

The bug is caused by how the `cache_status` property was accessed in our code. This fix should help us avoid the unexpected crashes. Here is a quick break down of the change.

**The Issue:**
We were directly accessing `cache_status` on the `specs[0]` object. If `specs[0]` or `cache_status` does not exist (in this case the card is filtered out by the search), the code would throw an error.

**The Fix:**
I've introduced optional chaining (`?.`) and the nullish coalescing operator (`??`)to address the nonexistent property issue:

Before:
```javascript
const cached = modelFormat === 'pytorch'
    ? specs[0].cache_status &&
      specs[0].cache_status === true
    : specs[0].cache_status &&
      specs[0].cache_status[index] === true
```
After:
```javascript
const cached = modelFormat === 'pytorch'
    ? specs[0]?.cache_status ?? false === true
    : specs[0]?.cache_status?.[index] ?? false === true;
```

**Brief Explaination**
- **Optional Chaining (`?.`):** This is like asking, "Hey, do you have this property? If not, it's cool, just be `undefined`." It stops JavaScript from freaking out and throwing errors when it doesn't find something.
- **Nullish Coalescing Operator (`??`):** This one's like a safety net. It catches `undefined` or `null` and replaces them with a default value, which in our case is `false`.

I have tested the updates locally and the UI is not crashing anymore. Please let me know if there are any additional errors that are not fixed.

### DIFF
--- a/xinference/web/ui/src/scenes/launch_model/modelCard.js
+++ b/xinference/web/ui/src/scenes/launch_model/modelCard.js
@@ -496,10 +496,8 @@ const ModelCard = ({ url, modelData, gpuAvailable, is_custom = false }) => {
 
                     const cached =
                       modelFormat === 'pytorch'
-                        ? specs[0].cache_status &&
-                          specs[0].cache_status === true
-                        : specs[0].cache_status &&
-                          specs[0].cache_status[index] === true
+                        ? specs[0]?.cache_status ?? false === true
+                        : specs[0]?.cache_status?.[index] ?? false === true
                     const displayedQuant = cached ? quant + ' (cached)' : quant
 
                     return (


Bojun-Feng:
Can we have this merged please?

aresnow1:
> Can we have this merged please?

Sure, will review soon.

ChengjieLi28 (approved):

PR #803, opened 2023-12-22
fix: error when check model_uid & model_name in restful_api

liunux4odoo:


qinxuye (approved):
LGTM, Thank you for your contribution.

PR #856, opened 2024-01-03
BUG: Fix typing_extension version problem in notebook

onesuper:
Only change one line


qinxuye:
I think we can limit typing_extension version in setup.cfg at https://github.com/xorbitsai/inference/blob/46e0de16fe3aabdee455cadfb305851a48f91a95/setup.cfg#L38

qinxuye (approved):
LGTM

PR #858, opened 2024-01-04
BUG: Fix generate of chatglm3

aresnow1:


xinference/core/model.py
```diff
@@ -375,7 +375,6 @@ def _wrapper():
 
         return await self._call_wrapper(_wrapper)
 
-    @log_async(logger=logger)
     async def next(
```
codingl2k1, line 378: Why removing this `log_async`? I will remove this `next` in https://github.com/xorbitsai/inference/pull/859. We can keep this decorator in this PR to avoid conflicts.

xinference/model/llm/pytorch/core.py
```diff
@@ -409,7 +409,7 @@ def _sanitize_generate_config(
     ) -> PytorchGenerateConfig:
         generate_config = super()._sanitize_generate_config(generate_config)
         if (
-            generate_config.get("stop", None) is None
+            (not generate_config.get("stop", None))
```
codingl2k1, line 412: If `generate_config` is a dict, then `.get(xxx)` is equal to `.get(xxx, None)`.

codingl2k1 (approved):

PR #881, opened 2024-01-11
ENH: make the size param of *_to_image more compatible

liunux4odoo:
the `size` param use '*' as seperator only. it should be better to support other characters such as 'x' or any other non-integers.


codingl2k1 (approved):
LGTM

Issue #897, opened 2024-01-15
Qwen-72B-Chat-Int4 custom register can't use tool-call


Zhangtiande:
```python
2024-01-15 15:44:36,375 xinference.api.restful_api 1583 ERROR    [address=0.0.0.0:38069, pid=4867] Model qwen-chat-test is not support tool calls.
Traceback (most recent call last):
  File "/root/miniconda3/envs/llm/lib/python3.10/site-packages/xinference/api/restful_api.py", line 859, in create_chat_completion
    data = await model.chat(prompt, system_prompt, chat_history, kwargs)
  File "/root/miniconda3/envs/llm/lib/python3.10/site-packages/xoscar/backends/context.py", line 227, in send
    return self._process_result_message(result)
  File "/root/miniconda3/envs/llm/lib/python3.10/site-packages/xoscar/backends/context.py", line 102, in _process_result_message
    raise message.as_instanceof_cause()
  File "/root/miniconda3/envs/llm/lib/python3.10/site-packages/xoscar/backends/pool.py", line 657, in send
    result = await self._run_coro(message.message_id, coro)
  File "/root/miniconda3/envs/llm/lib/python3.10/site-packages/xoscar/backends/pool.py", line 368, in _run_coro
    return await coro
  File "/root/miniconda3/envs/llm/lib/python3.10/site-packages/xoscar/api.py", line 306, in __on_receive__
    return await super().__on_receive__(message)  # type: ignore
  File "xoscar/core.pyx", line 558, in __on_receive__
    raise ex
  File "xoscar/core.pyx", line 520, in xoscar.core._BaseActor.__on_receive__
    async with self._lock:
  File "xoscar/core.pyx", line 521, in xoscar.core._BaseActor.__on_receive__
    with debug_async_timeout('actor_lock_timeout',
  File "xoscar/core.pyx", line 526, in xoscar.core._BaseActor.__on_receive__
    result = await result
  File "/root/miniconda3/envs/llm/lib/python3.10/site-packages/xinference/core/utils.py", line 35, in wrapped
    ret = await func(*args, **kwargs)
  File "/root/miniconda3/envs/llm/lib/python3.10/site-packages/xinference/core/model.py", line 79, in wrapped_func
    ret = await fn(self, *args, **kwargs)
  File "/root/miniconda3/envs/llm/lib/python3.10/site-packages/xinference/core/model.py", line 274, in chat
    return await self._call_async_wrapper(_async_wrapper)
[...]

PR #899, opened 2024-01-15
BUG: Fix tool calls on custom model

codingl2k1:
Fixes: https://github.com/xorbitsai/inference/issues/897


aresnow1:
In https://github.com/xorbitsai/inference/pull/861, we add `model_family` field when register models, we can use this field to check model type.

aresnow1 (approved):
LGTM

Issue #1021, opened 2024-02-21
 Failed to download model 'qwen1.5-chat' (size: 7, format: ggufv2) after multiple retries


isanwenyu:
### Describe the bug
2024-02-21 14:09:43,564 xinference.api.restful_api 57495 ERROR    [address=127.0.0.1:25907, pid=57568] Failed to download model 'qwen1.5-chat' (size: 7, format: ggufv2) after multiple retries
modelscope.hub.errors.NotExistError: The file path: qwen1.5-7b-chat-q5_k_m.gguf not exist in: qwen/Qwen1.5-7B-Chat-GGUF

PR #1027, opened 2024-02-22
BUG: Fix missing qwen 1.5 7b gguf

codingl2k1:
Fixes: https://github.com/xorbitsai/inference/issues/1021


aresnow1:
Other size models support more quantizations too, we can add supports in this PR, including huggingface.

aresnow1 (approved):

PR #1061, opened 2024-02-29
BUG: Fix early truncation due to `max_token` being default to `16` instead of `1024`

ZhangTianrong:
The `CreateCompletionLlamaCpp` model inherits from the signature of  [`llama_cpp.llama.create_completion`](https://github.com/abetlen/llama-cpp-python/blob/8c71725d5345b2ba325e6d28209875057c5873e6/llama_cpp/llama.py#L1384) which sets `max_tokens` to `16` by default. This invalidates the default value set in [`restful_api.py`](https://github.com/xorbitsai/inference/blob/d53c80a6a82ded81231d22984ff4d02eb3935ae5/xinference/api/restful_api.py#L848) which expects `max_tokens` to be `None` if not set in the request.

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -1137,7 +1137,7 @@ class RESTfulAPI:
         }
         kwargs = body.dict(exclude_unset=True, exclude=exclude)
 
-        # TODO: Decide if this default value overide is necessary #1061
+        # TODO: Decide if this default value override is necessary #1061
         if body.max_tokens is None:
             kwargs["max_tokens"] = max_tokens_field.default
 
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -63,6 +63,7 @@ from ..types import (
     CreateChatCompletion,
     CreateCompletion,
     ImageList,
+    max_tokens_field,
 )
 from .oauth2.auth_service import AuthService
 from .oauth2.types import LoginUserForm
@@ -844,6 +845,10 @@ class RESTfulAPI:
         }
         kwargs = body.dict(exclude_unset=True, exclude=exclude)
 
+        # TODO: Decide if this default value overide is necessary #1061
+        if body.max_tokens is None:
+            kwargs["max_tokens"] = max_tokens_field.default
+
         if body.logit_bias is not None:
             raise HTTPException(status_code=501, detail="Not implemented")
 
@@ -1131,6 +1136,10 @@ class RESTfulAPI:
             "user",
         }
         kwargs = body.dict(exclude_unset=True, exclude=exclude)
+
+        # TODO: Decide if this default value overide is necessary #1061
+        if body.max_tokens is None:
+            kwargs["max_tokens"] = max_tokens_field.default
 
         if body.logit_bias is not None:
             raise HTTPException(status_code=501, detail="Not implemented")
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -63,7 +63,6 @@ from ..types import (
     CreateChatCompletion,
     CreateCompletion,
     ImageList,
-    max_tokens_field,
 )
 from .oauth2.auth_service import AuthService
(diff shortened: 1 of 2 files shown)


aresnow1:
There are some problems with CI, and not related to your changes, we will solve it soon.

aresnow1:
CI should be fixed, could you rebase main branch and push again?

ZhangTianrong:
[`test_type.py`](https://github.com/xorbitsai/inference/blob/60c0c50c7f1330bf80bbf3fe9ee25b2250fa7901/xinference/core/tests/test_types.py#L108) seems to require `CreateCompletionTorch`, `CreateCompletionLlamaCpp` and `CreateCompletionCTransformers` to have the same default value of `1024` for `max_tokens`. Both the implementations before and after this PR are failing this test since `CreateCompletionLlamaCpp.__fields__["max_tokens"].default` was `16` and now `None`. To pass this test, all I need to do is to change the default value from `None` to `1024`. That means the part to handle unset `max_tokens` in `request.json` in [`restful_api.py`](https://github.com/xorbitsai/inference/blob/60c0c50c7f1330bf80bbf3fe9ee25b2250fa7901/xinference/api/restful_api.py#L848) is redundant and should be deleted. Is that the expected behavior?

ZhangTianrong:
`CreateCompletionOpenAI` is not covered in `test_types.py` and it has `None` as its default `max_toknes` according to [`openai.types.chat.completion_create_params.CompletionCreateParamsBase`](https://github.com/openai/openai-python/blob/main/src/openai/types/chat/completion_create_params.py#L109) (in fact, there is no default value for any filed and they all fallback to `None` in [`types.py`](https://github.com/xorbitsai/inference/blob/60c0c50c7f1330bf80bbf3fe9ee25b2250fa7901/xinference/types.py#L402)). Should we also set it to `1024`? `null`, which `None` translates to in `json` is also an accept value for the official OpenAI API except that it doesn't take any effect.

aresnow1:
> `CreateCompletionOpenAI` is not covered in `test_types.py` and it has `None` as its default `max_toknes` according to [`openai.types.chat.completion_create_params.CompletionCreateParamsBase`](https://github.com/openai/openai-python/blob/e41abf7b7dbc1e744d167f748e55d4dedfc0dca7/src/openai/types/chat/completion_create_params.py#L109) (in fact, there is no default value for any field and they all fallback to `None` in [`types.py`](https://github.com/xorbitsai/inference/blob/60c0c50c7f1330bf80bbf3fe9ee25b2250fa7901/xinference/types.py#L402)). Should we also set it to `1024`? `null`, which `None` translates to in `json` is also an accept value for the official OpenAI API except that it doesn't take any effect.

https://github.com/xorbitsai/inference/blob/d53c80a6a82ded81231d22984ff4d02eb3935ae5/xinference/api/restful_api.py#L848-L849

The code is for handling the case where the `max_tokens` parameter is set to None when using the OpenAI SDK, I think we can temporarily keep this logic as we haven't determined the best solution yet.

xinference/api/restful_api.py
```diff
@@ -845,6 +845,7 @@ async def create_completion(self, request: Request) -> Response:
         }
         kwargs = body.dict(exclude_unset=True, exclude=exclude)
 
+        # TODO: Decide if this default value overide is necessary #1061
```
qinxuye, line 848:
```suggestion
        # TODO: Decide if this default value override is necessary #1061
```

xinference/api/restful_api.py
```diff
@@ -1136,6 +1137,7 @@ async def create_chat_completion(self, request: Request) -> Response:
...
```
qinxuye, line 1140:
```suggestion
        # TODO: Decide if this default value override is necessary #1061
```

qinxuye (commented):
Spell checks failed for CI.

ChengjieLi28 (approved):

qinxuye:
Thanks for your contribution.

PR #1062, opened 2024-02-29
ENH: Add "[DONE]" to the end of stream generation for better openai SDK compatibility

### LINKED ISSUE #998: BUG 通过api调用，一句话完了好像没有返回[DONE]
### Describe the bug
通过api调用，一句话完了好像没有返回[DONE]

### To Reproduce
0.85 docker镜像跑的 千问 1.5的14B INT4量化版本

ZhangTianrong:
The official API from OpenAI always send a "[DONE]" before it terminates the stream. Although redundant, some clients like [ChatGPT.好友](https://help.chats.li/) rely on it instead of a non-empty `finish_reason` to determine the the end of the stream. So, it might be beneficial to abide by this convension. 

fixes #998

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -1256,6 +1256,7 @@ class RESTfulAPI:
                         self.handle_request_limit_error(re)
                     async for item in iterator:
                         yield item
+                    yield "[DONE]"
                 except Exception as ex:
                     logger.exception("Chat completion stream got an error: %s", ex)
                     await self._report_error_event(model_uid, str(ex))
--- a/xinference/client/common.py
+++ b/xinference/client/common.py
@@ -43,6 +43,8 @@ def streaming_response_iterator(
         line = line.strip()
         if line.startswith(b"data:"):
             json_str = line[len(b"data:") :].strip()
+            if json_str == b"[DONE]":
+                continue
             data = json.loads(json_str.decode("utf-8"))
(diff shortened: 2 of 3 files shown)


ZhangTianrong:
[test_client.py](https://github.com/xorbitsai/inference/actions/runs/8095006267/job/22128715572#step:5:539) requires all chunks in the stream generation to be a valid `json` object, but "[DONE]" is supposed to be a special case. It's worth modifying the test to handle the last chunk specifically.

qinxuye:
> [test_client.py](https://github.com/xorbitsai/inference/actions/runs/8095006267/job/22128715572#step:5:539) requires all chunks in the stream generation to be a valid `json` object, but "[DONE]" is supposed to be a special case. It's worth modifying the test to handle the last chunk specifically.

I think you can strip the `[DONE]` and make the tests passed.

ZhangTianrong:
It seems that the [assert](https://github.com/xorbitsai/inference/blob/18e183fd2a5a4fd7ecf8a6287d875ae77d4ecb9f/xinference/client/tests/test_client.py#L277) in `test_client.py` is not working. In `assert "content" or "role" in chunk["choices"][0]["delta"]`, since comparation operators like `in` takes precedence over logical operators like `or` and `"content"` is a non empty, the statement always evaluates to `True`.

ChengjieLi28 (approved):

PR #1073, opened 2024-03-04
ENH: Add a dropdown to the web UI to support adjusting GPU offload layers for llama.cpp loader

notsyncing:
This PR adds an "N GPU Layers" input to model card if you selected `ggufv2` format, to adjust the number of layers offloaded to GPU when using llama.cpp loader. It will replace the "N-GPU" dropdown (only on `ggufv2` selection).

Special values: `-1`means auto (the old `SIZE_TO_GPU_LAYERS` method), `0` means CPU only.

btw: I'm not sure if I should replace the N-GPU dropdown (Looks like the "N-GPU" selection won't have any effect for llama.cpp), or add a new row to the model card. Maybe we could use a wizard-like page for more configurable options in the future?

### DIFF
--- a/xinference/web/ui/src/scenes/launch_model/modelCard.js
+++ b/xinference/web/ui/src/scenes/launch_model/modelCard.js
@@ -125,8 +125,11 @@ const ModelCard = ({ url, modelData, gpuAvailable, is_custom = false }) => {
       quantization: quantization,
       n_gpu:
         nGPU === '0' ? null : nGPU === 'auto' ? 'auto' : parseInt(nGPU, 10),
-      n_gpu_layers: nGPULayers,
       replica: replica,
+    }
+
+    if (nGPULayers >= 0) {
+      modelDataWithID.n_gpu_layers = nGPULayers
     }
 
     // First fetcher request to initiate the model
@@ -495,7 +498,7 @@ const ModelCard = ({ url, modelData, gpuAvailable, is_custom = false }) => {
               </FormControl>
             </Grid>
             <Grid item xs={6}>
-              {modelFormat !== 'ggufv2' ? (
+              {modelFormat !== 'ggufv2' && modelFormat !== 'ggmlv3' ? (
                 <FormControl
                   variant="outlined"
                   margin="normal"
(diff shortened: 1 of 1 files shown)


notsyncing:
Just noticed #1070 :joy: I can rebase this PR on that if you want.

qinxuye:
> Just noticed #1070 😂 I can rebase this PR on that if you want.

You came cross to have the same idea haha, you can rebase this after that merged.

ChengjieLi28:
Hi, @notsyncing . Note that #1070 has been merged. Should just keep the frontend code in this PR.
Could you please  provide creenshots showing the effect of adding this option? We recently want to refactor the front-end, as there was hardly any place left to add new parameters where the LAUNCH model was.

notsyncing:
Screenshots:

If model format is not `ggufv2`, shows the following (original):
Screenshot_20240306_145001

If model format is `ggufv2`, change the `N-GPU` dropdown to `N GPU Layers` input:
Screenshot_20240306_145016

xinference/web/ui/src/scenes/launch_model/modelCard.js
```diff
@@ -493,39 +495,58 @@ const ModelCard = ({ url, modelData, gpuAvailable, is_custom = false }) => {
...
-                fullWidth
-                disabled={!modelFormat || !modelSize || !quantization}
-              >
-                <InputLabel id="n-gpu-label">N-GPU</InputLabel>
-                <Select
-                  labelId="n-gpu-label"
-                  value={nGPU}
-                  onChange={(e) => setNGPU(e.target.value)}
-                  label="N-GPU"
+              {modelFormat !== 'ggufv2' ? (
```
ChengjieLi28, line 498: Also include `ggmlv3`

xinference/web/ui/src/scenes/launch_model/modelCard.js
```diff
@@ -124,6 +125,7 @@ const ModelCard = ({ url, modelData, gpuAvailable, is_custom = false }) => {
       quantization: quantization,
       n_gpu:
         nGPU === '0' ? null : nGPU === 'auto' ? 'auto' : parseInt(nGPU, 10),
+      n_gpu_layers: nGPULayers,
```
ChengjieLi28, line 128: If the user does not choose model_format `ggufv2` / `ggmlv3`, better to not set this parameter. Because these extra parameters would be passed to the backend engine, which may lead to errors.

xinference/web/ui/src/scenes/launch_model/modelCard.js
notsyncing: fixed

xinference/web/ui/src/scenes/launch_model/modelCard.js
notsyncing: fixed

ChengjieLi28 (approved):

PR #1098, opened 2024-03-07
BUG: fix  _eval_qwen_chat_arguments parsing problem

channingxiao18:
when using qwen and qwen1.5 with tools calls.  There are are two problems:

1.  when the output has a Action and Action Input, _eval_qwen_chat_arguments will ignore "Thought",  but infact the thought has meanningful logic, it's better to output the thought to content.  After the getting the tool call output, when filling back to LLM, it fully matched how the original template designed:
Use the following format:

Question: the input question you must answer
Thought: you should always think about what to do
Action: the action to take, should be one of [{tool_names}]
Action Input: the input to the action
Observation: the result of the action
... (this Thought/Action/Action Input/Observation can be repeated zero or more times)
Thought: I now know the final answer
Final Answer: the final answer to the original input question

https://github.com/QwenLM/Qwen/blob/main/examples/react_prompt.md

2. The second problem is when there are some error to parse or there are no tool calls.  
   it reaches "        return text, None, None  ". it will output the full message directly, even with  Thoughts: now I know the final answer\nFinal Answer: xxxx.  
   
   here are some test case:
   output_with_tool = """\nThought: 我需要调用current_time API来获取当前时间\nAction: current_time\nAction Input: {}\nObservation: {"current_time": "2023-04-07 15:30:00"}\nThought: 我现在知道当前时间是2023年04月07日15点30分\nFinal Answer: 现在是2023年04月07日15点30分。"""
output_with_tool2 = """Thought: 我需要调用current_time API来获取当前时间\nAction: current_time\nAction Input: {}\nObservation:"""
output_with_tool3 = """Thought: 我需要调用current_time API来获取当前时间\nAction: current_time\nAction Input: {}"""

output_without_tool = """Thought: 我现在知道当前时间是2023年04月07日15点30分\nFinal Answer: 现在是2023年04月07日15点30分。"""
output_without_tool2 = """Final Answer: 现在是2023年04月07日15点30分。"""
output_without_tool3 = """现在是2023年04月07日15点30分。"""


qinxuye:
Lint failed, could you fix it?

qinxuye:
You can refer to https://inference.readthedocs.io/zh-cn/latest/development/contributing_codebase.html#autofixing-formatting-errors to fix formatting issue, it would be easier.

qinxuye:
Can you rebase the main branch to fix the CI?

xs818818:
同问这个问题，ReAct使用确实蛮需要的dify中貌似默认xinference使用工具时用ReAct模式,每次都失败

codingl2k1:
> 同问这个问题，ReAct使用确实蛮需要的dify中貌似默认xinference使用工具时用ReAct模式,每次都失败

这个bugfix能修掉你说的这个问题吗？

xinference/model/llm/utils.py
```diff
@@ -535,26 +535,39 @@ def _eval_qwen_chat_arguments(c, tools):
...
                 func_args = text[j + len("\nAction Input:") : k].strip()
+                content = text[
+                    t + len("\nThought:") : i
+                ].strip()  # len("\nThought:") and len("Thought:") both are OK since there is a space after :
             if func_name:
-                return None, func_name, json.loads(func_args)
-            z = text.rfind("\nFinal Answer: ")
-            if z >= 0:
-                text = text[z + len("\nFinal Answer: ") :]
+                return content, func_name, json.loads(func_args)
```
codingl2k1, line 558: Actually, as shown by the OpenAI API, the content should be null if the tool call is successful: https://platform.openai.com/docs/api-reference/chat/create
```
{
  "id": "chatcmpl-abc123",
  "object": "chat.completion",
  "created": 1699896916,
  "model": "gpt-3.5-turbo-0125",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": null,
        "tool_calls": [
          {
            "id": "call_abc123",
            "type": "function",
            "function": {
              "name": "get_current_weather",
              "arguments": "{\n\"location\": \"Boston, MA\"\n}"
            }
          }
        ]
      },
      "logprobs": null,
      "finish_reason": "tool_calls"
    }
  ],
  "usage": {
    "prompt_tokens": 82,
    "completion_tokens": 17,
    "total_tokens": 99
  }
}
```
@qinxuye I think we can merge this if it fixes some corner cases.

qinxuye: OK, @codingl2k1 can you help to fix the corner case?

codingl2k1: I mean some corner use cases require the thought content, but OpenAI's API just returns null. If we strictly follow OpenAI's API, then we should return null if the tool call is successful.

codingl2k1: So, I think we can merge this PR even if it mismatches the OpenAI's API.

codingl2k1 (approved):
LGTM

xinference/model/llm/utils.py
qinxuye: OK, I will merge it

PR #1102, opened 2024-03-07
FEAT: launch UI of audio model.

hainaweiben:
In fact, I found that most audio models are not available on ModelScope.


xinference/model/audio/model_spec_modelscope.json
```diff
@@ -0,0 +1,14 @@
+[
+    {
+        "model_name": "whisper-small.en",
+        "model_family": "whisper",
+        "model_id": "iic/speech_whisper-small_asr_english",
```
qinxuye, line 5: Is this model identical to whisper-small.en? It looks not.

hainaweiben: Actually, I'm not very sure about this whisper-small.en from modelscope as shown in the picture.
image
If not, I will delete the corresponding JSON content.

hainaweiben: https://modelscope.cn/models/iic/speech_whisper-small_asr_english/summary

qinxuye: You can see the description, it's different I think.

hainaweiben: I agree with your opinion and will delete the JSON file of this model.

xinference/model/audio/model_spec_modelscope.json
qinxuye, line 1: Is only this file enough? Can the audio model choose to download from modelscope now?

qinxuye:
@hainaweiben I think you can still add the support for audio models on the web UI, without considering the modelscope part.

ChengjieLi28 (approved):

PR #1127, opened 2024-03-11
ENH: Make GET /v1/models compatible with OpenAI API.

notsyncing:
This PR add some fields according to https://platform.openai.com/docs/api-reference/models/list while retaining its original fields. Now I'm able to use xinference with https://github.com/open-webui/open-webui

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -584,8 +584,22 @@ class RESTfulAPI:
 
     async def list_models(self) -> JSONResponse:
         try:
-            data = await (await self._get_supervisor_ref()).list_models()
-            return JSONResponse(content=data)
+            models = await (await self._get_supervisor_ref()).list_models()
+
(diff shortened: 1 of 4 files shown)


notsyncing:
Looks like that we must break some use cases of this API to be compatible with OpenAI API. Any ideas? Merge this PR and break (or completely move to OpenAI API), or close this PR and keep its original state?

qinxuye:
I think we should be compatible with OpenAI, if the behavior of CI is wrong, you can modify it.

notsyncing:
I decided to move it completely to OpenAI API format, because even we add the openai fields `data` and `object` to the response, it can interfere the users who are using `len(response)`. So, I think it's better to break it completely, rather than a strange response which duplicates the data twice.

Tested with `curl`, `xinference list` and `open-webui`.

ChengjieLi28 (approved):

Issue #1154, opened 2024-03-16
ENH: vLLM latest models support 


luweizheng:
vLLM supports more [models](https://github.com/vllm-project/vllm/releases) and we defined them in `xinference/model/llm/vllm/core.py` should we update the `VLLM_SUPPORTED_MODELS` and `VLLM_SUPPORTED_CHAT_MODELS`
in the following ways:

```
if VLLM_INSTALLED and vllm.__version__ >= "0.3.2":
    VLLM_SUPPORTED_CHAT_MODELS.append("gemma-it")
```

qinxuye:
Yes. In the 0.9.3 we just add support for vllm>0.3.3. Some models could be absent.

PR #1155, opened 2024-03-17
ENH: vLLM latest models support

1572161937:
add vLLM latest models ``gemma-it``, ``orion-chat``, ``orion-chat-rag``.
FIX https://github.com/xorbitsai/inference/issues/1154


luweizheng (approved):
Tested with `gemma-it` on my environment.

PR #1190, opened 2024-03-26
BLD: add autoawq in setup

utopia2077:
Support autoawq.
Otherwise, when inferring the AWQ quantization model, it will error that dependencies are missing.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -95,6 +95,7 @@ all =
     controlnet_aux
     orjson
     auto-gptq ; sys_platform!='darwin'
+    autoawq
     optimum
     sglang[all] ; sys_platform=='linux'
 intel =
@@ -115,6 +116,7 @@ transformers =
     einops
     tiktoken
     auto-gptq
+    autoawq
     optimum
     peft
 vllm =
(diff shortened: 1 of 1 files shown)


qinxuye:
I think you can add the autoawq to ALL and transformers.

see 

https://github.com/xorbitsai/inference/blob/2f1ee95bd9781d95ab97dc0e81d37fef0c1df3ae/setup.cfg#L97

and 

https://github.com/xorbitsai/inference/blob/2f1ee95bd9781d95ab97dc0e81d37fef0c1df3ae/setup.cfg#L117

Then when pip install "xinference[all]" it will be installed, so you don't need to modify the dockerfile.

utopia2077:
Modified as above.
Installing dependencies inside a Docker container is very inconvenient, so it is recommended to add it.

setup.cfg
```diff
@@ -95,6 +95,7 @@ all =
     controlnet_aux
     orjson
     auto-gptq ; sys_platform!='darwin'
+    autoawq
```
qinxuye, line 98:
```suggestion
    autoawq ; sys_platform!='darwin'
```

qinxuye (approved):
LGTM

PR #1195, opened 2024-03-27
FEAT: force to specify worker ip and gpu idx when launching models

ChengjieLi28:
- add `--worker-ip` and `--gpu-idx` in launch command line
- Change the GPU allocation algorithm to maintain the original policy when gpu_idx is not specified, and when gpu_idx is specified, it can be launched as long as the existing model on the GPU is not a vllm model, and there will be a corresponding log.


qinxuye (approved):
LGTM

PR #1200, opened 2024-03-27
Added vllm support for deepseek models

ivanzfb:
Added vllm support for deepseek models，including deepseek-chat, deepseek-coder-instruct


qinxuye (approved):
LGTM

Issue #1208, opened 2024-03-28
[Feature request] Support stream=True in tool calls when using vllm


zhanghx0905:
### Is your feature request related to a problem? Please describe

Refer to our previous discussion in the FastGPT issue area.

https://github.com/labring/FastGPT/issues/1046

With the release of Qwen1.5 last month, open-source LLMs now have significant tool calling capabilities. 
Currently, xinference does not support setting stream=True in tool calls.

### Describe the solution you'd like

Based on my research so far, https://github.com/xusenlinzy/api-for-open-llm has implemented a design that appears to be quite scalable, see:

https://github.com/xusenlinzy/api-for-open-llm/blob/master/api/vllm_routes/chat.py

To implement streaming processing for tool calls, xinference should consider rewriting the streaming response function: https://github.com/xorbitsai/inference/blob/5a83dd5774942e0409872f362e9d8eb52a95c6c3/xinference/model/llm/vllm/core.py#L319

qinxuye:
Do you have interest to implement this?

zhanghx0905:
> Do you have interest to implement this?

Of course, I will try to implement the feature as I need it. But I'm not sure when I'll be done, and if I am, I'll submit a pr for you guys!

qinxuye:
Thank you in advance.

BTW, you can comment with a single word `take` to assign this thread to yourself to tell others you are working on it.

yanmao2023:
这个功能是急需的，也是目前xinference的一个受限

zhanghx0905:
take

leoterry-ulrica:
> take

测试验证通过了么？

zhanghx0905:
> 测试验证通过了么？

仍然 WIP。如果你想试试，可以拉取 https://github.com/zhanghx0905/inference/tree/dev-streaming-tools，并执行 `pip install -e .`

lm970585581:
使用python3.11启动会报错
pkg_resources.UnknownExtra: typer 0.12.0 has no such extra feature 'all'

lm970585581:
> > 测试验证通过了么？
> 
> 仍然 WIP。如果你想试试，可以拉取 https://github.com/zhanghx0905/inference/tree/dev-streaming-tools 并执行 `pip install -e .`

用你的分支使用python3.10和3.11启动都会报错
pkg_resources.UnknownExtra: typer 0.12.0 has no such extra feature 'all'
这是什么原因呢？

[...]

PR #1215, opened 2024-03-29
FEAT: Enable streaming in tool calls for Qwen when using vllm

zhanghx0905:
#1208 
- This is a non-intrusive commit, and the control flow of the original functionality has not been modified in any way
- **Only vllm backend is supported**. Other backends (like pytorch, llama-cpp, sglang) are not supported.


xinference/api/restful_api.py
```diff
@@ -1344,9 +1344,9 @@ async def create_chat_completion(self, request: Request) -> Response:
                     status_code=400,
                     detail=f"Only {function_call_models} support tool messages",
                 )
-        if body.tools and body.stream:
+        if body.tools and body.stream and not model.is_vllm_backend():
             raise HTTPException(
-                status_code=400, detail="Tool calls does not support stream"
+                status_code=400, detail="Tool calls only support streaming when using VLLM"
```
ChengjieLi28, line 1349: Why tool call `only` supports streaming in this PR?

zhanghx0905: There seems to be an ambiguity. What about "Streaming support for tool calls is available only when using vLLM."

weipienlee:
Hi @zhanghx0905, i checked out you pull request since I'm eager to get it to work with langchain/opengpts. Are your changes "part of" the **OpenAI API compatibility** or only for the Xinference llm langchain adapter? I indeed do not get a "streaming not supported" message anymore for both qwen-chat and qwen1.5-chat, however, get the following seemly truncated llm output (might be a streaming end mistake):
image

And i used "standard" langchain tool binding.
image

weipienlee:
update: action string is complete with Tavily Search tool
```
Thought: I need to use the tavily_search_results_json API to get information about the weather in Hamburg.
Action: tavily_search_results_json
Action Input: {"query": "weather in hamburg"}
```
with params:
```
    _type: "openai-chat"

    model: "qwen-chat"

    model_name: "qwen-chat"

    n: 1

    stop: null

    stream: true

    temperature: 0

    tools

        0: {
          "type": "function",
          "function": {
            "name": "tavily_search_results_json",
            "description": "A search engine optimized for comprehensive, accurate, and trusted results. Useful for when you need to answer questions about current events. Input should be a search query.",
            "parameters": {
              "type": "object",
              "properties": {
                "query": {
                  "description": "search query to look up",
                  "type": "string"
                }
              },
              "required": [
                "query"
              ]
            }
          }
        }
 ```

**Comment by zhanghx0905** [comment]
> update: action string is complete with Tavily Search tool
> 
> ```

I haven't used this part of langchain's functionality, maybe you can provide me with a more detailed reproduction of the steps considering my unfamiliarity.

I am working on this PR mainly for fastgpt tools compatibility as I need it. OpenAI REST API compatibility is of course an important goal.
This PR is still under development, it lacks testing and some details still need to be discussed, so feel free to leave a comment!

**Comment by weipienlee** [comment]
You're right, that part of the [React agent design pattern](https://python.langchain.com/docs/modules/agents/agent_types/react). [Here](https://github.com/langchain-ai/langchain/blob/08c10bd66a93ca16b930ca2d7d7775c053241bbf/libs/langchain/langchain/agents/react/agent.py#L16) you have you that that is actually done by "prompt engineering" (tools description in context) instead of using the [function calling api](https://platform.openai.com/docs/guides/function-calling) of the llm. It's seems to be come a standard. Here you see how it is used in [Langchain](https://python.langchain.com/docs/modules/model_io/chat/function_calling) which makes it easier to create all sorts of other agent design pattern besides React. Just fyi

**Comment by zhanghx0905** [comment]
> You're right, that's part of the [React agent design pattern]


Based on my understanding, there is no conflict between the React-based Tool Call pattern and the OpenAI API. The keypoint lies in whether the model correctly initiates a tool call following "Observation".

I am also considering filtering out all tokens before outputting the "Final Answer:" for the Qwen model in the token stream, which should offer a better interaction experience for users.

Regardless, this PR targets only the Qwen and ChatGLM3 models, while the former significantly outperforming the latter.

**Comment by weipienlee** [comment]
> Based on my understanding, there is no conflict between the React-based Tool Call pattern and the OpenAI API. The keypoint lies in whether the model correctly initiates a tool call following "Observation".

The api documentation of [OpenAI's tool-choice](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice) (click in the **functions** in the right pane) shows that function specifications has become part of the API parameters instead of part of the messages parameter ("system of prompt"). Both request and response are now well defined structures (usually json). This strategy has been adapted by others like e.g. [llama function calling](https://docs.llama-api.com/essentials/function), [mistral.ai function calling](https://docs.mistral.ai/guides/function-calling/), where the structured request functions specifications and structured response are similar. Since the response is structured, there is no needs for a "stop sequence" like e.g. "observation" to intercept/detect function calling. In addition, for this purpose some of these llms are explicitly set to a "json mode".

My understanding is that some React agent implementations (e.g. [langchain react agent](https://github.com/langchain-ai/langchain/blob/d5c412b0a9a8af02754de60f93d93c3efa33ccdb/libs/langchain/langchain/agents/react/agent.py#L16)) use function specifications and instructions as part of the system message/prompt and not using the function calling api parameter. Part of those instructions is to respond in a thought, action, observation format ("unstructured" non-json/xml). Althought, it is still imaginable that the llm service like e.g Xorbits Inference tool "injects" these instructions, but usually it's done on application level. So with respect of not using the OpenAI **function calling** api, you're right that it will not conflict with the OpenAI api.  

This is my interpretation and I trying to learn, please let me know of there are flaws in my reasoning and assumptions.

**Comment by zhanghx0905** [comment]
@weipienlee 

I've filtered out Qwen ”React“ intermediate thought process, and now the feature should be much closer to what you'd expect. You can try it.

**Comment by weipienlee** [comment]
My excuses @zhanghx0905, I think we have a misunderstanding. I was not aiming for changing the "React" agent design pattern, rather, for an extra "OpenAI function calling/tool_choice mode/api". Having something "in the middle" might not beneficial for anybody. 

E.g. I tested your code `commit hash:5e712b127809aae887ef54c85195ef3867c1c7bf` with a React agent design pattern in streaming mode and that works perfectly for **non-quantized** Qwen-chat 1.8B and Qwen1.5-chat 0.5B pytorch models. I'm using the OpenAI rest API via LangChain ChatOpenAI class.

Once using any **quantized** models, however, Orbits Inference does not adhere to the stop sequence. I tried:
- "\nObservation" (default)
- 'Observation'
- 'Observation:'

but this might a be another bug.

I can test your last commits but I have to figure out how to checkout/see you last commits in my clone

**Comment by zhanghx0905** [comment]
> My excuses @zhanghx0905, I think we have a misunderstanding. I was not aiming for changing the "React" agent design pattern, rather, for an extra "OpenAI function calling/tool_choice mode/api". Having something "in the middle" might not beneficial for anybody (worried about breaking existing code).

I must remind you that the so-called function calling or tool calling is actually nothing more than prompting to drive LLM to output a JSON object. "React" is one way of organizing prompts. An LLM must be fine-tuned and aligned before it can correctly output this kind of json object and understand prompt. This is why very few open source models have "tool" capabilities (almost only Qwen series).

The larger the number of parameters of the model, the greater its ability to follow instructions, select the right tools, and generate the correct json payload. The number of model parameters of 0.5B and 1.8B is too small for a LLM. Their capabilities are actually very limited, let alone their quantified versions.

**Comment by weipienlee** [comment]
I tested commit hash 17df51595f08c7be2a0b84cfb132f44282251797 and seems not much changed, but it is still working. So I'm not sure what the goal was. Using the OpenAI api with React design pattern LangChain ChatOpenAI class:

**was** (result of another day, commit hash 5e712b127809aae887ef54c85195ef3867c1c7bf)
llm tool call:
I should use the duckduckgo_search tool to find out the current weather in Hamburg.
Action: duckduckgo_search
Action Input: "current weather in Hamburg

llm final anwers:
I now know the current weather in Hamburg.
Final Answer: The current weather in Hamburg is mostly cloudy with a chance of showers throughout the day. The temperature is ranging from 10°C to 14°C, with a light breeze coming from the north-west. There is no precipitation expected for the next 10 days.

**current**
llm tool call:
I should use the duckduckgo_search tool to find out the current weather in Hamburg.
Action: duckduckgo_search
Action Input: "current weather in Hamburg

llm final answer:
I now know the current weather in Hamburg.
Final Answer: The current weather in Hamburg is currently clear with a low chance of precipitation. The temperature is 6°C and feels like 4°C. There is a light breeze from the north east with gusts at 6 m/s. The long-range 12-day forecast shows that there is a high chance of precipitation near 90%. You can check the latest weather information by visiting the website or app of the city of Hamburg.

**Comment by weipienlee** [comment]
> > My excuses @zhanghx0905, I think we have a misunderstanding. I was not aiming for changing the "React" agent design pattern, rather, for an extra "OpenAI function calling/tool_choice mode/api". Having something "in the middle" might not beneficial for anybody (worried about breaking existing code).
> 
> I must remind you that the so-called function calling or tool calling is actually nothing more than prompting to drive LLM to output a JSON object. "React" is one way of organizing prompts. OpenAI's tool call API is just a layer of encapsulation on top of it. This is the purpose of this PR.
> 
> An LLM must be fine-tuned and aligned before it can correctly output this kind of json object and understand prompt. This is why very few open source models have "tool" capabilities (almost only Qwen series).
> 
> The larger the number of parameters of the model, the greater its ability to follow instructions, select the right tools, and generate the correct json payload. The number of model parameters of 0.5B and 1.8B is too small. Their capabilities are actually very limited, let alone their quantified versions.

Ok, understood and thanks for the effort to handle my feedback @zhanghx0905. I will try larger models as well, but have the aforementioned remaining end-sequence adherence problem with quantized models. Do you know if this this also due to the size of model (quantized degraded behaviour) or an actual "bug?

**Comment by weipienlee** [comment]
Another unrelated question @zhanghx0905, is there a way to run an Inference checkout without building an egg and install it first?

**Comment by zhanghx0905** [comment]
> > > My excuses @zhanghx0905, I think we have a misunderstanding. I was not aiming for changing the "React" agent design pattern, rather, for an extra "OpenAI function calling/tool_choice mode/api". Having something "in the middle" might not beneficial for anybody (worried about breaking existing code).
> > 
> > 
> > I must remind you that the so-called function calling or tool calling is actually nothing more than prompting to drive LLM to output a JSON object. "React" is one way of organizing prompts. OpenAI's tool call API is just a layer of encapsulation on top of it. This is the purpose of this PR.
> > An LLM must be fine-tuned and aligned before it can correctly output this kind of json object and understand prompt. This is why very few open source models have "tool" capabilities (almost only Qwen series).
> > The larger the number of parameters of the model, the greater its ability to follow instructions, select the right tools, and generate the correct json payload. The number of model parameters of 0.5B and 1.8B is too small. Their capabilities are actually very limited, let alone their quantified versions.
> 
> Ok, understood and thanks for the effort to handle my feedback @zhanghx0905. I will try larger models as well, but have the aforementioned remaining end-sequence adherence problem with **gguff** quantized models. Do you know if this this also due to the size of model (quantized degraded behaviour) or an actual "bug? I already tried e.g. 7B quantized as well but also failing. Maybe still not big enough?

This PR only supports vllm, if you don't use vllm you should get an 400 error and shouldn't get any output. There does seem to be a problem, thanks for the feedback!

**Comment by weipienlee** [comment]
i see, tnx

**Comment by weipienlee** [comment]
I used llama.cpp to run quantized models and Inference is using it. Is support on the roadmap?

**Comment by leoterry-ulrica** [comment]
image
Although fastgpt uses the tool without reporting errors, the content of the answer is incorrect

**Comment by zhanghx0905** [comment]
> Although fastgpt uses the tool without reporting errors, the content of the answer is incorrect

我只为 vllm backend 实现了该功能，其它的 backend 仍然返回 400 error，这个在最新的 commit 中应该已经保证了。

File: xinference/model/llm/utils.py [hunk 2]
```diff
@@ -446,6 +446,7 @@ def _to_chat_completion_chunk(cls, chunk: CompletionChunk) -> ChatCompletionChun
                     "index": i,
                     "delta": {
                         "content": choice["text"],
+                        "tool_calls": choice.get("tool_calls"),
```
ChengjieLi28, line 449: When there is no `tool_call` field, should we just not add this field the the chat completion, rather than a `None` here? (in order to be more compatible with OpenAI API)

xinference/model/llm/vllm/core.py
```diff
@@ -319,16 +323,42 @@ async def async_generate(
...
+
+                if tools:
+                    # only handle the first choice
+                    choice = chunk["choices"][0]
+                    if choice["finish_reason"] is not None:
+                        choice["text"] = previous_texts[0]
+                        _content, func, args = ChatModelMixin._eval_tool_arguments(
+                            self.model_family, chunk, tools
+                        )
+                        choice["text"] = ""
```
ChengjieLi28, line 347: Why `text` field becomes empty string when it is the `tool call` results? I think it would be good to keep the origin text information.

xinference/model/llm/utils.py
```diff
@@ -606,7 +606,34 @@ def _tool_calls_completion(cls, model_family, model_uid, c, tools):
                 f"Model {model_family.model_name} is not support tool calls."
             )
         logger.debug("Tool call content: %s, func: %s, args: %s", content, func, args)
+        return content, func, args
 
+    @classmethod
+    def _tools_token_filter(cls, model_family):
```
ChengjieLi28, line 612: I'm confused that what's this function for? It seems that if `family in ["qwen-chat", "qwen1.5-chat"]`, judge that whether `tokens.endswith("\nFinal Answer:")`. Why need this complicated function?

xinference/model/llm/vllm/core.py
```diff
@@ -319,16 +323,42 @@ async def async_generate(
...
+
                 for i, choice in enumerate(chunk["choices"]):
                     delta = choice["text"][len(previous_texts[i]) :]
                     previous_texts[i] = choice["text"]
                     choice["text"] = delta
+
+                if tools:
+                    # only handle the first choice
+                    choice = chunk["choices"][0]
+                    if choice["finish_reason"] is not None:
```
ChengjieLi28, line 342: Why use `choice["finish_reason"] is not None` as the condition?

xinference/model/llm/utils.py
zhanghx0905: I'd like to filter out Qwen's react thought process, i.e. "Thought, Action, Observe" and just keep what comes after "Final Answer:".

xinference/model/llm/vllm/core.py
zhanghx0905: Or choice["finish_reason"] == "stop"? Anyway we need a way to check if the model wants to call the tool, at least for Qwen `choice["finish_reason"] is not None` is feasible.

xinference/model/llm/utils.py
ChengjieLi28: Please make the code more cleaner. It is so complicated here. The code here is much less readable.

xinference/model/llm/vllm/core.py
zhanghx0905: The main reason for setting the content to empty ('') or None is because I was testing with FastGPT, which initially checks if the content is None. Only if the content is indeed None, does it proceed to determine whether a tool has been invoked.

https://github.com/labring/FastGPT/blob/5cfa43287f16b441f24ae5435e374fdf92c52153/packages/service/core/workflow/dispatch/agent/runTool/toolChoice.ts#L334

Of course their code may not make complete sense.

xinference/model/llm/utils.py
zhanghx0905: We can replace it with `**({"tool_calls": choice["tool_calls"]} if "tool_calls" in choice else {})`

xinference/model/llm/vllm/core.py
ChengjieLi28: The `finish_reason` comes from the vllm output here. Did you make sure this  reliable?

xinference/model/llm/utils.py
zhanghx0905: If you don't think this is necessary, we can simply remove this function, as well as the relevant lines in vllm/core.py

ChengjieLi28 (approved):

PR #1224, opened 2024-04-01
ENH: make function create_embedding could receive extra args

amumu96:


xinference/model/rerank/core.py
```diff
@@ -134,6 +134,7 @@ def rerank(
         top_n: Optional[int],
         max_chunks_per_doc: Optional[int],
         return_documents: Optional[bool],
+        **kwargs,
```
qinxuye, line 137: Should add a assert to disallow actual values in the kwargs.

qinxuye (approved):
LGTM

xinference/model/rerank/core.py
```diff
@@ -134,8 +134,11 @@ def rerank(
...
     ) -> Rerank:
         assert self._model is not None
+        if len(kwargs):
```
ChengjieLi28, line 140: Just `if kwargs`

ChengjieLi28 (approved):

PR #1227, opened 2024-04-01
BUG: Fix `opencv` issue in docker container

### LINKED ISSUE #1225: ImportError when running xprobe/xinference:latest (v.0.10.0) container with OpenCV dependency
### Bug Description
When attempting to run the `xprobe/xinference:latest` container, I encountered an `ImportError` related to the OpenCV (`cv2`) dependency not finding `libgthread-2.0.so.0`.

### Steps to Reproduce
1. Pull the latest image of `xprobe/xinference` from Docker Hub.
2. Run the container with the following command:
docker run -v D:\xinference_data:/xinference_home -e XINFERENCE_HOME=/xinference_home -p 9998:9997 --gpus all xprobe/xinference:latest xinference-local -H 0.0.0.0
3. Observe the error output that includes the following message:
ImportError: libgthread-2.0.so.0: cannot open shared object file: No such file or directory

### Expected Behavior
The container should run without encountering the `ImportError`, as the necessary libraries, including `libgthread-2.0.so.0`, should be available within the container environment.

### Actual Behavior
The container fails to run, and the error message indicates that the `libgthread-2.0.so.0` library is missing, which suggests that it is not part of the container's built image.

### Screenshots
image

### Environment
- Windows 10
- Docker version: 4.28.0

### Possible Solution
A potential solution would be to ensure that `libglib2.0-0` is included in the Docker image. This could be done by adding the following lines to the Dockerfile:
```Dockerfile
RUN apt-get update && \
 apt-get install -y libglib2.0-0 && \
 rm -rf /var/lib/apt/lists/*

ChengjieLi28:
Fixes #1225

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,7 +49,7 @@ install_requires =
     async-timeout
     peft
     timm
-    opencv-python
+    opencv-contrib-python
 
 [options.packages.find]
 exclude =
@@ -74,7 +74,7 @@ dev =
     flake8>=3.8.0
     black
     openai>1
-    opencv-python
+    opencv-contrib-python
     langchain
     orjson
     sphinx-tabs
@@ -146,7 +146,7 @@ doc =
     sphinx-design
     prometheus_client
     timm
-    opencv-python
+    opencv-contrib-python
 benchmark =
     psutil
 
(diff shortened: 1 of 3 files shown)


qinxuye (approved):
LGTM

PR #1231, opened 2024-04-02
BUG: Fix issue with supervisor not clearing information after worker exit

hainaweiben:
The problem arose in a distributed environment where upon the unexpected or manual exit (such as CTRL-C) of a worker, the supervisor failed to clear corresponding information. This resulted in the subsequent startup of the same model generating IDs with random strings.

### DIFF
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -870,9 +870,6 @@ class SupervisorActor(xo.StatelessActor):
                             address,
                             dead_models,
                         )
-                        for model_uid in dead_models:
-                            self._model_uid_to_replica_info.pop(model_uid[:-4], None)
-                            self._replica_model_uid_to_worker.pop(model_uid, None)
                         dead_nodes.append(address)
                     elif (
                         status.failure_remaining_count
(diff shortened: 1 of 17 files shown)


xinference/core/supervisor.py
```diff
@@ -870,6 +870,9 @@ async def _check_dead_nodes(self):
                             address,
                             dead_models,
                         )
+                        for model_uid in dead_models:
+                            self._model_uid_to_replica_info.pop(model_uid[:-4], None)
```
ChengjieLi28, line 874: Do not use `model_uid[:-4]` to get `model_uid` from `rep_model_uid`.
Use `parse_replica_model_uid` util to get real `model_uid`.

xinference/core/supervisor.py
```diff
@@ -870,6 +870,9 @@ async def _check_dead_nodes(self):
...
+                            self._replica_model_uid_to_worker.pop(model_uid, None)
```
ChengjieLi28, line 875: Please also consider to remove info in the function `remove_worker`. 
`remove_worker` is the logic of the supervisor when the worker is normally exited by ctrl-c. From there it is possible to remove worker related information directly. 
On the one hand, it forms a double insurance policy with the worker detection logic. 
On the other hand, some users may not turn on worker detection.

xinference/core/supervisor.py
```diff
@@ -979,6 +979,15 @@ async def add_worker(self, worker_address: str):
 
     @log_async(logger=logger)
     async def remove_worker(self, worker_address: str):
+        dead_models = []
```
ChengjieLi28, line 982: What's this `dead_models` used for?

xinference/core/supervisor.py
```diff
@@ -979,6 +979,15 @@ async def add_worker(self, worker_address: str):
...
+        for model_uid in self._replica_model_uid_to_worker:
+            if self._replica_model_uid_to_worker[model_uid].address == worker_address:
+                dead_models.append(model_uid)
+        for replica_model_uid in dead_models:
+            model_uid, _, _ = parse_replica_model_uid(replica_model_uid)
+            self._model_uid_to_replica_info.pop(model_uid, None)
+            self._replica_model_uid_to_worker.pop(replica_model_uid, None)
+
```
ChengjieLi28, line 990: You should do these things also in worker dead detection logic.

ChengjieLi28 (approved):

PR #1242, opened 2024-04-03
TST: Fix tests due to `llama-cpp-python` `v0.2.58`

ChengjieLi28:
See https://github.com/abetlen/llama-cpp-python/issues/1319


qinxuye (approved):
LGTM

PR #1254, opened 2024-04-07
REF: Refactor launch model for Web UI

yiboyasss:


yiboyasss:
The renderings are as follows

截屏2024-04-08 11 38 56

ChengjieLi28 (approved):

PR #1262, opened 2024-04-08
FEAT: Support Qwen MoE model for huggingface and modelscope

xiaodouzi666:


XiaoCC:
The checkpoint you are trying to load has model type `qwen2_moe` but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.

PR #1268, opened 2024-04-09
DOC: update readme

qinxuye:


setup.cfg
```diff
@@ -27,6 +27,7 @@ install_requires =
     xoscar>=0.3.0
     torch
     gradio>=3.39.0
+    typer[all]<0.12.0  # fix typer required by gradio
```
ChengjieLi28, line 30: Also add this to cpu dockerfile.

ChengjieLi28 (approved):

PR #1277, opened 2024-04-10
ENH: support int and str compare for model size

mikeshi80:
Currently, if LLM Spec's `model_size_in_billions` is int, but client send the string, or vice versa, the launch model command will fail.

For example, `model_size_in_billions` in LLM Spec is 32, but the parameter in launch model command is "32", it fails.

This PR makes this judgment smarter, so 32 will equal to "32", or "1.8" will equal to "1_8".


xinference/model/llm/llm_family.py
```diff
@@ -782,6 +785,32 @@ def get_user_defined_llm_families():
...
+def match_model_size(
+    model_size: Union[int, str], spec_model_size: Union[int, str]
+) -> bool:
+    if model_size == spec_model_size:
+        return True
+
+    if isinstance(model_size, str):
+        try:
+            ms = int(model_size)
+            ss = int(spec_model_size)
```
ChengjieLi28, line 797: What if `model_size` is `"1_8"` and `spec_model_size` is `"18"`?
There's a misjudge here.

mikeshi80: ok, fixed it and add some test cases.

xinference/model/llm/llm_family.py
```diff
@@ -782,6 +782,25 @@ def get_user_defined_llm_families():
         return UD_LLM_FAMILIES.copy()
 
 
+def match_model_size(
+    model_size: Union[int, str], spec_model_size: Union[int, str]
```
ChengjieLi28, line 786: Now what if `model_size` is a float value? For example, `model_size` is `1.8` and `spec_model_size` is `1_8`?

mikeshi80: OK, I will add the support. I noticed the type hint is `int` or `Union[int, str]` in many function signatures, I will modify them as well.

mikeshi80: I think only int and string will be passed as parameter of RESTful API endpoint for launch model after I read the source code, the float value will be converted to string by replace the dot to underscore like "1_8", so the function above just supports int and str is sufficient.

The RESTful endpoint for model launch will convert string to int if there is no underscore in it, it will make "1.8" format float string failed, so I added the support for float string with dot as delimiter.

And I found launch model APIs in `restful_api` only support int and string format value, so I added the support for float since to use float value is more intuitive.

ChengjieLi28 (approved):

PR #1350, opened 2024-04-22
BUG: Fix old rerank models use flag rerank issue

codingl2k1:
Old reranker model use `sentence_transformers`, only new reranker model use `FlagEmbedding`.

### DIFF
--- a/xinference/model/rerank/core.py
+++ b/xinference/model/rerank/core.py
@@ -112,16 +112,10 @@ class RerankModel:
         self._device = device
         self._model_config = model_config or dict()
         self._use_fp16 = use_fp16
-        self._use_sentence_transformers = model_spec.model_name in [
-            "bge-reranker-large",
-            "bge-reranker-base",
-            "bce-reranker-base_v1",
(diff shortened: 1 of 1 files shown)


qinxuye:
Can sentence-transformers support `bge-reranker-v2-m3`?

codingl2k1:
I will try it, but the document does not show this usage.

xinference/model/rerank/core.py
```diff
@@ -112,29 +112,52 @@ def __init__(
         self._device = device
         self._model_config = model_config or dict()
         self._use_fp16 = use_fp16
+        self._use_sentence_transformers = model_spec.model_name in [
```
qinxuye, line 115: How about for model that type == 'normal' or not specified type just using sentence_transformers? If user register rerank, what library will use?

codingl2k1: Good point. Currently, the most popular rerank model is the normal type.

qinxuye (approved):
LGTM, already tested, these models work well.

PR #1354, opened 2024-04-23
ENH: fix the max length of codeqwen-7B-chat

mikeshi80:
Although both in the official document and config.json, the max length is stated as 65536, however, it is 32768 after actual testing


qinxuye (approved):
LGTM

PR #1372, opened 2024-04-25
ENH: Support more quantization with VLLM

amumu96:


qinxuye:
IIRC, there are two match functions, besides, the docs need to be updated.

qinxuye (approved):
LGTM

xinference/model/llm/vllm/core.py
```diff
@@ -239,10 +239,17 @@ def match(
         if llm_spec.model_format == "pytorch":
             if quantization != "none" and not (quantization is None):
                 return False
-        if llm_spec.model_format in ["gptq", "awq"]:
-            # Currently, only 4-bit weight quantization is supported for GPTQ, but got 8 bits.
+        if llm_spec.model_format == "awq":
+            # Currently, only 4-bit weight quantization is supported for AWQ, but got 8 bits.
             if "4" not in quantization:
                 return False
+        if llm_spec.model_format == "gptq":
```
qinxuye, line 246: We should consider the version of vLLM, the int3 and int8 should be supported since v0.3.3.

Issue #1293, opened 2024-04-13
xinferrence  rewrite user's prompt cause unexpected behavior with tools parameter in Qwen1.5 series


onionknightdd:
### Describe the bug
xinferrence seems using `isinstance(tools,list)` to trigger the ReAct Prompt to implement the tool use instead of `True if  tools else False` when use Qwen1.5 LLM

### To Reproduce
To help us to reproduce this bug, please provide information below:
![image](https://github.com/xorbitsai/inference/assets/49679672/6a68b146-29b5-4f44-9668-da41f1fa23da)
![image](https://github.com/xorbitsai/inference/assets/49679672/9051669d-1fbb-486e-ac60-012d8edbefa9)
![image](https://github.com/xorbitsai/inference/assets/49679672/1a067372-d7ed-4bec-acd6-b800ec2cee1b)

when set tools=[]  xinference use ReAct Prompt rewrite the user's promtp and send to llm
when set tools=None  xinference behave like normal
when the tool has no parameters , xinference raise the error ,which should not, the tool use should aceept the tools/func without parms

1. Your Python version. --> 3.10
2. The version of xinference you use.  --->v0.10.1
3. Versions of crucial packages.   -->  xinference 
4. Full stack of the error.  --> n/a
5. Minimized code to reproduce the error.  use post request to access /v1/chat/completions with parameter `"tools" :[ ]`

### Expected behavior
should use `True if  tools else False` to implemet the tool use 
and when the tool has no parammters the api should accpet it instead of raise error

qinxuye:
Do you have interest to send a PR to address this issue?

PR #1381, opened 2024-04-25
BUG: Fix qwen tool call paramerter empty issue

codingl2k1:
Fixes: https://github.com/xorbitsai/inference/issues/1293

### DIFF
--- a/xinference/core/tests/test_restful_api.py
+++ b/xinference/core/tests/test_restful_api.py
@@ -663,7 +663,7 @@ def test_restful_api_for_gorilla_openfunctions_tool_calls(
 )
 @pytest.mark.skip(reason="Cost too many resources.")
 def test_restful_api_for_qwen_tool_calls(setup, model_format, quantization):
-    model_name = "qwen-chat"
+    model_name = "qwen1.5-chat"
 
     endpoint, _ = setup
     url = f"{endpoint}/v1/models"
@@ -677,7 +677,7 @@ def test_restful_api_for_qwen_tool_calls(setup, model_format, quantization):
     payload = {
         "model_uid": "test_tool",
         "model_name": model_name,
-        "model_size_in_billions": 7,
+        "model_size_in_billions": 4,
         "model_format": model_format,
         "quantization": quantization,
     }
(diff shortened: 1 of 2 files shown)


xinference/core/tests/test_restful_api.py
```diff
@@ -691,6 +691,67 @@ def test_restful_api_for_qwen_tool_calls(setup, model_format, quantization):
...
+    assert "content" in completion["choices"][0]["message"]
+    assert "tool_calls" == completion["choices"][0]["finish_reason"]
+    assert (
+        "google_search"
+        == completion["choices"][0]["message"]["tool_calls"][0]["function"]["name"]
+    )
+    arguments = completion["choices"][0]["message"]["tool_calls"][0]["function"][
+        "arguments"
+    ]
+    assert not json.loads(arguments)
```
qinxuye, line 753: I tested and change back to 7b, 4b will failure earlier, and for 7b it will fail at this line.
```
>       assert not json.loads(arguments)
E       assert not {'query': '周杰伦'}
E        +  where {'query': '周杰伦'} = <function loads at 0x7fd1171cfc40>('{"query": "\\u5468\\u6770\\u4f26"}')
E        +    where <function loads at 0x7fd1171cfc40> = json.loads
```

codingl2k1: I updated the ut and tested the latest 4b and 7b qwen 1.5:
- 4b
  - Can't pass the false tool call check (the model should not call any tools because there are no matched tools), so I removed this check.
  - The output varies between torch and ggufv2 models, so I removed this check.
- 7b
  - Seems does not support tool calls. All the tool calls are failed.
  - torch: Can't tool call with or without tool parameters.
  - ggufv2 Q5_K_M: Output 9!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

codingl2k1: I think the qwen 1.5 model has problems making tool calls. My env:

pip Name: transformers
Version: 4.39.1

Name: llama_cpp_python
Version: 0.2.64

qinxuye: How about just using 7b? Maybe 4b is not good enough.

codingl2k1: I have tested 7b, as shown above, it does not support tool calls. All the tests are failed. 4b is OK.

codingl2k1: I cleaned my local cache and downloaded the latest model to test.

qinxuye: The pytorch version can pass, but failed for gguf, is this PR ready for review?

qinxuye (approved):
LGTM

PR #1405, opened 2024-04-29
BLD: Use self-hosted aws machine to build docker image

ChengjieLi28:
See https://github.com/xorbitsai/inference/actions/runs/8872625046/job/24357238245


qinxuye (approved):
LGTM

PR #1420, opened 2024-05-01
BUG: Fix tool calls return invalid usage

codingl2k1:
The qwen model use ReAct to call tools, the token usage may larger than the expected.

- Fixes: https://github.com/xorbitsai/inference/issues/1412


qinxuye (approved):
LGTM

PR #1425, opened 2024-05-05
ENH: make qwen_vl support streaming output

Minamiyama:


fyrestone:
Could you add some unit tests? We don't run them in our CI because of resource limitations, but we can use them to check the code during development. https://github.com/xorbitsai/inference/blob/main/xinference/model/llm/tests/test_multimodal.py#L24

codingl2k1:
Could you add some unit tests? We don't run them in our CI because of resource limitations, but we can use them to check the code during development. https://github.com/xorbitsai/inference/blob/main/xinference/model/llm/tests/test_multimodal.py#L24

Minamiyama:
> Could you add some unit tests? We don't run them in our CI because of resource limitations, but we can use them to check the code during development. https://github.com/xorbitsai/inference/blob/main/xinference/model/llm/tests/test_multimodal.py#L24

how about this⬇️ record video？
https://github.com/xorbitsai/inference/assets/918617/5d975b58-cc02-4e27-b5bf-d059be8a4c7f

qinxuye (approved):
LGTM, excellent work.

PR #1435, opened 2024-05-07
CLN: Remove all speculative-related codes

ChengjieLi28:
Remove these because this interface has been severely mismatched with the current interface.
If you want to know about the impl details, please see #509


qinxuye (approved):
LGTM

PR #1461, opened 2024-05-09
BUG: fix top_k for vllm backend

sixsun10:


xinference/model/llm/vllm/core.py
```diff
@@ -216,6 +217,7 @@ def _sanitize_generate_config(
         )
         sanitized.setdefault("temperature", generate_config.get("temperature", 1.0))
         sanitized.setdefault("top_p", generate_config.get("top_p", 1.0))
+        sanitized.setdefault("top_k", generate_config.get("top_k", 10))
```
qinxuye, line 220: I checked the doc of vllm, the default value of top_k is -1, could set it to -1?

qinxuye:
Thanks for your contribution.

PR #1476, opened 2024-05-11
FEAT: Code completions

mikeshi80:
Added the support for code completions, support 3 kinds of completions:

 1. Common code completions, like generate model, it will continue to generate the code according inputted code above.
 2. Code infill completions, the more parameter "suffix" is provided, the model generates the content between prompt and suffix.
 3. Repository level code completions, just like common code completions, but the other files in the repository are provided too, to generate more precise content.

I added two RESTful APIs:

 1. `/v1/code/prompt`
 2. `v1/code/completions`

Users can get the generated prompt by call 1st API.

### Common Code Completion

POST the JSON object

```json
{
  "model": "code-completion-model-uid",
  "mode": "completion",
  "prompt": "#write a quick sort algorithm"
}
```

### Repository Level Code Completion
Users can provide files parameter , which is the mapping, key is file name or file path (it is defined in code prompt style), value is the file content.

Some model need the `repo_name`, that presents repository name.

```json
{
  "model": "code-completion-model-uid",
  "mode": "completion",
  "prompt": "#write a quick sort algorithm",
  "files": {
    "test.py": "test file content",
    "main.py": "main file content"
  },
  "repo_name": "my-test-repo"
}
```

Note: if the model does not support repository level completion, it works like common code completion, "files" and "repo_name" will be ignored.

### Infill Code Completion

Users can use infill mode by set mode to "infill", and provide suffix parameter.

```json
{
  "model": "code-completion-model-uid",
    "mode": "infill",
    "prompt": "def quick_sort(arr):\n    if len(arr) <= 1:\n        return arr\n    pivot = arr[0]\n    left = []\n    right = []\n",
    "suffix": "\n        if arr[i] < pivot:\n            left.append(arr[i])\n        else:\n            right.append(arr[i])\n    return quick_sort(left) + [pivot] + quick_sort(right)"
}
```

Note: if the model does not support infill mode, you will get the exception.

If `/v1/code/prompt` API works well, it will return the response like below:

```json
{
  "prompt": "<｜fim▁begin｜>def quick_sort(arr):\n    if len(arr) <= 1:\n        return arr\n    pivot = arr[0]\n    left = []\n    right = []\n<｜fim▁hole｜>\n        if arr[i] < pivot:\n            left.append(arr[i])\n        else:\n            right.append(arr[i])\n    return quick_sort(left) + [pivot] + quick_sort(right)<｜fim▁end｜>"
}
```

When you get the generated prompt, you can pass it to `/v1/completions` to get the response.

If you want to combine 2 requests into 1, you can just use the 2nd RESTful API `/v1/code/completions`, it just combined generating prompt and completions requests.


qinxuye:
There are conflicts to resolve.

PR #1500, opened 2024-05-15
ENH: Compatible with `huggingface_hub` >= `0.23.0`

Ago327:


setup.cfg
```diff
@@ -36,7 +36,7 @@ install_requires =
     pydantic
     fastapi
     uvicorn
-    huggingface-hub>=0.19.4,<0.23.0
```
ChengjieLi28, line 39: Also remove restrict in `xinference/deploy/docker/cpu.Dockerfile`

xinference/model/llm/llm_family.py
```diff
@@ -682,42 +682,121 @@ def cache_from_huggingface(
...
+            cache_dir_name = (
+                f"{llm_family.model_name}-{llm_spec.model_format}"
+                f"-{llm_spec.model_size_in_billions}b"
+            )
+            if quant_suffix:
+                cache_dir_name += f"-{quant_suffix}"
+            home_path = os.environ.get(XINFERENCE_ENV_HOME_PATH)
+            if home_path is None:
+                home_path = str(Path.home())
+            real_path = home_path + "/.cache/huggingface/hub/" + cache_dir_name
```
ChengjieLi28, line 710: Just use `XINFERENCE_CACHE_DIR`. Same as all other places.

xinference/model/llm/llm_family.py
```diff
@@ -682,42 +682,121 @@ def cache_from_huggingface(
...
+                    "model_format": llm_spec.model_format,
+                },
+                llm_spec.model_id,
+                revision=llm_spec.model_revision,
+                local_dir=real_path,
+            )
+            for subdir, dirs, files in os.walk(download_dir):
+                for file in files:
+                    relpath = os.path.relpath(os.path.join(subdir, file), download_dir)
+                    symlink_local_file(os.path.join(subdir, file), cache_dir, relpath)
```
ChengjieLi28, line 726: Can we just download to the realpath and do not create symlink?

PR #1528, opened 2024-05-21
BUG: Fix gpu_idx allocate error when set replica > 1 

amumu96:


qinxuye:
How about worker IP?

ChengjieLi28 (approved):

qinxuye:
I mean how to specify two replica with worker A card 1+2 and worker B card 0+3, for instance?

PR #1533, opened 2024-05-22
FEAT: add deepseek llm and coder base

qinxuye:
Deepseek llm and coder base is absent, but coder base is useful for code complete.


codingl2k1 (approved):
LGTM

PR #1534, opened 2024-05-22
BLD: compatible with torch 2.3.0

qinxuye:
Main reason is thirdparty of OmniLLM inited cuda in some imports, but I don't know what triggered, so I move the imports into functions.


codingl2k1 (approved):
LGTM

PR #1535, opened 2024-05-22
FEAT: add codeqwen1.5

qinxuye:
Base model is useful for code completion job, thus added codeqwen1.5 base model.


codingl2k1 (approved):
LGTM

Issue #1515, opened 2024-05-17
BUG: Fail to run model "bge-reranker-v2-minicpm-layerwise" with Xinference v0.11.0 version (docker images)


majestichou:
### Describe the bug
I downloaded the "bge-reranker-v2-minicpm-layerwise" model weights to the server and registered this model (the registered model name is "bge-reranker-v2-minicpm-layerwise-self") with Xinference v0.11.0 version (docker images). Then I launched this model. However, it crushed. The error information is as follows:
```
Traceback (most recent call last):
  File "/opt/conda/lib/python3.10/site-packages/xinference/api/restful_api.py", line 697, in launch_model
    model_uid = await (await self._get_supervisor_ref()).launch_builtin_model(
  File "/opt/conda/lib/python3.10/site-packages/xoscar/backends/context.py", line 227, in send
    return self._process_result_message(result)
  File "/opt/conda/lib/python3.10/site-packages/xoscar/backends/context.py", line 102, in _process_result_message
    raise message.as_instanceof_cause()
  File "/opt/conda/lib/python3.10/site-packages/xoscar/backends/pool.py", line 659, in send
    result = await self._run_coro(message.message_id, coro)
  File "/opt/conda/lib/python3.10/site-packages/xoscar/backends/pool.py", line 370, in _run_coro
    return await coro
  File "/opt/conda/lib/python3.10/site-packages/xoscar/api.py", line 384, in __on_receive__
    return await super().__on_receive__(message)  # type: ignore
  File "xoscar/core.pyx", line 558, in __on_receive__
    raise ex
  File "xoscar/core.pyx", line 520, in xoscar.core._BaseActor.__on_receive__
    async with self._lock:
  File "xoscar/core.pyx", line 521, in xoscar.core._BaseActor.__on_receive__
    with debug_async_timeout('actor_lock_timeout',
  File "xoscar/core.pyx", line 526, in xoscar.core._BaseActor.__on_receive__
    result = await result
  File "/opt/conda/lib/python3.10/site-packages/xinference/core/supervisor.py", line 836, in launch_builtin_model
    await _launch_model()
  File "/opt/conda/lib/python3.10/site-packages/xinference/core/supervisor.py", line 800, in _launch_model
    await _launch_one_model(rep_model_uid)
  File "/opt/conda/lib/python3.10/site-packages/xinference/core/supervisor.py", line 781, in _launch_one_model
    await worker_ref.launch_builtin_model(
[...]

PR #1538, opened 2024-05-23
FEAT: Auto detect rerank type for unknown rerank type

codingl2k1:
Auto detect rerank type based on the tokenizer name.

Fixes: https://github.com/xorbitsai/inference/issues/1515


qinxuye (approved):
LGTM

Issue #1517, opened 2024-05-18
docker cpu 版本，下载模型异常：Server error: 500 - [address=0.0.0.0:33434, pid=15] 'NoneType' object is not subscriptable


Jamel-jun:
### Describe the bug
下载minicpm-2b-dpo-bf16 出现错误。

### To Reproduce
使用的 docker compose 部署，cpu 版本。

### Expected behavior
期望正常下载部署模型。

### Additional context
错误日志：
<img width="1982" alt="image" src="https://github.com/xorbitsai/inference/assets/50609931/009ea9f8-ce69-4596-bbf9-ac41700066f5">
<img width="1044" alt="image" src="https://github.com/xorbitsai/inference/assets/50609931/b8196248-b5f8-4988-8ad7-667306f93121">
<img width="1995" alt="image" src="https://github.com/xorbitsai/inference/assets/50609931/5f30bbea-655f-416c-9462-004ae79a54a2">

页面提示：
<img width="1226" alt="image" src="https://github.com/xorbitsai/inference/assets/50609931/2418fd36-a8a1-48bd-bdcb-412cbfc1be22">

docker compose 部署文件
```yaml
version: '3.4'
services:
  xinference:
    image: xprobe/xinference:latest-cpu
    container_name: xinference
    hostname: xinference
    command: xinference-local -H 0.0.0.0 --log-level debug
    privileged: true
    user: root
    environment:
      - TZ=Asia/Shanghai
      - XINFERENCE_MODEL_SRC=modelscope
      - XINFERENCE_HOME=/home/data
    volumes:
      - ./data:/home/data
    networks:
      - diy_network
    ports:
      - 9997:9997
    logging:
      driver: "json-file"
      options:
        max-size: "500m"

networks:
  diy_network:
    external: true
```

ciekawy:
I'm getting the same trying to run on macos  with `xinference-local --host 0.0.0.0 --port 9997` (non docker)

wencan:
+1

zhubinn:
+1

wencan:
The bug is not present in v0.11.0.

PR #1539, opened 2024-05-23
BUG: Fix start worker failed due to None device name

codingl2k1:
Fixes: https://github.com/xorbitsai/inference/issues/1517


qinxuye (approved):
LGTM, tested it, can work

qinxuye:
I modified because previous change cannot pass CI, now if device env name is None, just use CUDA_VISIBLE_DEVICES which is identical to previous behavior.

codingl2k1:
> I modified because previous change cannot pass CI, now if device env name is None, just use CUDA_VISIBLE_DEVICES which is identical to previous behavior.

Thanks.

PR #1548, opened 2024-05-28
FEAT: Continuous batching for chat model on transformers backend

ChengjieLi28:
- Continous batching impl based on huggingface transformers
- abort_request interface only for batching
- ~~TODO: UT~~

Known limitations:
- Only for chat model for now, since not all the generate model has batch generate ability.
- Only for `PytorchChatModel` in xinference. Some models like `chatglm3`, although it is loaded by `transformers` backend, it is launched by `ChatglmPytorchChatModel` class which directly uses `chat` interface and it cannot be applied to this feature.


xinference/core/model.py
```diff
@@ -181,9 +199,27 @@ def __init__(
...
+
+            self._scheduler_ref = await xo.create_actor(
+                SchedulerActor,
+                address=self.address,
+                uid=SchedulerActor.gen_uid(self.model_uid(), self._model.rep_id),
+            )
+            self._isolation = Isolation(asyncio.new_event_loop(), threaded=True)
+            self._isolation.start()
+            asyncio.run_coroutine_threadsafe(
+                self._scheduler_ref.run(), loop=self._isolation.loop
```
codingl2k1, line 220: Why not call the `run()` in the `__post_create__` of `SchedulerActor`? I think that could be cleaner.

xinference/core/model.py
```diff
@@ -235,8 +271,22 @@ def is_vllm_backend(self) -> bool:
...
+        return (
+            XINFERENCE_TRANSFORMERS_ENABLE_BATCHING
+            and isinstance(self._model, PytorchChatModel)
+            and self._model.__class__.__name__ == PytorchChatModel.__name__
+        )
+
+    async def load(self):
         self._model.load()
+        if self.allow_batching():
+            await self._scheduler_ref.set_model(self._model)
```
codingl2k1, line 286: Maybe we can make the model as a init param of `SchedulerActor` and create the `SchedulerActor` here.

ChengjieLi28: I tried this method. And I found that if passing self._model as an init parameter to an actor init method. The model will be copied.

codingl2k1: I got it. I have optimized the actor call in the same process to be a normal function call, with no data serialization. But the `create_actor` API uses the old protocol which will serialize/deserialize the input arguments.

codingl2k1: The full call stack in `create_actor`, we can see it go through the socket channel.

```
  File "/Users/codingl2k1/Work/inference/xinference/core/worker.py", line 650, in launch_builtin_model
    model_ref = await xo.create_actor(
    ^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.6/lib/python3.11/site-packages/xoscar/api.py", line 78, in create_actor
    return await ctx.create_actor(actor_cls, *args, uid=uid, address=address, **kwargs)
      ^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.6/lib/python3.11/site-packages/xoscar/backends/context.py", line 141, in create_actor
    future = await self._call(address, create_actor_message, wait=False)
    ^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.6/lib/python3.11/site-packages/xoscar/backends/context.py", line 77, in _call
    return await self._caller.call(
    ^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.6/lib/python3.11/site-packages/xoscar/backends/core.py", line 181, in call
    return await self.call_with_client(client, message, wait)
    ^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.6/lib/python3.11/site-packages/xoscar/backends/core.py", line 126, in call_with_client
    await client.send(message)
    ^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.6/lib/python3.11/site-packages/xoscar/backends/communication/base.py", line 278, in send
    return await self.channel.send(message)
    ^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.6/lib/python3.11/site-packages/xoscar/backends/communication/socket.py", line 77, in send
    buffers = await serializer.run()
    ^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.6/lib/python3.11/site-packages/xoscar/serialization/aio.py", line 71, in run
    return await self._get_buffers()
    ^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.6/lib/python3.11/site-packages/xoscar/serialization/aio.py", line 40, in _get_buffers
    headers, buffers = await serialize_with_spawn(
    ^^^^^^^^^^^^^^^^^
  File "xoscar/serialization/core.pyx", line 795, in serialize_with_spawn
    result = _serialize_with_stack(
    ^^^^^^^^^^^^^^^^^
  File "xoscar/serialization/core.pyx", line 689, in xoscar.serialization.core._serialize_with_stack
    serialized, subs, final = _serial_single(
    ^^^^^^^^^^^^^^^^^
  File "xoscar/serialization/core.pyx", line 627, in xoscar.serialization.core._serial_single
    ret_serial = serializer.serial(obj, context)
    ^^^^^^^^^^^^^^^^^
  File "xoscar/serialization/core.pyx", line 285, in xoscar.serialization.core.PickleSerializer.serial
    return (), pickle_buffers(obj), True
    ^^^^^^^^^^^^^^^^^
  File "xoscar/serialization/core.pyx", line 244, in xoscar.serialization.core.pickle_buffers
    buffers[0] = cloudpickle.dumps(
    ^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.6/lib/python3.11/site-packages/cloudpickle/cloudpickle_fast.py", line 73, in dumps
    cp.dump(obj)
  File "/Users/codingl2k1/.pyenv/versions/3.11.6/lib/python3.11/site-packages/cloudpickle/cloudpickle_fast.py", line 632, in dump
    return Pickler.dump(self, obj)
    ^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/Work/inference/xinference/model/llm/pytorch/core.py", line 462, in __reduce__
    raise Exception("Forbid serialization")
    ^^^^^^^^^^^^^^^^^
Exception: [address=localhost:60075, pid=57865] Forbid serialization
```

xinference/core/scheduler.py
```diff
@@ -0,0 +1,358 @@
...
+                waiting_list.append(req)
+                if len(running_list) + len(waiting_list) == max_num_seqs:
+                    break
+        # must waiting_list in front
+        return waiting_list + running_list
+
+    async def step(self):
+        from ..model.llm.pytorch.utils import empty_cache
+
+        req_list = self._handle_request()
```
codingl2k1, line 259: Actually, the `step` and `_handle_request` methods would be run in a dedicated thread. This is different from the threads used for other member methods, such as `add_request` and `abort_request`. Luckily, many data structures can be shared among threads without a lock.

xinference/core/scheduler.py
```diff
@@ -0,0 +1,358 @@
...
+            logger.info(f"Request id: {req_id} not found. No-op for xinference.")
+            return AbortRequestMessage.NOT_FOUND.name
+        else:
+            self._abort_req_ids.add(req_id)
+            logger.info(f"Request id: {req_id} found to be aborted.")
+            return AbortRequestMessage.DONE.name
+
+    async def run(self):
+        while True:
+            await self.step()
```
codingl2k1, line 356: We'd better add a try-catch block and logs in the while loop to avoid the thread exiting silently. For example, if an exception is raised, it will cause the thread to exit.

xinference/core/scheduler.py
ChengjieLi28: Currently, no thread safety issues have been identified in testing. Reopen this comment if error happens.

xinference/core/model.py
```diff
@@ -359,24 +399,82 @@ async def generate(self, prompt: str, *args, **kwargs):
...
+                        prompt, queue, *args, **kwargs
+                    )
+                    gen = self._to_json_async_gen(ret)
+                    self._current_generator = weakref.ref(gen)
+                    return gen
+                else:
+                    from .scheduler import XINFERENCE_NON_STREAMING_ABORT_FLAG
+
+                    assert self._loop is not None
+                    future = self._loop.create_future()
```
codingl2k1, line 456: This is an `asyncio.Future`, it is not thread safe: https://docs.python.org/3/library/asyncio-future.html#asyncio.Future You are awaiting for the future in the add_request thread and set it in the run thread. Another way is to create a concurrent future and wrap future.

```python
fut = concurrent.futures.Future()
# Pass off to your other thread somehow
# and return a wrapped future for asyncio
return asyncio.futures.wrap_future(fut, loop = loop)
```

await for the wrapped asyncio future and pass the concurrent future to another thread.

ChengjieLi28: Fix

qinxuye (approved):
LGTM

PR #1550, opened 2024-05-29
Correct ModelActor import path in worker & supervisor

frostyplanet:
It seems not necessary to install ModelActor in xinference.core, 
using original path core.model.ModelActor instead.

### DIFF
--- a/xinference/core/__init__.py
+++ b/xinference/core/__init__.py
@@ -11,5 +11,3 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-from .model import ModelActor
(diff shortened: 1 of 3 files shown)


ChengjieLi28:
Why should change this? What's the benefit of your PR?
Or what would be the problem with not changing it to that? 
Give me a valid reason for your PR to be signed off.

frostyplanet:
1. For better clearify dependency because supervisor & worker often deploy in seperate docker image
2. There's no point for supervisor import ModelActor since it only call it by RPC. Removing ModelActor from xinference.core.init is reasonable. This PR will not affect normal usage.
3. xinference cli is slow because it will import transformer & pytorch & init cuda. This is the prerequisite change.

ChengjieLi28:
> 1. For better clearify dependency because supervisor & worker often deploy in seperate docker image
> 2. There's no point for supervisor import ModelActor since it only call it by RPC. Removing ModelActor from xinference.core install process is reasonable. This PR will not affect normal usage.
> 3. xinference cli is slow because it will import transformers & pytorch which will init cuda. This is the prerequisite change to improve user interaction speed

Next time it would be better to put these directly in the description of the PR to avoid wasting time.

ChengjieLi28 (approved):

PR #1551, opened 2024-05-29
FEAT: Support new model CogVLM

amumu96:
image


xinference/model/llm/pytorch/cogvlm2.py
```diff
@@ -0,0 +1,255 @@
...
+                return text, None
+            elif len(images) == 1:
+                return text, images
+            else:
+                raise RuntimeError(
+                    "Only one image per message is supported by CogVLM2."
+                )
+        return content, None
+
+    def _history_content_to_intern(
```
codingl2k1, line 135: Is this method useful for CogVLM? It shows that `_history_content_to_intern`

amumu96: yes, just forget to rename func,  ill fix it

qinxuye (approved):
LGTM

fire17:
can you guys add CogVLM2 to the BrillAI platform?
I want to test its pixel coordinate functionality - ie
return the location of the button named "retry" in [[x0,y0,x1,y1]] format
it should return the bounding box in values between 1-1000 in reference to the image,
locating / pinpointing vlm is not something i have seen yet but cogvlm2 seems to be able to do this
Thanks :) and all the best

Issue #1288, opened 2024-04-12
FEAT: support text-to-audio open model


weili20:
### Is your feature request related to a problem? Please describe
No.

### Describe the solution you'd like
ModelScope has some text-to-audio open models. here is the link: https://modelscope.cn/models?page=1&tasks=text-to-speech&type=audio

### Describe alternatives you've considered
No

### Additional context
No

qinxuye:
Thank you for your suggestion, text-to-audio is definitely on our roadmap, we will support it in an appropriate time. Meanwhile, contribution welcomed for this feature.

liduang:
希望尽快支持文本转语音的模型，十分感谢

PR #1578, opened 2024-06-02
FEAT: Support ChatTTS

codingl2k1:
- Basic support `speech` audio API
- Compatible with OpenAI speech API
- Support ChatTTS
- Fixes: https://github.com/xorbitsai/inference/issues/1288

### DIFF
--- a/xinference/client/restful/restful_client.py
+++ b/xinference/client/restful/restful_client.py
@@ -684,6 +684,37 @@ class RESTfulAudioModelHandle(RESTfulModelHandle):
         response_data = response.json()
         return response_data
 
+    def speech(
+        self,
+        input: str,
+        voice: str = "",
+        response_format: str = "mp3",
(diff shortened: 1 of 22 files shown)


xinference/client/restful/restful_client.py
```diff
@@ -684,6 +684,37 @@ def translations(
...
+    def speech(
+        self,
+        input: str,
+        voice: str = "",
+        response_format: str = "mp3",
+        speed: float = 1.0,
+    ):
+        """Generates audio from the input text.
+
+        :param input: The text to generate audio for. The maximum length is 4096 characters.
```
qinxuye, line 696: Could you use the same comment style as the other functions?

codingl2k1: OK

qinxuye (approved):
LGTM

Issue #1596, opened 2024-06-06
A BUG when use qwen with stream and tools.


idiotTest:
### Describe the bug
When i use tools and stream in qwen vllm,I found the result was truncated. I mean many results lost.
### To Reproduce
I depoly the model ,which is qwen1.5-14b-chat,and i use the vllm.
```shell
curl --location 'http://124.70.213.108:7009/v1/chat/completions' \
--header 'Content-Type: application/json' \
--data '{
    "model": "qwen1.5-14b-chat",
    "messages": [
        {
            "content": "\n            你是一位智能助手，请通过你自己的知识和调用工具解决问题,全部回答请用中文,如果可以调用工具解决，请调用工具，如果不可以则用自己的理解回答。\n            ",
            "role": "system"
        },
        {
            "content": "中国有几个省份",
            "role": "user"
        }
    ],
    "tools": [
        {
            "type": "function",
            "function": {
                "name": "search_wheather",
                "description": "search_wheather(input: str) - 当你需要对某个城市的天气进行查询时,使用这个工具",
                "parameters": {
                    "type": "object",
                    "properties": {
                        "input": {
                            "type": "string"
                        }
                    },
                    "required": [
                        "input"
                    ]
                }
            }
        }
    ],
    "stream": true,
    "n": 1,
    "temperature": 0.0
}' 
```
### Expected behavior
The model stream result seems to be truncated.Only the last a few message received.

DCJsenior:
I have also been troubled by this bug for a long time when I use langchain AgentExecutor. I hope the developers can solve it, I would appreciate it if it can be solved.

zhanghx0905:
For streaming responses to qwen's tool calls, xinference's current implementation performs some filtering operations, in order to filter out the ReAct "thought" process. This operation seems somewhat problematic. Maybe the filtering step should be removed.

idiotTest:
> For streaming responses to qwen's tool calls, xinference's current implementation performs some filtering operations, in order to filter out the ReAct "thought" process. This operation seems somewhat problematic. Maybe the filtering step should be removed.

Thinks for your reply.
I want to test your  methods，But because of some reason，not available for now.
Did you test question and your idea?

Jiehuishun:
> For streaming responses to qwen's tool calls, xinference's current implementation performs some filtering operations, in order to filter out the ReAct "thought" process. This operation seems somewhat problematic. Maybe the filtering step should be removed.

[...]

PR #1598, opened 2024-06-06
BUG: Filtering Step for Streaming Responses to Qwen's Tool Calls when using vLLM

zhanghx0905:
In the current implementation of xinference, a filtering operation is performed on streaming responses to Qwen's tool calls to filter out the ReAct "thought" process. This operation appears to be problematic and may be causing issues with the response stream.

https://github.com/xorbitsai/inference/issues/1596#issue-2338176418
https://github.com/labring/FastGPT/issues/1668#issue-2330074252

Suggested Fix:
- Redesign the filtering step for streaming responses to Qwen's tool calls to ensure accurate response delivery.


qinxuye:
Will this cause other errors?

qinxuye:
@codingl2k1 Can you look at this PR?

DCJsenior:
Have you tested your modification？Can it works and no other side-effect?

zhanghx0905:
> Have you tested your modification？Can it works and no other side-effect?


No other parts call the removed _tools_token_filter function. I am certain that changes are made only when qwen, streaming, vllm, and tools are all enabled at the same time.

However, I am not sure if this PR will effectively address both issues. Perhaps removing the filtering step is not a good choice. I will upload the screenshot showing the effect of this PR later.

Jiehuishun:
Having Tested this bug fix. Maybe not prefect, but enough! Hope maintainer review it asap.

zhanghx0905:
image

xinference/model/llm/utils.py
```diff
@@ -670,35 +670,6 @@ def _eval_tool_arguments(cls, model_family, c, tools):
...
-            # Encapsulating function to reset 'found' after each call
-            found = False
-
-            def process_token(tokens: str):
-                nonlocal found
-                # Once "Final Answer:" is found, future tokens are allowed.
-                if found:
-                    return True
-                # Check if the token ends with "\nFinal Answer:" and update `found`.
-                if tokens.endswith("\nFinal Answer:"):
```
codingl2k1, line 693: This part similar to the official `openai_api.py`: https://github.com/QwenLM/Qwen/blob/main/openai_api.py#L346 Are you sure it's OK to remove it?

GabrielXie:
It seems that this preserves the ReAct "thought" process, but this is not user-friendly. Is it possible to use the method of the official Qwen function call?
https://qwen.readthedocs.io/zh-cn/latest/framework/function_call.html

zhanghx0905:
> It seems that this preserves the ReAct "thought" process, but this is not user-friendly. Is it possible to use the method of the official Qwen function call? https://qwen.readthedocs.io/zh-cn/latest/framework/function_call.html

Yeah. This will be much better.

xinference/model/llm/utils.py
zhanghx0905: > This part is similar to the official `openai_api.py`: https://github.com/QwenLM/Qwen/blob/main/openai_api.py#L346 Are you sure it's OK to remove it?

Can Qwen offical `openai_api.py` support tools with stream response? I'm not sure.

DCJsenior:
Now that the output is a mix of thought, action, actionInput, and finalAnswer, can we just return finalAnswer and decide where the code is for these.

zhanghx0905:
> Now that the output is a mix of thought, action, actionInput, and finalAnswer, can we just return finalAnswer and decide where the code is for these.

I hope so too, the problem is that I don't know why _tools_token_filter doesn't work correctly. 

https://github.com/xorbitsai/inference/blob/1ea34120ddea40ccc93cf576f3bc82fc6aae2070/xinference/model/llm/utils.py#L674

If you know how to fix it, please give me some tips or submit a new pr.

xinference/model/llm/utils.py
codingl2k1: From the qwen official `openai_api.py`: https://github.com/QwenLM/Qwen/blob/main/openai_api.py#L410 It says that function calling does not work in stream mode.

codingl2k1: image

codingl2k1: Is it possible that the `\nFinal Answer:` is not in one token?

codingl2k1: For example, `Final` is token a,  `Answer` is token b. If this is True, we can use a two-tokens buffer to check the  `Final Answer` string.

zhanghx0905: After debugging, I realized that the reason for the error was most likely that the model was outputting `Final answer` instead of `Final Answer`.

zhanghx0905: vllm's LLMEngine doesn't return tokens one by one, it may return a bunch of tokens at a time

zhanghx0905: I committed the new implementation, which should fix the bug and keep the filter. @codingl2k1

codingl2k1: Nice work! Could you add a test for it?

zhanghx0905:
I've added a test to test the effect of tools streaming calls. But I've also found it difficult to fully validate the correctness of this PR, as the bug happens randomly.

codingl2k1 (approved):
LGTM

okwinds:
> It seems that this preserves the ReAct "thought" process, but this is not user-friendly. Is it possible to use the method of the official Qwen function call? https://qwen.readthedocs.io/zh-cn/latest/framework/function_call.html


推理引擎应该只做推理引擎该做的事情，对于用户体验是否友好的判断，这是业务领域的决策。这样的逻辑，在架构上，不应该放在模型层的推理引擎去帮业务做决定，这样的设计在分域治理的架构上是不可取，希望官方谨慎处理这样的推理逻辑。可能会因为穿透性的给最终用户“良好体验”，而给更多的推理业务带来风险。

The inference engine should only do what an inference engine is supposed to do. The judgment of whether the user experience is friendly is a decision in the business domain. Such logic should not be placed in the model layer's inference engine to make decisions for the business in terms of architecture. This kind of design is undesirable in a domain-oriented governance architecture. It is hoped that the official will handle such inference logic with caution. It may bring risks to more inference businesses because it penetrates to provide a "good experience" to the end users.

PR #1610, opened 2024-06-10
ENH: quantization for glm-4v

Minamiyama:
fix: #1609

### DIFF
--- a/xinference/model/llm/pytorch/glm4v.py
+++ b/xinference/model/llm/pytorch/glm4v.py
@@ -62,7 +62,7 @@ class Glm4VModel(PytorchChatModel):
         device = self._pytorch_model_config.get("device", "auto")
         self._device = select_device(device)
 
-        kwargs = { "device_map": self._device }
+        kwargs = {"device_map": self._device}
         quantization = self.quantization
         if quantization != "none":
             if self._device == "cuda" and self._is_linux():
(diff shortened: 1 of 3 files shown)


xinference/model/llm/pytorch/glm4v.py
```diff
@@ -56,19 +56,29 @@ def match(
...
+        kwargs = {"device_map": self._device}
+        quantization = self.quantization
+        if quantization != "none":
+            if self._device == "cuda" and self._is_linux():
+                kwargs["device_map"] = "auto"
+                self._device = "auto"
+                if quantization == "4-bit":
+                    kwargs["load_in_4bit"] = True
+                elif quantization == "8-bit":
+                    kwargs["load_in_8bit"] = True
```
codingl2k1, line 74: In the else block, we failed to load the quantized model silently. Could you please add some logs to provide more information about the error?

Minamiyama: > In the else block, we failed to load the quantized model silently. Could you please add some logs to provide more information about the error?

4v-4bit
4v-8bit

I can launch quantized modes successfully from web ui

ChengjieLi28: @Minamiyama What @codingl2k1 means is that If your environment is not `self._device == "cuda" and self._is_linux()`, when the users choose `4-bit` or `8-bit`, the model is still loaded via `none` quantization, which may lead to misunderstanding. You should add a log or just raise error to show this case.

amumu96 (approved):

PR #1621, opened 2024-06-12
ENH: Supports `generate` interface for continuous batching

ChengjieLi28:
1. Support `generate` interface for continuous batching
2. Fix streaming chunk id issue. All the chunk id should be same
See: https://platform.openai.com/docs/api-reference/chat/streaming


qinxuye (approved):
LGTM

PR #1627, opened 2024-06-12
FEAT: Added a method to download models from csghub. 

hainaweiben:
This can be utilized by setting the environment variable `XINFERENCE_MODEL_SRC="csghub"`. Additionally, you need to set `XINFERENCE_CSG_TOKEN` to your csghub login token. Please note that currently, the qwen2 model library on csghub has been removed and is temporarily unavailable for download.


qinxuye (approved):
LGTM

qinxuye (approved):
LGTM

KinglyWayne:
Maybe we should change csghub endpoint from "https://hub-stg.opencsg.com" to "https://hub.opencsg.com" for production usage

qinxuye:
> Maybe we should change csghub endpoint from "https://hub-stg.opencsg.com" to "https://hub.opencsg.com" for production usage

Could you open a PR to fix it?

I think it could be a environment as well, we can use production as default.

hainaweiben:
> Maybe we should change csghub endpoint from "https://hub-stg.opencsg.com" to "https://hub.opencsg.com" for production usage

According to this document, the endpoint should be "https://hub-stg.opencsg.com/".

PR #1642, opened 2024-06-14
FEAT: Add Tools Support for Qwen Series MOE Models

zhanghx0905:
This PR adds tools support for the Qwen series Mixture of Experts (MOE) models. The Qwen Chat series MOE models possess similar tool call capabilities as other dense models, enhancing their utility and consistency.

### DIFF
--- a/xinference/model/llm/llm_family.json
+++ b/xinference/model/llm/llm_family.json
@@ -2290,7 +2290,8 @@
       "zh"
     ],
     "model_ability": [
-      "chat"
+      "chat",
+      "tools"
     ],
     "model_description": "Qwen1.5-MoE is a transformer-based MoE decoder-only language model pretrained on a large amount of data.",
     "model_specs": [
@@ -2595,7 +2596,8 @@
       "zh"
     ],
     "model_ability": [
-      "chat"
+      "chat",
+      "tools"
     ],
     "model_description": "Qwen2 is the new series of Qwen large language models. ",
     "model_specs": [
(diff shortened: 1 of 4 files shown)


xinference/model/llm/llm_family.json
```diff
@@ -2290,7 +2290,8 @@
       "zh"
     ],
     "model_ability": [
-      "chat"
```
qinxuye, line 2293: Can you add the tools to `llm_family_modelscope.json`?

zhanghx0905:
Also, the size of the qwen MOE model is incorrectly labeled; the actual number of parameters for Qwen1.5-MoE-A2.7B-Chat is 14B, whereas the actual number of parameters for Qwen2-57B-A14B is 57B.

qinxuye:
> Also, the size of the qwen MOE model is incorrectly labeled; the actual number of parameters for Qwen1.5-MoE-A2.7B-Chat is 14B, whereas the actual number of parameters for Qwen2-57B-A14B is 57B.

That depends on how the size is defined, we’d prefer to treat it as activated parameters size.

zhanghx0905:
> > Also, the size of the qwen MOE model is incorrectly labeled; the actual number of parameters for Qwen1.5-MoE-A2.7B-Chat is 14B, whereas the actual number of parameters for Qwen2-57B-A14B is 57B.
> 
> That depends on how the size is defined, we’d prefer to treat it as activated parameters size.

In this case, model memory calculation will also conduct according to the number of active parameters. The memory estimation for MOE models will be wrong.

qinxuye:
Looks like #1648 added moe in it as well, can you look at it?

qinxuye (approved):
LGTM

PR #1645, opened 2024-06-14
BUG: show error when user launch quantized model without device supported

### LINKED ISSUE #1641: BUG: quantization for glm-4v
### Describe the bug
ENH: quantization for glm-4v #1610 

If environment is not self._device == "cuda" and self._is_linux(), when the users choose 4-bit or 8-bit, the model is still loaded via none quantization, this may lead to misunderstanding. we should add a log or just raise error to show this case.

we will fix at next release.

Minamiyama:
FIxes #1641 .


qinxuye (approved):
LGTM

PR #1659, opened 2024-06-17
ENH: Continuous batching supports all the models with `transformers` backend

ChengjieLi28:
Currently, three models are implemented separately for inference:
- internlm2-chat
- glm
- falcon

1. For falcon, after testing, there is no need to write a single function to handle its inference process. So I delete them.
2. For internlm2-chat, this model is sensitive to the inference parameters like `temperature`, `top_p`, `top_k`. We cannot set them by default if use does not pass these.
3. For glm4,  the inference process lacks the processing of `attention_mask` and `position_ids`, which are very important in glm4 inference. ~~However, glm4's batch inference seems to be problematical. See https://github.com/THUDM/GLM-4/issues/25~~
4. Currently, the `chatglm` series may have issues with duplicate outputs.

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -109,7 +109,7 @@ jobs:
             sudo rm -rf "/usr/local/share/boost"
             sudo rm -rf "$AGENT_TOOLSDIRECTORY"
           fi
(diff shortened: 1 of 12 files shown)


qinxuye (approved):
LGTM

PR #1666, opened 2024-06-18
ENH: Set the CSG Hub endpoint as an environment variable.

hainaweiben:
with the default value being: "https://hub-stg.opencsg.com/"

fix #1627


xinference/model/llm/llm_family.py
```diff
@@ -691,7 +692,9 @@ def cache_from_csghub(
                 "model_format": llm_spec.model_format,
             },
             llm_spec.model_id,
-            endpoint="https://hub-stg.opencsg.com",
+            endpoint=os.environ.get(
+                XINFERENCE_ENV_CSG_ENDPOINT, "https://hub-stg.opencsg.com/"
```
qinxuye, line 696: Shall it be `https://hub.opencsg.com/` by default?

xinference/model/llm/llm_family.py
```diff
@@ -711,7 +714,9 @@ def cache_from_csghub(
                 },
                 llm_spec.model_id,
                 file_name=filename,
-                endpoint="https://hub-stg.opencsg.com",
+                endpoint=os.environ.get(
+                    XINFERENCE_ENV_CSG_ENDPOINT, "https://hub-stg.opencsg.com/"
```
qinxuye, line 718: Since the endpoint is used twice, we can add an variable to `constants.py`, for instance, `CSG_ENDPOINT`.

xinference/model/llm/llm_family.py
hainaweiben: According to this [document](https://git-devops.opencsg.com/csg-hub-sdk/csghub-sdk
), the endpoint should be "https://hub-stg.opencsg.com/".

qinxuye: But according to https://github.com/xorbitsai/inference/pull/1627#issuecomment-2172112270 they said the stg is not for production.

hainaweiben: I have tested and it should be using "https://hub-stg.opencsg.com/".

qinxuye (approved):
LGTM

PR #1671, opened 2024-06-19
FEAT: Add support for Flexible Model

shellc:
Flexible Model is a customizable inference pipeline within xinference, supporting a range of tasks including image segmentation, traditional machine learning regression, and more.

Mechanism:

Create the model object with custom definitions and initialization functions.
Use xinference.model.flexible.launchers.transformers to load transformer models.
Invoke the model via the /v1/flexible/infers endpoint.
Access the model using the SDK’s infer interface.

Examples:
```
from xinference.client import Client

client = Client("http://localhost:9997")

model = client.get_model("flexible-model")

result = model.infer(inputs="happy")
```

image


qinxuye:
@yiboyasss Could you review the frontend part of this PR?

xinference/model/flexible/core.py
```diff
@@ -0,0 +1,223 @@
...
+        """
+        raise NotImplementedError("infer method not implemented.")
+
+    @property
+    def model_uid(self):
+        return self._model_uid
+
+    @property
+    def model_path(self):
+        return self.model_path
```
qinxuye, line 190:
```suggestion
        return self._model_path
```

xinference/model/flexible/tests/test_flexible_models.py
```diff
@@ -0,0 +1,63 @@
...
+            model_uri="mock",
+            launcher="xinference.model.flexible.launchers.transformers",
+        ),
+        task="mock",
+    )
+
+    model.load()
+
+    result = model.infer(inputs="hello world")
+    # assert result == {"inputs": "hello world"}
```
qinxuye, line 60: Could remove this line.

qinxuye: And the result is not tested?

qinxuye (commented):
I left some comments

xinference/api/restful_api.py
```diff
@@ -1344,6 +1354,40 @@ async def create_variations(
...
+            raise HTTPException(status_code=500, detail=str(e))
+
+        try:
+            result = await model.infer(**kwargs)
+            return Response(result, media_type="application/json")
+        except RuntimeError as re:
+            logger.error(re, exc_info=True)
+            await self._report_error_event(model_uid, str(re))
+            self.handle_request_limit_error(re)
+            raise HTTPException(status_code=400, detail=str(re))
```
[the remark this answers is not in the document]
shellc: Maintain consistency with other APIs, such as create_embedding, create_rerank, etc.

The RuntimeError may stem from a user request that triggers a manageable runtime error. It is advisable to address this issue uniformly in future implementations.

xinference/model/flexible/core.py
shellc: fix it

xinference/model/flexible/tests/test_flexible_models.py
shellc: There are Several assertions after this line.

This is a MockModel and cannot perform actual computations. The behavior of this model is to echo the inputs.
The purpose of the tests here is to evaluate the initialization and invocation of the model.

yiboyasss:
1. Input boxes can undergo some JSON format validation to ensure that correct JSON data can be passed.
截屏2024-07-09 17 15 51


2. Add a check between lines 75 and 190 in the registerModel.js file to ensure the correct echo of the editing mode data.
截屏2024-07-09 17 17 04

yiboyasss:
BUG: When the registered model name already exists, an error message appears when registering again, but the previously registered model with the same name will also disappear, which is not reasonable. Example: When a model with the name 'flexible-model' already exists, registering the same name again will cause the previous' flexible-model 'model to disappear.

qinxuye (approved):
LGTM, thanks for your contribution.

liyingya-123:
怎么通过curl来访问这个部署呢？

PR #1680, opened 2024-06-20
ENH: Add guard for model launching

frostyplanet:
```
    worker: Add guard for model launching
    
    Because model launching is a long process (download model, loading into GPU).
    client might encounter network error in the middle while worker is processing,
    add a guard the prevent duplicate operation with the same model_uid.
    
    Provide an rpc call check_model_launch_status() to return LuanchStatus,
    to determine whether worker is still working on this model_uid.
```

### DIFF
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -73,6 +73,9 @@ class WorkerActor(xo.StatelessActor):
         self._main_pool.recover_sub_pool = self.recover_sub_pool
 
         # internal states.
+        # temporary placeholder during model launch process:
+        self._model_uid_launching_guard: Dict[str, bool] = {}
+        # attributes maintained after model launched:
         self._model_uid_to_model: Dict[str, xo.ActorRefType["ModelActor"]] = {}
         self._model_uid_to_model_spec: Dict[str, ModelDescription] = {}
         self._gpu_to_model_uid: Dict[int, str] = {}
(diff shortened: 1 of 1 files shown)


xinference/core/worker.py
```diff
@@ -594,10 +597,14 @@ async def launch_builtin_model(
         launch_args.pop("kwargs")
         launch_args.update(kwargs)
 
-        event_model_uid, _, __ = parse_replica_model_uid(model_uid)
+        try:
+            origin_uid, _, _ = parse_replica_model_uid(model_uid)
+        except Exception as e:
+            logger.exception(e)
```
qinxuye, line 603: This seems useless.

xinference/core/worker.py
```diff
@@ -640,50 +647,55 @@ async def launch_builtin_model(
         assert model_uid not in self._model_uid_to_model
         self._check_model_is_valid(model_name, model_format)
 
-        subpool_address, devices = await self._create_subpool(
-            model_uid, model_type, n_gpu=n_gpu, gpu_idx=gpu_idx
-        )
+        if self.check_model_launch_status(model_uid) is not None:
```
qinxuye, line 650: `check` seems an action without return value. Maybe can call it get?

xinference/core/worker.py
```diff
@@ -694,10 +706,13 @@ async def launch_builtin_model(
 
     @log_async(logger=logger)
     async def terminate_model(self, model_uid: str):
-        event_model_uid, _, __ = parse_replica_model_uid(model_uid)
+        # Terminate model while its launching is not allow
+        if model_uid in self._model_uid_launching_guard:
+            raise ValueError(f"{model_uid} is launching")
+        origin_uid, _, __ = parse_replica_model_uid(model_uid)
```
qinxuye, line 712: Why changed the name of the variable?

qinxuye (commented):
I left some comments.

xinference/core/worker.py
frostyplanet: Just in case parse wrong replica_model_uid causing exception,  @log_async does not log exception

xinference/core/worker.py
frostyplanet: there were duplicated calls, just unify the name
```
event_model_uid, _, __ = parse_replica_model_uid(model_uid)
...
origin_uid, _, __ = parse_replica_model_uid(model_uid)
 ```

xinference/core/worker.py
frostyplanet: OK

qinxuye (approved):
LGTM

Issue #1685, opened 2024-06-21
It seems that this preserves the ReAct "thought" process, but this is not user-friendly. Is it possible to use the method of the official Qwen function call?


okwinds:
It seems that this preserves the ReAct "thought" process, but this is not user-friendly. Is it possible to use the method of the official Qwen function call?
https://qwen.readthedocs.io/zh-cn/latest/framework/function_call.html

_Originally posted by @GabrielXie in https://github.com/xorbitsai/inference/issues/1598#issuecomment-2159605576_

https://github.com/xorbitsai/inference/pull/1598
推理引擎应该只做推理引擎该做的事情，对于用户体验是否友好的判断，这是业务领域的决策。官方的Qwen function call的方法，不应该隐藏react “thought”的推理过程。隐藏“thought”信息的逻辑，这在架构上，不应该放在模型层的推理引擎去帮业务做决定。这样的设计在分域治理的架构上是不可取的，希望官方谨慎考虑处理这样的推理逻辑。这可能会因为穿透性的给最终用户“良好体验”，反而给更多的推理业务带来风险和困扰。这在架构上应该解耦业务逻辑。

The inference engine should stick to its own duties, and the judgment of whether the user experience is friendly should be a decision made in the business domain. The official method of calling the Qwen function should not obscure the reasoning process of the "thought" in React. The logic of concealing "thought" information should not be incorporated into the model layer's inference engine to assist the business in making decisions. Such a design is undesirable in the architecture of domain-based governance, and it is hoped that the official will carefully consider the handling of such reasoning logic. This could potentially lead to a "good experience" for the end user through penetration, but in turn, it may bring risks and troubles to more inference operations.

This should decouple the business logic in the architecture.

zhanghx0905:
I think it's useful to use a global variable to control whether the feature takes effect.

zhanghx0905:
take

okwinds:
> I think it will be useful to have a global variable to control whether the feature takes effect.

这也许是一个折中妥协的解决方案。基于你所提出的方案我能给出的建议是，默认显示react的“thought”内容，通过变量配置选择隐藏，以适配一些特殊需要。谢谢你的工作

Yep, this might be a compromise solution. Based on the proposal you have put forward, my suggestion is to display the "thought" content of React by default, and provide the option to hide it through variable configuration to accommodate some special needs. Thanks a lot for ur work.

okwinds:
[...]

PR #1717, opened 2024-06-25
ENH: Add Environment Variable to Disable QWEN vllm Tools Response Filter

zhanghx0905:
This PR introduces a new environment variable, XINFERENCE_DISABLE_QWEN_REACT_RESPONSE_FILTER, to disable the QWEN vllm's tool call response filtering mechanism.

Xinference will by default enable the QWEN vllm's tool call ReAct response filtering.
Users will only see what comes after the "Final Answer" in QWEN's ReAct response.

Setting this environment to 1 will disable the response filtering mechanism, 
allowing raw responses from QWEN vllm tools to be returned without additional processing.

To disable filtering:

```sh
export XINFERENCE_DISABLE_QWEN_REACT_RESPONSE_FILTER=1
```

#1685

image


fyrestone:
Does this work well with continuous chat, as all the output will be contained in the next chat messages? For example, chat with the previous tool call result.

codingl2k1:
Does this work well with continuous chat, as all the output will be contained in the next chat messages? For example, chat with the previous tool call result.

zhanghx0905:
> Does this work well with continuous chat, as all the output will be contained in the next chat messages? For example, chat with the previous tool call result.

You're right, enabling this option will cause more information to be included in the history. It does have the potential to affect subsequent tool calls, especially for smaller models.

But as suggested in the cited issue, users can choose how to use chat history on the client side, and they can implement their own filtering policies.

zhanghx0905:
I suddenly realized there are actually multiple scenarios to consider. For example, what kind of reasoning backend to use (vllm or not) and whether to enable stream response or not. My PR only considered the case where vllm is used and stream is enabled so it shouldn't be merged now.

PR #1719, opened 2024-06-26
FEAT: Delete cluster

hainaweiben:
Use the command 'xinference stop-cluster' to exit the cluster.


xinference/api/restful_api.py
```diff
@@ -552,6 +552,36 @@ def serve(self, logging_conf: Optional[dict] = None):
...
+                if self.is_authenticated()
+                else None
+            ),
+        )
+        self._router.add_api_route(
+            "/v1/clusters",
+            self.abort_cluster,
+            methods=["DELETE"],
+            dependencies=(
+                [Security(self._auth_service, scopes=["clusters:list"])]
```
ChengjieLi28, line 580: Just `admin` scopes for these three interfaces. We do not define these three scopes for now. Users may not understand this.

amumu96 (approved):

PR #1724, opened 2024-06-27
ENH: Continuous batching supports `vision` model ability

ChengjieLi28:
Refactor impl.

Currently supports three multimodal models:
- qwen-vl-chat
- cogvlm2
- glm-4v


qinxuye (approved):
LGTM

PR #1771, opened 2024-07-03
ENH: Add more log modules

ChengjieLi28:
1. for uvicorn log, the console would not show these logs while the file would remain these.
2. for transformers and vllm log, just remain them on console and file.


qinxuye (approved):
LGTM

PR #1844, opened 2024-07-11
REF: Remove `chatglm-cpp` and Fix latest `llama-cpp-python` issue

ChengjieLi28:
1. Remove `chatglm-cpp` support.
2. Remove `chatglm`, `chatglm2`, `chatglm3` gglmv3 models.
3. Add some glm4-chat gguf models.
4. Remove `create_embedding` task for LLM models.
5. Compatible with the latest `llama-cpp-python`
  - GPU docker can have the latest version
  - CPU docker should remain `v0.2.77` version, since it has import issue for the latest version.


qinxuye (approved):
LGTM

PR #1994, opened 2024-08-01
FEAT: Support gte-Qwen2-7B-instruct and multi gpu deploy

amumu96:


xinference/model/embedding/core.py
```diff
@@ -281,12 +322,21 @@ def encode(
 
             return all_embeddings, all_token_nums
 
-        all_embeddings, all_token_nums = encode(
```
qinxuye, line 284: Is it possible to use model name?

qinxuye (approved):
LGTM

Issue #1995, opened 2024-08-01
TTS curl Internal Server Error


tsiens:
### System Info / 系統信息

cuda12.1

### Running Xinference with Docker? / 是否使用 Docker 运行 Xinfernece？

- [X] docker / docker
- [ ] pip install / 通过 pip install 安装
- [ ] installation from source / 从源码安装

### Version info / 版本信息

0.13.3

### The command used to start Xinference / 用以启动 xinference 的命令

docker

### Reproduction / 复现过程

Xinference python success
but curl error
```
curl -X 'POST' \
>   'http://127.0.0.1:9997/v1/audio/speech' \
>   -H 'accept: application/json' \
>   -H 'Content-Type: application/json' \
>   -d '{
>     "model": "ChatTTS",
>     "text": "你好",
>     "voice": "中文女"
>   }'
Internal Server Error
```
```
curl -X 'POST' \
>   'http://127.0.0.1:9997/v1/audio/speech' \
>   -H 'accept: application/json' \
>   -H 'Content-Type: application/json' \
>   -d '{
>     "model": "CosyVoice-300M-SFT",
>     "text": "你好",
>     "voice": "中文女"
>   }'
Internal Server Error
```

### Expected behavior / 期待表现

return result as 
```
curl -X 'POST' \
>   'http://127.0.0.1:9997/v1/chat/completions' \
>   -H 'accept: application/json' \
>   -H 'Content-Type: application/json' \
>   -d '{
>     "model": "qwen2-instruct",
>     "messages": [
>         {
>             "role": "system",
>             "content": "You are a helpful assistant."
>         },
>         {
>             "role": "user",
>             "content": "What is the largest animal?"
>         }
>     ],
>     "max_tokens": 512,
>     "temperature": 0.7
>   }'
[...]

PR #2009, opened 2024-08-03
ENH: Improve internal server error

codingl2k1:
For 500 internal server error:
- Response error string.
- Log error traceback.

Fix doc.

Fixes: https://github.com/xorbitsai/inference/issues/1995

### DIFF
--- a/doc/source/models/model_abilities/audio.rst
+++ b/doc/source/models/model_abilities/audio.rst
@@ -245,7 +245,7 @@ Basic usage, launch model ``CosyVoice-300M-SFT``.
       -H 'Content-Type: application/json' \
       -d '{
         "model": "<MODEL_UID>",
-        "text": "<The text to generate audio for>",
+        "input": "<The text to generate audio for>",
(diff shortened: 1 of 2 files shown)


doc/source/models/model_abilities/audio.rst
```diff
@@ -245,7 +245,7 @@ Basic usage, launch model ``CosyVoice-300M-SFT``.
       -H 'Content-Type: application/json' \
       -d '{
         "model": "<MODEL_UID>",
-        "text": "<The text to generate audio for>",
+        "input": "<The text to generate audio for>",
```
qinxuye, line 248: I copied the curl above: https://inference.readthedocs.io/en/latest/models/model_abilities/audio.html#speech

It might need to be modified as well?

xinference/api/restful_api.py
```diff
@@ -1320,6 +1327,7 @@ async def create_speech(
             None, media_type="application/octet-stream"
         ),
     ) -> Response:
+        print("create_speech")
```
qinxuye, line 1330: Remove print.

qinxuye (approved):
LGTM

PR #2025, opened 2024-08-06
FEAT: Add support for llama-3.1-instruct 405B model

frostyplanet:
Add support for llama-3.1-instruct 405B model


qinxuye (approved):
LGTM

PR #2043, opened 2024-08-08
REF: Mark `Deprecate` for `prompt`, `system_prompt` and `chat_history` parameters in `chat` client interface

ChengjieLi28:
image


qinxuye (approved):
LGTM

PR #2079, opened 2024-08-13
Feat: Support internvl2 and internvl stream

amumu96:


xinference/model/llm/pytorch/intern_vl.py
```diff
@@ -346,7 +363,121 @@ def chat(
...
+                return_history=False,
+                generation_config=generation_config,
             ),
         )
-        return self._to_chat_completion(chunk)
+        thread.start()
+
+        completion_id = str(uuid.uuid1())
+        prompt_tokens, completion_tokens, total_tokens = 0, 0, 0
+        prompt_tokens = self._get_input_tokens(content, history, num_patches_list)
```
qinxuye, line 409: Is it possible that we generate the prompt, and turn it into token ids, so we can calculate the prompt_tokens, the benefit is that we don't need to tokenize twice.

xinference/model/llm/vllm/core.py
```diff
@@ -28,6 +29,8 @@
     Union,
 )
 
+from transformers import AutoTokenizer
```
qinxuye, line 32: `transformers` is an optional dependency, so we may need to put the import into function.

xinference/model/llm/vllm/core.py
```diff
@@ -520,6 +527,11 @@ class VLLMChatModel(VLLMModel, ChatModelMixin):
     def match(
         cls, llm_family: "LLMFamilyV1", llm_spec: "LLMSpecV1", quantization: str
     ) -> bool:
+        if (
+            llm_family.model_name in VLLM_SUPPORTED_VISION_MODEL_LIST
```
qinxuye, line 531: Will internvl be matched by chat model?

xinference/model/llm/vllm/core.py
```diff
@@ -606,3 +618,133 @@ async def async_chat(
...
+        logger.info(
+            f"Loading {self.model_uid} with following model config: {self._model_config}"
+        )
+
+        engine_args = AsyncEngineArgs(
+            model=self.model_path,
+            **self._model_config,
+        )
+        self._engine = AsyncLLMEngine.from_engine_args(engine_args)
+        self._tokenizer = AutoTokenizer.from_pretrained(
```
qinxuye, line 678: Only add a tokenizer beyond the VLLMModel? Why not calling the base load method first?

xinference/model/llm/vllm/core.py
```diff
@@ -606,3 +618,133 @@ async def async_chat(
...
+        if llm_spec.model_format == "pytorch":
+            if quantization != "none" and not (quantization is None):
+                return False
+        if isinstance(llm_family, CustomLLMFamilyV1):
+            if llm_family.model_family not in VLLM_SUPPORTED_VISION_MODEL_LIST:
+                return False
+        else:
+            if llm_family.model_name not in VLLM_SUPPORTED_VISION_MODEL_LIST:
+                return False
+        if "chat" not in llm_family.model_ability:
```
qinxuye, line 697: Shall we consider the `vision` ability?

xinference/model/llm/vllm/core.py
```diff
@@ -606,3 +618,133 @@ async def async_chat(
...
+        system_prompt: Optional[str] = None,
+        chat_history: Optional[List[ChatCompletionMessage]] = None,
+        generate_config: Optional[Dict] = None,
+    ) -> Union[ChatCompletion, AsyncGenerator[ChatCompletionChunk, None]]:
+        # only support single image, waiting vllm support multi images
+        assert self.model_family.prompt_style is not None
+        prompt_style = self.model_family.prompt_style.copy()
+        chat_history = chat_history or []
+        messages, images = self.get_prompt(prompt, chat_history, prompt_style)
+        prompt = self._tokenizer.apply_chat_template(  # type: ignore
```
qinxuye, line 727: The prompt is not generated by us?

Issue #2074, opened 2024-08-13
glm4 function调用报错


Alex-Zuo-One:
### System Info / 系統信息

推理框架transformers

### Running Xinference with Docker? / 是否使用 Docker 运行 Xinfernece？

- [X] docker / docker
- [ ] pip install / 通过 pip install 安装
- [ ] installation from source / 从源码安装

### Version info / 版本信息

14.0

### The command used to start Xinference / 用以启动 xinference 的命令

docker run

### Reproduction / 复现过程

传参:
```
{
    "model": "glm-4-9b-chat",
    "messages": [
      {
        "role": "system",
        "content": "你是一个小助手"
      },
      {
        "role": "user",
        "content": "北京的天气怎么样"
      }
    ],
    "tools": [
      {
        "type": "function",
        "function": {
          "name": "find_sports_shoes",
          "description": "查找商场内可用的运动鞋信息",
          "parameters": {
            "type": "object",
            "properties": {
              "brand": {
                "type": "string",
                "description": "鞋子的品牌"
              },
              "size": {
                "type": "string",
                "description": "鞋子的尺码"
              }
            },
            "required": ["brand"]
          },
          "strict": false
        }
      }
    ],
    "tool_choice": "auto",
    "stream": false
}
```
返回:
```
{
    "detail": "[address=0.0.0.0:45971, pid=242] 'type'"
}
```

报错:
```
2024-08-13 04:04:15,718 xinference.api.restful_api 155 ERROR    [address=0.0.0.0:45971, pid=242] 'type'
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/xinference/api/restful_api.py", line 1710, in create_chat_completion
    data = await model.chat(
  File "/usr/local/lib/python3.10/dist-packages/xoscar/backends/context.py", line 231, in send
    return self._process_result_message(result)
  File "/usr/local/lib/python3.10/dist-packages/xoscar/backends/context.py", line 102, in _process_result_message
    raise message.as_instanceof_cause()
  File "/usr/local/lib/python3.10/dist-packages/xoscar/backends/pool.py", line 656, in send
    result = await self._run_coro(message.message_id, coro)
  File "/usr/local/lib/python3.10/dist-packages/xoscar/backends/pool.py", line 367, in _run_coro
    return await coro
  File "/usr/local/lib/python3.10/dist-packages/xoscar/api.py", line 384, in __on_receive__
    return await super().__on_receive__(message)  # type: ignore
[...]

Issue #2055, opened 2024-08-09
v0.14.0.post1 glm-4-9b-chat (as custom model) functioncall not working


hujunwei:
### System Info / 系統信息

v0.14.0.post1

### Running Xinference with Docker? / 是否使用 Docker 运行 Xinfernece？

- [X] docker / docker
- [ ] pip install / 通过 pip install 安装
- [ ] installation from source / 从源码安装

### Version info / 版本信息

- v0.14.0.post1
- glm4-9b-chat 注册为custom model （离线环境，需从路径加载）

```json
{
  "version": 1,
  "context_length": 128000,
  "model_name": "glm-4-9b-chat",
  "model_lang": [
      "en",
      "zh"
  ],
  "model_ability": [
      "chat",
      "tools"
  ],
  "model_description": "glm-4-9b-chat",
  "model_family": "glm4-chat",
  "model_specs": [
      {
          "model_format": "pytorch",
          "model_size_in_billions": 9,
          "quantizations": [
              "none"
          ],
          "model_id": null,
          "model_hub": "huggingface",
          "model_uri": "/root/.xinference/glm-4-9b-chat",
          "model_revision": null
      }
  ],
  "prompt_style": {
      "style_name": "CHATGLM3",
      "system_prompt": "",
      "roles": [
          "user",
          "assistant"
      ],
      "intra_message_sep": "",
      "inter_message_sep": "",
      "stop": [
          "<|endoftext|>",
          "<|user|>",
          "<|observation|>"
      ],
      "stop_token_ids": [
          151329,
          151336,
          151338
      ]
  },
  "is_builtin": false
}
```

### The command used to start Xinference / 用以启动 xinference 的命令

**docker 启动命令**

`docker run --name="xinference" -v /home/glm4-9b/model:/root/.xinference -e XINFERENCE_HOME=/root/.xinference -p 9997:9997 --gpus all xprobe/xinference:v0.14.0.post1 xinference-local -H 0.0.0.0 --log-level debug`

**launch 命令**

`docker exec <container-id> xinference launch --model-engine vLLM -n glm-4-9b-chat -f pytorch --quantization none --n-gpu auto --replica 1 --dtype half --gpu_memory_utilization 0.9`

### Reproduction / 复现过程

**请求payload**

```
{
    "model": "glm-4-9b-chat",
    "messages": [
        {
            "role": "user",
            "content": "西安现在多少度?"
        }
    ],
    "tools": [
        {
            "type": "function",
[...]

PR #2081, opened 2024-08-13
BUG: Fix custom glm4 & remove tool calls of ChatGLM3

codingl2k1:
- Fix custom glm4 issue.
- Update glm4 hf to the latest, compatible with transformers 4.44
- Disable tool call for glm3.

Fixes: https://github.com/xorbitsai/inference/issues/2074 
Fixes: https://github.com/xorbitsai/inference/issues/2055


qinxuye (approved):
LGTM

Issue #2111, opened 2024-08-19
benchmark异步并行测试报错


zhangyanbo2007:
### System Info / 系統信息

内部使用引擎：vllm==0.5.3.post1

### Running Xinference with Docker? / 是否使用 Docker 运行 Xinfernece？

- [ ] docker / docker
- [X] pip install / 通过 pip install 安装
- [ ] installation from source / 从源码安装

### Version info / 版本信息

部署框架：xinference==0.14.0

### The command used to start Xinference / 用以启动 xinference 的命令

python benchmark/benchmark_serving.py --dataset xinference_project/data/ShareGPT_V3_unfiltered_cleaned_split.json --tokenizer /data/public/hub_repository/modelscope/hub/qwen/Qwen2-72B-Instruct  --num-prompt 100 --model-uid qwen--Qwen2-72B-Instruct --concurrency 50  --port 9999

### Reproduction / 复现过程

执行：
python benchmark/benchmark_serving.py --dataset xinference_project/data/ShareGPT_V3_unfiltered_cleaned_split.json --tokenizer /data/public/hub_repository/modelscope/hub/qwen/Qwen2-72B-Instruct  --num-prompt 100 --model-uid qwen--Qwen2-72B-Instruct --concurrency 50  --port 9999
结果：
python benchmark/benchmark_serving.py --dataset xinference_project/data/ShareGPT_V3_unfiltered_cleaned_split.json --tokenizer /data/public/hub_repository/modelscope/hub/qwen/Qwen2-72B-Instruct  --num-prompt 100 --model-uid qwen--Qwen2-72B-Instruct --concurrency 50
Namespace(host='localhost', port=9997, dataset='xinference_project/data/ShareGPT_V3_unfiltered_cleaned_split.json', tokenizer='/data/public/hub_repository/modelscope/hub/qwen/Qwen2-72B-Instruct', num_prompts=100, prompt_len_limit=1024, concurrency=50, request_rate=inf, seed=0, trust_remote_code=False, model_uid='qwen--Qwen2-72B-Instruct')
INFO:__main__:Preparing for benchmark.
Special tokens have been added in the vocabulary, make sure the associated word embeddings are fine-tuned or trained.
Token indices sequence length is longer than the specified maximum sequence length for this model (221043 > 131072). Running this sequence through the model will result in indexing errors
INFO:__main__:Benchmark starts.
Traceback (most recent call last):
[...]

PR #2113, opened 2024-08-19
BUG: fix asyncio.Queue error in benchmark

Dawnfz-Lenfeng:
- fix https://github.com/xorbitsai/inference/issues/2111
- Previously, `asyncio.Queue` was created in the `__init__` method rather than in the `run()` function. In lower versions of Python (e.g., 3.9), this could cause the queue to be attached to an incorrect event loop. 
- To address this issue, the creation of the queue has been moved to the warm_up method.
- Additionally, a minor error in a log message has been corrected.


qinxuye (approved):
LGTM

PR #2145, opened 2024-08-22
FEAT: Support LMDeploy for internvl2 and fix finish reasion miss at internvl stream

amumu96:


xinference/model/llm/lmdeploy/core.py
```diff
@@ -0,0 +1,554 @@
...
+        except ImportError:
+            error_message = "Failed to import module 'lmdeploy'"
+            installation_guide = [
+                "Please make sure 'lmdeploy' is installed. ",
+                "You can install it by `pip install lmdeploy`\n",
+            ]
+
+            raise ImportError(f"{error_message}\n\n{''.join(installation_guide)}")
+
+        chat_template_config = ChatTemplateConfig("internvl-internlm2")
```
qinxuye, line 166: This looks too specified for internvl only?

xinference/model/llm/llm_family_modelscope.json
```diff
@@ -4908,6 +4908,122 @@
         ]
     }
   },
+  {
```
qinxuye, line 4911: Seems duplicated.

xinference/model/llm/llm_family_modelscope.json
```diff
@@ -4778,7 +4778,7 @@
             "model_revision": "master"
         },
         {
-            "model_format": "pytorch",
+            "model_format": "awq",
             "model_size_in_billions": 2,
             "quantizations": [
               "none"
```
qinxuye, line 4784: Quantization of AWQ should be Int4.

xinference/model/llm/lmdeploy/core.py
```diff
@@ -0,0 +1,550 @@
...
+    def match(
+        cls, llm_family: "LLMFamilyV1", llm_spec: "LLMSpecV1", quantization: str
+    ) -> bool:
+        if llm_spec.model_format == "awq":
+            # Currently, only 4-bit weight quantization is supported for AWQ, but got 8 bits.
+            if "4" not in quantization:
+                return False
+        if llm_family.model_name not in LMDEPLOY_SUPPORTED_CHAT_MODELS:
+            return False
+        return True
```
qinxuye, line 184: Shall check lmdeploy installed?

amumu96: we check it when load func

qinxuye: But the users will find lmdeploy on the launch page, but actually it does not work. If match False, it will not show.

amumu96: ok, add `LMDEPLOY_INSTALLED` to check it

ChengjieLi28 (approved):

PR #2165, opened 2024-08-26
ENH: support padding for sd inpainting model

wxiwnd:


xinference/model/image/stable_diffusion/core.py
```diff
@@ -306,6 +310,17 @@ def inpainting(
             model = self._model
 
         width, height = map(int, re.split(r"[^\d]+", size))
+
+        if padding_image_to_multiple := kwargs.pop("padding_image_to_multiple", None):
+            # Model like SD3 image to image requires image's height and width is times of 16
```
qinxuye, line 315:
```suggestion
            # Model like SD3 inpainting requires image's height and width is times of 16
```

qinxuye (approved):
LGTM

PR #2172, opened 2024-08-27
ENH: Fix callback status when model die

frostyplanet:
Mark model status ERROR instead of TERMINATED after model die,
to distinguish with the behavior when model terminated by user.

由于 model die 的时候，先 terminate 再 re-launch,  用 TERMINATED 可能会在语义上与正常模型停止有冲突. 并且假如 model 多次 relaunch 失败放弃重试, 用 termianted 就不会在 supervisor 中留下状态. 所以用 ERROR 更好


frostyplanet:
Tested with sending kill signal to model process

qinxuye (approved):
LGTM

PR #2182, opened 2024-08-28
ENH: solve the problem of health check for image-to-text model

luhairong11:
solve the problem of health check for image-to-text model


qinxuye (approved):
LGTM

PR #2189, opened 2024-08-29
BUG: fix register model list error

amumu96:
Fix the bug that even if the registration fails, the model is still written to the registration list


qinxuye (approved):
LGTM

PR #2229, opened 2024-09-04
ENH: unpad for image2image/inpainting model

wxiwnd:
Padding for Image2Image model will cause unnecessary black edge in image output.


qinxuye (approved):
LGTM

PR #2231, opened 2024-09-05
CHORE: Compatible with `openai >= 1.40`

ChengjieLi28:


.github/workflows/python.yaml
```diff
@@ -142,7 +142,7 @@ jobs:
           MODULE: ${{ matrix.module }}
         run: |
           if [ "$MODULE" == "gpu" ]; then
-            ${{ env.SELF_HOST_PYTHON }} -m pip install -U "openai>1,<1.40"
```
qinxuye, line 145: How about set it to `openai>1.40`? or it won't update the openai library we may not find issue.

qinxuye (approved):
LGTM

PR #2249, opened 2024-09-08
ENH: Update CosyVoice Huggingface

codingl2k1:


qinxuye:
The tests changed to modelscope, we can change back now?

https://github.com/xorbitsai/inference/blob/e2618be96293f112709c9ceed639a3443455a0e7/xinference/model/audio/tests/test_cosyvoice.py#L26

qinxuye (approved):
LGTM

PR #2290, opened 2024-09-12
BUG: fix issue with model launch failing when .safetensors file is missing (#2094)

Charmnut:
**Reason:**
The `openai/whisper` model provides weights in the `.safetensors` format, but the `xinference`-supported `BELLE-2/Belle-whisper-large-v3-zh` model only provides weights in the `pytorch_model.bin` format. Setting `use_safetensors=True` causes errors when running the `Belle-whisper-large-v3-zh` model.

**Solution:**
Added a check to determine if `.safetensors` files are present in the model directory.

**Effect:**
Both `whisper` and `Belle-whisper` models now launch and inference correctly on our company's private server.

### DIFF
--- a/xinference/model/audio/whisper.py
+++ b/xinference/model/audio/whisper.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
+import os
+from glob import glob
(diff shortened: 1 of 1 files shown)


qinxuye (approved):
LGTM

PR #2294, opened 2024-09-12
FEAT: Support yi-coder-chat

Jun-Howie:


qinxuye:
Could you add support for the base/generate model? and the lint should be fixed.

Jun-Howie:
> Could you add support for the base/generate model? and the lint should be fixed.

Completion of base model support

PR #2409, opened 2024-10-09
FEAT: add whisper-large-v3-turbo audio model

hwzhuhao:
1.add whisper-large-v3-turbo audio model
audio-1
audio-2
audio-3


qinxuye (approved):
LGTM, thanks.

PR #2411, opened 2024-10-09
FEAT: Add environment variable setting to increase the retry attempts after model download failures

hwzhuhao:
1.Add environment variable setting to increase the retry attempts after model download failures
2.Remove unuse MAX_ATTEMPTS variable in audio, image and vedio module


qinxuye (approved):
LGTM, thanks.

PR #2495, opened 2024-10-29
BUG: fix bge-reranker-v2-minicpm-layerwise rerank issue

hustyichi:
The return value of compute_score() in the bge-reranker-v2-minicpm-layerwise model is a two-dimensional array. Currently, an error message `only integer scalar arrays can be converted to a scalar index` is reported during execution, so additional processing is required

### DIFF
--- a/xinference/model/rerank/core.py
+++ b/xinference/model/rerank/core.py
@@ -268,7 +268,11 @@ class RerankModel:
             similarity_scores = self._model.compute_score(sentence_combinations)
             if not isinstance(similarity_scores, Sequence):
                 similarity_scores = [similarity_scores]
-            if similarity_scores and isinstance(similarity_scores[0], Sequence):
+            elif (
+                isinstance(similarity_scores, list)
+                and len(similarity_scores) > 0
+                and isinstance(similarity_scores[0], Sequence)
+            ):
                 similarity_scores = similarity_scores[0]
 
         sim_scores_argsort = list(reversed(np.argsort(similarity_scores)))
--- a/xinference/model/rerank/core.py
+++ b/xinference/model/rerank/core.py
@@ -268,6 +268,8 @@ class RerankModel:
             similarity_scores = self._model.compute_score(sentence_combinations)
             if not isinstance(similarity_scores, Sequence):
                 similarity_scores = [similarity_scores]
+            if similarity_scores and isinstance(similarity_scores[0], Sequence):
+                similarity_scores = similarity_scores[0]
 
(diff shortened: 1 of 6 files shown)


hustyichi:
Test code 
```from FlagEmbedding import LayerWiseFlagLLMReranker

RERANK_MODEL = "BAAI/bge-reranker-v2-minicpm-layerwise"

rerank_model = LayerWiseFlagLLMReranker(RERANK_MODEL, use_fp16=True, device="cuda")

question = "What is the meaning of life?"
passages = [
    "This is a test passage.",
    "This is another test passage.",
]

scores = rerank_model.compute_score(
    [[question, passage] for passage in passages],
    cutoff_layers=[28],
    normalize=True,
)

print(scores)
print(type(scores), type(scores[0]))
```

The log is:

```
[[5.225226840386987e-05, 3.120191139396651e-05]]
<class 'list'> <class 'list'>
```

xinference/model/rerank/core.py
```diff
@@ -268,6 +268,8 @@ def rerank(
             similarity_scores = self._model.compute_score(sentence_combinations)
             if not isinstance(similarity_scores, Sequence):
                 similarity_scores = [similarity_scores]
+            if similarity_scores and isinstance(similarity_scores[0], Sequence):
```
qinxuye, line 271: This looks like just the opposite operation of 269-270.

if similarity_scores is not a sequence, wrap into a list, then check if similarity_scores[0] is a sequence, unwrap it.

hustyichi: Yes, there are different strange structures in the return values ​​of bge series models

qinxuye: So I just wonder if there is more clear implementation?

hustyichi: This is the simplest, most painless fix I can think of, maybe you guys can have a more elegant solution to fix this problem

qinxuye:
```
if not isinstance(similarity_scores, Sequence):
    similarity_scores = [similarity_scores]
elif isinstance(similarity_scores, list) and len(similarity_scores) > 0  and isinstance(similarity_scores[0], Sequence):
```

I think `elif` is more clear because they cannot happen at the same time. About the condition I modified a bit, you can see if it can work.

hustyichi: Optimized the implementation of the code based on your suggestions

qinxuye: Thanks, have you verified it can work?

qinxuye (approved):
LGTM

PR #2503, opened 2024-10-31
BLD: Remove Python 3.8 & Support Python 3.12

ChengjieLi28:
Known issues:
1. auto-gptq does not support Python 3.12, which means docker image cannot be based on vllm:latest
2. For macos, eva-decord does not support Python 3.12

Docker build test:
https://github.com/xorbitsai/inference/actions/runs/11625742751


qinxuye (approved):
LGTM

PR #2509, opened 2024-11-04
ENH: add normalize to rerank model

hustyichi:
The current Rerank model return value is not normalized. FlagEmbedding provides a normalization option. It is recommended to return the normalized value.

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -1316,11 +1316,6 @@ class RESTfulAPI(CancelMixin):
         payload = await request.json()
         body = RerankRequest.parse_obj(payload)
         model_uid = body.model
-        kwargs = {
-            key: value
-            for key, value in payload.items()
-            if key not in RerankRequest.__annotations__.keys()
-        }
 
         try:
             model = await (await self._get_supervisor_ref()).get_model(model_uid)
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -115,7 +115,7 @@ class RerankRequest(BaseModel):
     return_documents: Optional[bool] = False
     return_len: Optional[bool] = False
     max_chunks_per_doc: Optional[int] = None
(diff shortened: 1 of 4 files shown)


hustyichi:
The current return value is as follows：

qinxuye:
Should the If normalize, what's the result?

hustyichi:
The structure of the result is the same, except that the values ​​are normalized to be between 0 and 1

The corresponding implementation in FlagEmbedding is as follows
```
def sigmoid(x):
    return 1 / (1 + np.exp(-x))

for i in range(len(all_scores)):
    all_scores[i] = [all_scores[i][idx] for idx in np.argsort(length_sorted_idx)]
    if normalize:
          all_scores[i] = [sigmoid(score) for score in all_scores[i]]
```

qinxuye:
OK, looks like this will not effect the result, but I don't if this will cause any side effect. @codingl2k1 What's your opinion?

qinxuye:
> > OK, looks like this will not effect the result, but I don't know if this will cause any side effect. @codingl2k1 What's your opinion?
> 
> This will be a breaking change. Since FlagEmbedding sets normalize to False by default, how about we keep normalize set to False by default and allow users to set normalize in kwargs?

OK, @hustyichi What's your idea? is it possible to add kwargs to rerank and allow setting normalize to True?

hustyichi:
> > > OK, looks like this will not effect the result, but I don't know if this will cause any side effect. @codingl2k1 What's your opinion?
> > 
> > 
> > This will be a breaking change. Since FlagEmbedding sets normalize to False by default, how about we keep normalize set to False by default and allow users to set normalize in kwargs?
> 
> OK, @hustyichi What's your idea? is it possible to add kwargs to rerank and allow setting normalize to True?

This is a good idea, but requires more code

hustyichi:
I added the always_normalize option, which is not enabled by default

qinxuye (approved):
LGTM

PR #2542, opened 2024-11-11
ENH: Support fish speech reference audio

codingl2k1:
- Fixes: https://github.com/xorbitsai/inference/issues/2532

### DIFF
--- a/doc/source/models/model_abilities/audio.rst
+++ b/doc/source/models/model_abilities/audio.rst
@@ -331,7 +331,7 @@ Clone voice, launch model ``CosyVoice-300M``.
 
     zero_shot_prompt_text = ""
     # The zero shot prompt file is the voice file
-    # the words said in the file shoule be identical to zero_shot_prompt_text
+    # the words said in the file should be identical to zero_shot_prompt_text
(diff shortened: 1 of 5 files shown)


qinxuye:
Is `reference_audio` equivalent to `prompt_speech` for cosyvoice? I wonder if it's possible to reuse the option?

qinxuye:
> > Is `reference_audio` in fish equivalent to `prompt_speech` in cosyvoice? I wonder if it's possible to reuse the option?
> 
> We can reuse the option, then users will need to pass `prompt_speech` instead of `reference_audio`. I am not sure which name is better.

Yeah, we can unify the APIs since `prompt_speech` is added already, we can add it in doc that it can be used as `reference_audio` for fish speech.

qinxuye:
I had some comments:

1. We may set `enable_reference_audio` to True if `prompt_speech` specified.
2. We can support passing  `compile=True` for model loading which give significant performance improvement.

qinxuye:
Looks like our CI cannot run with `compile=True`.

qinxuye (approved):
LGTM

PR #2600, opened 2024-11-28
BUG: Correct the input bytes data by langchain_openai  #2589

xiyuan-lee:


qinxuye:
Lint failed, please install pre-commit and run `pre-commit install` at the project directory, this will do pre-check before committing.

xinference/api/restful_api.py
```diff
@@ -1292,6 +1293,33 @@ async def create_embedding(self, request: Request) -> Response:
         payload = await request.json()
         body = CreateEmbeddingRequest.parse_obj(payload)
         model_uid = body.model
+
+        # 检查 body.input 是否是一个二维整数列表
```
qinxuye, line 1297: Could you write the comments in English?

qinxuye: Below are the same for a couple of comments.

xiyuan-lee: OK

qinxuye:
I just found this part is implemented in rest API, maybe it's better to put them into the embedding model? 

https://github.com/xorbitsai/inference/blob/0d4cb9c9ba3289b7616d963e0c68a2a6bb352422/xinference/model/embedding/core.py#L226

qinxuye:
> I just found this part is implemented in rest API, maybe it's better to put them into the embedding model?
> 
> https://github.com/xorbitsai/inference/blob/0d4cb9c9ba3289b7616d963e0c68a2a6bb352422/xinference/model/embedding/core.py#L226

Since rest API may not be in charge of any computation including the decoding.

xiyuan-lee:
> I just found this part is implemented in rest API, maybe it's better to put them into the embedding model?
> 
> https://github.com/xorbitsai/inference/blob/0d4cb9c9ba3289b7616d963e0c68a2a6bb352422/xinference/model/embedding/core.py#L226
> 
> Since rest API may not be in charge of any computation including the decoding.

Ok, I move the changes to the embedding model.

xinference/model/embedding/core.py
```diff
@@ -224,6 +226,28 @@ def to(self, *args, **kwargs):
             )
 
     def create_embedding(self, sentences: Union[str, List[str]], **kwargs):
+        # Check if sentences is a two-dimensional list of integers
+        if isinstance(sentences, list) and all(
```
qinxuye, line 230: I recommend to put all these codes into a function to explicitly explain what it is dong.

```
def _fix_langchain_openai_inputs(self, sentences: xxx):
    import tiktoken  # put import here, tiktoken is not a must-have dependency

    # check xxx
    if isinstance xxx
    
```

xiyuan-lee: ok

qinxuye: The import needs to be moved like I said, or all CI would fail.

xiyuan-lee: OK

qinxuye:
I slightly modified your code, check all the data if all of them is int is not so necessary IMHO, I think we could only check if it's list of list of int, and check the first int is OK, or this check may costs too much CPU.

qinxuye (approved):
LGTM

PR #2602, opened 2024-11-28
FEAT: support QwQ-32B-Preview

qinxuye:


amumu96 (approved):

tangtianyi19980709:
can we support qwq quantization ?

qinxuye:
Yes, awq is included already, but hf only.

tangtianyi19980709:
thanks for reply!

PR #2623, opened 2024-12-03
ENH: Optimize error message when user parameters are passed incorrectly

namecd:
issue： [# 2539](https://github.com/xorbitsai/inference/issues/2539)

More information is provided for the situation described in issue (#2539), and a test example is added to ensure that the error parameters of the test message are correct

Problem description:

1. When a user specifies a parameter but does not specify the parameter value, the error message is not obvious
2. After investigation, it is found that the problem is the 'click' library of the parsing parameter passed, and the parameter name that indicates the value of the next parameter is taken as the parameter value

Solution:

- Adds an error message and indicates the location of the parameter passing error

### DIFF
--- a/xinference/deploy/test/test_cmdline.py
+++ b/xinference/deploy/test/test_cmdline.py
@@ -23,6 +23,7 @@ from ..cmdline import (
     list_model_registrations,
     model_chat,
     model_generate,
+    model_launch,
     model_list,
     model_terminate,
     register_model,
@@ -311,3 +312,58 @@ def test_remove_cache(setup):
 
     assert result.exit_code == 0
     assert "Cache directory qwen1.5-chat has been deleted."
+
+
+def test_launch_error_in_passing_parameters():
+    runner = CliRunner()
+
+    # Known parameter but not provided with value.
+    result = runner.invoke(
+        model_launch,
+        [
+            "--model-engine",
+            "transformers",
+            "--model-name",
+            "qwen2.5-instruct",
+            "--model-uid",
+            "-s",
+            "0.5",
+            "-f",
+            "gptq",
+            "-q",
+            "INT4",
+            "111",
+            "-l",
+        ],
+    )
+    assert result.exit_code == 1
+    assert (
+        "You must specify extra kwargs with `--` prefix. There is an error in parameter passing that is 0.5."
+        in str(result)
+    )
+
+    # Unknown parameter
+    result = runner.invoke(
+        model_launch,
+        [
+            "--model-engine",
+            "transformers",
+            "--model-name",
+            "qwen2.5-instruct",
+            "--model-uid",
+            "123",
+            "-s",
+            "0.5",
+            "-f",
+            "gptq",
+            "-q",
+            "INT4",
+            "-l",
+            "111",
+        ],
+    )
+    assert result.exit_code == 1
+    assert (
+        "You must specify extra kwargs with `--` prefix. There is an error in parameter passing that is -l."
+        in str(result)
+    )
--- a/xinference/deploy/cmdline.py
+++ b/xinference/deploy/cmdline.py
@@ -846,7 +846,9 @@ def model_launch(
     kwargs = {}
     for i in range(0, len(ctx.args), 2):
         if not ctx.args[i].startswith("--"):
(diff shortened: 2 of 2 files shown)


qinxuye:
The lint failed, please install `pre-commit` and do `pre-commit install` at the project level, then commit the code will be formated automatically,  and if flake8 and so forth failed, you will know before pushing to github.

namecd:
With `pre-commit` already passed, except for `deploy/cmdline.py` and `deploy/test/test_error_message.py`, which are the main changes, the rest are formatting changes, which are done automatically by `black`

qinxuye:
Please revert the changes that are irrelevant.

xinference/deploy/test/test_error_message.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+    print(result.output)
+    assert result.exit_code == 1
+    assert (
+        "You must specify extra kwargs with `--` prefix. There is an error in parameter passing that is -l."
+        in str(result)
+    )
+
+
+if __name__ == "__main__":
```
qinxuye, line 64: This is not needed.

xinference/deploy/test/test_error_message.py
qinxuye, line 1: Better to name this file test_cmd.py

xinference/deploy/test/test_error_message.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+    print(result.output)
+    assert result.exit_code == 1
+    assert (
+        "You must specify extra kwargs with `--` prefix. There is an error in parameter passing that is 0.5."
+        in str(result)
+    )
+
+
+def test_error_parameters():
```
qinxuye, line 35: What's the differences for these two tests? IMHO, they can be merged into one.

You need to add  comment to tell others what you are testing.

xinference/deploy/test/test_error_message.py
```diff
@@ -0,0 +1,65 @@
...
+            "0.5",
+            "-f",
+            "gptq",
+            "-q",
+            "INT4",
+            "111",
+            "-l",
+        ],
+    )
+    print(result.output)
```
qinxuye, line 27: Remove print.

xinference/deploy/test/test_error_message.py
```diff
@@ -0,0 +1,65 @@
...
+            "0.5",
+            "-f",
+            "gptq",
+            "-q",
+            "INT4",
+            "-l",
+            "111",
+        ],
+    )
+    print(result.output)
```
qinxuye, line 56: Ditto.

xinference/deploy/test/test_cmdline.py
```diff
@@ -17,6 +17,8 @@
 import pytest
 from click.testing import CliRunner
 
+from xinference.deploy.cmdline import model_launch
```
qinxuye, line 20: Use relative import.

xinference/deploy/test/test_error_message.py
namecd: OK， I went through all the comments and made the changes

xinference/deploy/test/test_cmdline.py
```diff
@@ -311,3 +313,58 @@ def test_remove_cache(setup):
 
     assert result.exit_code == 0
     assert "Cache directory qwen1.5-chat has been deleted."
+
+
+def test_launch_error_in_passing_parameters():
+    runner = CliRunner()
+
+    # Known parameter is not specified when error
```
qinxuye, line 321:
```suggestion
    # Known parameter but not provided with value.
```

xinference/deploy/test/test_cmdline.py
```diff
@@ -311,3 +313,58 @@ def test_remove_cache(setup):
...
+            "-l",
+        ],
+    )
+    assert result.exit_code == 1
+    assert (
+        "You must specify extra kwargs with `--` prefix. There is an error in parameter passing that is 0.5."
+        in str(result)
+    )
+
+    # Unknown parameter is specified when error
```
qinxuye, line 346:
```suggestion
    # Unknown parameter
```

namecd: OK，I revised according to your suggestion

PR #2641, opened 2024-12-09
FIX: fix launching bge-m3 with hybrid mode

pengjunfeng11:


xinference/model/embedding/core.py
```diff
@@ -208,12 +208,15 @@ def to(self, *args, **kwargs):
                 ]
                 raise ImportError(f"{error_message}\n\n{''.join(installation_guide)}")
 
-            model_kwargs = {"torch_dtype": torch_dtype} if torch_dtype else None
+            if torch_dtype and torch_dtype == torch.float16:
+                model_kwargs = {"use_fp16": True}
+            else:
+                model_kwargs = None
```
qinxuye, line 214: Should it be an empty dict?

qinxuye (approved):
LGTM

PR #2540, opened 2024-11-11
FEAT: support sparse vector for bge-m3

pengjunfeng11:
### 支持bge-m3模型的稀疏向量生成功能，
调用方式为：```model.create_embedding(text, return_sparse=True)```

### 新增convert_ids_to_tokens方法
该方法可将token_id转换为人类可读文字，调用方式为
```
from xinference.client import Client

client = Client("http://ip:port")
model = client.get_model(model_name)
seq = model.convert_ids_to_tokens(key_list)
```
该方法返回类型为List[str]，如传入List[str]，将按顺序返回值

Fixes #2527 .

PR #2649, opened 2024-12-10
BUG: fix ImportError when optional dependency FlagEmbedding is not installed

zjuyzj:
#2540 introduces `FlagEmbedding` package to support sparse vector for bge-m3. However, `FlagEmbedding` is an optional dependency for embedding model (but required for rerank):
https://github.com/xorbitsai/inference/blob/70a008138be7fa13706964124d342f194f7f584d/setup.cfg#L184-L187
In the case `FlagEmbedding` is not installed, an `ImportError` will be raised even if embedding models only based on `sentence-transformers` are used, due to the logic related to `BGEM3FlagModel`. This PR fixes this problem.

By the way, maybe adding `FlagEmbedding` to required dependency for embedding is not suitable now until upstream solve the dependency, since it requires `transformers==4.44.2`(https://github.com/FlagOpen/FlagEmbedding/blob/master/setup.py#L18) and will make models like Qwen2-VL unusable (https://huggingface.co/Qwen/Qwen2-VL-7B-Instruct/discussions/5).

### DIFF
--- a/xinference/model/embedding/core.py
+++ b/xinference/model/embedding/core.py
@@ -261,154 +261,165 @@ class EmbeddingModel:
     def create_embedding(self, sentences: Union[str, List[str]], **kwargs):
         sentences = self._fix_langchain_openai_inputs(sentences)
 
-        from FlagEmbedding import BGEM3FlagModel
         from sentence_transformers import SentenceTransformer
 
         kwargs.setdefault("normalize_embeddings", True)
 
-        @no_type_check
-        def _encode_bgem3(
-            model: Union[SentenceTransformer, BGEM3FlagModel],
-            sentences: Union[str, List[str]],
-            batch_size: int = 32,
-            show_progress_bar: bool = None,
-            output_value: str = "sparse_embedding",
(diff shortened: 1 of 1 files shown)


xinference/model/embedding/core.py
```diff
@@ -261,154 +263,155 @@ def _fix_langchain_openai_inputs(self, sentences: Union[str, List[str]]):
...
-            length_sorted_idx = np.argsort([-_text_length(sen) for sen in sentences])
-            sentences_sorted = [sentences[idx] for idx in length_sorted_idx]
-
-            for start_index in trange(
-                0,
-                len(sentences),
-                batch_size,
-                desc="Batches",
-                disable=not show_progress_bar,
+        if self.with_flag_embedding:
```
qinxuye, line 270: Maybe just try to import `FlatEmbedding`, if not installed, skip the definition of `_encode_bgem3 `.

qinxuye:
Tests seem failed.

qinxuye (approved):
LGTM

PR #2665, opened 2024-12-13
ENH: upgrade sglang in docker

amumu96:


xinference/deploy/docker/Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN pip install --upgrade -i "$PIP_INDEX" pip && \
     # use pre-built whl package for llama-cpp-python, otherwise may core dump when init llama in some envs
     pip install "llama-cpp-python>=0.2.82" -i https://abetlen.github.io/llama-cpp-python/whl/cu124 && \
     pip install -i "$PIP_INDEX" --upgrade-strategy only-if-needed -r /opt/inference/xinference/deploy/docker/requirements.txt && \
-    pip install -i "$PIP_INDEX" --no-deps sglang && \
+    pip install -i "$PIP_INDEX" "sglang[srt]" && \
```
qinxuye, line 32: Shall us fix a version in case there is new issue when sglang upgrade.

xinference/deploy/docker/Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN pip install --upgrade -i "$PIP_INDEX" pip && \
     # use pre-built whl package for llama-cpp-python, otherwise may core dump when init llama in some envs
     pip install "llama-cpp-python>=0.2.82" -i https://abetlen.github.io/llama-cpp-python/whl/cu124 && \
     pip install -i "$PIP_INDEX" --upgrade-strategy only-if-needed -r /opt/inference/xinference/deploy/docker/requirements.txt && \
-    pip install -i "$PIP_INDEX" --no-deps sglang && \
+    pip install -i "$PIP_INDEX" --no-deps "sglang[srt]" && \
```
qinxuye, line 32: no deps seems it doesn't matter it has extra or not. so I think the change of this file  could be reverted.

PR #2684, opened 2024-12-17
ENH: Update cosyvoice 2

codingl2k1:


qinxuye:
For hf we can use https://huggingface.co/mrfakename/CosyVoice2-0.5B/tree/main before the official repo is created

qinxuye (approved):
LGTM

PR #2711, opened 2024-12-26
FEAT: Reduce code redundancy by setting default values

pengjunfeng11:
Reduce code redundancy by setting default values


qinxuye (approved):
LGTM

PR #2673, opened 2024-12-14
ENH: Guided Decoding OpenAIClient compatibility

wxiwnd:
FIX: https://github.com/xorbitsai/inference/issues/2636
also https://github.com/xorbitsai/inference/issues/2620

cc @qinxuye 

```python
client = openai.Client(api_key="not empty", base_url="http://127.0.0.1:9997/v1")
class TSchema(BaseModel):
    test: str
    num: int
schema = TSchema.model_json_schema()
messages = [
{"role": "system", "content": "you are a helpful assistant"},
{"role": "user", "content": "give me an example json, fill a random number"}
]
completion = client.chat.completions.create(
model="qwen2-instruct",
messages = messages,
extra_body={
"guided_json": schema,
"guided_decoding_backend": "outlines"
})
```
Output:
```
{ "test": "Example JSON", "num": 42 }
```
Guided Choice also works fine.

PR #2740, opened 2025-01-06
FEAT: Support cogagent-9b

amumu96:
image

### DIFF
--- a/xinference/types.py
+++ b/xinference/types.py
@@ -300,6 +300,16 @@ class PytorchGenerateConfig(TypedDict, total=False):
     lora_name: Optional[str]
     stream_options: Optional[Union[dict, None]]
     request_id: Optional[str]
+    platform: Optional[Literal["Mac", "WIN", "Mobile"]]
+    format: Optional[
+        Literal[
+            "(Answer in Action-Operation-Sensitive format.)",
+            "(Answer in Status-Plan-Action-Operation format.)",
+            "(Answer in Status-Action-Operation-Sensitive format.)",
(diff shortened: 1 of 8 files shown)


xinference/types.py
```diff
@@ -300,6 +300,16 @@ class PytorchGenerateConfig(TypedDict, total=False):
     lora_name: Optional[str]
     stream_options: Optional[Union[dict, None]]
     request_id: Optional[str]
+    platform: Optional[Literal["Mac", "WIN", "Mobile"]]
```
qinxuye, line 303: Looks like this is for cogagent only, is it possible to put them into `**kwargs` stuff?

xinference/types.py
```diff
@@ -430,6 +443,19 @@ class CreateChatModel(BaseModel):
 CreateChatCompletionLlamaCpp: BaseModel = CreateCompletionLlamaCpp
 
 
+class CreateExtraChatCompletion(BaseModel):
```
qinxuye, line 446: This seems not used?

amumu96: 2819 resultful_api.py `body = CreateChatCompletion.parse_obj(raw_body)` needs this code, or it will throw error of parsing

qinxuye: IC. Can this be more flexible? [I mean it can ](https://github.com/xorbitsai/inference/pull/2673/files) introduced extra body, the new format and platform is too specific for cogagent only, is it possible to throw them into extra body?

qinxuye (approved):
LGTM

PR #2734, opened 2025-01-03
FEAT: support cline for vllm engine

hwzhuhao:
support cline with vllm engine, i have test with vllm 0.6.1.post2.
1. vllm version
image
2. gradio diaglog
image
image
3. vscode+cline
image
image

close #2659

Issue #2659, opened 2024-12-12
Supporting Cline requests style.


gahoo:
### Feature request / 功能建议

`Cline` requests are quite different in the `content` section which might be a `list` instead of `str`.

```json
{
	"model": "qwen-2.5-coder-instruct",
	"messages": [{
		"role": "system",
		"content": "You are Cline, a highly skilled software engineer with extensive knowledge in many programming languages, frameworks, design patterns, and best practices......"
	}, {
		"role": "user",
		"content": [{
			"type": "text",
			"text": "<task>......</task>"
		}, {
			"type": "text",
			"text": "<environment_details>.....</environment_details>"
		}, {
			"type": "text",
			"text": "[TASK RESUMPTION] ....."
		}, {
			"type": "text",
			"text": "<environment_details>.......</environment_details>"
		}, {
			"type": "text",
			"text": "[TASK RESUMPTION] ......"
		}, {
			"type": "text",
			"text": "<environment_details>......</environment_details>"
		}]
	}],
	"temperature": 0,
	"stream": true,
	"stream_options": {
		"include_usage": true
	}
}
```

 So the default `chat_template` might cause the following error:

```
TypeError: [address=0.0.0.0:33357, pid=544] can only concatenate str (not "list") to str
```

### Motivation / 动机

It would be great that Cline or other similar tools can use Xinference's API.

Similar issues happened in DeepSeek API, and seems fixed. see https://github.com/cline/cline/issues/230 .

### Your contribution / 您的贡献

None

qinxuye:
I have no clue about cline, to be honest, this is my first time hearing of it.

But welcome to contribute on this feature.

gahoo:
Cline is a VS code extension, previous named Claude Dev.

I think it would be a lot of work since every LLM model has it's own `chat_template`. Do you think custom `chat_template` is a solution?

qinxuye:
Does Cline has its own API router like openai ones /v1, if so, we can tell the message is from Cline, and we can write a function to convert the messages.

gahoo:
> Cline supports API providers like OpenRouter, Anthropic, OpenAI, Google Gemini, AWS Bedrock, Azure, and GCP Vertex. You can also configure any OpenAI compatible API, or use a local model through LM Studio/Ollama. If you're using OpenRouter, the extension fetches their latest model list, allowing you to use the newest models as soon as they're available.

[...]

PR #2763, opened 2025-01-15
ENH: support cline style messages for all backend engines

liunux4odoo:
As described in #2659 the Cline extension uses a specific messages format.
#2734 supports cline for vLLM engine.
This PR applies it to all other engines.
Test on windows+transformers is success.


qinxuye (approved):
LGTM

PR #2777, opened 2025-01-22
ENH: Some improvements for Xavier

ChengjieLi28:
1. Xavier supports model recovery. Due to the special nature of rank 0, a separate, simple model actor is created for rank 0. This actor is responsible only for constructing the collective communication world. A new role, `collective_manager`, is introduced. In the recovery logic, the `collective_manager` instructs all ranks to reconstruct the collective communication world.
2. If an error occurs during the transmission process (e.g., the sender is undergoing recovery), vllm's scheduling supports rollback, where it reverts to forcing computation instead of transmission.
3. Xavier's transmission process only applies to the PREFILL stage, and no data transmission occurs during the DECODE stage. This is because data transmission in the DECODE stage only affects the last token of a block when it is about to be filled with tokens. This would be counterproductive and negatively impact overall throughput, as it introduces additional RPC call times for many queries.
4. Fix the error when multiple models are enabled with Xavier simultaneously. This is because the block tracker should be one-to-one with the model, rather than having just one block tracker for all models.

### DIFF
--- a/doc/source/locale/zh_CN/LC_MESSAGES/user_guide/vllm_enhancement.po
+++ b/doc/source/locale/zh_CN/LC_MESSAGES/user_guide/vllm_enhancement.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Xinference \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2025-01-10 14:44+0800\n"
+"POT-Creation-Date: 2025-01-23 14:46+0800\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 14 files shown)


ChengjieLi28:
> 1. Xavier supports model recovery. Due to the special nature of rank 0, a separate, simple model actor is created for rank 0. This actor is responsible only for constructing the collective communication world. A new role, `collective_manager`, is introduced. In the recovery logic, the `collective_manager` instructs all ranks to reconstruct the collective communication world.
> 2. If an error occurs during the transmission process (e.g., the sender is undergoing recovery), vllm's scheduling supports rollback, where it reverts to forcing computation instead of transmission.
> 3. Xavier's transmission process only applies to the PREFILL stage, and no data transmission occurs during the DECODE stage. This is because data transmission in the DECODE stage only affects the last token of a block when it is about to be filled with tokens. This would be counterproductive and negatively impact overall throughput, as it introduces additional RPC call times for many queries.
> 4. Fix the error when multiple models are enabled with Xavier simultaneously. This is because the block tracker should be one-to-one with the model, rather than having just one block tracker for all models.

qinxuye (approved):
LGTM

PR #2779, opened 2025-01-22
FEAT: create_embedding add field model_replica

zhoudelong:


xinference/model/embedding/core.py
```diff
@@ -693,7 +693,8 @@ def base64_to_image(base64_str: str) -> Image.Image:
                 if not is_bge_m3_flag_model and not kwargs.get("return_sparse")
                 else "dict"
             ),
-            model=self._model_uid,
+            model="",
```
qinxuye, line 696: How about just let the model be the model name? Empty looks weird, what's you opinion? @codingl2k1 @amumu96

qinxuye:
All tests failed.

xinference/api/restful_api.py
```diff
@@ -1330,8 +1330,10 @@ async def create_embedding(self, request: Request) -> Response:
             raise HTTPException(status_code=500, detail=str(e))
 
         try:
-            embedding = await model.create_embedding(body.input, **kwargs)
-            return Response(embedding, media_type="application/json")
+            embedding_bytes = await model.create_embedding(body.input, **kwargs)
+            embedding = json.loads(embedding_bytes.decode("utf-8"))
+            embedding["model"] = model_uid  # type: ignore
```
qinxuye, line 1335: Try to pass original model uid into model, instead of parsing, adding info then serializing, this is too ugly.

qinxuye (changes requested):

qinxuye:
Lint failed, please fix it.

xinference/core/tests/test_restful_api.py
```diff
@@ -356,6 +356,7 @@ def test_restful_api_for_embedding(setup):
 
     assert "embedding" in embedding_res["data"][0]
     assert len(embedding_res["data"][0]["embedding"]) == model_spec.dimensions
+    assert "model_replica" in embedding_res.keys()
```
qinxuye, line 359:
```
assert embedding_res["model_replica"] is not None
```

Not only to check if the key exists, but also check if the value exists in real.

And, did you check the `model` ? That's the original problem to solve, to make sure `model` is identical to model uid.

qinxuye (changes requested):

xinference/core/tests/test_restful_api.py
```diff
@@ -356,6 +356,9 @@ def test_restful_api_for_embedding(setup):
...
```
qinxuye, line 359:
```suggestion
    assert "model_replica" in embedding_res
```

xinference/core/tests/test_restful_api.py
```diff
@@ -353,9 +353,13 @@ def test_restful_api_for_embedding(setup):
     }
     response = requests.post(url, json=payload)
     embedding_res = response.json()
+    print("embedding_res: ", embedding_res)
```
qinxuye, line 356: Please remove print.

qinxuye (approved):
LGTM

PR #2820, opened 2025-02-08
BUG: Use `Cache` class instead of raw `tuple` for transformers continuous batching, compatible with latest transformers

ChengjieLi28:


qinxuye (approved):
LGTM

PR #2850, opened 2025-02-13
FEAT: [UI] support form parameters and command line conversion.

yiboyasss:
截屏2025-02-13 11 48 36

### DIFF
--- a/xinference/web/ui/src/components/MenuSide.js
+++ b/xinference/web/ui/src/components/MenuSide.js
@@ -100,6 +100,7 @@ const MenuSide = () => {
           boxSizing: 'border-box',
         },
       }}
+      style={{ zIndex: 1 }}
     >
       {/* Title */}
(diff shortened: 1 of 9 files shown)


qinxuye:
I tried mlx engine, and the copied content is 

```
xinference launch --model-name qwen2.5-instruct --model-type LLM --model-engine MLX --model-format mlx --model-size-in-billions 7 --quantization 4bit --n-gpu auto --replica 1 --request-limits  --gpu-idx
```

There is no value set for `--gpu-idx`.

qinxuye:
And the icons are too small maybe the users would never discover them, you can try to highlight them in some ways.

qinxuye:
It should be `--size-in-billions` instead of `--model-size-in-billions`.

qinxuye (approved):
LGTM

PR #2877, opened 2025-02-18
FEAT: support distributed inference for sglang

qinxuye:
With this PR, now deepseek-r1 FP8 can run on 2 8xH100/H20 workers.

To use this feature, launch model with extra param: n_worker: 2.

Web UI will add this option soon, command line needs to be updates as well.


amumu96 (approved):

Zephyr69:
what about distributed vllm inference?

PR #2896, opened 2025-02-20
Fix FLUX scheduler

shuaiqidezhong:
Fix FLUX scheduler

### DIFF
--- a/xinference/model/image/stable_diffusion/core.py
+++ b/xinference/model/image/stable_diffusion/core.py
@@ -416,8 +416,12 @@ class DiffusionModel(SDAPIDiffusionModelMixin):
     @contextlib.contextmanager
     def _reset_when_done(model: Any, sampler_name: str):
         assert model is not None
+        name_or_path: Optional[str] = getattr(model, "_name_or_path", None)
+        is_flux_sampler = "flux" in name_or_path.casefold() if name_or_path else False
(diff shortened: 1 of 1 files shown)


xinference/model/image/stable_diffusion/core.py
```diff
@@ -416,8 +416,12 @@ def _get_scheduler(model: Any, sampler_name: str):
     @contextlib.contextmanager
     def _reset_when_done(model: Any, sampler_name: str):
         assert model is not None
+        name_or_path: Optional[str] = getattr(model, "_name_or_path", None)
+        is_flux_sampler = "flux" in name_or_path.casefold() if name_or_path else False
```
qinxuye, line 420: I think we'd better judge flux by `model_spec`,  so we may set this method to normal function instead of static method. Because we cannot ensure that the model path contains flux.

xinference/model/image/stable_diffusion/core.py
```diff
@@ -416,8 +416,12 @@ def _get_scheduler(model: Any, sampler_name: str):
...
         scheduler = DiffusionModel._get_scheduler(model, sampler_name)
-        if scheduler:
+        # When the model is not flux
+        should_swap_scheduler = scheduler is not None and not is_flux_sampler
```
qinxuye, line 423: Better to name a better one, maybe `ignore_setting_scheduler`.

PR #2905, opened 2025-02-21
FEAT: support vllm reasoning content

amumu96:


qinxuye (approved):
LGTM

qinxuye:
This work needs to followup works, including:

1. Other engines. Now only vllm added.
2. Gradio support. With `reasoning_content` on, the gradio will not reflect the think.

Issue #2559, opened 2024-11-18
Got an Error while using Langchain Chatchat with internlm2.5.


frankSARU:
### System Info / 系統信息

CUDA 12.4 
Transformer 4.44

### Running Xinference with Docker? / 是否使用 Docker 运行 Xinfernece？

- [ ] docker / docker
- [X] pip install / 通过 pip install 安装
- [ ] installation from source / 从源码安装

### Version info / 版本信息

Version:Release: v0.16.0

### The command used to start Xinference / 用以启动 xinference 的命令

HF_ENDPOINT=https://hf-mirror.com xinference-local --host 10.11.238.110 --port 9997

### Reproduction / 复现过程

在使用chat功能时，xinf平台会出现如下报错（使用kb_chat功能，即知识库功能，并不会报错，其他知识库应用也未报错，排查了Langchain Chatchat相关的设置，chat.py脚本自上一次成功运行以来未做修改，在xinf页面报issue只是为了排查具体是哪个模块的问题）：
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
2024-11-18 17:24:13,778 xinference.api.restful_api 3537005 ERROR    Chat completion stream got an error: [address=10.11.238.110:36475, pid=3545480] int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
Traceback (most recent call last):
  File "/home/pei/.conda/envs/xinf/lib/python3.11/site-packages/xinference/api/restful_api.py", line 1926, in stream_results
    async for item in iterator:
  File "/home/pei/.conda/envs/xinf/lib/python3.11/site-packages/xoscar/api.py", line 340, in __anext__
    return await self._actor_ref.__xoscar_next__(self._uid)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pei/.conda/envs/xinf/lib/python3.11/site-packages/xoscar/backends/context.py", line 231, in send
    return self._process_result_message(result)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pei/.conda/envs/xinf/lib/python3.11/site-packages/xoscar/backends/context.py", line 102, in _process_result_message
    raise message.as_instanceof_cause()
  File "/home/pei/.conda/envs/xinf/lib/python3.11/site-packages/xoscar/backends/pool.py", line 656, in send
[...]

PR #2962, opened 2025-02-27
BUG: fix Error while using Langchain-chatchat, because the parameter [max_tokens] passed is None

William533036:
BUG: fix Error while using Langchain Chatchat ，because The parameter [max_tokens] passed is None

Fixes https://github.com/xorbitsai/inference/issues/2559


qinxuye (approved):
LGTM

PR #2992, opened 2025-03-04
BUG: Fix tokenizer error in create_embedding

shuaiqidezhong:
Fixed the issue where model_uid was incorrectly passed to tokenizer in create_embedding. Removed the problematic code and tests passed.


shuaiqidezhong (commented):
extract model_uid from kwargs before encode to avoid conflict with encode parameters while preserving it for Embedding creation

qinxuye (approved):
LGTM

PR #2997, opened 2025-03-04
ENH: Support xllamacpp

codingl2k1:
Disabled by default, can be enabled by setting env `USE_XLLAMACPP=1`

### DIFF
--- a/xinference/client/tests/test_client.py
+++ b/xinference/client/tests/test_client.py
@@ -165,7 +165,6 @@ def test_RESTful_client(setup):
     assert len(client.list_models()) == 0
 
 
-@pytest.mark.skipif(os.name == "nt", reason="Skip windows")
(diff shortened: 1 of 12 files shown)


xinference/client/tests/test_client.py
```diff
@@ -161,6 +165,11 @@ def _check(stream=False):
     assert len(client.list_models()) == 0
 
 
+@pytest.mark.skipif(os.name == "nt", reason="Skip windows")
```
qinxuye, line 168: Why skip windows?

qinxuye (commented):
Really excellent work, I left some comments.

qinxuye:
I found that now the lock would exist for Xllamacpp, and I added a commit to release it.

qinxuye (approved):
LGTM

PR #3007, opened 2025-03-06
Support Baichuan-M1-14B-Instruct

Jun-Howie:
VLLM is not currently supported
  https://github.com/vllm-project/vllm/issues/14259


Jun-Howie (commented):
fix:Failed to download model 'QwQ-32B' (size: 32, format: ggufv2) after multiple retries

PR #3013, opened 2025-03-07
FEAT: all engine support reasoning_content

amumu96:
> Model ID: deepseek-r1-distill-qwen
> Model Size: 15 Billion Parameters
> Model Format: pytorch
> Model Quantization: none

vLLM without reasoning：
image

vLLM with reasoning：
image

Transformers without reasoning：
image

Transformers with reasoning：
image

SGLang without reasoning：
image

SGLang with reasoning：
image

llama.cpp without reasoning：
pass

llama.cpp with reasoning：
pass

MLX without reasoning：
pass

MLX with reasoning：
pass

> Model ID: qwen2.5-instruct
> Model Size: 5 Billion Parameters
> Model Format: pytorch
> Model Quantization: none

Transformers：
image

vLLM：
image

SGLang：
image

llama.cpp:
image

MLX:
pass

### DIFF
--- a/xinference/model/llm/core.py
+++ b/xinference/model/llm/core.py
@@ -57,6 +57,7 @@ class LLM(abc.ABC):
         self.model_spec = model_spec
         self.quantization = quantization
         self.model_path = model_path
+        self.reasoning_parser = None
         if args:
             raise ValueError(f"Unrecognized positional arguments: {args}")
         if kwargs:
--- a/xinference/model/llm/core.py
+++ b/xinference/model/llm/core.py
@@ -120,7 +120,7 @@ class LLM(abc.ABC):
     ) -> bool:
         raise NotImplementedError
 
-    def should_stream_with_reasoning_parsing(self, reasoning_content):
+    def prepare_parse_reasoning_content(self, reasoning_content):
         # Initialize reasoning parser if model has reasoning ability
         if "reasoning" in self.model_family.model_ability and reasoning_content:
             module_name = self.model_family.model_family or self.model_family.model_name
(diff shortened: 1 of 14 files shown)


xinference/model/llm/core.py
```diff
@@ -117,6 +120,16 @@ def match(
     ) -> bool:
         raise NotImplementedError
 
+    def should_stream_with_reasoning_parsing(self, reasoning_content):
```
qinxuye, line 123: Could rename to `prepare_parse_reasoning_content`.

xinference/model/llm/utils.py
```diff
@@ -565,7 +566,14 @@ def _eval_tool_arguments(cls, model_family, c):
         return result
 
     @classmethod
-    def _tool_calls_completion_chunk(cls, model_family, model_uid, c, chunk_id=None):
+    def _tool_calls_completion_chunk(
```
qinxuye, line 569: This function is not only for tool call, maybe we can call it `_post_process_completion_chunk`

xinference/model/llm/utils.py
```diff
@@ -616,7 +635,13 @@ def _tool_calls_completion_chunk(cls, model_family, model_uid, c, chunk_id=None)
         }
 
     @classmethod
-    def _tool_calls_completion(cls, model_family, model_uid, c):
+    def _tool_calls_completion(
```
qinxuye, line 638: And this function could be renamed to `_post_process_completion` as well

qinxuye (approved):
LGTM

PR #3065, opened 2025-03-13
DOC: add auth usage guide for http request

Minamiyama:


qinxuye:
Why can't github recognize diff? I can see the whole file is modified.

Minamiyama:
> Why can't github recognize diff? I can see the whole file is modified.

due to CRLF in windows, I've changed to LF now

qinxuye (approved):
LGTM

PR #3102, opened 2025-03-21
FEAT: Gemma-3-it that supports vision

qinxuye:
Due to gemma-3 1b and the others are different at:

* 1b is text only, but the others are multimodel
* 1b has 32k context length, the others are 131k

Thus this PR separates them into two models, gemma-3-1b-it and gemma-3-it.


amumu96 (approved):
LGTM

PR #3115, opened 2025-03-24
BUG: fix llm stream response

amumu96:
1. Modify xinference/client/tests/test_client.py: For chunk where finish_reason is not None, assert that delta = {"content": ""}.

2. Modify xinference/model/llm/llama_cpp/core.py: Filter out keys in the returned results that do not belong to ChatCompletionChunk.

3. Modify xinference/model/llm/reasoning_parser.py: Fix the issue where both reasoning_content="" and content="".

4. Modify xinference/model/llm/utils.py:  Ensure that chunk with finish_reason not being None includes values for both content and reasoning_content.

### DIFF
--- a/xinference/client/tests/test_client.py
+++ b/xinference/client/tests/test_client.py
@@ -83,7 +83,6 @@ def test_RESTful_client(setup):
             generate_config={"stream": True, "max_tokens": 5},
         )
         for chunk in streaming_response:
-            print(chunk)
             assert "finish_reason" in chunk["choices"][0]
             finish_reason = chunk["choices"][0]["finish_reason"]
             if finish_reason is None:
--- a/xinference/client/tests/test_client.py
+++ b/xinference/client/tests/test_client.py
@@ -90,7 +90,7 @@ def test_RESTful_client(setup):
                     "role" in chunk["choices"][0]["delta"]
                 )
             else:
-                assert chunk["choices"][0]["delta"] == {}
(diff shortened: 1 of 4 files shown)


xinference/client/tests/test_client.py
```diff
@@ -83,14 +83,15 @@ def _check_stream():
             generate_config={"stream": True, "max_tokens": 5},
         )
         for chunk in streaming_response:
+            print(chunk)
```
qinxuye: Please remove this.

xinference/model/llm/llama_cpp/core.py
```diff
@@ -302,7 +302,8 @@ def _to_iterator():
                 while (r := q.get()) is not _Done:
                     if type(r) is _Error:
                         raise Exception("Got error in chat stream: %s", r.msg)
-                    yield r
+                    chunk_keys = ChatCompletionChunk.__annotations__.keys()
```
qinxuye, line 305: Could you add some comment?

e.g. the chunk contains some key like `system_fingerprint` this may cause xxx.

thus we need to filter it.

qinxuye: And you don't need to get keys()

```
key in ChatCompletionChunk.__annotations__
```

is more clean and efficient.

qinxuye (approved):
LGTM

PR #3120, opened 2025-03-25
FEAT: distributed inference for vLLM

qinxuye:


xinference/model/llm/vllm/core.py
```diff
@@ -247,16 +254,62 @@ def __init__(
...
     def load(self):
         try:
             import vllm
+            from vllm.config import VllmConfig
             from vllm.engine.arg_utils import AsyncEngineArgs
             from vllm.engine.async_llm_engine import AsyncLLMEngine
+            from vllm.executor.executor_base import ExecutorBase
             from vllm.lora.request import LoRARequest
+
+            from .distributed_executor import XinferenceDistributedExecutor
```
amumu96, line 312: Maybe need check vllm version in here, beacuse `distributed_executor `  import DistributedExecutorBase which was defined in vllm 0.7.0

qinxuye: I added vllm version check, I have moved this import into the condition.

PR #3131, opened 2025-03-27
BUG: Fix PyTorch TypeError: Make _ModelWrapper Inherit from nn.Module

JamesFlare1212:
### Problem Description
An error occurs when assigning an instance of `_ModelWrapper` to a sub-module in a PyTorch model:

```
TypeError: cannot assign '_ModelWrapper' as child module 'model'; (torch.nn.Module or None expected)
```

### Cause of the Problem
PyTorch requires all sub-modules to inherit from `nn.Module`, but the original implementation of `_ModelWrapper` does not.

### Solution
- Have `_ModelWrapper` class inherit from `nn.Module`
- Call `super().__init__()` in the constructor
- Rename the existing `__call__` method to conform with PyTorch’s standard naming convention, which is called `forward`

The modifications are minimal and do not introduce any further changes beyond necessary adjustments for compatibility.

### How to Verify
- The original TypeError no longer appears after reloading the model  
- The model runs as expected, and the logic of counting tokens by `n_tokens` remains unchanged

### DIFF
--- a/xinference/model/rerank/core.py
+++ b/xinference/model/rerank/core.py
@@ -106,8 +106,9 @@ def generate_rerank_description(model_spec: RerankModelSpec) -> Dict[str, List[D
     return res
 
 
-class _ModelWrapper:
+class _ModelWrapper(nn.Module):
     def __init__(self, module: nn.Module):
+        super().__init__()
         self._module = module
         self._local_data = threading.local()
 
@@ -119,15 +120,13 @@ class _ModelWrapper:
     def n_tokens(self, new_n_tokens):
         self._local_data.n_tokens = new_n_tokens
 
-    def __getattr__(self, attr):
-        return getattr(self._module, attr)
-
-    def __call__(self, **kwargs):
+    def forward(self, **kwargs):
         attention_mask = kwargs["attention_mask"]
-        # when batching, the attention mask 1 means there is a token
-        # thus we just sum up it to get the total number of tokens
         self.n_tokens += attention_mask.sum().item()
         return self._module(**kwargs)
+
+    def __getattr__(self, attr):
+        return getattr(self._module, attr)
 
 
 class RerankModel:
--- a/xinference/model/rerank/core.py
+++ b/xinference/model/rerank/core.py
@@ -109,7 +109,7 @@ def generate_rerank_description(model_spec: RerankModelSpec) -> Dict[str, List[D
 class _ModelWrapper(nn.Module):
     def __init__(self, module: nn.Module):
         super().__init__()
-        self._module = module
+        self.model = module
         self._local_data = threading.local()
 
     @property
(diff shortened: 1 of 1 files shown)


xinference/model/rerank/core.py
```diff
@@ -119,16 +120,14 @@ def n_tokens(self):
     def n_tokens(self, new_n_tokens):
         self._local_data.n_tokens = new_n_tokens
 
-    def __getattr__(self, attr):
-        return getattr(self._module, attr)
-
-    def __call__(self, **kwargs):
+    def forward(self, **kwargs):
         attention_mask = kwargs["attention_mask"]
-        # when batching, the attention mask 1 means there is a token
```
qinxuye, line 127: Could you add back the comment?

JamesFlare1212:
I found this change will cause other problems in more tests. Other parts rely on the properties of this class. Please do not merge the changes yet.

qinxuye:
> I found this change will cause other problems in more tests. Other parts rely on the properties of this class. Please do not merge the changes yet.

Yes, I noticed that many tests failed. Are you aware of what's happening?

JamesFlare1212:
I patched up the issues with each attribute according to the error messages, but the original access logic cannot be used directly; this is also one of the reasons for the errors. However, there have been significant changes, and I am not very familiar with the project design. Could you please review it? If everything is okay, let's figure out how to merge the changes.

```python
class _ModelWrapper(nn.Module):
    def __init__(self, module: nn.Module):
        super().__init__()
        self.model = module
        self._local_data = threading.local()

    @property
    def n_tokens(self):
        return getattr(self._local_data, "n_tokens", 0)

    @n_tokens.setter
    def n_tokens(self, value):
        self._local_data.n_tokens = value

    def forward(self, **kwargs):
        attention_mask = kwargs.get("attention_mask")
        # when batching, the attention mask 1 means there is a token
        # thus we just sum up it to get the total number of tokens
        if attention_mask is not None:
            self.n_tokens += attention_mask.sum().item()
        return self.model(**kwargs)

    def __getattr__(self, attr):
        try:
            return super().__getattr__(attr)
        except AttributeError:
            return getattr(self.model, attr)
 ```

JamesFlare1212:
> > I found this change will cause other problems in more tests. Other parts rely on the properties of this class. Please do not merge the changes yet.
> 
> Yes, I noticed that many tests failed. Are you aware of what's happening?

Long stroy short, it is assigning `_ModelWrapper` (which is not a subclass of `torch.nn.Module`) to `self._model.model`, and PyTorch enforces that all attributes under a `torch.nn.Module` must also be instances of `torch.nn.Module` or `None`. At first, I just changed the way it pass the call and init. But other parts form this project use some properties with the old way.

The previous changes caused __getattr__ to result in infinite recursion. In essence, it accidentally goes back to itself, whereas previously it did not return itself. It will cost too much to change all of them. So, I tried to pass the properties to the orignal class. like `wrapper.config` will return `self.model.config`. So, other part from this project can remains the same.

qinxuye (approved):
LGTM

PR #3150, opened 2025-03-30
FEAT: sglang vision

Minamiyama:


xinference/model/llm/sglang/core.py
```diff
@@ -108,6 +109,9 @@ class SGLANGGenerateConfig(TypedDict, total=False):
     "deepseek-v3",
     "deepseek-r1",
 ]
+SGLANG_SUPPORTED_VISION_MODEL_LIST = [
```
qinxuye, line 112: `gemma-3-it` which is a vision model could be moved here.

qinxuye (approved):
LGTM

PR #3159, opened 2025-03-31
feat: fix the quantization parameter in the vLLM engine cannot work

amumu96:
Add model_quantization parameter to replace the quantization parameter in the vLLM engine


qinxuye (approved):
LGTM

PR #3191, opened 2025-04-06
REF: simplify transformers model registration with decorators.

Minamiyama:
define two decorators for self registering only in new model code file, keep codes loose coupling

image
image


qinxuye (approved):
LGTM

PR #3198, opened 2025-04-07
ENH: Use xllamacpp by default

codingl2k1:
export USE_XLLAMACPP=0 to switch back to the llama-cpp-python backend.


qinxuye:
About the codes related to llama-cpp-python. Like:

https://github.com/xorbitsai/inference/blob/aae95e7d6498917c74e9ac34b4cbe36c0e0a8b0b/xinference/types.py#L406-L421

Shall us process in this PR?

fyrestone:
> About the codes related to llama-cpp-python. Like:
> 
> https://github.com/xorbitsai/inference/blob/aae95e7d6498917c74e9ac34b4cbe36c0e0a8b0b/xinference/types.py#L406-L421
> 
> Shall us process in this PR?

There will be another PR to remove the llama-cpp-python.

qinxuye (approved):
LGTM

PR #3202, opened 2025-04-08
BUG: fix stop hang for vllm engine

qinxuye:
Before the processes(subpool) may be terminated for many times, this may cause hang when stopping vllm workers.

In this PR, vllm worker actors are just destroyed instead of killed, and prevent shutdown called repeatedly in `__del__`.


amumu96 (approved):
LGTM

PR #3233, opened 2025-04-11
FEAT: support getting download progress and cancel download

qinxuye:
This PR basically did two things:

1. Support getting progress when launching models
2. Support cancelling when downloading.

This main idea is to patch `tqdm.update`. All the hub libraries like huggingface_hub and modelscope all used tqdm as progress bar vendor. Important part is to identify which tqdm is the main progress( like downloading 11 files) and the file progress(for a file 300M of 500M downloaded), the progress is calculated from them. Further more, when user cancels the launching, the downloader will know and raise error in `tqdm.update`, thus the downloading can be cancelled as expected.

This PR only supports command lines, web UI needs to adapt new APIs about getting progress and cancelling.

Demo:

https://github.com/user-attachments/assets/99905b25-33d3-4e57-b8df-4027377aba3b

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -469,6 +469,26 @@ class RESTfulAPI(CancelMixin):
             ),
         )
         self._router.add_api_route(
+            "/v1/models/{model_uid}/progress",
+            self.get_launch_model_progress,
+            methods=["GET"],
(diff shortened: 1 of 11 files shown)


amumu96 (approved):
LGTM

PR #3259, opened 2025-04-15
FEAT: add ggufv2 support for vLLM

harryzwh:
Add ggufv2 support for vLLM. 
A 671B Deepseek-R1 model from [here](https://huggingface.co/unsloth/DeepSeek-R1-GGUF/tree/main/DeepSeek-R1-UD-Q2_K_XL) has been tested.

Due to limitation from vLLM, currently only supports single gguf model (merge by llama.cpp's tool) and requires loading config and tokenizer from config files. More details can be found in [here](https://github.com/vllm-project/vllm/pull/13167).

The testing config is attached below.
微信截图_20250415175050

Note that this feature will be available only when vllm>=0.8.2, thus requires manual update of vllm.

### DIFF
--- a/xinference/model/llm/__init__.py
+++ b/xinference/model/llm/__init__.py
@@ -57,7 +57,7 @@ from .llm_family import (
 
 def check_format_with_engine(model_format, engine):
     # only llama-cpp-python support and only support ggufv2
-    if model_format in ["ggufv2"] and engine != "llama.cpp":
+    if model_format in ["ggufv2"] and engine not in ["llama.cpp", "vLLM"]:
         return False
     if model_format not in ["ggufv2"] and engine == "llama.cpp":
         return False
--- a/xinference/model/llm/llm_family.py
+++ b/xinference/model/llm/llm_family.py
@@ -17,7 +17,6 @@ import os
 from threading import Lock
 from typing import Any, Dict, List, Optional, Set, Tuple, Type, Union
 
-import huggingface_hub
 from typing_extensions import Annotated, Literal
 
 from ..._compat import (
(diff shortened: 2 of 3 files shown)


Jun-Howie:
Thank you for your contribution. I noticed that vllm does not support multiple gguf files and needs to be merged manually. In addition, we will consider automatically downloading and merging gguf files.

harryzwh:
> Thank you for your contribution. I noticed that vllm does not support multiple gguf files and needs to be merged manually. In addition, we will consider automatically downloading and merging gguf files.

In addition to gguf files, vLLM now still need to load somethings from the original json files instead of that embedded in the gguf file. So, we may want to see if vLLM going to fully support gguf format or we need to implement the download of config file as well. Alternatively, have a huggingface/modelscope repo that contains a single gguf file and required config files will make the implemation easier.

As my GPU server is fully disconnected from Internet, I just manually download the needed files and place them in the corresponding directories accordingly so that they can be recognized as cached models.

suntao2015005848:
+1

qinxuye:
We may not be able to catch up with upcoming v1.5.0.

I will go back here next week and see how we can make an effort to optimize this in order to merge it.

qinxuye:
I made some modification, if the users do not provide with `tokenizer`, Xinference will try to download the corresponding huggingface model (only tokenizer and config.json), and specify the params for the user.

suntao2015005848:
When will the version be released?

qinxuye:
> When will the version be released?

Should be April 30.

amumu96 (approved):
LGTM

PR #3278, opened 2025-04-17
ENH: Set gradio default concurrency to cpu count

codingl2k1:
Fixes: https://github.com/xorbitsai/inference/issues/3093


qinxuye (approved):
LGTM

PR #3311, opened 2025-04-21
BLD: support flash-attn at Dockerfile

amumu96:
support flash-attn at Dockerfile to accelerate transformers inference


qinxuye (approved):
LGTM

PR #3361, opened 2025-04-30
BUG: fix failure of clearing resources when loading model failed

qinxuye:
When model loads in a thread, e.g vllm xinf distributed executor, when loading failed which raised in `wait_for_load`, the worker status is not recorded yet, thus the `terminate_model` will skip, resources is occupied on workers already, this PR fixed it.


Damonpkl:
目前最新版本，好像仍然有这个问题，model is already  in model list,无法通过xinference terminal 去删掉模型 @qinxuye

Issue #3208, opened 2025-04-09
安装 CosyVoice2-0.5B 报错提示[Errno 2] No such file or directory: '/home/tank/xinference/cache/CosyVoice2-0.5B/cosyvoice.yaml


zhangguohan:
### System Info / 系統信息

python-3.11.1 

### Running Xinference with Docker? / 是否使用 Docker 运行 Xinfernece？

- [ ] docker / docker
- [x] pip install / 通过 pip install 安装
- [ ] installation from source / 从源码安装

### Version info / 版本信息

xinference, version 1.4.1

### The command used to start Xinference / 用以启动 xinference 的命令

可以直接安装

### Reproduction / 复现过程

    ^^^^^^^^^^^^^^^^^
  File "/home/tank/miniconda3/envs/xinference/lib/python3.11/site-packages/ruamel/yaml/constructor.py", line 183, in construct_non_recursive_object
    data = constructor(self, tag_suffix, node)
    ^^^^^^^^^^^^^^^^^
  File "/home/tank/miniconda3/envs/xinference/lib/python3.11/site-packages/hyperpyyaml/core.py", line 472, in _construct_object
    raise ImportError("There is no such class as %s" % callable_string)
    ^^^^^^^^^^^^^^^^^
ImportError: [address=0.0.0.0:41645, pid=2553233] There is no such class as cosyvoice.flow.decoder.CausalConditionalDecoder
2025-04-09 10:52:42,104 uvicorn.access 2417451 INFO     14.23.51.58:55585 - "POST /v1/models HTTP/1.1" 500
2025-04-09 10:55:39,221 xinference.core.worker 2417588 INFO     [request 1e562210-14ee-11f0-859d-bcfce7637c8f] Enter launch_builtin_model, args: <xinference.core.worker.WorkerActor object at 0x7fe165b38350>, kwargs: model_uid=CosyVoice2-0.5B-0,model_name=CosyVoice2-0.5B,model_size_in_billions=None,model_format=None,quantization=None,model_engine=None,model_type=audio,n_gpu=auto,request_limits=None,peft_model_config=None,gpu_idx=None,download_hub=modelscope,model_path=None,xavier_config=None
2025-04-09 11:08:48,926 xinference.core.model 2553471 INFO     Start requests handler.
2025-04-09 11:08:49,121 xinference.model.audio.cosyvoice 2553471 INFO     Loading CosyVoice model, compile=False...
[...]

PR #3365, opened 2025-04-30
ENH: Update cosyvoice

codingl2k1:
- Please use the latest model from modelscope, as the one available on huggingface is outdated.
- When loading the model for streaming inference, set use_flow_cache=True; the default mode is non-stream.
- Cosyvoice code version: https://github.com/FunAudioLLM/CosyVoice/commit/587604b2b433bc350c344b4b181b47249b54faf2
- Forward the exception from the model to the RESTful API handler to prevent the stream request from hanging during a server exception.

Fixes: https://github.com/xorbitsai/inference/issues/3208


qinxuye:
Can we add a pre check to tell the users maybe he/she forgot to add the option when launching model for streaming.

qinxuye (approved):
LGTM

PR #3368, opened 2025-05-01
add support for olmocr

harryzwh:
Add model support for [olmocr](https://huggingface.co/allenai/olmOCR-7B-0225-preview/tree/main) which is finetuned from Qwen2-VL-7B-Instruct.

### DIFF
--- a/xinference/model/llm/llm_family.json
+++ b/xinference/model/llm/llm_family.json
@@ -10152,5 +10152,39 @@
         "numpy==1.26.4"
       ]
     }
+  },
+  {
+    "version":1,
+    "context_length":32768,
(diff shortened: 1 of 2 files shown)


xinference/model/llm/llm_family.json
```diff
@@ -10152,5 +10152,39 @@
         "numpy==1.26.4"
       ]
     }
+  },
+  {
+    "version":1,
+    "context_length":32768,
+    "model_name":"olmocr-0225-preview",
+    "model_famil":"qwen2-vl-instruct",
```
Minamiyama, line 10160: here's a typo, should be `model_family`

PR #3400, opened 2025-05-07
Modify the reference transformer version to transformers>=4.51.3

Zhoneym:
Modify the reference transformer version to transformers>=4.51.3 for compatibility with Qwen3


qinxuye:
This may cause more dependency conflicts if we limit the version more strictly.

We recommend to solve version by [model virtualenv](https://inference.readthedocs.io/en/latest/models/virtualenv.html) in the future

PR #3412, opened 2025-05-08
ENH: llama.cpp backend use xllamacpp

codingl2k1:


qinxuye:
IIRC, the `types` needs to be fixed.

PR #3442, opened 2025-05-13
FEAT: llama.cpp backend support multimodal

codingl2k1:
- llama.cpp backend support mutimodal projectors
- gemma3 gguf support multimodal
- makes the inference error clear

Fixes: https://github.com/xorbitsai/inference/issues/3416


xinference/model/llm/core.py
```diff
@@ -152,6 +152,7 @@ def __init__(
         llm_spec: "LLMSpecV1",
         quantization: Optional[str],
         model_path: Optional[str] = None,
+        multimodal_projector: Optional[str] = None,
```
qinxuye, line 155: This seemed not used? This options may be illustrated in `to_dict` and then we can expose this option to users on web UI.

xinference/model/llm/llama_cpp/core.py
```diff
@@ -131,6 +132,10 @@ def load(self):
                 legacy_model_file_path = os.path.join(self.model_path, "model.bin")
                 if os.path.exists(legacy_model_file_path):
                     model_path = legacy_model_file_path
+            if self.model_spec.multimodal_projectors:
+                mmproj = os.path.join(
+                    self.model_path, self.model_spec.multimodal_projectors[0]
```
qinxuye, line 137: Maybe we can let the users pass `multimodel_projector` to specify the exact one?

xinference/api/restful_api.py
```diff
@@ -1014,6 +1015,7 @@ async def launch_model(
             "replica",
             "n_gpu",
             "request_limits",
+            "multimodal_projector",
```
qinxuye, line 1018: This param is too specific, can we just put it into the kwargs? and let the llama.cpp engine handle it.

xinference/client/restful/restful_client.py
```diff
@@ -970,6 +970,7 @@ def launch_model(
         replica: int = 1,
         n_worker: int = 1,
         n_gpu: Optional[Union[int, str]] = "auto",
+        multimodal_projector: Optional[str] = None,
```
qinxuye, line 973: Ditto.

qinxuye (commented):
I think we'd better put `multimodal_projector` into extra params.

Besides, the UI has added this option.

xinference/web/ui/src/locales/zh.json
```diff
@@ -53,6 +53,7 @@
     "modelFormat": "模型格式",
     "modelSize": "模型大小",
     "quantization": "量化",
+    "multimodelProjector": "多模型投影器",
```
qinxuye, line 56:
```suggestion
    "multimodelProjector": "多模态投影器",
```

qinxuye (approved):
LGTM

PR #3447, opened 2025-05-14
FEAT: Support Skywork-OR1

Jun-Howie:


qinxuye:
Since skywork-or1 is out, shall us just remove preview?

Jun-Howie:
The preview version and the official version are both based on the DeepSeek-R1-Distill-Qwen series. I think we can remove the preview version

Jun-Howie:
Delete in the next version

qinxuye (approved):
LGTM

PR #3527, opened 2025-05-27
ENH: remove check when specified GPU index for vllm

### LINKED ISSUE #1848: 当加载了llm模型Qwen1.5-14B-Chat-GPTQ-int4后，就不能加载m3e-base向量模型了
错误截图：
图片

错误：
2024-07-11 21:53:09 RuntimeError: [address=0.0.0.0:54846, pid=45] User specified GPU index 0 has been occupied with a vLLM model: Qwen1.5-14B-Chat-GPTQ-int4-1-0, therefore cannot allocate GPU memory for a new model.

操作：当加载了llm模型Qwen1.5-14B-Chat-GPTQ-int4后，就不能加载m3e-base向量模型了

欢迎大家加q群讨论一下：27831318

kota-iizuka:
closes #1848 #2260 #2708 #3041

In the current implementation, after a user deploys a vLLM model, it is not possible to deploy another model using the same GPU. In particular, if you deploy a vLLM model to a specific device by setting the `--gpu_memory_utilization` parameter to a small value, an error in Xinference will stop the launch, even though the model can actually be launched in the remaining memory. This is unnatural behavior considering that, for example, you can deploy a vLLM model on the same GPU after deploying a Transformers model (the result is the same, but the order of deployment changes whether it can be deployed or not).

In this PR, we have removed this error only when the user specifies the GPU ID. This allows advanced users who want to launch multiple models on the same GPU to set it, while leaving the behavior unchanged for general users who do not intend to customize GPU control.

## Steps to reproduce the error

```
$ docker run -e XINFERENCE_ENABLE_VIRTUAL_ENV=1 -p 9997:9997 --gpus all xprobe/xinference:v1.6.0.post1 xinference-local -H 0.0.0.0

$ xinference launch --model-name qwen3 --model-type LLM --model-engine vLLM --model-format pytorch --size-in-billions 0_6 --quantization none --n-gpu auto --replica 1 --n-worker 1 --gpu-idx 0 --gpu_memory_utilization 0.4

$ xinference launch --model-name qwen3 --model-type LLM --model-engine Transformers --model-format pytorch --size-in-billions 0_6 --quantization none --n-gpu auto --replica 1 --n-worker 1 --gpu-idx 0

RuntimeError: [address=0.0.0.0:46013, pid=93] GPU index 0 has been occupied with a vLLM model: qwen3-0, therefore cannot allocate GPU memory for a new model.
```

### DIFF
--- a/xinference/core/tests/test_worker.py
+++ b/xinference/core/tests/test_worker.py
@@ -389,9 +389,9 @@ async def test_launch_model_with_gpu_idx(setup_pool):
     await worker.launch_builtin_model(
         "embedding_3", "mock_model_name", None, None, None, "embedding", n_gpu=1
     )
-    # should be on gpu 1
+    # should be on gpu 0
     await worker.launch_builtin_model(
-        "rerank_4", "mock_model_name", None, None, None, "rerank", gpu_idx=[1]
+        "rerank_4", "mock_model_name", None, None, None, "rerank", gpu_idx=[0]
     )
     # should be on gpu 2
     await worker.launch_builtin_model(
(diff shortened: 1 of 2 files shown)


kota-iizuka:
@qinxuye I have revised the test scenario to Proposal 3. Since the expected deployment destination of rerank_7 has changed from GPU2 to GPU1, I have also revised the related tests.

One of the tests has stopped due to a timeout. The other tests have passed, so I believe there is no problem with the logic. Could you please re-run the test? (I probably cannot run it due to insufficient permissions.)

ChengjieLi28 (approved):

PR #3601, opened 2025-06-10
FEAT: support deepseek-r1-0528 Mixed quantization

Jun-Howie:
This repository delivers an Int4 + selectively-Int8 GPTQ model: only layers that are highly sensitive to quantization remain in Int8, while the rest stay Int4—preserving generation quality with minimal file-size overhead.DeepSeek-R1-0528

Preliminary trials show that converting the entire model to pure Int4 (AWQ/GPTQ) under the quantization layout used in vLLM’s current DeepSeek-R1 implementation degrades inference accuracy and can produce faulty outputs. Layer-wise fine-grained quantization substantially mitigates this issue.


qinxuye (approved):
LGTM

PR #3605, opened 2025-06-10
DOC: Update CosyVoice doc

codingl2k1:


doc/source/models/model_abilities/audio.rst
```diff
@@ -293,6 +293,12 @@ Use the fixed tone color of ``seed_2155`` to generate speech.
 CosyVoice Usage
 ~~~~~~~~~~~~~~~
 
+CosyVoice has three different models:
```
qinxuye, line 296: Maybe here we can tell users there are two versions of cosyvoice, or they might ignore there is cosyvoice 2.0

PR #3665, opened 2025-06-18
BUG: Fix max_tokens value in Qwen3 Reranker

yasu-oh:
## Description
Corrects the max_tokens setting for Qwen3 Reranker from the mistaken 40960 to the intended 32768.

## Why
The inflated limit could mask overflow errors and misrepresent the model’s true context window.

## Change
config: max_tokens = 32768 (was 40960)

No other code or interface changes.


qinxuye (approved):
LGTM

PR #3666, opened 2025-06-18
Add 4-sample micro-batching to Qwen-3 reranker to reduce GPU memory

yasu-oh:
### Motivation

The current Qwen-3 code concatenates *all* documents in a single batch, tokenises them and runs one forward pass.  For long contexts (32 768 tokens) this inflates the KV cache linearly with the number of documents and results in out-of-memory errors on commodity GPUs.

### What this PR does

* **Fixed micro-batch size of 4**
  Splits `documents` into slices of four before tokenisation / inference and aggregates logits afterwards.
* **No API change**
  Call-sites remain untouched; the change is internal to `RerankModel.rerank()`.

```python
micro_bs = 4
for i in range(0, len(documents), micro_bs):
    ...
    similarity_scores.extend(self.compute_logits(inputs))
```

### Impact

| Scenario               | Peak VRAM (before) | Peak VRAM (after) | Δ     |
| ---------------------- | ------------------ | ----------------- | ----- |
| 100 docs × 32 k tokens | ~ 351 GB | ~ 14 GB | -96 % |

### DIFF
--- a/xinference/model/rerank/core.py
+++ b/xinference/model/rerank/core.py
@@ -370,13 +370,18 @@ class RerankModel:
                 )
                 return output
 
-            pairs = [
-                format_instruction(kwargs.get("instruction", None), query, doc)
-                for doc in documents
-            ]
-            # Tokenize the input texts
-            inputs = self.process_inputs(pairs)
-            similarity_scores = self.compute_logits(inputs)
+            # reduce memory usage.
+            micro_bs = 4
+            similarity_scores = []
+            for i in range(0, len(documents), micro_bs):
+                sub_docs = documents[i : i + micro_bs]
+                pairs = [
+                    format_instruction(kwargs.get("instruction", None), query, doc)
+                    for doc in sub_docs
+                ]
+                # Tokenize the input texts
+                inputs = self.process_inputs(pairs)
+                similarity_scores.extend(self.compute_logits(inputs))
         else:
             # Related issue: https://github.com/xorbitsai/inference/issues/1775
             similarity_scores = self._model.compute_score(


yasu-oh:
In addition to this change, I’d like to discuss something.
While it’s true that Qwen3-Reranker 0.6B, 4B, and 8B are models with a context length of 32k, in practical use, specifying 32k leads to very high memory consumption.
Wouldn't it be better to set max_length = 8192, as shown in the Qwen3-Reranker reference code?

qinxuye:
> In addition to this change, I’d like to discuss something. While it’s true that Qwen3-Reranker 0.6B, 4B, and 8B are models with a context length of 32k, in practical use, specifying 32k leads to very high memory consumption. Wouldn't it be better to set max_length = 8192, as shown in the Qwen3-Reranker reference code?

A few solutions I think.

1. Support passing extra param like `max_tokens=8192` when launching model, this might require some modification of the code in which passing the extra params in `self.model_config` to the model.
2. The main branch has updated, now we can register a custom model with specifying the `max_tokens`, you can create a custom model.

yasu-oh:
Yes, I think allowing users to change `max_tokens` and `micro_bs` parameters through the web GUI would greatly improve usability!

qinxuye:
Would you like to add support for those arguments in this PR?

qinxuye (approved):
LGTM

qinxuye:
Let's merge first, we can do the optimization in the other PRs.

yasu-oh:
Thank you for merging the PR.
Due to a high workload on my end, I have been unable to allocate personal development time, and therefore could not include the WEB UI changes in this PR.

PR #3693, opened 2025-06-23
FEAT: support MLU device

nan9126:
added support for MLU, tested LLM(vllm), embedding(transformers), rerank(transformers)


qinxuye:
Could you install pre-commit via pip and run `pre-commit install` at the project level, this will install the tools for lint.

Then when you commit it will amend many things automatically, e.g. black, you can check what you can do to pass the lint.

nan9126:
Thanks for the feedback, I've installed `pre-commit`, fixed the lint issues, and force-pushed the updated commits.

PR #3712, opened 2025-06-25
ENH: Limit default n_parallel for llama.cpp backend

codingl2k1:


xinference/model/llm/llama_cpp/core.py
```diff
@@ -162,7 +162,7 @@ def load(self):
             if self.model_family.chat_template:
                 params.chat_template = self.model_family.chat_template
             # This is the default value, could be overwritten by _llamacpp_model_config
-            params.n_parallel = os.cpu_count()
+            params.n_parallel = max(8, os.cpu_count() or 1)
```
qinxuye, line 165: Should it be `min`, if cpu_count is 32, this is still 32?

PR #3718, opened 2025-06-27
BUG: fix `<think>` wrongly added when set `chat_template_kwargs {"enable_thinking": False}`

### LINKED ISSUE #3647: Qwen3: enable_thinking: false still returns <think> tags
### Feature request / 功能建议

When using Qwen3 models with chat_template_kwargs: {"enable_thinking": false}, the expectation is that the model will completely skip the thinking phase and avoid returning <think> tags.

However, even when enable_thinking is explicitly set to false, the model still returns <think> tags in the output. This causes issues for downstream tasks that expect no thinking content.

I propose updating the logic so that when enable_thinking is set to false, the model entirely omits any <think> tag or related content.

### Motivation / 动机

In certain use cases, such as structured task execution or programmatic pipelines, the presence of <think> tags causes unexpected behavior or crashes. Allowing a reliable way to disable the thinking mechanism at runtime via chat_template_kwargs would provide better control and flexibility.

This would also align with the intended purpose of the enable_thinking parameter, ensuring that users can dynamically toggle this feature without needing to modify templates or fork logic.

### Your contribution / 您的贡献

I can help test the fix once available.

qinxuye:
Fixes #3647

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -130,7 +130,7 @@ jobs:
             conda install -c conda-forge "ffmpeg<7"
             pip install "mlx>=0.22.0"
             pip install mlx-lm
-            pip install mlx-vlm
+            pip install "mlx-vlm<0.2.0"
             pip install mlx-whisper
             pip install f5-tts-mlx
             pip install qwen-vl-utils!=0.0.9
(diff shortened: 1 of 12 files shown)


amumu96 (commented):
LGTM

amumu96 (approved):

PR #3809, opened 2025-07-18
BUG: pre check to prevent from list index out of range for FunASR family models

leslie2046:
ret = await asyncio.to_thread(fn, *args, **kwargs)
  File "/usr/lib/python3.10/asyncio/threads.py", line 25, in to_thread
    return await loop.run_in_executor(None, func_call)
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/opt/inference/xinference/model/audio/funasr.py", line 140, in transcriptions
    text = rich_transcription_postprocess(result[0]["text"])
IndexError: list index out of range

### DIFF
--- a/xinference/model/audio/funasr.py
+++ b/xinference/model/audio/funasr.py
@@ -137,6 +137,10 @@ class FunASRModel:
             result = self._model.generate(  # type: ignore
                 input=f.name, cache={}, language=language, **kw
             )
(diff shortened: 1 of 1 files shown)


qinxuye (approved):
LGTM

PR #3844, opened 2025-07-25
add mlu device check

nan9126:
This PR adds support for detecting MLU (Cambricon) devices in the inference logic.

#### Changes:
- `xinference/model/llm/core.py`:
  - Added `_has_mlu_device()` static method using `cnmon info` command to detect MLU presence.

- `xinference/model/llm/vllm/core.py`:
  - Integrated MLU device check by calling `LLM._has_mlu_device()` where applicable.
  - Ensures the model only attempts to run when valid MLU hardware is present.


qinxuye (approved):
LGTM

PR #3911, opened 2025-08-05
Replace @torch.no_grad() with @torch.inference_mode() in Qwen3-Reranker

yasu-oh:
This PR updates the decorator from @torch.no_grad() to @torch.inference_mode() in the Qwen3-Reranker implementation.

torch.inference_mode() is a more optimized context for inference-only workloads, as it disables autograd and activates additional internal optimizations to reduce CPU overhead and improve execution speed. This change should not affect numerical results but can lead to better runtime performance and lower memory usage during inference.


qinxuye (approved):
LGTM, thanks!

PR #3939, opened 2025-08-12
feat: sglang support streaming function call

aniya105:
sglang support streaming function call

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -2249,8 +2249,9 @@ class RESTfulAPI(CancelMixin):
                 )
         if body.tools and body.stream:
             is_vllm = await model.is_vllm_backend()
+            is_sglang = await model.is_sglang_backend()
             if not (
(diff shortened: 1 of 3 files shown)


qinxuye:
I am wondering can we make function calling more engine irrelevant? Maybe we can support streaming tool calls for all engines. What's your opinion?

aniya105:
> I am wondering can we make function calling more engine irrelevant? Maybe we can support streaming tool calls for all engines. What's your opinion?

That's a good idea, but I'm not familiar with other engines and can't be of further assistance. I'm sorry.

qinxuye:
> > I am wondering can we make function calling more engine irrelevant? Maybe we can support streaming tool calls for all engines. What's your opinion?
> 
> That's a good idea, but I'm not familiar with other engines and can't be of further assistance. I'm sorry.

That's OK, let's focus sglang first, we can refactor later.

qinxuye (approved):
LGTM

PR #3983, opened 2025-08-20
ENH:Add Template for Qwen3 Reranker when model_engine = vllm

zhcn000000:
Template from https://docs.vllm.ai/en/latest/examples/offline_inference/qwen3_reranker.html


qinxuye (approved):
LGTM

PR #3988, opened 2025-08-21
ENH: support chat_template_kwargs for llama.cpp

qinxuye:


xinference/model/llm/llama_cpp/core.py
```diff
@@ -312,6 +321,8 @@ def _handle_chat_completion():
                     "tools": tools,
                 }
             )
+            if chat_template_kwargs:
+                data["chat_template_kwargs"] = chat_template_kwargs
```
codingl2k1, line 325: There is a `default_template_kwargs` in CommonParams, it's better to set the kwargs during the initialization of the server.

qinxuye: This option could be a generate config.

codingl2k1 (approved):

PR #4019, opened 2025-08-28
DOC:Update version 2 about custom models

OliverBryant:
Replace the json block about how to define a custom LLM model


qinxuye (approved):
LGTM

PR #4022, opened 2025-08-29
FEAT: [Model] Support DeepSeek-V3.1 Quantization and tool

Jun-Howie:
```bash
# ❗there are glitches with vllm 0.10.1.1, still looking for resolutions❗
# ❗downgrade vllm for now ❗
pip install vllm==0.9.2 transformers==4.53.0

SITE_PACKAGES=$(pip -V | awk '{print $4}' | sed 's/\/pip$//')
# ❗patch up AWQ MoE quant config, otherwise some modules cannot be properly loaded❗
cp awq_marlin.py "$SITE_PACKAGES/vllm/model_executor/layers/quantization/awq_marlin.py"
# ❗patch up for fp32 e_score_correction_bias, see https://www.github.com/vllm-project/vllm/pull/23640❗
cp deepseek_v2.py "$SITE_PACKAGES/vllm/model_executor/models/deepseek_v2.py"
```

### DIFF
--- a/xinference/model/llm/llm_family.json
+++ b/xinference/model/llm/llm_family.json
@@ -5314,6 +5314,116 @@
   {
     "version": 2,
     "context_length": 131072,
+    "model_name": "Deepseek-V3.1",
+    "model_lang": [
+      "en",
+      "zh"
+    ],
+    "model_ability": [
+      "chat",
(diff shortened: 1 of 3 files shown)


qinxuye (approved):
LGTM

Issue #3991, opened 2025-08-22
Add support for the Anthropic API format


gahoo:
### Feature request / 功能建议

The Anthropic API ecosystem is gaining popularity. Adding support for the Anthropic API format can help integrate xinference into this ecosystem, such as with tools like Claude Code.

### Motivation / 动机

Use Claude Code with alternative models.

### Your contribution / 您的贡献

None

qinxuye:
Thanks, this is important to make claude code running on top of Xinference.

PR #4037, opened 2025-09-04
FEAT: add support for the Anthropic API format

OliverBryant:
fix: #3991


qinxuye:
Could you add a unittest to test if anthropic API works as expected?

PR #4038, opened 2025-09-05
FEAT: [model] support hexgrad/Kokoro-82M-v1.1-zh

JavisPeng:
Support Kokoro-82M-v1.1-zh

测试Kokoro-82M对纯中文支持比较好，但若中文中包含英文字母，模型基本会跳过不发音。项目需要本人只修改了kokoro.py源码以支持使用 Kokoro-82M-v1.1-zh。

### DIFF
--- a/xinference/model/audio/kokoro.py
+++ b/xinference/model/audio/kokoro.py
@@ -81,7 +81,7 @@ class KokoroModel:
         logger.info("Launching Kokoro model with language code: %s", lang_code)
         self._model = KPipeline(
             lang_code=lang_code,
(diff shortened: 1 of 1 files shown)


qinxuye:
Please install pre-commit, and run `pre-commit install` at your project level to enable it. It will format the code or do lint check before you commit.

I only see the implementation but you seems not add this model to model_spec.json?

And you need to modify to let the model be passed to the right implementation.

https://github.com/xorbitsai/inference/blob/55e5f73d75d607c6440034c05d5314dbbc93cf35/xinference/model/audio/core.py#L184-L187

PR #4063, opened 2025-09-15
BUG: Optimize rerank model lookup logic and add support for video model type

amumu96:
/v1/model_registrations/{model_type}/{model_name} GET
- optimize rerank model lookup logic
- support for video model type


qinxuye (approved):
LGTM

PR #4076, opened 2025-09-17
BUG: fix register_model when model name is duplicated

llyycchhee:
check model name is duplicate before worker register model

### DIFF
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -883,9 +883,9 @@ class SupervisorActor(xo.StatelessActor):
                     pass
                 else:
                     raise e
-            except Exception as e:
-                logger.error(f"Get model registration failed. Error: {e}")
-                raise e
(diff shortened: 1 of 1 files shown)


xinference/core/supervisor.py
```diff
@@ -867,6 +867,26 @@ async def register_model(
...
+                        f"Model {model_spec.model_name} already registered"
+                    )
+            except ValueError as e:
+                if "not found" in str(e):
+                    pass
+                else:
+                    raise e
+            except Exception as e:
+                logger.error(f"Get model registration failed. Error: {e}")
+                raise e
```
qinxuye, line 888:
```suggestion
            except Exception:
                logger.error("Get model registration failed.", exc_info=True)
                raise
```

qinxuye: {e} will only record error without traceback.

Issue #4067, opened 2025-09-16
注册了灵活模型后无法显示已经注册的灵活模型


leslie2046:
### System Info / 系統信息

Ubuntu 22.04 LTS

### Running Xinference with Docker? / 是否使用 Docker 运行 Xinfernece？

- [x] docker / docker
- [ ] pip install / 通过 pip install 安装
- [ ] installation from source / 从源码安装

### Version info / 版本信息

v1.10.0

### The command used to start Xinference / 用以启动 xinference 的命令

services:
  xinference-supervisor:
    image: "${IMAGE}"
    container_name: xinference-supervisor
    network_mode: host
    environment:
      - NVIDIA_VISIBLE_DEVICES=all
      - XINFERENCE_HOME=/data
      - XINFERENCE_MODEL_SRC=modelscope
    volumes:
      - /root/data:/data
    command: >
      xinference-supervisor -H 192.168.1.33 --port 9997 --log-level DEBUG

services:
  xinference-worker1:
    image: "${IMAGE}"
    container_name: xinference-worker1
    restart: always
    network_mode: host
    environment:
      - XINFERENCE_HOME=${XINFERENCE_HOME}
      - XINFERENCE_MODEL_SRC=${XINFERENCE_MODEL_SRC}
    volumes:
      - ${CACHE_PATH}:/data
    command: >
      xinference-worker -e "${ENDPOINT}" -H "$HOST" --log-level ${LOG_LEVEL}
    stdin_open: true
    tty: true
  xinference-worker2:
    image: "${IMAGE}"
    container_name: xinference-worker2
    restart: always
    network_mode: host
    environment:
      - XINFERENCE_HOME=${XINFERENCE_HOME}
      - XINFERENCE_MODEL_SRC=${XINFERENCE_MODEL_SRC}
    volumes:
      - ${CACHE_PATH}:/data
    command: >
      xinference-worker -e "${ENDPOINT}" -H "$HOST" --log-level ${LOG_LEVEL}
    stdin_open: true
    tty: true

### Reproduction / 复现过程

Traceback (most recent call last):
  File "/opt/conda/lib/python3.11/site-packages/xinference/core/utils.py", line 93, in wrapped
    ret = await func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/xinference/core/supervisor.py", line 610, in list_model_registrations
[...]

PR #4085, opened 2025-09-19
FEAT: support list flexible model via webui and cmdline

leslie2046:
close https://github.com/xorbitsai/inference/issues/4067


qinxuye:
LInt failed, needs to be fixed.

qinxuye (approved):
LGTM

PR #4102, opened 2025-09-26
FEAT: [UI] add model feature in Launch Model list.

yiboyasss:


xinference/model/utils.py
```diff
@@ -709,3 +724,446 @@ def _wrapper(self, *args, **kwargs):
...
+    return register_builtin_models_unified(
+        model_type="llm",
+        flatten_func=flatten_quantizations,
+        model_class=LLMFamilyV2,
+        builtin_registry={},  # Special handling
+        special_handling=llm_special_handling,
+    )
+
+
+def register_embedding_builtin_models():
```
qinxuye: Maybe these functions should be put into the specified model type dir?

Issue #4168, opened 2025-10-23
add DeepSeek-OCR


yz342:
### Feature request / 功能建议

add DeepSeek-OCR, with v1/images/ocr api support

### Motivation / 动机

best OCR

### Your contribution / 您的贡献

NA

fsea:
同求

dmarkubex:
need too

PR #4187, opened 2025-10-30
FEAT: [model] DeepSeek-OCR

OliverBryant:
fix: #4168

### DIFF
--- a/xinference/ui/gradio/media_interface.py
+++ b/xinference/ui/gradio/media_interface.py
@@ -1301,7 +1301,7 @@ class MediaInterface:
                                 "error", "OCR visualization failed"
                             )
                             # Return formatted error message for Markdown
-                            error_md = f"**错误**: {error_msg}"
+                            error_md = f"**Error**: {error_msg}"
                             return error_md, "", ""
                     elif isinstance(response, str):
                         # Handle string response from original model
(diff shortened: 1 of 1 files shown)


xinference/api/restful_api.py
```diff
@@ -2271,6 +2271,11 @@ async def create_ocr(
         self,
         model: str = Form(...),
         image: UploadFile = File(media_type="application/octet-stream"),
+        prompt: str = Form(None),
+        model_size: str = Form(None),
+        save_results: bool = Form(None),
+        save_dir: Optional[str] = Form(None),
```
qinxuye: `save_dir` is a directory that exists in server, this should not be exposed to users.

xinference/api/restful_api.py
```diff
@@ -2271,6 +2271,11 @@ async def create_ocr(
         self,
         model: str = Form(...),
         image: UploadFile = File(media_type="application/octet-stream"),
+        prompt: str = Form(None),
```
qinxuye: I think all the added arguments could be put into kwargs.

xinference/client/restful/async_restful_client.py
```diff
@@ -619,6 +668,75 @@ async def ocr(self, image: Union[str, bytes], **kwargs):
         await _release_response(response)
         return response_data
 
+    async def visualize_ocr(
```
qinxuye: This method is not needed now.

xinference/client/restful/restful_client.py
```diff
@@ -535,11 +535,60 @@ def inpainting(
         response_data = response.json()
         return response_data
 
-    def ocr(self, image: Union[str, bytes], **kwargs):
+    def ocr(
+        self,
+        image: Union[str, bytes],
+        prompt: str = "<image>\nFree OCR.",
```
qinxuye: Ditto, this is too specific, I think we don't need to change the APIs, just put everything else into kwargs.

qinxuye (commented):
I left some comments.

xinference/ui/gradio/media_interface.py
```diff
@@ -1255,19 +1259,248 @@ def extract_text_from_image(
...
+                    if isinstance(response, dict):
+                        if response.get("success"):
+                            text_result = response.get("text", "No text extracted")
+
+                            # Debug: Check if text is empty
+                            if not text_result or not text_result.strip():
+                                logger.warning("OCR returned empty text")
+                                logger.warning(f"Full response: {response}")
+                                # Return a helpful message instead of empty result
+                                text_result = """**OCR 识别完成，但未检测到文本内容**
```
qinxuye: The language in gradio should be English.

qinxuye (approved):
LGTM

PR #4352, opened 2025-12-11
ENH: support lauching with --device cpu

hubutui:


qinxuye:
Have you tested it?

hubutui:
yes, it works as expect.

PR #4355, opened 2025-12-12
BUG: [UI] fix the bug in the copy function.

yiboyasss:
Fix the bug where copying fails in MUI Dialog and Drawer components when the browser is in an insecure context.


qinxuye (approved):
LGTM

PR #4357, opened 2025-12-12
ENH: add glm-4.5 tool calls support && vllm StructuredOutputsParams support

OliverBryant:
fix: #4266


xinference/model/llm/vllm/core.py
```diff
@@ -1233,6 +1296,14 @@ async def async_generate(
             request_id = str(uuid.uuid1())
 
         assert self._engine is not None
+        start_wall_time = time.time()
+        start_perf = time.perf_counter()
+        logger.info(
```
qinxuye: This could be logger.debug.

PR #4358, opened 2025-12-13
FEAT: multi replicas on a single GPU && add launch strategy

OliverBryant:
Removed the restriction that only one instance could be launched per GPU && Added launch strategy
fix: #4104


suercxx:
Server error: 400 - Invalid input. Allocated gpu must be a multiple of replica.
我有多卡，想在一个指定的卡上部署，多个embing模型副本、但是报错、如上，该怎么解决

PR #4403, opened 2025-12-23
FEAT: continuous batching support for MLX chat models

qinxuye:
This PR added support for batching MLX chat models.

Note:

1. Distributed inference is not supported for batching for now.
2. VLM can enable continuous batching only when images have same shape, this is almost useless, hence, vision models are excluded as well.


amumu96 (approved):
LGTM

Issue #4402, opened 2025-12-23
not support minicpm4.5 vllm?


worldlanguage:
**not support minicpm4.5 vllm?**

worldlanguage:
only support minicpm4.5

worldlanguage:
@XprobeBot @qinxuye

qinxuye:
OK, looks like vllm 0.10.2 supported it, @OliverBryant could you add the support?

PR #4405, opened 2025-12-24
FEAT: Add the architectures field for llm model launch

OliverBryant:
FEAT: Add the architectures field to determine which model function to launch.
FEAT: add minicpm 4.5 vllm support
fix: #4402


qinxuye (approved):
LGTM

PR #4414, opened 2025-12-26
FEAT: support rerank async batch

llyycchhee:
After testing, the time required to concurrently send 1,000 requests was reduced from 43 seconds to 6 seconds.


qinxuye (approved):
LGTM

PR #4425, opened 2026-01-04
FEAT: Support MThreads (MUSA) GPU

yeahdongcn:
This PR adds support for Moore Threads (MUSA) GPU platform, expanding Xinference's hardware compatibility.

### DIFF
--- a/xinference/device_utils.py
+++ b/xinference/device_utils.py
@@ -63,6 +63,7 @@ def is_mlu_available() -> bool:
 def is_musa_available() -> bool:
     try:
         import torch
+        import torch_musa  # noqa: F401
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -50,6 +50,8 @@ install_requires =
     peft<=0.17.1
     timm
     setproctitle
+    mthreads-ml-py>=2.2.8  # For MUSA support
```
qinxuye: We should add a tag for MUSA instead of putting this requirements into must-have requirements.

yeahdongcn: Updated. Thanks.

PR #4429, opened 2026-01-04
FEAT: support auto batch for sentence_transformers rerank

llyycchhee:
The time spent using sentence_transformers engine is reduced by 50%.


qinxuye (approved):
LGTM

PR #4438, opened 2026-01-05
ENH: update model "MiniMax-M2" JSON

OliverBryant:
Automated sync for model "MiniMax-M2" (llm) by user OliverBryant.


qinxuye:
Reap model should not be added to the original one, IMO the original model hub config needs to be reverted.

PR #4454, opened 2026-01-08
ENH:  support vllm mtp & rope scaling

ZhikaiGuo960110:
fix https://github.com/xorbitsai/inference/issues/4453
add two configs and parse logic for extracting dict from webui input json, 
now user can add speculative_config or rope_scaling if needed from webui directly, launch more easily.


ZhikaiGuo960110:
fix rope_scaling issue

ZhikaiGuo960110:
also fix an error in read gpu count

PR #4575, opened 2026-02-04
BUG: vllm reranker model gpu release error

OliverBryant:
After uninstalling the reranker model, the GPU is not released.
fix: #4558


qinxuye (approved):
LGTM

PR #4627, opened 2026-02-26
BLD: [CI] fix windows runner SSL can't found

llyycchhee:
Try to reduce concurrency.


qinxuye:
I don't think it's related to parallelism.

PR #4654, opened 2026-03-04
ENH: update model "qwen3.5" JSON

llyycchhee:
Automated sync for model "qwen3.5" (llm) by user llyycchhee.


qinxuye:
Why are there two PRs for qwen3.5

Issue #2402, opened 2024-10-08
集群模式部署，如果重启supervisor，必须重启所有woker吗？


paradin:
### System Info / 系統信息

docker

### Running Xinference with Docker? / 是否使用 Docker 运行 Xinfernece？

- [ ] docker / docker
- [X] pip install / 通过 pip install 安装
- [ ] installation from source / 从源码安装

### Version info / 版本信息

0.15.2

### The command used to start Xinference / 用以启动 xinference 的命令

分布式场景，正常启动 supervisor 和 worker
supervisor启动指定了supervisor-port
在worker上启动一个模型，如：bge-m3

### Reproduction / 复现过程

重启supervisor，前端无法查看正在运行的模型 bge-m3；模型服务不可用；

### Expected behavior / 期待表现

1. supervisor重启后，已经运行的模型正常
2. 模型服务正常

pkunight:
我也发现了这个问题, 必须先启动supervisor, 后启动worker, 而且此后连接不能中断. 否则即使supervisor成功重启了, worker依然会持续报错连不上supervisor的ip地址.

ak47947:
我也发现这个问题了，如果这个问题不解决，是没法真正集群使用的

nasircsms:
The similar issue I have faced is.. if a worker goes down, the cluster starts giving error.. then have to restart the supervisor, make the workers join again and then run models again. 

Ideally the cluster should keep on running even if a worker worker goes down. If one model was deployed on two workers and one of them is down then it should keep serving from the other worker. If both are down then that model should not be available for serving. 

But cluster should remain available in all cases and continue serving from other workers. 

(using v1.15.0)

leslie2046:
> The similar issue I have faced is.. if a worker goes down, the cluster starts giving error.. then have to restart the supervisor, make the workers join again and then run models again.
> 
> Ideally the cluster should keep on running even if a worker worker goes down. If one model was deployed on two workers and one of them is down then it should keep serving from the other worker. If both are down then that model should not be available for serving.
> 
> But cluster should remain available in all cases and continue serving from other workers.
> 
> (using v1.15.0)

wait for this version and see if solved your problem?

nasircsms:
> > The similar issue I have faced is.. if a worker goes down, the cluster starts giving error.. then have to restart the supervisor, make the workers join again and then run models again.
[...]

PR #4731, opened 2026-03-27
FEAT: reconnect and reconstrcut model replicas after restart supervisor

leslie2046:
close https://github.com/xorbitsai/inference/issues/2402

## Problem solved: 
Resolving the issue that workers must restart and the model must be redeployed after the supervisor restarts.
## Technical solution: 
After the supervisor restarts, the worker automatically reconnects to the supervisor and uploads its own status and model replica status information.

### DIFF
--- a/xinference/core/tests/test_worker.py
+++ b/xinference/core/tests/test_worker.py
@@ -30,9 +30,9 @@ class MockWorkerActor(WorkerActor):
     def __init__(
         self,
         supervisor_address: str,
-        supervisor_endpoint: Optional[str],
         main_pool: MainActorPoolType,
         cuda_devices: List[int],
+        supervisor_endpoint: Optional[str] = None,
     ):
         super().__init__(
             supervisor_address=supervisor_address,


leslie2046:
@qinxuye

nasircsms:
@leslie2046 can you please have a look at https://github.com/xorbitsai/inference/issues/2402#issuecomment-4143539089

qinxuye:
> @leslie2046 can you please have a look at [#2402 (comment)](https://github.com/xorbitsai/inference/issues/2402#issuecomment-4143539089)

Maybe it's a different issue, the solution might be if some replica failed, it should be blacklisted.

nasircsms:
> > @leslie2046 can you please have a look at [#2402 (comment)](https://github.com/xorbitsai/inference/issues/2402#issuecomment-4143539089)
> 
> Maybe it's a different issue, the solution might be if some replica failed, it should be blacklisted.

Yes.. I have tested it with v2.4.0 as well. As soon as a worker becomes unavailable then whole cluster fails. Even if that worker joins again on the same port but still cluster remains unavailable. Need to restart supervisor and all the workers to make the cluster available again.

qinxuye:
> > > @leslie2046 can you please have a look at [#2402 (comment)](https://github.com/xorbitsai/inference/issues/2402#issuecomment-4143539089)
> > 
> > 
> > Maybe it's a different issue, the solution might be if some replica failed, it should be blacklisted.
> 
> Yes.. I have tested it with v2.4.0 as well. As soon as a worker becomes unavailable then whole cluster fails. Even if that worker joins again on the same port but still cluster remains unavailable. Need to restart supervisor and all the workers to make the cluster available again.

I guess the worker is not dead right? Maybe just because the heartbeat is not sent just in time.

nasircsms:
> > > > @leslie2046 can you please have a look at [#2402 (comment)](https://github.com/xorbitsai/inference/issues/2402#issuecomment-4143539089)
> > > 
> > > 
> > > Maybe it's a different issue, the solution might be if some replica failed, it should be blacklisted.
> > 
> > 
> > Yes.. I have tested it with v2.4.0 as well. As soon as a worker becomes unavailable then whole cluster fails. Even if that worker joins again on the same port but still cluster remains unavailable. Need to restart supervisor and all the workers to make the cluster available again.
> 
> I guess the worker is not dead right? Maybe just because the heartbeat is not sent just in time.

In my case worker machine is powered-off or restarted. Although I am running xinference as service so on boot it auto-connect to supervisor again on same port, but even if worker comes back or not the cluster is done for.

qinxuye:
> > > > > @leslie2046 can you please have a look at [#2402 (comment)](https://github.com/xorbitsai/inference/issues/2402#issuecomment-4143539089)
> > > > 
> > > > 
> > > > Maybe it's a different issue, the solution might be if some replica failed, it should be blacklisted.
> > > 
> > > 
> > > Yes.. I have tested it with v2.4.0 as well. As soon as a worker becomes unavailable then whole cluster fails. Even if that worker joins again on the same port but still cluster remains unavailable. Need to restart supervisor and all the workers to make the cluster available again.
> > 
> > 
> > I guess the worker is not dead right? Maybe just because the heartbeat is not sent just in time.
> 
> In my case worker machine is powered-off or restarted. Although I am running xinference as service so on boot it auto-connect to supervisor again on same port, but even if worker comes back or not the cluster is done for.

You can try this PR, it might address this issue.

PR #4739, opened 2026-03-29
FEAT(webui): add localstorage management for model deploy configuration

leslie2046:
## Summary
- Improve launch configuration caching so each model_name can retain multiple saved configurations keyed by model_uid, while still overwriting the latest cache for the same model_name + model_uid pair.
- Add a configuration cache picker in the launch drawer so users can view, load, and delete cached model launch settings directly from the UI.
- Extend i18n strings for the new cache management flow, including cache list, load/delete actions, last-updated metadata, and invalid-cache messaging.
图片


leslie2046:
@qinxuye

qinxuye:
@yiboyasss please review this if you got time.

PR #4747, opened 2026-03-31
bug: fix qwen3 reranker vllm precision

ZhikaiGuo960110:
to fix this issue, and be consistent with vllm reranker dataset precision score, remove this special template for qwen3 reranker family. 
fix https://github.com/xorbitsai/inference/issues/4744

### DIFF
--- a/xinference/model/rerank/vllm/core.py
+++ b/xinference/model/rerank/vllm/core.py
@@ -105,41 +105,11 @@ class VLLMRerankModel(RerankModel, BatchMixin):
         else:
             query_list = query
 
-        if self.model_family.model_name in {
-            "Qwen3-Reranker-0.6B",
-            "Qwen3-Reranker-4B",
-            "Qwen3-Reranker-8B",
-        }:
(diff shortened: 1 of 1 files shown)


XiaoliChan:
@qinxuye Reference to this issue: https://github.com/QwenLM/Qwen3-Embedding/issues/71, include template in VLLM is better than without the template, so I think we shouldn't remove the template... (maybe have a variable to control it)

qinxuye:
@ZhikaiGuo960110 Can you look at this?

ZhikaiGuo960110:
image
after testing, result relevance score is indeed a big difference, without template, the score is indeed close and not precise, we should not remove template.  However, in the scenario of Arguana Rerank dataset, I use evalscope to evaluate Qwen3 0.6B reranker model(with template, score is 0.699)(without template, score is 0.736, in consistent with vllm serve behavior and precision score), so the better way to approach this issue is to have variable control template opening. 
with template, 
image
without template, 
image

PR #4752, opened 2026-04-01
fix: add variable to control template for Qwen3 Reranker Family

ZhikaiGuo960110:
add variable to control template for Qwen3 Reranker Family.  default behavior is enable template

### DIFF
--- a/xinference/model/rerank/vllm/core.py
+++ b/xinference/model/rerank/vllm/core.py
@@ -115,7 +115,7 @@ class VLLMRerankModel(RerankModel, BatchMixin):
         }:
             if enable_qwen3_rerank_template is False:
                 pass
-            if enable_qwen3_rerank_template is True or QWEN3_RERANK_TEMPLATE:
+            elif enable_qwen3_rerank_template is True or QWEN3_RERANK_TEMPLATE:
                 instruction = "Given a web search query, retrieve relevant passages that answer the query"
                 prefix = (
                     "<|im_start|>system\nJudge whether the Document meets the requirements based on"


ZhikaiGuo960110:
https://github.com/QwenLM/Qwen3-Embedding/issues/71  this issue shows why we should still have template. but the discussion in https://github.com/xorbitsai/inference/pull/4747 this shows why is better to have a variable controling template.

XiaoliChan:
Great work!

ZhikaiGuo960110:
template is strongly related to certain types of rerank task. for some task, adding template may have a bad side effect.

qinxuye:
How about adding an extra option to enable this for each request?

and, for environment variable, could you add a doc here: https://github.com/xorbitsai/inference/blob/main/doc/source/getting_started/environments.rst

ZhikaiGuo960110:
For extra option on per request level,  even if the server side has global env variable, not enable qwen3 template, 
XINFERENCE_QWEN3_RERANK_TEMPLATE=0 xinference-local -H 0.0.0.0 -p 9997

but we can pass extra kwargs to make template applied. like this 

import requests
import json
response = requests.post(
    "http://localhost:9997/v1/rerank",
    json={
        "model": "Qwen3-Reranker-0.6B",
        "query": "A man is eating pasta.",
        "documents": [
            "A man is eating food.",
            "A man is eating a piece of bread.",
            "The girl is carrying a baby.",
            "A man is riding a horse.",
            "A woman is playing violin."
        ],
        "kwargs": json.dumps({
            "enable_qwen3_rerank_template": True
        }),
    }
)
print("Rerank 结果:", response.json())

ZhikaiGuo960110:
if in curl format, should be like this
curl -X 'POST' \
  'http://localhost:9997/v1/rerank' \
  -H 'accept: application/json' \
  -d '{
    "model": "Qwen3-Reranker-0.6B",
    "query": "A man is eating pasta.",
    "documents": [
        "A man is eating food.",
        "A man is eating a piece of bread.",
        "The girl is carrying a baby.",
        "A man is riding a horse.",
        "A woman is playing violin."
    ],
    "kwargs": "{\"enable_qwen3_rerank_template\": true}"
  }'

PR #4787, opened 2026-04-10
fix: support JSON object parameters in CLI

### LINKED ISSUE #4760: [Bug] Unable to pass a JSON object in the xinference-local command line
### System Info / 系統信息

...

- [ ] docker / docker

### Version info / 版本信息

2.4.0

...

### Reproduction / 复现过程

Pass a JSON in command line will be treated as a string.

- Example 1: [Vllm's compilation_config](https://www.modelscope.cn/models/tclf90/Qwen3.5-27B-AWQ/feedback)
- Example 2: [Vllm's kv_transfer_config](https://docs.vllm.ai/en/latest/api/vllm/distributed/kv_transfer/kv_connector/v1/flexkv_connector/)

### Expected behavior / 期待表现

It should work.

Ricardo-M-L:
## Summary
- Fixes #4760 — CLI parameters like `--compilation_config '{"mode": 3}'` were treated as plain strings instead of parsed JSON objects
- Added JSON auto-detection to `handle_click_args_type()` in `xinference/deploy/utils.py`: tries `json.loads()` for args starting with `{` or `[`, falls back to string on failure
- Added comprehensive test suite (`xinference/deploy/test/test_utils.py`) covering JSON objects, arrays, nested structures, malformed JSON fallback, and backwards compatibility with existing type conversions

## Details
The root cause is that `handle_click_args_type()` only handled `None`, `bool`, `int`, `float`, and plain strings. When vLLM engine parameters like `compilation_config`, `kv_transfer_config`, or `speculative_config` were passed as JSON strings via the CLI, they were forwarded as raw strings instead of dicts, causing downstream failures.

The fix adds a JSON parsing step **after** int/float checks (to preserve priority) but **before** the string fallback. It only attempts parsing when the arg starts with `{` or `[` to avoid unnecessary overhead.

## Test plan
- [x] Unit tests for all three real-world examples from the issue (compilation_config, kv_transfer_config, speculative_config)
- [x] Tests for nested JSON objects
- [x] Tests for JSON arrays
- [x] Tests for malformed JSON gracefully falling back to string
- [x] Tests for backwards compatibility (None, bool, int, float, plain string)
- [x] All 30 test cases pass


qinxuye (approved):
LGTM

PR #4788, opened 2026-04-10
fix: support Jina API task parameters for jina-embeddings-v4

Ricardo-M-L:
## Summary

- Map Jina-style `task` parameter values (`retrieval.passage`, `retrieval.query`, `text-matching`, `code`) to SentenceTransformer `prompt_name`, restoring the query/passage distinction lost when Xinference collapsed them into a single `retrieval`
- Maintain backward compatibility: plain `retrieval` still works (defaults to `retrieval.passage`)
- Also supports `jina-embeddings-v3` which uses the same task types

Fixes #4773

## Problem

The official Jina AI API accepts `task` values like `retrieval.passage` and `retrieval.query` that select different embedding prompts. Xinference only supported `retrieval`, losing the important distinction between query and passage embeddings. Users migrating from the Jina API had to modify their application code.

## Changes

**`xinference/model/embedding/sentence_transformers/core.py`**
- Added `JINA_V4_TASK_TO_PROMPT_NAME` mapping table and `JINA_TASK_SUPPORTED_MODELS` set
- Added `_resolve_jina_task()` helper that maps Jina API task values to SentenceTransformer `prompt_name`
- Modified `_create_embedding()` to extract `task` from kwargs and pass resolved `prompt_name` to the encode call for both the jina/clip code path and the default code path

**`xinference/model/embedding/sentence_transformers/tests/test_jina_task_mapping.py`** (new)
- 20+ unit tests covering all valid task values, backward compatibility, error handling, case insensitivity, and mapping table completeness

## Test plan

- [x] Unit tests for `_resolve_jina_task()` cover all task values
- [x] Backward compatibility tested: `retrieval` maps to `retrieval.passage`
- [x] Invalid task raises clear `ValueError` with valid options
- [x] Non-Jina models unaffected (returns `None`, no prompt_name set)
- [ ] Integration test with actual jina-embeddings-v4 model (requires GPU)


qinxuye (approved):
LGTM

PR #4822, opened 2026-04-21
fix: handle missing 'cpu' key in get_cluster_device_info to prevent KeyError 500

m199369309:
## Summary                                                                                                         
                                                                                                                   
  - `get_cluster_device_info(detailed=True)` raises `KeyError: 'cpu'` when a worker's status dict is missing the
  `cpu` key (e.g. incomplete reporting, collection timeout), causing the management API to return 500
  - Replace `worker_status.status["cpu"]` with defensive `status.get("cpu")` + `isinstance(cpu_raw, ResourceStatus)`
  type check
  - When cpu info is missing or has unexpected type, set `cpu_available`, `cpu_count`, `mem_used`, `mem_available`,
  `mem_total` to `None` instead of crashing
  - Add `logger.debug` for missing key and `logger.warning` for unexpected type for observability

  ## Context

  The same method already handles GPU keys defensively with `if k != "cpu"`, but the `detailed=True` branch still
  hard-indexes `status["cpu"]`. This aligns the CPU branch with the existing defensive style.

  ## Test plan

  - [x] Added 3 unit tests in `xinference/core/tests/test_cluster_metrics.py`:
    - Worker missing `cpu` key but having GPUs — GPU fields computed correctly, CPU/mem fields are `None`
    - Worker with completely empty status dict
    - Mixed scenario: one normal worker + one worker missing `cpu` key coexist in the same result
  - [x] Existing `test_supervisor_cluster_device_info_includes_gpu_utilization_average` still passes (regression)    
  - [ ] `pytest xinference/core/tests/test_cluster_metrics.py -v`

### DIFF
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -734,12 +734,32 @@ class SupervisorActor(xo.StatelessActor):
                 "gpu_vram_total": total,
             }
             if detailed:
-                cpu_info = worker_status.status["cpu"]
-                info["cpu_available"] = cpu_info.total * (1 - cpu_info.usage)
-                info["cpu_count"] = cpu_info.total
-                info["mem_used"] = cpu_info.memory_used
-                info["mem_available"] = cpu_info.memory_available
-                info["mem_total"] = cpu_info.memory_total
+                cpu_raw = worker_status.status.get("cpu")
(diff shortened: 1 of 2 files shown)


qinxuye (approved):
LGTM

PR #4823, opened 2026-04-21
fix: venv concurrent creation race, cold-start lock dir, and jina-embeddings-v4 torch mismatch

m199369309:
## Summary                                                                                                                                       
                                                                                                                                                   
  Three related fixes for multi-replica virtualenv deployment failures:                                                                            
                                                                                                                                                   
  1. **Concurrent venv creation race** (`worker.py`): When deploying models with `replica>=2` on the same worker, concurrent `asyncio.to_thread`
  calls to `_create_virtual_env_manager` race on the same virtualenv path, causing `uv venv` to fail with "directory already exists". Added
  `_exclusive_venv_path_lock` context manager using `fcntl.flock` to serialize venv creation + `.pth` injection per normalized `env_path`.

  2. **Cold-start FileNotFoundError** (`worker.py`): After deleting the entire venv tree, `os.open` for the lock file fails because the parent
  directory doesn't exist yet. Added `os.makedirs(lock_dir, exist_ok=True)` before `os.open` in `_exclusive_venv_path_lock`.

  3. **torch/torchvision mismatch** (`model_spec.json`): `jina-embeddings-v4` declared `#system_torchvision#` but not `#system_torch#`, causing the
   child venv to install a different torch version. The version mismatch triggers `RuntimeError: operator torchvision::nms does not exist`. Added
  `"#system_torch# ; #engine# == \"sentence_transformers\""` to the virtualenv packages.                                                           
                                                                                                                                                 
  ## Files changed                                                                                                                               
                                                                                                                                                 
  - `xinference/core/worker.py` — `_exclusive_venv_path_lock` + lock wrapping in `_create_virtual_env_manager`
  - `xinference/model/embedding/model_spec.json` — `jina-embeddings-v4` virtualenv packages

  ## Test plan

  - [ ] Deploy `jina-embeddings-v4` with `replica>=2` on same worker — no `uv venv` exit status 2
  - [ ] Delete entire venv directory, redeploy — no `FileNotFoundError` on lock file
  - [ ] Multi-replica deployment — no `operator torchvision::nms does not exist`
  - [ ] Single replica deployment behaves identically to before
  - [ ] Different models with different venv paths remain parallel

### DIFF
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -110,6 +110,9 @@ def _exclusive_venv_path_lock(env_path: str):
     logical virtualenv path. Avoids TOCTOU races when multiple model replicas
     call ``create_env`` concurrently (see Xinference virtualenv v4 layout).
 
+    Ensures the lock file's parent directory exists before ``os.open`` so cold
+    starts work after the entire venv tree was removed.
+
     Uses a sibling lock file ``{realpath(env_path)}.xinference-venv.lock``.
     On Windows this is a no-op (``fcntl`` unavailable / different semantics).
     """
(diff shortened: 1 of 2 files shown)


qinxuye (approved):
LGTM

PR #4825, opened 2026-04-21
feat: parallelize multi-replica terminate_model and improve UI delete UX

m199369309:
## Summary

  ### Backend (`supervisor.py`)
  - Replace serial `for await` loop with `asyncio.gather(return_exceptions=True)` in `terminate_model`, aligning with existing
  `cancel_launch_builtin_model` pattern
  - Wall-clock time for cross-worker multi-replica termination reduces from sum-of-replicas to max-of-replicas
  - Use `pop(..., None)` instead of `del` for `_replica_model_uid_to_worker`, making concurrent/repeated deletes idempotent (eliminates
  `KeyError` seen in production logs)

  ### Frontend (`running_models/index.js` + locales)
  - Confirmation dialog before terminate, showing model name and replica count with warning about expected duration
  - Prevent duplicate clicks: button disabled + `CircularProgress` spinner while terminating
  - Async polling with exponential backoff (2s → 4s → 8s, max 48 attempts) until model disappears from list
  - Error hint guiding user to refresh rather than re-clicking
  - i18n: 6 new `runningModels.terminate*` keys added to en/zh/ja/ko

  ## Context

  In multi-model + multi-replica clusters, deleting a model took excessively long (e.g. 950s for a 2-replica model) because replicas on  
  different workers were terminated sequentially. Users would click delete repeatedly, compounding the issue with duplicate requests that
   eventually hit `KeyError`.                                                                                                            
                                                                                                                                       
  ## Test plan                                                                                                                         
                                                                                                                                         
  - [ ] Cross-worker `replica>=2` terminate: wall-clock time shorter (close to slowest replica, not sum)
  - [ ] Same-worker multi-replica: no regression (worker-side still serial, gather only parallelizes Supervisor dispatch)
  - [ ] Repeated DELETE on same model: no `KeyError`, `pop` handles idempotently
  - [ ] UI: confirmation dialog shows correct replica count
  - [ ] UI: button disables during terminate, shows spinner
  - [ ] UI: polling removes model from list automatically after completion
  - [ ] UI: error toast shows refresh hint on failure
  - [ ] Locales: all 4 languages display correctly


qinxuye (approved):
LGTM

PR #4833, opened 2026-04-22
fix: remove last message role restriction in chat completion endpoint

amumu96:
## Summary
- Remove the validation that required the last message role to be `"user"`, `"system"`, or `"tool"` in the `create_chat_completion` endpoint
- Only keep the `messages` empty check, aligning with OpenAI API behavior
- Fixes 400 errors from frameworks like LangGraph that pass conversation history including assistant messages

## Test plan
- [ ] Verify chat completion works with assistant message as the last message
- [ ] Verify empty messages still returns 400
- [ ] Verify normal user/system/tool messages still work as before

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -2001,7 +2001,7 @@ class RESTfulAPI(CancelMixin):
 
         messages = body.messages and list(body.messages) or None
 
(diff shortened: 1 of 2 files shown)


qinxuye:
Tests failed due to this change.

qinxuye (approved):
LGTM

PR #4845, opened 2026-04-24
feat: TCP keepalive, reverse-channel probe, list_models cache, .pth atomic write, launch exemption

m199369309:
## Summary
- **Require xoscar>=0.9.6** for native TCP keepalive support (xorbitsai/xoscar#189), removing the xinference-side monkey-patch
- Add supervisor→worker **reverse-channel ping probe** in `_check_dead_nodes` to detect dead reverse channels even when worker→supervisor heartbeat is healthy
- Add `list_models` **cache degradation**: return cached results when a worker is temporarily unreachable instead of returning empty dict (prevents UI from showing no models)
- Reduce `XINFERENCE_GET_MODEL_RPC_TIMEOUT` from **120s to 30s** to limit actor lock contention during worker communication failures
- Fix `.pth` file **concurrent write race** in virtualenv setup by using atomic write (`os.replace`) and skipping writes when content is unchanged
- **Exempt workers from reverse-channel dead detection** while launching models, as long-running model downloads can starve the actor event loop causing false-positive dead detection
- Limit `huggingface_hub` download concurrency to **2** (`XINFERENCE_MODEL_DOWNLOAD_WORKERS`) to reduce GIL contention during model downloads

## Changed Files
| File | Change |
|------|--------|
| `setup.cfg` | xoscar>=0.9.3 → xoscar>=0.9.6 (native TCP keepalive) |
| `xinference/core/supervisor.py` | Reverse-channel probe, list_models cache, `_workers_launching` tracking |
| `xinference/core/worker.py` | `ping()` method, `.pth` atomic write, download concurrency limit |
| `xinference/constants.py` | RPC timeout 120→30, new `XINFERENCE_MODEL_DOWNLOAD_WORKERS` |

## Test plan
- [ ] CI pipeline passes
- [ ] Verify TCP keepalive active via xoscar native: `ss -tnop | grep <supervisor-port>` shows `timer:(keepalive,...)`
- [ ] Verify reverse-channel detection: simulate network partition triggers "Worker reverse-channel timeout" within 25s
- [ ] Verify list_models degradation: worker unreachable → UI still shows cached models
- [ ] Verify launching exemption: deploying large model does not trigger false "Worker dead"
- [ ] Verify .pth race fix: multi-replica deploy on existing venv succeeds on first attempt


xinference/core/xoscar_keepalive.py
qinxuye, line 1: This is a bit hacky, I recommend to contribute this code into xoscar itself.

qinxuye (approved):
LGTM

PR #4846, opened 2026-04-24
refactor(device_utils): replace if/elif chains with DeviceSpec registry

amumu96:
## Summary
- Introduce `DeviceSpec` dataclass and `DEVICE_REGISTRY` list to replace scattered if/elif chains in `device_utils.py`
- Public functions (`get_available_device`, `empty_cache`, `gpu_count`, `is_device_available`, etc.) are now data-driven via the registry
- Adding support for a new device now only requires a single `DeviceSpec` entry instead of touching multiple functions

## Test plan
- [x] All existing import paths verified (`get_available_device`, `empty_cache`, `gpu_count`, `get_nvidia_gpu_info`, etc.)
- [x] `gather_node_info()` in `xinference/core/resource.py` works correctly
- [x] Pre-commit hooks pass (black, flake8, isort, mypy, codespell)
- [x] GPU detection returns correct results on CUDA machine

### DIFF
--- a/xinference/core/resource.py
+++ b/xinference/core/resource.py
@@ -17,7 +17,7 @@ from typing import Dict, Union
 
 import psutil
 
-from ..device_utils import get_nvidia_gpu_info
+from ..device_utils import get_gpu_info
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


qinxuye (approved):
LGTM

PR #4849, opened 2026-04-25
fix: normalize dashes to underscores in CLI extra kwargs

### LINKED ISSUE #4830: [Bug] Cannot use mmproj when starting gguf models via CLI
### System Info / 系統信息

Python 3.12
xllamacpp 0.2.14

- [ ] docker / docker

### Version info / 版本信息

2.5.0

```
xinference launch --model-name qwen3.5 --model-type LLM --n-gpu auto --replica 1 --disable-virtual-env --model-engine llama.cpp --model-format ggufv2 --size-in-billions 9 --quantization UD-Q4_K_XL --multimodal-projector mmproj-F16.gguf --model-uid qwen3 --n_ctx 10000
```

### Reproduction / 复现过程

1. Start Qwen3.5 GGUFs in CLI
2. Open the Gradio web page
3. Upload an image, enter text and press Enter

```
Error
[address=0.0.0.0:43969, pid=2028473] Got error in chat stream: image input is not supported - hint: if this is unexpected, you may need to provide the mmproj
```

### Expected behavior / 期待表现

It should work. 

It did work when you start in web page. The command line is copied from web page.

octo-patch:
Fixes #4830

## Problem

When launching a gguf model with `--multimodal-projector` (dashes, as the web UI's "Copy CLI command" feature generates), the projector is silently ignored, resulting in:

```
image input is not supported - hint: if this is unexpected, you may need to provide the mmproj
```

**Root cause:** The `xinference launch` command passes unrecognized options through `ctx.args`. The extra-kwargs parser extracted the key verbatim (`ctx.args[i][2:]`), preserving dashes. So `--multimodal-projector` became the dict key `"multimodal-projector"`, while downstream code looks for `"multimodal_projector"` (underscore).

The web UI's `commandBuilder.js` converts underscore keys to dash format (`key.replace(/_/g, '-')`), so the generated CLI command always uses dashes — making this a consistent failure for any user who copies from the UI.

## Solution

Normalize dashes to underscores in the extra-kwargs key, mirroring the standard behavior of Click's built-in options:

```python
param_name = ctx.args[i][2:].replace("-", "_")
```

This makes `--multimodal-projector` and `--multimodal_projector` equivalent, consistent with how Click handles its native options.

## Testing

Manually verified that:
- `--multimodal-projector mmproj-F16.gguf` now correctly maps to `kwargs["multimodal_projector"]`
- `--multimodal_projector mmproj-F16.gguf` (underscore) still works as before
- The existing `--model_path` backward-compat check still works correctly

### DIFF
--- a/xinference/deploy/cmdline.py
+++ b/xinference/deploy/cmdline.py
@@ -925,7 +925,7 @@ def model_launch(
                 f"You must specify extra kwargs with `--` prefix. "
                 f"There is an error in parameter passing that is {ctx.args[i]}."
             )
-        param_name = ctx.args[i][2:]
+        param_name = ctx.args[i][2:].replace("-", "_")
         param_value = handle_click_args_type(ctx.args[i + 1])
         if param_name == "model_path":
             # fix for --model_path which is the old fashion to set model_path,


qinxuye:
I think the frontend should be fixed instead of command line.

Because underscore is only for those non-internal options, some internal options are dashes, e.g. --n-gpu

octo-patch:
@qinxuye Good point — I should have looked closer at the dashes-as-internal-option case before reaching for a CLI-side fix.

Youʼre right that blanket-normalizing dashes to underscores in `cmdline.py` is too broad: it would silently break the `--n-gpu`-style internal options that legitimately rely on the dashed form. The correct place to fix this is the source of the conversion in `commandBuilder.js`, where the regex `key.replace(/_/g, -)` rewrites every kwarg unconditionally.

Iʼll close this PR and put up a frontend-side fix that either:
1. Skips the underscore→dash rewrite for keys not in a known-dashed allowlist, or
2. Mirrors how Click resolves its own options (only normalize on the params actually exposed by `xinference launch`).

Let me know if either approach is preferable, otherwise Iʼll go with (1) since itʼs the smallest change that preserves the current UX for `--n-gpu` users.

qinxuye:
Maybe for frontend, it's not a global replace, we can fix this specific param.

PR #4861, opened 2026-04-27
fix: correct get_context_length default fallback behavior

Ricardo-M-L:
## Summary

Correct get_context_length default fallback behavior to properly handle cases where context length is not explicitly set.

🤖 Generated with Claude Code

### DIFF
--- a/xinference/model/llm/transformers/utils.py
+++ b/xinference/model/llm/transformers/utils.py
@@ -52,19 +52,22 @@ def get_context_length(config) -> int:
     ):
         max_sequence_length = config.max_sequence_length
     else:
-        max_sequence_length = 2048
+        max_sequence_length = None
     if hasattr(config, "seq_length") and config.seq_length is not None:
         seq_length = config.seq_length
     else:
-        seq_length = 2048
+        seq_length = None
     if (
         hasattr(config, "max_position_embeddings")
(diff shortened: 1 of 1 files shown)


qinxuye:
There is conflict, please resolve it.

qinxuye:
This logic has been modified in #4768

Ricardo-M-L:
Polite bump 🙏 — this is a small, contained fix for the \`get_context_length\` fallback when a config doesn't expose any of the canonical context fields. CI is fully green.

The previous behavior silently returned a hardcoded default that was wrong for several modern long-context models. The fix walks the config in the documented precedence order and only falls back at the very end with a logged warning.

Ricardo-M-L:
Closing — looks like this got resolved by refactor 🎉

Upstream extracted the body of `get_context_length()` into a shared helper `xinference/model/llm/utils.py:get_context_length_from_config()`, which already handles the fallback correctly:

```python
candidates = _collect_context_length_candidates(config, nested_attrs)
if not candidates:
    return 2048
return max(candidates)
```

The `transformers/utils.py:get_context_length()` now just delegates to it. So the `max(*[], default=2048)` corner-case my PR was guarding against is gone — no candidates → `return 2048` happens at the new helper. Closing as superseded.

PR #4786, opened 2026-04-10
fix: replace eval() with safe alternatives to prevent RCE in tool parsers

Ricardo-M-L:
## Summary

- Replace dangerous `eval()` calls with `json.loads()` + `ast.literal_eval()` fallback in Llama3 tool parsing code
- `eval()` with empty globals/locals is still exploitable via `__class__.__bases__` chains, enabling arbitrary code execution from untrusted LLM output
- Add comprehensive security and functional tests (25+ test cases)

## Changes

| File | Description |
|------|-------------|
| `xinference/model/llm/tool_parsers/llama3_tool_parser.py` | Replace `eval()` with `json.loads()` primary + `ast.literal_eval()` fallback |
| `xinference/model/llm/utils.py` | Same fix in `_eval_llama3_chat_arguments()` |
| `xinference/model/llm/tool_parsers/tests/test_llama3_tool_parser.py` | 18 new tests: normal parsing, RCE prevention, edge cases |
| `xinference/model/llm/tests/test_utils.py` | 7 new tests for `_eval_llama3_chat_arguments` security |

## Why `eval(x, {}, {})` is not safe

```python
# This executes successfully even with empty globals/locals:
eval("().__class__.__bases__[0].__subclasses__()", {}, {})
```

The `__builtins__` module is accessible through object introspection, allowing full code execution.

## Fix approach

1. **`json.loads()`** as primary parser (handles standard JSON from most LLMs)
2. **`ast.literal_eval()`** as fallback (handles Python literal formats like `True`/`False`/`None` and single-quoted strings that Llama3 may output)
3. Proper `KeyError`/`TypeError` handling for missing `name`/`parameters` keys

`ast.literal_eval()` only evaluates literal structures (strings, numbers, tuples, lists, dicts, booleans, None) and raises `ValueError` for anything else, making it safe against code injection.

## Audit of other parsers

All 7 other tool parser files were audited — they already use `json.loads()`:
- `qwen_tool_parser.py` — `json.loads()` 
- `deepseek_v3_tool_parser.py` — `json.loads()` 
- `deepseek_v3_1_tool_parser.py` — `json.loads()` 
- `deepseek_r1_tool_parser.py` — `json.loads()` 
- `glm4_tool_parser.py` — `json.loads()` 
- `minimax_tool_parser.py` — `json.loads()` 

## Test plan

- [x] Normal JSON tool call parsing works correctly
- [x] Python literal format (True/False/None, single quotes) works correctly
[...]

PR #4872, opened 2026-05-04
fix(ocr): replace eval() with ast.literal_eval in deepseek_ocr coordinate parsing

Ricardo-M-L:
### What

\`xinference/model/image/ocr/deepseek_ocr.py\` had two \`eval()\` calls in the coordinate-extraction helpers:

| Site | Old code |
| --- | --- |
| \`extract_coordinates_and_label\` (L231) | \`cor_list = eval(ref_text[2])\` |
| \`extract_text_blocks\` (L399) | \`coords = eval(f\"[{coords_str}]\")\` |

In both, the string passed to \`eval()\` is a substring extracted from the **OCR model's text output**. OCR output is LLM-generated and attacker-influenceable through the source document or prompt, so an OCR response shaped like

\`\`\`
<|ref|>x<|/ref|><|det|>[[__import__('os').system('id')]]<|/det|>
\`\`\`

would have its payload **executed by the host process** when the result is parsed for downstream rendering / block extraction.

### Why this matters

Same RCE shape as the merged tool-parser fix #4786 — that PR covered \`xinference/model/llm/\` but did not reach the image OCR path. Anyone running the deepseek_ocr model on attacker-influenced documents would silently be executing whatever Python the model emits inside \`<|det|>...<|/det|>\` brackets.

### Fix

Replace both calls with \`ast.literal_eval\`, which only accepts Python literal structures (lists, tuples, numbers, strings, …) and refuses calls, attribute access, and imports. The expected coordinate strings are list literals like \`[[10, 20, 30, 40]]\`, which \`literal_eval\` parses **identically** to \`eval\`, so behavior on legitimate input is unchanged.

Exception handling is also tightened from a bare \`except Exception\` to the specific \`ValueError\` / \`SyntaxError\` that \`literal_eval\` raises (plus \`IndexError\` / \`TypeError\` in the first helper, which indexes \`ref_text[1]\`/\`ref_text[2]\`).

\`\`\`python
# Before
cor_list = eval(ref_text[2])

# After
# OCR output is LLM-generated and attacker-influenceable. literal_eval
# only accepts Python literal structures and rejects code execution.
cor_list = ast.literal_eval(ref_text[2])
\`\`\`

### Tests

Adds \`xinference/model/image/ocr/tests/test_deepseek_ocr_safe_eval.py\` with cases that pin the security guarantee:

- ✅ Legitimate single and multi coordinate lists still parse.
- ✅ A payload of \`__import__('pathlib').Path(...).write_text(...)\` is rejected with **no side effect** (verified by checking a tmp_path sentinel file is never created).
- ✅ Attribute-walking payloads (\`().__class__.__bases__[0]...\`) are rejected.
- ✅ Malformed truncated input does not crash callers.
- ✅ A malicious block followed by a valid block in the same OCR response still allows the valid block to surface (parser does not abort on a bad block).

The test module \`pytest.importorskip\`s \`torch\` / \`PIL\` / \`torchvision\` so it is safely skipped in environments without the OCR runtime, but runs fully in CI.

### Notes

- Same fix shape as merged PR #4786 (\`fix: replace eval() with safe alternatives to prevent RCE in tool parsers\`). This PR closes the remaining \`eval()\` site that #4786 didn't reach.
- No external behavior change — \`literal_eval\` is a strict subset of \`eval\` for the expected inputs.

\`ruff check\` passes on all touched files.

### DIFF
--- a/xinference/model/image/ocr/deepseek_ocr.py
+++ b/xinference/model/image/ocr/deepseek_ocr.py
@@ -390,9 +390,13 @@ def extract_text_blocks(text: str) -> List[Dict[str, Any]]:
     if not isinstance(text, str):
         return []
 
-    # Pattern to extract text and coordinates
+    # Pattern to extract text and coordinates. The coordinates group keeps
+    # the surrounding `[[...]]` so the resulting structure is nested
+    # (`[[x1, y1, x2, y2], ...]`) — matching extract_coordinates_and_label
+    # and the `bbox = coords[0] if len(coords) == 1 else coords` logic
+    # below, which assumes a list-of-lists.
     block_pattern = (
-        r"<\|ref\|>(.*?)<\|/ref\|><\|det\|>\[\[(.*?)\]\]<\|/det\|>(.*?)(?=<\|ref\|>|$)"
+        r"<\|ref\|>(.*?)<\|/ref\|><\|det\|>(\[\[.*?\]\])<\|/det\|>(.*?)(?=<\|ref\|>|$)"
     )
 
     blocks = []
@@ -405,7 +409,7 @@ def extract_text_blocks(text: str) -> List[Dict[str, Any]]:
             # Use ast.literal_eval instead of eval — the coordinate string is
             # OCR model output (LLM-generated). literal_eval only accepts
             # Python literal structures and rejects code execution.
-            coords = ast.literal_eval(f"[{coords_str}]")
+            coords = ast.literal_eval(coords_str)
             if isinstance(coords, list) and len(coords) > 0:
                 blocks.append(
                     {
--- a/xinference/model/image/ocr/deepseek_ocr.py
+++ b/xinference/model/image/ocr/deepseek_ocr.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import ast
 import logging
 import os
 import re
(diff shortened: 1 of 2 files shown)


qinxuye:
Tests failed.

Ricardo-M-L:
Pushed \`9f76e30d\` to fix the failing tests in this PR's own test file. Root cause was an unrelated pre-existing bug in the regex above: \`\[\[(.*?)\]\]\` was stripping the outer \`[[...]]\` brackets, so \`ast.literal_eval(f"[{coords_str}]")\` produced a *flat* \`[1,2,3,4]\` instead of the nested \`[[1,2,3,4]]\` that:

1. \`extract_coordinates_and_label\` returns (sibling helper, same file)
2. The immediately-following \`bbox = coords[0] if len(coords) == 1 else coords\` line assumes (only makes sense for list-of-lists)
3. The new safe-eval tests in this PR check for

Fix: capture the outer brackets in the regex and pass directly to \`literal_eval\` — same shape as \`extract_coordinates_and_label\`.

Verified locally with the exact test fixtures: well-formed → \`[[1,2,3,4]]\`, multiple blocks → each nested, code payload → rejected by \`literal_eval\` (\`ValueError\`).

Ready for re-review 🙏

qinxuye (approved):
LGTM

PR #4875, opened 2026-05-06
fix: Set default enable-thinking option in command line to False

### LINKED ISSUE #4886: xinference 2.7.0 版本命令行启动qwen3.6模型无法关闭思考
大佬们，
    我使用xinference 2.7.0 版本，页面启动qwen3.6模型，可以正常关闭思考（默认就是关闭）。但是我把命令行代码复制出来运行，启动的模型，思考就没法关闭了。

Image

**复制出来的命令行：**
```
xinference launch --model-name qwen3.6 --model-type LLM --n-gpu auto --replica 1 --reasoning_content false --n-worker 1 --model-engine SGLang --model-format fp8 --size-in-billions 35 --quantization FP8 --model-uid qwen3 --gpu-idx 4,5,6,7 --download_hub modelscope --disable-virtual-env
```

Image

这是什么情况呢？

la1ty:
Close #4886 .

Fix the wrong default thinking behaviour when starting a model in command line without `--enable-thinking` option.

How to reproduce:

1. Start Qwen3.5 in command line without `--enable-thinking`.
2. Make a chat completion request to it:

```
from openai import OpenAI


API_BASE_URL = "http://your.xinference:port/v1"
API_KEY = "EMTPY"
LLM_MODEL = "qwen3.5"

client = OpenAI(
    base_url=API_BASE_URL,
    api_key=API_KEY
)

user_msg = "Hi"
result = client.chat.completions.create(
    messages=[
        {"role": "user", "content": user_msg},
    ],
    model=LLM_MODEL,
    # max_tokens=10,
)
print(result.model_dump_json())
```

### DIFF
--- a/xinference/deploy/cmdline.py
+++ b/xinference/deploy/cmdline.py
@@ -859,7 +859,7 @@ def remove_cache(
     "--enable-thinking",
     "enable_thinking",
     flag_value=True,
-    default=None,
+    default=False,
     help="Enable thinking mode for hybrid reasoning LLMs (e.g., Qwen3).",
 )
 @click.option(


la1ty:
I suddenly realized that this is a dirty fix, but I haven't found the root reason yet. 

If you start the model in web page, the behaviour is normal. So I modify the command line option.

qinxuye:
enable_thinking should be default as true by design, this might be not related.

la1ty:
> enable_thinking should be default as true by design, this might be not related.

But currently, the model's command line args only include `--enable-thinking`. 

If the launch command has an `--enable-thinking`, `enable_thinking` will be set to `True`, else it will be set to `None`, which causes the wrong default behaviour.

la1ty:
It seems that the front end and the command line have a completely opposite logic on `enable_thining`:

- Front end: set default value to `true`, unless you disable it manually.
- Command line: set default value to `None`, unless you pass the `--enable-thinking` argument.

qinxuye:
> It seems that the front end and the command line have a completely opposite logic on `enable_thining`:
> 
> * Front end: set default value to `true`, unless you disable it manually.
> * Command line: set default value to `None`, unless you pass the `--enable-thinking` argument.

Yeah, a bit confusing.

What's your option on this?

la1ty:
My personal thought is to modify the command-line side to be consistent with the web UI (normal option that accepts value, default to `True`), but this is a breaking change and it will invalid many existing shell scripts. 

If such a change is too big, simply setting the default value to False can also temporarily solve this issue, but we need to ensure that this parameter only affects hybrid models' behaviours.

There's also another option for smooth transition: deprecate `--enable-thinking` flag option and use `--no-enable-thinking` flag option instead, then set the default value to `True`.

qinxuye:
Let's merge this first.

qinxuye (approved):
LGTM

PR #4879, opened 2026-05-07
feat: add modular vLLM post-install patch framework with hybrid KV cache fix

m199369309:
## Summary

- Add `xinference/model/llm/vllm/patches/` module with auto-discovery registry for applying post-install patches to vLLM site-packages
- Implement hybrid KV cache page size patch (LCM-based padding) for Qwen3.5/3.6/3.6-MoE models, working around [vllm-project/vllm#37121](https://github.com/vllm-project/vllm/issues/37121)
- Use subprocess-based vLLM path resolution to correctly handle uv-managed virtual environments and `.pth`-injected parent site-packages
- Filter patches by model **architectures** (not model names), automatically covering fine-tuned and custom-registered models

## Motivation

Qwen3.5/3.6 models use hybrid attention (linear_attention + full_attention) with non-divisible KV cache page sizes. vLLM 0.20.x raises `NotImplementedError` during `unify_kv_cache_spec_page_size()`. The official fix (vllm-project/vllm#40128) has not been merged.

Previous monkey-patching approach fails because vLLM's EngineCore uses `spawn` multiprocessing (due to CUDA initialization), which doesn't inherit in-memory patches. File-level patching solves this by modifying the source directly in site-packages.

## Design

```
xinference/model/llm/vllm/patches/
├── __init__.py                          # Auto-discovery registry + apply_vllm_patches()
├── _base.py                             # VllmPatch dataclass
└── hybrid_kv_cache_page_size.py         # Qwen3.5/3.6 KV cache patch
```

- **Auto-discovery**: `pkgutil.iter_modules` scans for modules exposing `PATCH` attribute
- **Architecture filtering**: Each patch declares `architectures` set; uses `_resolve_architectures()` to cover fine-tuned models that inherit from built-in families
- **Idempotent**: Marker-based detection prevents double-patching
- **Subprocess path resolution**: Uses venv's own Python interpreter to locate vLLM via `importlib.util.find_spec`, handling `.pth` injections and uv-managed envs
- **Fallback**: Glob pattern for standard venv layouts

## Changes

| File | Change |
|------|--------|
| `xinference/model/llm/vllm/patches/__init__.py` | New: registry + entry point |
| `xinference/model/llm/vllm/patches/_base.py` | New: VllmPatch dataclass |
| `xinference/model/llm/vllm/patches/hybrid_kv_cache_page_size.py` | New: KV cache patch with subprocess path resolution |
| `xinference/core/worker.py` | Add `model_name`/`architectures` params to `_prepare_virtual_env`, call `apply_vllm_patches` after install |

## Removal condition

When vLLM officially merges the hybrid KV cache fix (tracking: vllm-project/vllm#37121, #38041), delete the patch file. The framework (`__init__.py`, `_base.py`) can remain for future patches or be removed entirely (worker.py has `ImportError` guard).

## Test plan

- [ ] Deploy Qwen3.6-35B-A3B-FP8 with vLLM 0.20.x on multi-GPU worker
- [ ] Verify worker logs show: `Applied hybrid KV cache patch to .../kv_cache_utils.py`
- [ ] Verify logs show: `Hybrid KV cache: padding max page size 1073152 -> 1081344`
- [ ] Verify no `NotImplementedError` during model initialization
- [ ] Verify non-hybrid models (e.g., Qwen2.5) are unaffected (patch skipped)
- [ ] Verify fine-tuned model with `model_family="qwen3.6"` gets patch applied via architecture inheritance

### DIFF
--- a/xinference/model/llm/vllm/patches/__init__.py
+++ b/xinference/model/llm/vllm/patches/__init__.py
@@ -4,6 +4,7 @@ vLLM post-install patches registry.
 Patches are auto-discovered: any module in this package that exposes a
 top-level ``PATCH`` attribute (instance of VllmPatch) is registered automatically.
 """
+
 import importlib
 import logging
 import pkgutil
--- a/xinference/model/llm/vllm/patches/__init__.py
+++ b/xinference/model/llm/vllm/patches/__init__.py
@@ -1,0 +1,56 @@
+"""
+vLLM post-install patches registry.
+
+Patches are auto-discovered: any module in this package that exposes a
+top-level ``PATCH`` attribute (instance of VllmPatch) is registered automatically.
+"""
+import importlib
+import logging
+import pkgutil
+from typing import List, Optional
+
+from ._base import VllmPatch
+
+logger = logging.getLogger(__name__)
+
+# Auto-discover patch modules
+_PATCHES: List[VllmPatch] = []
+for _info in pkgutil.iter_modules(__path__):
+    if _info.name.startswith("_"):
+        continue
+    _mod = importlib.import_module(f".{_info.name}", __name__)
+    if hasattr(_mod, "PATCH"):
+        _PATCHES.append(_mod.PATCH)
+
+
+def apply_vllm_patches(
+    env_path: str,
+    model_name: Optional[str] = None,
+    model_family: Optional[str] = None,
+) -> None:
+    """
+    Apply applicable vLLM patches for the given model.
+
+    Args:
+        env_path: Path to the virtual environment root.
+        model_name: Model name (for logging).
+        model_family: Model family string, used to filter patches.
+    """
+    for patch in _PATCHES:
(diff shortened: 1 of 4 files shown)


xinference/model/llm/vllm/patches/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    Apply applicable vLLM patches for the given model.
+
+    Args:
+        env_path: Path to the virtual environment root.
+        model_name: Model name (for logging).
+        model_family: Model family string, used to filter patches.
+    """
+    for patch in _PATCHES:
+        if patch.model_families and model_family:
```
qinxuye: How about using model architecture instead of model family in terms of judging if apply?

qinxuye (approved):
LGTM

qinxuye:
Tests failed, seemed that the other model types are effected.

PR #4880, opened 2026-05-07
fix(tool_parser): modify qwen3.5 check logic

llyycchhee:


qinxuye:
Just commit into 4870

PR #4887, opened 2026-05-08
bld: Restrict non-main release tags to admins

qinxuye:


.github/workflows/docker-cd.yaml
```diff
@@ -12,6 +12,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
```
rogercloud: ### Issue 2 (Major): Potential permission mismatch

`permissions: contents: read` restricts the `GITHUB_TOKEN`. The `getCollaboratorPermissionLevel` REST endpoint's docs do not explicitly state that `contents: read` covers it. If insufficient, this causes all releases to fail.

**Fix:** Move the permission check to a separate `check-permission` job without explicit `permissions`, then let the build job depend on it (`needs: check-permission`) and use `permissions: contents: read`.

.github/workflows/release.yaml
```diff
@@ -9,6 +9,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
```
rogercloud: ### Issue 2 (Major): Potential permission mismatch

Same concern as `docker-cd.yaml`: `permissions: contents: read` may not grant access to `github.rest.repos.getCollaboratorPermissionLevel`.

**Fix:** Use a two-job design: a `check-permission` job with default token permissions, and `build-publish` with `permissions: contents: read` that `needs: check-permission`.

.github/workflows/docker-cd.yaml
```diff
@@ -12,6 +12,9 @@ concurrency:
...
+  contents: read
+
 jobs:
   build:
     timeout-minutes: 240
```
rogercloud, line 27: ### Issue 4 (Minor): Mutable action reference

`actions/github-script@v7` is a floating tag. Per GitHub's security hardening guide, pinning to a full commit SHA is the only way to guarantee an immutable release.

**Fix:** `uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1`

.github/workflows/release.yaml
```diff
@@ -21,6 +23,51 @@ jobs:
```
rogercloud, line 20: ### Issue 4 (Minor): Mutable action reference

Same as `docker-cd.yaml`: prefer pinning `actions/github-script` to a commit SHA to prevent supply-chain attacks via tag movement.

**Fix:** `uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1`

.github/workflows/docker-cd.yaml
```diff
@@ -26,6 +29,51 @@ jobs:
...
 
+      - name: Check release tag permission
+        if: startsWith(github.ref, 'refs/tags/')
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { execFileSync, spawnSync } = require('child_process');
+
+            execFileSync('git', ['fetch', '--no-tags', 'origin', 'main:refs/remotes/origin/main'], {
+              stdio: 'inherit',
```
rogercloud: ### Issue 3 (Minor): Hidden error diagnostics

`stdio: 'ignore'` discards `stderr` when `git merge-base` fails unexpectedly, making production incidents very hard to debug.

**Fix:**
```javascript
const onMain = spawnSync('git', [
  'merge-base', '--is-ancestor', taggedCommit, 'origin/main',
], { stdio: 'pipe', encoding: 'utf-8' });

if (onMain.status !== 0 && onMain.status !== 1) {
  core.error(`git merge-base stderr: ${onMain.stderr}`);
  core.setFailed('Failed to determine whether the release tag is on main.');
  return;
}
```

.github/workflows/release.yaml
```diff
@@ -21,6 +23,51 @@ jobs:
...
+        with:
+          fetch-depth: 0
+      - name: Check release tag permission
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { execFileSync, spawnSync } = require('child_process');
+
+            execFileSync('git', ['fetch', '--no-tags', 'origin', 'main:refs/remotes/origin/main'], {
+              stdio: 'inherit',
```
rogercloud: ### Issue 3 (Minor): Hidden error diagnostics

Same as `docker-cd.yaml`: `stdio: 'ignore'` hides valuable stderr output from `git merge-base` when it fails unexpectedly.

**Fix:** Switch to `stdio: 'pipe'` and surface stderr in the failure message.

rogercloud (changes requested):
## Security & Maintainability Review

Thank you for this important security improvement. The intent to restrict non-main releases to admins is well-justified. After cross-checking with GitHub's official documentation, I've identified **4 valid issues** that should be addressed before merging.

### Issue 1 (Major): Duplicated permission-check logic across two workflows

The **~45 lines of JavaScript** in the `Check release tag permission` step are **identically duplicated** in both `docker-cd.yaml` and `release.yaml`. This violates DRY and creates a maintenance burden: any future change to the check (e.g., adding `maintain` permission, changing branch names) must be manually synced across both files.

**Recommended fix:** Extract a composite action.

Create `.github/actions/check-release-permission/action.yaml`:

```yaml
name: 'Check Release Tag Permission'
description: 'Ensure only admins can release from non-main branches'

runs:
  using: 'composite'
  steps:
    - uses: actions/checkout@v3
      with:
        fetch-depth: 0
    - name: Check release tag permission
      uses: actions/github-script@v7
      with:
        script: |
          const { execFileSync, spawnSync } = require('child_process');

          if (!context.ref.startsWith('refs/tags/')) {
            core.info('Not a tag push, skipping permission check.');
            return;
          }

          execFileSync('git', ['fetch', '--no-tags', 'origin', 'main:refs/remotes/origin/main'], {
            stdio: 'inherit',
          });

          const taggedCommit = process.env.GITHUB_SHA;
          const onMain = spawnSync('git', [
            'merge-base', '--is-ancestor', taggedCommit, 'origin/main',
          ], { stdio: 'ignore' });

          if (onMain.status === 0) {
            core.info(`Release tag ${context.ref} points to a commit on main.`);
            return;
          }
          if (onMain.status !== 1) {
            core.setFailed('Failed to determine whether the release tag is on main.');
            return;
          }

          const { data } = await github.rest.repos.getCollaboratorPermissionLevel({
            owner: context.repo.owner,
            repo: context.repo.repo,
            username: context.actor,
          });

          if (data.permission !== 'admin') {
            core.setFailed(
              `Release tag ${context.ref} points to a commit outside main. ` +
              `Only repository admins may release from non-main branches; ` +
              `${context.actor} has ${data.permission} permission.`
            );
          }
```

Then replace the inline steps in both workflows with:
```yaml
- uses: ./.github/actions/check-release-permission
```

### Issue 2 (Major): `permissions: contents: read` may not cover the collaborator permission API

Per GitHub docs, when the `permissions` key is present, **all unspecified permissions are set to `none`**. The `permissions` key does **not** include an `administration` option. GitHub's REST API docs for `GET /repos/{owner}/{repo}/collaborators/{username}/permission` do **not** explicitly map this endpoint to a `permissions` key value.

If `contents: read` is insufficient, the `getCollaboratorPermissionLevel` call will fail with `403`/`404`, causing **all releases (even admin ones) to be blocked**.

**Recommended fix:** Split the permission check into a separate job with default token permissions.

```yaml
jobs:
  check-permission:
    runs-on: ubuntu-latest
    steps:
      - uses: ./.github/actions/check-release-permission

  build:  # or build-publish
    needs: check-permission
    permissions:
      contents: read
    # ... rest of build steps ...
```

The `check-permission` job uses the **default** `GITHUB_TOKEN` (no explicit `permissions` key), which GitHub docs guarantee has sufficient permissions for in-repo REST API calls. The actual build job can then safely restrict to `contents: read`.

### Issue 3 (Minor): `spawnSync` hides error diagnostics with `stdio: 'ignore'`

When `git merge-base` fails for unexpected reasons (network, incomplete fetch, git environment issues), `stderr` is discarded, making production incidents hard to debug.

**Recommended fix:**
```javascript
const onMain = spawnSync('git', [
  'merge-base', '--is-ancestor', taggedCommit, 'origin/main',
], { stdio: 'pipe', encoding: 'utf-8' });

if (onMain.status !== 0 && onMain.status !== 1) {
  core.error(`git merge-base stderr: ${onMain.stderr}`);
  core.setFailed('Failed to determine whether the release tag is on main.');
  return;
}
```

### Issue 4 (Minor): `actions/github-script@v7` uses a mutable tag reference

GitHub's security hardening guide recommends pinning actions to a **full-length commit SHA** to mitigate supply-chain risks (tags can be moved or deleted).

**Recommended fix:**
```yaml
uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
```

Dependabot can automatically update SHA-pinned actions.

---

### Suggested implementation order
1. Extract composite action for the permission check.
2. Split into two jobs (`check-permission` without `permissions` + build job with `permissions: contents: read`).
3. Change `stdio: 'ignore'` to `stdio: 'pipe'` and log stderr on unexpected exit.
4. Pin `actions/github-script` to a commit SHA.

.github/workflows/docker-cd.yaml
qinxuye: Accepted and fixed in 974b6c00. The permission check now runs in a separate `check-release-permission` job without an explicit `permissions` override, and the Docker build job keeps `permissions: contents: read`.

.github/workflows/release.yaml
qinxuye: Accepted and fixed in 974b6c00. The PyPI release workflow now uses the same two-job structure: default-token permission check first, then `build-publish` with `permissions: contents: read`.

.github/workflows/release.yaml
qinxuye: Accepted and fixed in 974b6c00 by moving the check into the shared composite action, where `actions/github-script` is pinned to the full v7.0.1 commit SHA.

.github/workflows/docker-cd.yaml
qinxuye: Accepted and fixed in 974b6c00. The shared composite action pins the newly introduced `actions/github-script` usage to the full v7.0.1 commit SHA.

qinxuye:
Addressed the review feedback in 974b6c00:

1. Accepted: extracted the duplicated release tag permission logic into `.github/actions/check-release-permission/action.yaml`.
2. Accepted: split both workflows into a default-token `check-release-permission` job and a `contents: read` build job that depends on it.
3. Accepted: changed `git merge-base` to capture stderr and log it on unexpected failures.
4. Accepted for the newly introduced action: pinned `actions/github-script` to the full v7.0.1 commit SHA.

.github/workflows/release.yaml
qinxuye: Accepted and fixed in 974b6c00 via the shared action: `merge-base` stderr is now captured and surfaced on unexpected failures.

.github/workflows/docker-cd.yaml
qinxuye: Accepted and fixed in 974b6c00. `merge-base` now uses piped stdio with UTF-8 encoding and logs stderr before failing on unexpected exit statuses.

rogercloud (approved):
## Re-review: All issues resolved ✅

Thank you for addressing all the feedback promptly. I've verified each item against the latest commit.

### Issue 1 (Major): Code duplication — ✅ FIXED

The ~45 lines of duplicated permission-check logic have been extracted into `.github/actions/check-release-permission/action.yaml`. Both `docker-cd.yaml` and `release.yaml` now reference it via `uses: ./.github/actions/check-release-permission`.

### Issue 2 (Major): `permissions: contents: read` API coverage — ✅ FIXED

The permission check is now in a dedicated `check-release-permission` job **without** explicit `permissions`, guaranteeing the default `GITHUB_TOKEN` has sufficient scopes for `getCollaboratorPermissionLevel`. The build jobs (`build` and `build-publish`) use `needs: check-release-permission` and restrict to `permissions: contents: read`.

### Issue 3 (Minor): `stdio: 'ignore'` hiding diagnostics — ✅ FIXED

Changed to `stdio: 'pipe'` with `encoding: 'utf-8'`, and `onMain.stderr` is now surfaced via `core.error()` on unexpected exit.

### Issue 4 (Minor): Mutable action tag — ✅ FIXED

`actions/github-script` is now pinned to the full commit SHA `60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1`.

---

### Additional observations (non-blocking)

- The composite action also handles the non-tag skip logic internally (`if (!context.ref.startsWith('refs/tags/'))`), which is cleaner than the previous step-level `if` condition.
- Separating the permission check onto GitHub-hosted runners (`ubuntu-latest`) while keeping the actual build on `self-hosted` is a nice security side-effect.

LGTM. Approving.

PR #4899, opened 2026-05-12
feat: backport monitoring metrics and fix stream label / serve_count bugs

m199369309:
## Summary

- **Monitoring metrics backport (2.7.0 → 2.8.0)**: Restores 5 customized metric changes that were overwritten during the 2.8.0 upgrade:
  - `model_last_load_duration_seconds` Gauge with load timing in worker
  - `_WORKER_ONLY_METRICS` set + Supervisor-side deregister to avoid empty HELP/TYPE headers
  - image/audio/video engine/format label differentiation (empty labels instead of "unknown")
  - `grafana_alert_datasource` in UI config API + grafanaUtils.js URL params
  - `@request_limit` decorator on `image_to_image`, `inpainting`, `ocr` methods
  - Exception protection for `record_metrics`

- **Stream label fix**: `request_limit` decorator now recognizes xoscar `IteratorWrapper` as a stream type, fixing `stream="true"` label never being recorded for LLM requests

- **serve_count double-decrement fix**: Added stream type detection guard (`isasyncgen`/`isgenerator`/`IteratorWrapper`) to 3 `stream_results()` finally blocks, ensuring `decrease_serve_count()` is only called when the decorator has deferred the decrement to the caller. This prevents `model_serve_count` from going negative in two scenarios:
  - `model.chat()` fails before iterator is assigned (`iterator is None`)
  - `model.chat()` returns a non-stream value on `stream=True` path (iterator is not a generator/IteratorWrapper)

## Test plan

- [ ] Deploy and verify `model_last_load_duration_seconds` is reported after model load
- [ ] Verify Supervisor `/metrics` endpoint no longer shows worker-only metric HELP/TYPE headers
- [ ] Send stream + non-stream LLM requests, confirm `model_request_total{stream="true"}` and `stream="false"` are both recorded correctly
- [ ] Verify `model_serve_count` stays ≥ 0 after failed stream requests
- [ ] Verify `grafana_alert_datasource` appears in `/v1/cluster/ui_config` response
- [ ] Verify image/audio model metrics show empty engine/format labels instead of "unknown"

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -929,7 +929,8 @@ class RESTfulAPI(CancelMixin):
                     yield dict(data=json.dumps({"error": str(ex)}))
                     return
                 finally:
-                    await model.decrease_serve_count()
+                    if iterator is not None:
+                        await model.decrease_serve_count()
 
             return EventSourceResponse(
                 stream_results(), ping=XINFERENCE_SSE_PING_ATTEMPTS_SECONDS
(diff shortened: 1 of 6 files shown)


qinxuye (approved):
LGTM

PR #4902, opened 2026-05-12
feat: add Log Center UI with Elasticsearch integration

m199369309:
## Summary

- Add backend `GET /v1/cluster/logs` API that proxies Elasticsearch queries via `aiohttp` (compatible with ES 7.x/8.x/9.x, zero new Python dependencies)
- Add `es_enabled` field to `ui_config` response, controlled by `XINFERENCE_ES_URL` environment variable
- Add new **Log Center** page (`/#/logs`) with full-text search, log level/type/node filtering, time range picker, pagination, and search keyword highlighting
- Conditionally show "Log Center" sidebar menu item only when ES is configured
- Add i18n support for en/zh/ja/ko

### Backend (`xinference/api/routers/admin.py`)
- `get_ui_config()`: added `es_enabled` field based on `XINFERENCE_ES_URL`
- New `search_logs()` handler: builds parameterized ES `bool` query (prevents injection), supports `q`/`level`/`module`/`node`/`log_type`/`time_from`/`time_to`/`size`/`page_from` params, 10s timeout, Basic Auth / API Key auth
- Route registration: `GET /v1/cluster/logs` with `admin` scope

### Frontend
- **New file** `scenes/logs/index.js`: Log Center page with toolbar (search, level chips, log type chips, node select, time range popover, refresh), MUI Table with expandable detail rows, `from/size` pagination, `<mark>` search highlighting
- `router/index.js`: added `/logs` route
- `components/MenuSide.js`: added `ArticleOutlined` icon, `ApiContext` import, `esEnabled` state fetched from `ui_config`, conditional "Log Center" menu item
- `locales/{en,zh,ja,ko}.json`: added `menu.logs` and `logs.*` i18n keys

### Environment Variables
| Variable | Description | Example |
|----------|-------------|---------|
| `XINFERENCE_ES_URL` | Elasticsearch URL | `http://es:9200` |
| `XINFERENCE_ES_INDEX` | Log index pattern | `xinference-logs-*` |
| `XINFERENCE_ES_AUTH` | Auth (optional): `user:pass` or `ApiKey <base64>` | `elastic:changeme` |

## Test plan
- [ ] Set `XINFERENCE_ES_URL` and verify "Log Center" menu item appears
- [ ] Without `XINFERENCE_ES_URL`, verify menu item is hidden
- [ ] Test full-text search, level/type/node filtering, time range selection
- [ ] Test pagination (prev/next), search highlight in message column
- [ ] Test expandable row detail (module, pid, request_id, etc.)
- [ ] Verify admin auth is enforced on `/v1/cluster/logs`
- [ ] Test with ES 7.x and 8.x for compatibility

### DIFF
--- a/xinference/api/routers/admin.py
+++ b/xinference/api/routers/admin.py
@@ -2,11 +2,13 @@
 
 from __future__ import annotations
 
+import asyncio
 import logging
 import os
 import signal
 from typing import TYPE_CHECKING, Any
 
+import aiohttp
 from fastapi import Depends, HTTPException, Query, Request, Security
 
 from ..._version import get_versions
@@ -284,8 +286,6 @@ async def search_logs(
     size: int = 200,
     page_from: int = 0,
 ) -> JSONResponse:
-    import aiohttp
-
     es_url = os.environ.get("XINFERENCE_ES_URL", "")
     if not es_url:
         raise HTTPException(status_code=503, detail="Elasticsearch is not configured")
@@ -356,10 +356,11 @@ async def search_logs(
                         status_code=502, detail="Elasticsearch query failed"
                     )
                 data = await resp.json()
-    except aiohttp.ClientError as e:
-        logger.error("ES connection error: %s", e)
+    except (aiohttp.ClientError, asyncio.TimeoutError) as e:
+        logger.error("ES connection error or timeout: %s", e)
         raise HTTPException(
-            status_code=502, detail="Failed to connect to Elasticsearch"
+            status_code=502,
+            detail="Failed to connect to Elasticsearch or query timed out",
         )
 
     hits = [hit["_source"] for hit in data.get("hits", {}).get("hits", [])]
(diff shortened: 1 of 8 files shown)


xinference/api/routers/admin.py
```diff
@@ -268,10 +270,108 @@ async def get_ui_config() -> JSONResponse:
...
+) -> JSONResponse:
+    es_url = os.environ.get("XINFERENCE_ES_URL", "")
+    if not es_url:
+        raise HTTPException(status_code=503, detail="Elasticsearch is not configured")
+
+    es_index = os.environ.get("XINFERENCE_ES_INDEX", "xinference-logs-*")
+    es_auth = os.environ.get("XINFERENCE_ES_AUTH", "")
+
+    size = max(1, min(size, 500))
+    page_from = max(0, min(page_from, 10000))
```
qinxuye: This clamp still allows `page_from=10000` while `size` can be up to 500. Elasticsearch's default `index.max_result_window` requires `from + size <= 10000`, so the UI can reach `from=10000,size=200` and get a 400 from ES. Please clamp based on the requested size, e.g. `page_from = max(0, min(page_from, 10000 - size))`, and align the frontend next-page disabling with the same boundary.

qinxuye (commented):
Reviewed the Log Center changes and left one inline comment for a pagination boundary issue.

qinxuye (approved):
LGTM

PR #4906, opened 2026-05-13
feat: add collapsible sidebar with persistent state

m199369309:
## Summary
- Add collapse/expand toggle button to the sidebar navigation
- Collapsed mode shows only icons with tooltips, expanding to full width on toggle
- Sidebar state is persisted in localStorage across page reloads
- Smooth CSS transitions for width, logo, and toolbar elements

## Test plan
- [ ] Verify sidebar renders normally on page load (expanded by default)
- [ ] Click the collapse toggle button at the bottom — sidebar shrinks to icon-only mode
- [ ] Hover nav items in collapsed mode — tooltips appear with labels
- [ ] Click toggle again — sidebar expands back to full width
- [ ] Refresh the page — sidebar remembers its collapsed/expanded state
- [ ] Test on different screen widths to ensure responsive behavior is preserved

### DIFF
--- a/xinference/ui/web/ui/src/components/MenuSide.js
+++ b/xinference/ui/web/ui/src/components/MenuSide.js
@@ -187,11 +187,9 @@ const MenuSide = () => {
   }, [])
 
   const toggleCollapsed = () => {
-    setCollapsed((prev) => {
-      const next = !prev
-      writeCollapsed(next)
-      return next
-    })
+    const next = !collapsed
+    setCollapsed(next)
+    writeCollapsed(next)
   }
 
   const handleNavClick = (item) => {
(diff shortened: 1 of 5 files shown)


xinference/ui/web/ui/src/components/MenuSide.js
```diff
@@ -187,74 +211,132 @@ const MenuSide = () => {
...
+      <Box
+        display="flex"
+        justifyContent={collapsed ? 'center' : 'flex-end'}
+        px={1}
+        pb={1}
+      >
+        <Tooltip
+          title={
+            collapsed
+              ? t('menu.expand') || 'Expand'
```
qinxuye: This fallback will not be used when the key is missing: i18next returns the key string (`menu.expand`) by default, which is truthy. Since the locale files do not define `menu.expand` / `menu.collapse`, the tooltip will show the raw key. Please either add these keys to all locales or use i18next's default value form, e.g. `t('menu.expand', 'Expand')` / `t('menu.collapse', 'Collapse')`.

qinxuye (commented):
Reviewed the sidebar collapse change and left one inline comment for a small i18n issue.

qinxuye (approved):
LGTM

PR #4910, opened 2026-05-13
fix: harden docs workflow against branch injection

qinxuye:
## Summary
- validate PR head branch names before pushing generated docs updates
- validate fork repository names before constructing the push remote
- pass untrusted PR metadata through quoted shell variables instead of interpolating it directly into run commands

## Validation
- Parsed .github/workflows/pr_auto_run_gen_docs.yaml with PyYAML
- Simulated allowed and rejected branch/repository names
- Ran git diff --check


rogercloud (approved):
Reviewed against the security PoC in #4909. This fixes the direct branch/repo metadata injection path by validating the PR head ref/repository and passing them through quoted shell variables before use in git commands.

There are broader hardening opportunities for the pull_request_target workflow, but they are follow-up scope rather than blockers for this fix.

PR #4913, opened 2026-05-14
fix(ci): harden gen docs workflow

### LINKED ISSUE #4912: Harden pr_auto_run_gen_docs workflow against untrusted PR data
## Background

PR #4910 fixed the direct shell injection risk reported by #4909 by validating PR head branch/repository metadata and avoiding direct interpolation into `run:` commands.

This issue tracks follow-up hardening for `.github/workflows/pr_auto_run_gen_docs.yaml`. The workflow uses `pull_request_target` with write permissions, checks out PR content, runs trusted `gen_docs.py`, and pushes generated documentation back to the PR branch. Even though #4910 addresses the immediate branch-name injection path, the workflow still processes untrusted PR data in a privileged context.

## Goals

Reduce the amount of untrusted PR-controlled content that can affect:

- executed code or templates
- generated output paths
- committed files
- push targets
- secret exposure

## Proposed hardening

1. Validate PR metadata early

Move branch/repository validation to an early step before checkout/install/generation. Reject unexpected branch names before running the expensive or privileged parts of the workflow.

2. Enforce changed-file allowlist

Do not rely only on `pull_request_target.paths`. That filter decides whether the workflow runs, but it does not guarantee the PR changed only those files.

The workflow should compare base/head and fail unless changed files are limited to:

```text
xinference/model/llm/llm_family.json
xinference/model/embedding/model_spec.json
xinference/model/rerank/model_spec.json
xinference/model/image/model_spec.json
xinference/model/audio/model_spec.json
xinference/model/video/model_spec.json
```

3. Generate docs from a sanitized workspace

Instead of running `gen_docs.py` directly inside the PR checkout, create a workspace from the trusted base branch and copy only the allowed JSON files from the PR checkout into it.

This keeps PR-controlled files such as Python code, templates, workflow files, and other repo content out of the generation path.

4. Ensure templates always come from the base branch

`doc/source/gen_docs.py` loads templates from `../templates`, which resolves to `doc/templates` relative to the current working directory. When running inside the PR checkout, this can use PR-controlled templates if the PR modifies them.

The workflow should ensure templates are copied from or resolved against the trusted base checkout.

5. Validate generated output paths

`gen_docs.py` derives output filenames from model names in PR-controlled JSON. Add validation so generated filenames cannot contain path separators, `..`, control characters, or other unsafe characters.

For example, model-generated filenames should be limited to a safe pattern such as:

```text
^[A-Za-z0-9._-]+\.rst$
```

6. Validate the final staged diff before committing

Before committing and pushing, fail unless the staged diff only touches expected generated documentation paths, such as:

```text
doc/source/models/builtin/**
doc/source/getting_started/installation.rst
doc/source/user_guide/backends.rst
doc/source/user_guide/metrics.rst
```

This prevents unexpected files from being committed even if generation behavior changes.

7. Consider disabling automatic push to fork PRs

The workflow currently has logic to push back to fork PRs using `PUSH_TOKEN`. Since this workflow is intended for trusted model-sync branches, it may be safer to support automatic push only for same-repository bot branches and skip fork PRs with a clear message.

8. Split read-only and write steps

Consider separating validation/generation from the push step:

- validation/generation job: `contents: read`
- push job: `contents: write`, only after validation and diff allowlists pass

This reduces the amount of workflow execution that runs with write permissions.

## Suggested target flow

```text
1. Validate PR branch/repository metadata.
2. Checkout trusted base branch.
3. Fetch PR head.
4. Verify changed files match the JSON allowlist.
5. Create a sanitized workspace from base.
6. Copy only allowed PR JSON files into the sanitized workspace.
7. Run base-branch gen_docs.py using base-branch templates.
8. Validate generated/staged diff matches the output allowlist.
9. Commit and push only to a validated same-repo chore/models-sync branch.
```

rogercloud:
## Summary

- validate PR metadata and changed files before running the gen_docs workflow
- run gen_docs from a sanitized base-branch workspace with only allowed PR JSON inputs copied in
- validate generated docs paths before committing and skip automatic fork PR push

Fixes #4912

## Validation

- Parsed .github/workflows/pr_auto_run_gen_docs.yaml with PyYAML
- Ran git diff --check
- Verified staging only includes generated docs paths in a local git simulation
- Tested the workflow end-to-end in rogercloud/inference#1: Auto run gen_docs.py and commit changes to PR completed successfully

### DIFF
--- a/.github/workflows/pr_auto_run_gen_docs.yaml
+++ b/.github/workflows/pr_auto_run_gen_docs.yaml
@@ -13,13 +13,28 @@ on:
 
 permissions:
   contents: write
-  pull-requests: write
+  pull-requests: read
 
 jobs:
   run-gen-docs-and-commit:
     if: startsWith(github.event.pull_request.head.ref, 'chore/models-sync/')
     runs-on: ubuntu-latest
     steps:
+      - name: Validate PR metadata
+        env:
+          HEAD_REF: ${{ github.event.pull_request.head.ref }}
+          HEAD_REPO: ${{ github.event.pull_request.head.repo.full_name }}
+        run: |
+          set -euo pipefail
+          if ! printf '%s' "$HEAD_REF" | grep -Eq '^chore/models-sync/[A-Za-z0-9._/-]+$'; then
+            echo "Unsafe branch name: $HEAD_REF"
+            exit 1
+          fi
+          if ! printf '%s' "$HEAD_REPO" | grep -Eq '^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$'; then
+            echo "Unsafe repository name: $HEAD_REPO"
+            exit 1
+          fi
+
       - name: Checkout base branch (trusted code)
         uses: actions/checkout@v4
         with:
@@ -50,6 +65,15 @@ jobs:
           fi
           echo "run=true" >> $GITHUB_OUTPUT
 
+      - name: Validate changed files
+        if: steps.decide.outputs.run == 'true'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR_FILES_URL: ${{ github.event.pull_request.url }}/files
+        run: |
+          set -euo pipefail
+          python base/.github/scripts/validate_gen_docs_pr.py changed-files
+
       - name: Set up Python
         if: steps.decide.outputs.run == 'true'
         uses: actions/setup-python@v5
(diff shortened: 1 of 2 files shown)


qinxuye (commented):
Left one inline comment about the remaining generated-path boundary.

.github/workflows/pr_auto_run_gen_docs.yaml
```diff
@@ -61,51 +137,92 @@ jobs:
...
+          for path in \
+            doc/source/models/builtin \
+            doc/source/getting_started/installation.rst \
+            doc/source/user_guide/backends.rst \
+            doc/source/user_guide/metrics.rst
+          do
+            if [ -e "gen-docs-work/$path" ]; then
+              mkdir -p "pr/$(dirname "$path")"
+              rm -rf "pr/$path"
+              cp -a "gen-docs-work/$path" "pr/$path"
```
qinxuye, line 140: This is much safer than the previous version, but there is still one path boundary worth tightening: `gen_docs.py` derives builtin doc filenames from PR-controlled `model_name` values before this copy step. The staged-path check below catches unexpected builtin output paths, but a traversal-style model name could still write into one of the explicitly copied/allowed docs paths before validation. Can we reject model names containing path separators or `..` before running `gen_docs.py`, or sanitize filenames inside `gen_docs.py` before writing them?

rogercloud: Addressed in the latest push. I moved the validation logic out of the workflow heredocs into `base/.github/scripts/validate_gen_docs_pr.py`, and added a `model-names` validation step before `gen_docs.py` runs. It reads the sanitized workspace JSON files and rejects any `model_name` that is not a string matching `^[A-Za-z0-9._-]+$`, so path separators, `..`, spaces, and other unsafe path characters are rejected before any generated files are written.

I also kept the staged-path check as a final defense after generation.

qinxuye (approved):
LGTM

PR #4917, opened 2026-05-15
feat(logs): Kibana-style enhancements for Log Center

m199369309:
## Summary

- **DetailRow refactor**: Three-column table (Action/Field/Value) with Table/JSON tabs, field-level filtering (`+`/`-` buttons), multi-keyword highlighting across collapsed and expanded rows, Level/LogType chip highlight sync
- **JSON copy button**: Clipboard API with `execCommand` fallback for HTTP environments, visual feedback with auto-reset
- **Surrounding documents view**: New `GET /v1/cluster/logs/context` backend API with dual ES queries; ContextDialog with editable load count, newest-first ordering, row expansion, isolated filter state, and in-dialog anchor switching
- **ContextDialog enhancements**: Enter key to trigger loading, filter chips bar with client-side data filtering (OR within same key, AND across keys), anchor row always visible
- **i18n**: 18 new keys across en/zh/ja/ko locales

## Test plan

- [ ] Expand a log row → verify three-column Table/JSON tabs render correctly
- [ ] Click `+`/`-` action buttons → verify filter chips appear in toolbar, data filters, and highlighting works in both collapsed and expanded rows
- [ ] Click Level/LogType chips → verify highlighting syncs in DetailRow
- [ ] Switch to JSON tab → verify copy button works (test in both HTTPS and HTTP)
- [ ] Click "View surrounding documents" → verify ContextDialog opens with correct anchor
- [ ] Edit load count and click Load / press Enter → verify newer/older documents load
- [ ] Click a row's "View surrounding documents" inside ContextDialog → verify anchor switches (no nested dialog)
- [ ] Use `+`/`-` filters inside ContextDialog → verify filter chips and data filtering work independently from main page
- [ ] Test with ES not configured → verify graceful 503 error
- [ ] Verify all 4 languages display correctly

### DIFF
--- a/xinference/ui/web/ui/src/scenes/logs/index.js
+++ b/xinference/ui/web/ui/src/scenes/logs/index.js
@@ -848,12 +848,9 @@ const Logs = () => {
     params.set('time_to', timeRange.to)
     params.set('size', String(PAGE_SIZE))
     params.set('page_from', String(pageFrom))
-    if (fieldFilters.length) {
-      params.set(
-        'filters',
-        fieldFilters.map((f) => `${f.op}${f.key}:${String(f.value)}`).join(',')
-      )
-    }
+    fieldFilters.forEach((f) => {
+      params.append('filters', `${f.op}${f.key}:${String(f.value)}`)
+    })
 
     const token = sessionStorage.getItem('token')
     const headers = { 'Content-Type': 'application/json' }
--- a/xinference/ui/web/ui/src/scenes/logs/index.js
+++ b/xinference/ui/web/ui/src/scenes/logs/index.js
@@ -5,8 +5,10 @@ import 'dayjs/locale/zh-cn'
 import {
   AccessTime,
   AddCircleOutline,
+  ArticleOutlined,
   CalendarToday,
   Check as CheckIcon,
+  Close as CloseIcon,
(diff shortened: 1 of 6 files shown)


qinxuye (commented):
Left one inline comment about filter value serialization.

xinference/ui/web/ui/src/scenes/logs/index.js
```diff
@@ -212,6 +848,12 @@ const Logs = () => {
     params.set('time_to', timeRange.to)
     params.set('size', String(PAGE_SIZE))
     params.set('page_from', String(pageFrom))
+    if (fieldFilters.length) {
+      params.set(
+        'filters',
+        fieldFilters.map((f) => `${f.op}${f.key}:${String(f.value)}`).join(',')
```
qinxuye: This comma-delimited encoding loses data when a filter value itself contains a comma, which is common for `message`. For example, clicking `+` on `message = "foo, bar"` sends `filters=+message:foo, bar`; the backend then splits on `,` and only applies `message=foo`, silently dropping `bar`. Can we encode filters as repeated query params or JSON, or otherwise escape each field/value pair before joining?

qinxuye (approved):
LGTM

PR #4919, opened 2026-05-18
BLD: Split Docker builds into separate jobs

qinxuye:
## Summary
- Split Docker CD image publishing into separate aarch64, GPU, and CPU build jobs while keeping the original sequential order.
- Add per-job Docker cache cleanup before and after each build.
- Move latest tag publishing into a separate job that runs after all image builds succeed.
- Open or update a GitHub issue automatically when Docker CD image publishing fails.

## Tests
- python yaml.safe_load for .github/workflows/docker-cd.yaml
- git diff --check

### DIFF
--- a/.github/workflows/docker-cd.yaml
+++ b/.github/workflows/docker-cd.yaml
@@ -80,7 +80,7 @@ jobs:
 
   build-gpu:
     name: Build GPU Docker image
-    needs: check-release-permission
+    needs: build-aarch64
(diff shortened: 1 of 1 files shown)


rogercloud (approved):
Reviewed the Docker CD workflow split at head 9893835. I did not find any blocking issues in the workflow changes. The currently failing metal/GPU CI jobs appear unrelated to this Docker CD-only change.

PR #4933, opened 2026-05-21
feat(web): new UI (partially implemented)

maoyuehui:
1. Framework: Next.js, UI: @radix-ui
2. Global: i18n, request, theme, app-init, constants, hooks, etc.
3. Pages: login, cluster-information, register-model (partial)


qinxuye:
To clarify, tech stack used is aligned with Xagent, and will not be default until version 3.0

frontend/eslint.config.mjs
```diff
@@ -0,0 +1,25 @@
...
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+const compat = new FlatCompat({
+  baseDirectory: __dirname,
+});
+
+const eslintConfig = [
+  ...compat.extends("next/core-web-vitals", "next/typescript", "prettier"),
```
qinxuye, line 13: [P1] The new frontend build cannot load this config because `prettier` is extended here but `eslint-config-prettier` is not listed in `package.json`. Running `npm run build` in `frontend/` currently fails with `ESLint: Failed to load config "prettier" to extend from`. Please either add the missing dependency or remove this extend so production builds can complete.

frontend/src/components/ui/markdown-renderer.tsx
```diff
@@ -0,0 +1,191 @@
...
+          )
+        }
+        return (
+          <a href={href || undefined} title={title || undefined} {...props}>
+            {children}
+          </a>
+        )
+      },
+      img({ node: _node, src, alt, title, ...props }) {
+        if (src && src.startsWith('file:')) {
```
qinxuye: [P1] This currently breaks the TypeScript step in `npm run build`: `react-markdown` types `img.src` as `string | Blob`, so calling `.startsWith()` directly fails with `Property 'startsWith' does not exist on type 'string | Blob'`. Please narrow `src` with `typeof src === 'string'` before using string methods, and avoid passing a Blob through to the plain `img` `src` prop below.

qinxuye (commented):
I found two current build blockers that are not covered by the existing review threads. Also, `git diff --check upstream/main...HEAD` currently fails due to trailing whitespace in several frontend files, so please run the formatter or clear those whitespace errors as well.

qinxuye (approved):

PR #4949, opened 2026-05-27
feat(security): add IP/Key ban and rate limiting for brute-force protection

m199369309:
## Summary

- Add rate limiter with configurable thresholds for login and API endpoints
- Add IP and API Key ban management with automatic expiry
- Add security settings UI for managing bans and viewing blocked attempts
- Add Prometheus metrics for security events (failed logins, bans, rate limits)

## Dependencies

> ⚠️ This PR depends on #4948 (OIDC Auth).
> Please merge #4948 first. I will rebase this branch after it's merged.

## Changes

- `xinference/api/oauth2/advanced/rate_limiter.py` (new): Token bucket rate limiter with IP/Key tracking
- `xinference/api/oauth2/advanced/security_routes.py` (new): Ban management API endpoints
- `xinference/api/oauth2/advanced/auth_service.py`: Integrate rate limiting into auth flow
- `xinference/core/metrics.py`: Security event Prometheus counters
- `xinference/constants.py`: Security-related environment variable constants
- `xinference/ui/web/ui/src/scenes/security_settings/index.js` (new): Ban management UI
- `xinference/ui/web/ui/src/components/MenuSide.js`: Add security settings menu entry
- `xinference/ui/web/ui/src/router/index.js`: Add security settings route
- `xinference/ui/web/ui/src/locales/*.json`: i18n keys for security features

## How to Review

- Review the diff assuming #4948 (OIDC Auth) is already merged
- Conflicts with #4948 in shared files (constants.py, auth_service.py, locales) will resolve after rebase

## Testing

- Verified rate limiting triggers after threshold exceeded
- Verified IP/Key ban and automatic unban after expiry
- Verified security metrics exported to Prometheus


qinxuye (commented):
Found several non-duplicate blockers in the current security-hardening branch.

xinference/api/oauth2/advanced/auth_service.py
```diff
@@ -278,26 +275,7 @@ async def __call__(
...
-
-            def record_audit_event(**kwargs):
-                pass
-
-            def classify_endpoint(endpoint):
-                return ""
-
-            def resolve_model_info(model_id):
-                return "", ""
+        from .audit import classify_endpoint, record_audit_event, should_skip_audit
```
qinxuye: [P1] This reintroduces a hard dependency on `.audit`, but this PR and current `main` do not include `xinference/api/oauth2/advanced/audit.py`. I reproduced the advanced auth dependency path locally and it raises `ModuleNotFoundError: No module named 'xinference.api.oauth2.advanced.audit'` before it can validate the token. Please either include the audit module in this PR/dependency chain or keep the fallback used by the OIDC branch.

xinference/api/oauth2/advanced/security_routes.py
```diff
@@ -0,0 +1,205 @@
...
+async def unban_key_all(key_id: int, request: Request) -> JSONResponse:
+    rl = _get_rate_limiter(request)
+    count = rl.unban_key_all(key_id)
+    return JSONResponse(content={"unbanned": count})
+
+
+# --- Route registration ---
+
+
+def register_security_routes(api: "RESTfulAPI") -> None:
```
qinxuye, line 138: [P1] `register_security_routes` is never called. `RESTfulAPI` registers the normal advanced auth routes, but neither `restful_api.py` nor `advanced/routes.py` imports or invokes this function, so every new `/v1/admin/security/*` endpoint added by this file remains unregistered and the new UI will hit 404s. Please wire this registration into app startup when advanced auth is enabled.

xinference/ui/web/ui/src/router/index.js
```diff
@@ -5,6 +5,7 @@ import { useNavigate } from 'react-router-dom'
 import { ApiContext } from '../components/apiContext'
 import Layout from '../scenes/_layout'
 import ApiKeyManagement from '../scenes/apikey_management'
+import AuditLog from '../scenes/audit_log'
```
qinxuye: [P1] This imports `../scenes/audit_log`, but this PR does not add that scene and current `main` does not have it. The frontend bundle will fail module resolution as soon as this route file is built. If audit log is not part of this PR, please remove the audit route/menu addition from this branch or include the missing scene as part of the declared dependency.

xinference/constants.py
```diff
@@ -87,9 +87,7 @@ def get_xinference_home() -> str:
 XINFERENCE_CACHE_DIR = os.path.join(XINFERENCE_HOME, "cache")
 XINFERENCE_TENSORIZER_DIR = os.path.join(XINFERENCE_HOME, "tensorizer")
 XINFERENCE_MODEL_DIR = os.path.join(XINFERENCE_HOME, "model")
-XINFERENCE_LOG_DIR = os.environ.get(
-    "XINFERENCE_LOG_DIR", os.path.join(XINFERENCE_HOME, "logs")
-)
+XINFERENCE_LOG_DIR = os.path.join(XINFERENCE_HOME, "logs")
```
qinxuye: [P2] This drops the `XINFERENCE_LOG_DIR` environment override that is already present on `main` and reverts the log directory back to a fixed path. That looks unrelated to rate limiting and regresses deployments that configure log location through `XINFERENCE_LOG_DIR`. Please preserve the existing env-based definition while adding the new rate-limit constants.

xinference/api/oauth2/advanced/rate_limiter.py
```diff
@@ -0,0 +1,278 @@
...
+        now = time.time()
+        if now - self._last_cleanup < CLEANUP_INTERVAL:
+            return
+        self._last_cleanup = now
+        pairs: List[Tuple[Dict, RateLimitConfig]] = [
+            (self._ip_records, self._ip_config),
+            (self._key_records, self._key_config),
+        ]
+        for records, config in pairs:
+            stale_keys = [
```
qinxuye, line 78: This cleanup still leaves expired ban records behind because it only removes records with `not r.banned_until`. Once an IP/key has been banned, `banned_until` remains non-zero forever, so after the ban expires `_maybe_cleanup()` will not delete that stale record. I reproduced this locally by setting a 1s ban/window and forcing cleanup after expiry; the IP record remains in `_ip_records`. Please treat expired bans as removable, e.g. `not (r.banned_until and r.banned_until > now)`, plus the stale timestamp condition.

xinference/api/oauth2/advanced/auth_service.py
```diff
@@ -37,6 +37,15 @@
 
 logger = logging.getLogger(__name__)
 
+
+def _get_client_ip(request: Request) -> str:
```
qinxuye, line 56: This now trusts `X-Forwarded-For` / `X-Real-IP` unconditionally. If Xinference is exposed directly or through an untrusted hop, clients can spoof this header and rotate the apparent IP to bypass the new rate limiter. Please gate proxy-header handling behind an explicit trusted-proxy setting or only honor it when the immediate peer is trusted.

xinference/ui/web/ui/src/components/MenuSide.js
```diff
@@ -160,6 +161,13 @@ const MenuSide = () => {
       : []),
     ...(authAdvanced
       ? [
+          {
+            text: 'audit_log',
```
qinxuye: The router import/route for `audit_log` was removed, but this menu entry still navigates to `/audit_log`. Clicking it now falls through to the wildcard route instead of opening a valid page. Please remove this menu item as well, or keep it only in the PR that adds the audit route/page.

qinxuye (commented):
Found a few remaining issues after the latest update.

qinxuye (approved):

PR #4951, opened 2026-05-27
feat(audit): add comprehensive audit logging system

m199369309:
## Summary

- Add audit log module with structured event recording for all sensitive operations
- Add audit API endpoints for querying, filtering, and exporting audit events
- Add audit log UI with Kibana-style search, time range filter, and CSV export
- Add Elasticsearch audit index template for centralized log storage
- Integrate audit logging into admin operations and API key management

## Dependencies

> ⚠️ This PR depends on #4948 (OIDC Auth).
> Please merge #4948 first. I will rebase this branch after it's merged.

## Changes

- `xinference/api/oauth2/advanced/audit.py` (new): Audit event recorder and query service
- `xinference/api/restful_api.py`: Audit middleware integration
- `xinference/api/routers/admin.py`: Audit log query endpoints
- `xinference/constants.py`: Audit-related environment variable constants
- `xinference/ui/web/ui/src/scenes/audit_log/index.js` (new): Audit log viewer UI
- `xinference/ui/web/ui/src/components/MenuSide.js`: Add audit log menu entry
- `xinference/ui/web/ui/src/router/index.js`: Add audit log route
- `xinference/ui/web/ui/src/locales/*.json`: i18n keys for audit features
- `monitor/filebate/es-audit-index-template.json` (new): ES index template

## How to Review

- Review the diff assuming #4948 (OIDC Auth) is already merged
- Conflicts in shared files will resolve after rebase

## Testing

- Verified audit events recorded for login, API key CRUD, model operations
- Verified audit log UI filtering and export functionality
- Verified ES index template creates correct mappings


qinxuye (commented):
Found two blocking issues that are not covered by the existing review threads.

xinference/ui/web/ui/src/router/index.js
```diff
@@ -13,6 +14,7 @@ import Logs from '../scenes/logs'
 import Monitoring from '../scenes/monitoring'
 import RegisterModel from '../scenes/register_model'
 import RunningModels from '../scenes/running_models'
+import SecuritySettings from '../scenes/security_settings'
```
qinxuye: This imports `../scenes/security_settings`, but this PR does not add `xinference/ui/web/ui/src/scenes/security_settings` and the PR only declares a dependency on #4948. As-is the frontend bundle fails module resolution. Please either remove the security settings route/menu from this PR, or make the PR explicitly depend on the PR that adds that scene and rebase after it lands.

xinference/api/oauth2/advanced/audit.py
```diff
@@ -0,0 +1,171 @@
...
+    global _audit_logger
+    if _audit_logger is not None:
+        return _audit_logger
+
+    _audit_logger = logging.getLogger("xinference.audit")
+    _audit_logger.setLevel(logging.INFO)
+    _audit_logger.propagate = False
+
+    handler = create_rotating_handler(
+        filename=os.path.join(XINFERENCE_LOG_DIR, "audit.log"),
```
qinxuye, line 39: This writes audit events to a local `audit.log`, but the PR only adds an Elasticsearch index template. The existing bundled Filebeat ES config tails worker/supervisor logs and writes `xinference-logs-*`, while the new search endpoint queries `xinference-audit-*`, so the audit UI/API will have no records with the provided deployment path. Please add/update the ingestion config so `audit.log` is collected and indexed into the audit index, or otherwise wire the search path to where these events are actually shipped.

monitor/filebate/filebeat-to-es.yml
```diff
@@ -31,6 +31,19 @@ filebeat.inputs:
       cluster: xinference
     fields_under_root: true
 
+  - type: log
+    id: xinference-audit
+    enabled: true
+    paths:
+      - /data/logs/audit.log
```
qinxuye, line 38: This still does not match where the audit logger writes by default. `record_audit_event()` writes to `XINFERENCE_LOG_DIR/audit.log`, whose default is `$XINFERENCE_HOME/logs/audit.log`, but this Filebeat input watches `/data/logs/audit.log` and `${XINFERENCE_HOME}/audit.log`; the compose file also only mounts `/data/logs/worker` and `/data/logs/supervisor`. With the provided deployment config, Filebeat will still miss the audit log unless users manually change paths. Please watch/mount the actual `logs/audit.log` path or update the logger/config to use the same path.

xinference/api/routers/admin.py
```diff
@@ -539,6 +541,255 @@ async def fetch(body: dict, name: str) -> dict:
 # --- Route registration ---
 
 
+def _parse_relative_time(expr: str) -> Optional[datetime]:
+    """Parse ES-style relative time like 'now-1h', 'now-7d'."""
+    import re
+
+    if expr == "now":
+        return datetime.now(timezone.utc)
+    m = re.match(r"now-(\d+)([mhdw])", expr)
```
qinxuye, line 550: The local-file fallback only parses `now` / `now-<n><unit>` values. The UI sends epoch millisecond strings for custom absolute ranges, so when Elasticsearch is not configured, custom time filtering silently becomes unbounded because `_parse_relative_time()` returns `None`. Please also parse numeric epoch-ms values (or ISO timestamps) here so the fallback behaves like the ES query path.

qinxuye (commented):
Found two remaining issues after the latest update.

monitor/filebate/filebeat-to-es.yml
```diff
@@ -31,6 +31,19 @@ filebeat.inputs:
...
+      - ${XINFERENCE_HOME}/logs/audit.log
```
qinxuye: This path now matches the default `XINFERENCE_LOG_DIR` layout, so the filename issue is fixed. The remaining problem is deployment wiring: the provided `docker-compose.yml` still only mounts `/data/logs/worker` and `/data/logs/supervisor`, and it does not set/mount `XINFERENCE_HOME`. With that compose setup, Filebeat still cannot see `${XINFERENCE_HOME}/logs/audit.log`. Please add the audit log directory mount (or set/mount `XINFERENCE_HOME`) so the documented compose deployment actually collects audit events.

qinxuye (commented):
One of the two items is fixed; the remaining Filebeat deployment path still needs a small follow-up.

qinxuye (approved):
LGTM

PR #4963, opened 2026-05-29
fix(auth): return specific error messages for expired/disabled API keys

m199369309:
## Summary
- Return `401 API key has expired` when an expired API key is used, instead of the generic "Could not validate credentials"
- Return `401 API key is disabled` when a disabled API key is used
- Improves user experience by clearly indicating why authentication failed

## Test plan
- [ ] Use an expired API key → verify response is `401` with detail `"API key has expired"`
- [ ] Use a disabled API key → verify response is `401` with detail `"API key is disabled"`
- [ ] Use a valid API key → verify normal behavior unchanged
- [ ] Use an invalid/unknown API key → verify still returns generic credentials error

### DIFF
--- a/xinference/api/oauth2/advanced/auth_service.py
+++ b/xinference/api/oauth2/advanced/auth_service.py
@@ -437,7 +437,20 @@ class AdvancedAuthService:
                         )
                     except ImportError:
                         pass
-                raise credentials_exception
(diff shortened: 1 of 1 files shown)


qinxuye (approved):
LGTM

PR #4964, opened 2026-05-29
fix(monitor): periodic refresh for security gauges and ban remaining API

m199369309:
## Summary
- Add `update_security_gauges()` to periodically refresh `api_keys_active_total`, `api_keys_expired_total`, `banned_ips_total`, and `banned_keys_total` in the existing 15s metrics update loop
- Previously these gauges only had values when triggered by CRUD or ban events; after a restart they showed no data
- Add `get_ip_ban_remaining()` and `get_key_ban_remaining()` methods to RateLimiter for querying remaining ban duration

## Test plan
- [ ] Start xinference with advanced auth enabled, verify all 4 gauges have values within 15s without any CRUD/ban events
- [ ] Create/expire API keys → verify `api_keys_active_total` and `api_keys_expired_total` update correctly
- [ ] Trigger IP/key ban → verify `banned_ips_total` and `banned_keys_total` reflect correct counts
- [ ] Restart service → verify gauges recover values within 15s

### DIFF
--- a/xinference/core/metrics.py
+++ b/xinference/core/metrics.py
@@ -524,10 +524,14 @@ def update_security_gauges(auth_service) -> None:
             if not k.get("enabled", 1):
                 continue
             expires_at = k.get("expires_at")
-            if expires_at and datetime.fromisoformat(expires_at) < now:
-                expired += 1
-            else:
-                active += 1
+            if expires_at:
+                try:
+                    if datetime.fromisoformat(expires_at) < now:
+                        expired += 1
+                        continue
+                except ValueError:
+                    continue
+            active += 1
         api_keys_active_total.set({}, active)
         api_keys_expired_total.set({}, expired)
     except Exception:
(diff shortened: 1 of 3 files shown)


xinference/core/metrics.py
```diff
@@ -507,6 +507,55 @@ def update_cluster_metrics(
...
+    try:
+        keys = auth_service._db.list_api_keys()
+        active = 0
+        expired = 0
+        now = datetime.utcnow()
+        for k in keys:
+            if not k.get("enabled", 1):
+                continue
+            expires_at = k.get("expires_at")
+            if expires_at and datetime.fromisoformat(expires_at) < now:
```
qinxuye: This refresh can be disabled by a single malformed `expires_at` value. Unlike `_refresh_key_gauges()` and `ApiKeyCache._build_entry()`, this calls `datetime.fromisoformat(expires_at)` without catching `ValueError`; if one API key has a bad timestamp, the outer `except` skips the entire key-count update and the security gauges remain stale/empty, which is the state this PR is trying to fix. Please handle invalid timestamps per key and continue counting the rest.

qinxuye (approved):
LGTM

PR #4966, opened 2026-05-29
feat:  new ui (register json view, formInstance, launch model list, cache/env …)

maoyuehui:
1. register json view
2. formInstance extension
3. launch model list
- model list
- model update
- cache management
- env management


qinxuye (commented):
I found a few additional issues that do not overlap with the existing review threads.

frontend/src/app/running-model/page.tsx
```diff
@@ -0,0 +1,6 @@
+const RunningModel = ()=>{
```
qinxuye, line 1: 这里已经通过侧边栏暴露了 `/running-model` 入口，但页面现在只是一个空的占位 div。用户点击 Running Models 会看到空白页，建议要么实现运行中模型列表，要么先不要暴露这个入口。

frontend/src/components/pages/launch-model/index.tsx
```diff
@@ -0,0 +1,465 @@
...
+
+    if (target) {
+      router.push(`/launch-model/${target.path}`);
+    }
+  };
+  const updateModels = async () => {
+    setUpdateLoading(true);
+    try {
+      await request.post('/v1/models/update_type', { model_type: refreshType.toLowerCase() });
+      onTabChange(refreshType);
```
qinxuye: 在 Custom tab 里点击 Update 时，`refreshType` 会是 LLM/embedding 等具体类型，调用 `onTabChange(refreshType)` 会跳到内置模型页，例如 `/launch-model/llm`，而不是留在 Custom tab 刷新当前自定义模型类型。这里建议在 custom route 下只刷新当前 custom type，避免更新后切走页面。

frontend/src/components/pages/register-model/model-specs.tsx
```diff
@@ -0,0 +1,124 @@
...
+                  </FormField>
+                  {isLLM && (
+                    <FormField
+                      name={['model_specs', field.name, 'model_size_in_billions']}
+                      label={t('registerModel.modelSizeBillions')}
+                      rules={[
+                        { required: true },
+                        {
+                          pattern: /^\d+(\.\d+)?$/,
+                          message: t('registerModel.ente.rNumberGreaterThanZero'),
```
qinxuye: 这个 i18n key 拼错了，实际 locale 里是 `registerModel.enterNumberGreaterThanZero`。现在校验失败时会显示错误 key 或 fallback 文本，需要改成正确的 key。

frontend/src/app/running-model/page.tsx
maoyuehui: 好的，我先把导航的入口给关掉，到时这个页面开发完成后在打开

frontend/src/components/pages/launch-model/index.tsx
maoyuehui: 好的

frontend/package.json
```diff
@@ -40,6 +40,7 @@
     "marked": "^5.1.2",
     "next": "15.2.3",
     "next-themes": "^0.4.6",
+    "nunjucks": "^3.2.4",
```
qinxuye, line 43: Adding `nunjucks` here requires updating the npm lockfile as well. Right now `package-lock.json` is not in sync, so `npm ci` fails with missing `nunjucks`, `@types/nunjucks`, and transitive entries. Since the frontend CI uses npm install/checks, please update `package-lock.json` or avoid changing `package.json` in this staged PR.

frontend/src/components/pages/launch-model/launch-dialog.tsx
```diff
@@ -0,0 +1,158 @@
...
+    ),
+  };
+  const modelTypeFields: Partial<Record<ModelType, (keyof typeof fields)[]>> = {
+    [ModelType.LLM]: [
+      'modelEngine',
+      'modelFormat',
+      //  'model_size_in_billions', 'quantization'
+    ],
+  };
+  const handleLaunch = async (values: FormValues) => {
```
qinxuye, line 108: If launch is not part of this stage, the exposed Launch action should be disabled or hidden. As written, `handleLaunch` only logs values and the real `/v1/models` request is still commented out, so users can submit the dialog but nothing launches. Also remove the button `onClick={handleLaunch}` and rely on `type="submit"` so validation is not bypassed by a click event.

frontend/src/components/pages/register-model/auto-fill.tsx
```diff
@@ -1,16 +1,113 @@
...
+
+  const handleSubmit = async (values: Record<string, unknown>) => {
+    setLoading(true);
+    try {
+      const res = await request.post('/v1/models/llm/auto-register', values);
+      let newValues = { ...(res || {}) };
+      if (Array.isArray(newValues?.model_specs)) {
+        newValues.model_specs = newValues.model_specs.map((item: any) => ({
+          ...item,
+          model_size_in_billions: item?.model_size_in_billions?.includes('_')
```
qinxuye: This still assumes `model_size_in_billions` is a string. If the API returns a number, `.includes('_')` will throw. Please guard the type before calling `.includes`, e.g. only replace when `typeof size === 'string'`.

frontend/src/components/pages/launch-model/cache-management.dialog.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const { t } = useI18n();
+  const [open, setOpen] = useState(false);
+  const [dataSource, setDataSource] = useState<ModelCachedItem[]>([]);
+  const [pendingDeleteItem, setPendingDeleteItem] = useState<ModelCachedItem>();
+  const [deletingVersion, setDeletingVersion] = useState('');
+
+  const handleCopy = async (value: string) => {
+    if (!value) return;
+
+    await navigator.clipboard?.writeText(value);
```
qinxuye: This still shows a success toast when `navigator.clipboard` is unavailable, because `await undefined` is a no-op. Please explicitly check `navigator.clipboard?.writeText` before calling it and show an error or skip the success toast when copy is unsupported.

frontend/src/components/pages/launch-model/env-management-dialog.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const { t } = useI18n();
+  const [open, setOpen] = useState(false);
+  const [dataSource, setDataSource] = useState<ModelEnvItem[]>([]);
+  const [pendingDeleteItem, setPendingDeleteItem] = useState<ModelEnvItem>();
+  const [deletingLoading, setDeletingLoading] = useState(false);
+
+  const handleCopy = async (value: string) => {
+    if (!value) return;
+
+    await navigator.clipboard?.writeText(value);
```
qinxuye: Same clipboard issue here: when the Clipboard API is unavailable, this code still reports copy success. Please guard `navigator.clipboard?.writeText` before showing `copySuccess`.

qinxuye (commented):
I found a few issues that still matter even if this is a staged implementation. I am not blocking on incomplete future pages, only on exposed behavior or CI-breaking changes.

frontend/src/components/ui/auto-complete.tsx
```diff
@@ -0,0 +1,361 @@
...
+          disabled &&
+            "cursor-not-allowed opacity-50"
+        )}
+        onClick={() => {
+          if (disabled) return
+
+          inputRef.current?.focus()
+        }}
+      >
+        <input
```
qinxuye, line 236: The input still does not call `handleBlurBehavior` on blur. Tabbing away can leave the dropdown open and, when `allowCustomValue` is true, the typed value is not committed unless the user presses Enter or clicks outside.

frontend/src/lib/form.ts
```diff
@@ -0,0 +1,244 @@
...
+  let current = cloned;
+
+  for (
+    let i = 0;
+    i < path.length - 1;
+    i++
+  ) {
+    const key = path[i];
+
+    if (
```
qinxuye, line 101: This still only initializes missing nested paths when the current value is `undefined`. If `current[key]` is `null` or a primitive, the next `current = current[key]` can make the following assignment throw. Please also replace `null` and non-object values with `{}` or `[]` based on the next path segment.

frontend/src/components/pages/launch-model/launch-dialog.tsx
maoyuehui: ok， Outer entry click has been canceled

frontend/src/components/pages/launch-model/index.tsx
```diff
@@ -0,0 +1,469 @@
...
+            >
+              <ExternalLink className="size-3.5" />
+              Details
+            </button>
+          )}
+          <Button
+            size="sm"
+            onClick={(event) => {
+              event.stopPropagation();
+              // setSelectedModel(model);
```
qinxuye: If launching is out of scope for this stage, please hide or disable this button instead of leaving it clickable. Right now the handler only stops propagation and the actual `setSelectedModel(model)` call is commented out, so users can click **Launch** and get no response.

qinxuye (commented):
Left one remaining comment.

qinxuye (approved):
LGTM

PR #4965, opened 2026-05-29
feat(monitor): add per-model GPU memory usage metrics

m199369309:
## Summary
- Add real-time per-model GPU memory monitoring via Prometheus Gauge `xinference:model_gpu_memory_used_bytes`
- Uses pynvml to query per-process GPU memory with a PID tattoo mechanism to associate GPU worker processes (vLLM/SGLang) with their parent model
- Supports multi-GPU and multi-replica scenarios with labels: `model_uid`, `model_name`, `model_type`, `gpu_index`, `worker_address`, `replica_index`

## Changes
- `xinference/device_utils.py` — add `get_per_process_gpu_memory()` via pynvml
- `xinference/core/model.py` — add `get_pid()` method to ModelActor
- `xinference/core/worker.py` — GPU PID snapshot before/after model load + report per-model GPU memory in status
- `xinference/core/supervisor.py` — aggregate model_gpu_memory from workers
- `xinference/core/metrics.py` — new Gauge definitions + collection logic with stale label cleanup

## Test plan
- [ ] Load a model on GPU → verify `model_gpu_memory_used_bytes` appears in `/metrics` with correct labels
- [ ] Load multiple models → verify each model has distinct metrics
- [ ] Unload a model → verify its metric drops to 0 (stale cleanup)
- [ ] Test without GPU (CPU-only) → verify no errors, metric simply not populated
- [ ] Test with vLLM engine → verify GPU worker subprocess memory is attributed to correct model

PR #4969, opened 2026-05-29
feat(monitor): update Grafana dashboards with GPU memory panels

m199369309:
## Summary
- Add per-model GPU memory usage panels (timeseries + table)
- Enhance existing dashboard panels across all 4 language variants (en, ja, ko, zh-CN)
- 1321 insertions, 209 deletions across 4 dashboard JSON files

## Test plan
- [ ] Import updated dashboards into Grafana and verify all panels render correctly
- [ ] Verify per-model GPU memory panels show data when models are loaded


monitor/dashboard/xinference-grafana-dashboard-en.json
```diff
@@ -2715,13 +2715,291 @@
...
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "editorMode": "code",
+          "expr": "xinference:model_gpu_memory_used_bytes{worker_address=~\"$worker_address\", model_name=~\"$model_name\"}",
```
qinxuye, line 2808: The new panels query `xinference:model_gpu_memory_used_bytes`, but this PR does not add that metric and it is not present in the current codebase. As a result, these panels will render empty after merge unless the metric is added first or the panels are changed to use an existing metric. This applies to the same query in all localized dashboard files.

monitor/dashboard/xinference-grafana-dashboard-zh-CN.json
```diff
@@ -2715,13 +2715,291 @@
...
+      "type": "table"
+    },
     {
       "collapsed": false,
       "gridPos": {
         "h": 1,
         "w": 24,
         "x": 0,
-        "y": 131
+        "y": 140
```
qinxuye: The two new panels start at `y=131` with height `10`, so the next row should start at `y=141`. Here it starts at `y=140`, which overlaps the new panels by one grid row in the zh-CN dashboard. The other localized dashboards already use `y=141` here.

monitor/dashboard/xinference-grafana-dashboard-en.json
m199369309: The metric `xinference:model_gpu_memory_used_bytes` was already added in PR #4965 (merged). This dashboard PR depends on that metric being available at runtime.

qinxuye (approved):
LGTM

PR #4972, opened 2026-05-30
feat: persist launch model configuration history server-side

m199369309:
## Summary

The "Launch Model" drawer's deployment configuration history (engine, format, quantization, replica count, etc.) is currently stored only in the browser's `localStorage` (`historyArr`). This means the history is lost when switching browsers, using incognito mode, or moving to another machine.

This PR persists that history server-side in a standalone SQLite store so it is shared across all clients, and records the creator's username (`created_by`) when authentication is enabled.

### Backend
- New `xinference/core/launch_history_store.py`: SQLite-backed store (WAL, thread lock, UPSERT keyed on `(model_name, model_uid)`). `created_by` is preserved across later updates; `updated_by`/`updated_at` change on update. Timestamps are emitted as ISO-8601 with explicit `Z` so clients parse them as UTC.
- New `xinference/api/routers/launch_history.py`: `GET`/`POST`/`DELETE /v1/launch_history`. When authentication is enabled, routes are guarded by `models:list` / `models:write` scopes and `created_by`/`updated_by` are filled from the authenticated user; when auth is off the fields are empty strings.
- New env var `XINFERENCE_LAUNCH_HISTORY_DB_PATH` (default `<XINFERENCE_HOME>/launch_history.db`), independent of the auth database, so it works in any deployment mode.
- Unit tests in `xinference/api/tests/test_launch_history.py` and docs in `environments.rst`.

### Frontend
- `launchModelDrawer.js` now treats the API as the source of truth while keeping `localStorage` as an offline fallback: locally cached entries render immediately on open for a fast paint, then the server set replaces them; successful launches sync to the server; deletions are removed on the server. Each history item displays the creator (`created_by`).
- Adds the `createdBy` label to `en` / `zh` / `ja` / `ko` locales.

### Notes / design decisions
- The history is **globally shared** (all authenticated users see all stored configurations). This is intentional for the team-shared use case and was confirmed with maintainers beforehand.
- The store is deliberately separate from the auth DB so it functions whether or not authentication is enabled.

## Test plan
- [ ] `pytest -vv xinference/api/tests/test_launch_history.py`
- [ ] Deploy a model in a normal browser, confirm the configuration is written server-side (row in `launch_history.db`).
- [ ] Open the same page in an incognito window (empty localStorage), confirm history loads from the API.
- [ ] With authentication enabled, deploy as user A and confirm `created_by` shows A for user B.
- [ ] With authentication disabled, confirm no creator row is rendered.
- [ ] Restart the service, confirm history persists.
- [ ] `cd xinference/ui/web/ui && npm ci && npx eslint . && npx prettier --check . && npm run build`

### DIFF
--- a/xinference/ui/web/ui/src/scenes/launch_model/components/launchModelDrawer.js
+++ b/xinference/ui/web/ui/src/scenes/launch_model/components/launchModelDrawer.js
@@ -210,11 +210,18 @@ const LaunchModelDrawer = ({
 
     const modelUID = data.model_uid || item.model_uid || ''
 
+    const rawUpdatedAt = item.updated_at
+    const updatedAt =
+      typeof rawUpdatedAt === 'number'
+        ? rawUpdatedAt
+        : Date.parse(rawUpdatedAt)
+
     return {
       cache_key: buildHistoryKey(modelName, modelUID),
       model_name: modelName,
       model_uid: modelUID,
-      updated_at: Number(item.updated_at) || 0,
+      updated_at: Number.isFinite(updatedAt) ? updatedAt : 0,
+      created_by: item.created_by || data.created_by || '',
       data,
     }
   }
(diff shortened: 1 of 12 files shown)


xinference/ui/web/ui/src/scenes/launch_model/components/launchModelDrawer.js
```diff
@@ -247,12 +252,68 @@ const LaunchModelDrawer = ({
...
+    } catch (error) {
+      console.error('Failed to load launch history from server:', error)
+      // Offline fallback: use whatever is cached locally.
+      return getHistoryEntriesForModel()
+    }
+  }
+
+  const syncHistoryToServer = async (data) => {
+    try {
+      await fetchWrapper.post('/v1/launch_history', {
```
qinxuye, line 294: This persists the full launch payload to the server, including user-provided fields such as `envs` produced by `getFinalFormData()` below. The backend GET endpoint only requires `models:list`, so in auth-enabled deployments any user with model-list permission can read other users' saved launch configurations for the same model, potentially including environment variables or tokens. This changes the previous localStorage-only visibility semantics. Please either scope launch history by `created_by`, or at least redact sensitive fields such as `envs` before persisting the payload.

qinxuye (commented):
I found one remaining issue around the persisted launch history scope.

xinference/ui/web/ui/src/scenes/launch_model/components/launchModelDrawer.js
m199369309: Thanks, good catch. Fixed in 357d027a by scoping launch history per user (the stronger of the two options you suggested), so secrets never leave their owner.

**Store** (`launch_history_store.py`)
- Unique key changed `(model_name, model_uid)` → `(model_name, model_uid, created_by)`, so each user owns a separate row; `upsert` keeps the **full** payload for the owner (no params lost across devices). The table is new in this PR, so there's no migration concern.
- `list(username=...)` redacts rows owned by **other** users down to a `SHAREABLE_KEYS` whitelist before returning. I went with a whitelist rather than a denylist on `envs` because `getFinalFormData()` flattens arbitrary `custom` keys into the top-level payload, so a user could inject a secret under any key name (`api_key`, `hf_token`, ...) that a fixed denylist would miss.
- Whitelist = structural deploy params only: `model_uid/name/type/engine/format, model_size_in_billions, quantization, n_worker, n_gpu, n_gpu_layers, replica, request_limits, gpu_idx, download_hub, reasoning_content, gguf_quantization, lightning_version, cpu_offload, enable_thinking, multimodal_projector, enable_virtual_env, virtual_env_packages`. Everything else (`envs`, `quantization_config`, `peft_model_config`, `*_model_path`, `worker_ip`, and any unknown `custom` key) is stripped for non-owners.
- `delete(model_name, model_uid, username)` is scoped to `created_by == username`, so a user can only delete their own row.
- `username is None` means a trusted/internal call → no redaction; auth-disabled deployments pass `username=""`, all rows have `created_by=""` → treated as owner → full payload (same behavior as before auth).

**Router** GET/DELETE now capture the authenticated user and forward `username`, mirroring the POST auth on/off wrappers (so FastAPI doesn't treat `user` as a body field and 422).

**Frontend** `cache_key` now includes `created_by` so per-user entries don't collide in the local mirror / list rendering.

Net effect: owner sees everything (cross-device), teammates see only the non-sensitive structural config + `created_by`, secrets never leak. Covered by new tests: `test_list_redacts_other_users_sensitive_fields`, `test_list_returns_full_data_for_owner`, `test_upsert_distinct_users_create_separate_rows`, `test_delete_scoped_to_owner`.

qinxuye (approved):
LGTM

PR #4977, opened 2026-05-31
fix(monitor): capture vLLM/SGLang GPU workers via deferred PID tattoo

m199369309:
## Summary

Per-model GPU memory metrics added in #4965 attribute usage by walking the `ModelActor`'s psutil process tree. This works for embedding/rerank models (loaded in-process via Transformers) but **misses vLLM/SGLang tensor-parallel engines**: their GPU worker / `EngineCore` processes are spawned

- **outside** the `ModelActor` process tree (independent multiprocessing / Ray spawn), and
- **after** `model_ref.load()` returns.

As a result `psutil.Process(...).children(recursive=True)` never finds them, and LLM models report no `xinference:model_gpu_memory_used_bytes`.

## Approach

Deferred GPU-PID "tattoo":

1. Before `model_ref.load()`, snapshot the set of GPU PIDs (`get_per_process_gpu_memory().keys()`).
2. In `wait_for_load()` — invoked per replica after the engine is fully up — take a fresh snapshot and diff it against the pre-load set. The new PIDs are the model's GPU workers.
3. Tattoo those PIDs to the model and merge them into the `report_status` collection loop alongside the existing process-tree PIDs.

Snapshot calls are serialized with an `asyncio.Lock`; all per-model state is cleaned up on load failure and on `terminate_model`. The change is purely additive (+47/-1, single file) and builds only on primitives already introduced by #4965.

## Known limitation

Concurrent launches may cross-attribute newly spawned PIDs (the `B - A` diff can include another model's workers if it spawns within the same window). This affects **monitoring attribution only**, not functionality; the lock narrows but does not fully eliminate the window.

## Test plan

- [ ] Deploy a vLLM LLM (e.g. qwen3, TP=2); worker log shows `GPU PID tattoo for ...`.
- [ ] `/metrics` reports `xinference:model_gpu_memory_used_bytes` for the LLM (~per-GPU VRAM).
- [ ] embedding/rerank model metrics unchanged.
- [ ] Concurrent launch of two models attributes PIDs to the correct model.
- [ ] Load failure / terminate leaves no leaked snapshot or tattoo state.

Related: #4965

### DIFF
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -217,9 +217,11 @@ class WorkerActor(xo.StatelessActor):
         self._model_uid_to_recover_count: Dict[str, Optional[int]] = {}
         self._model_uid_to_launch_args: Dict[str, Dict] = {}
         self._model_uid_to_pid: Dict[str, int] = {}
-        self._model_uid_to_gpu_pids: Dict[str, set] = {}
-        self._model_uid_to_gpu_snapshot_before: Dict[str, set] = {}
-        self._gpu_snapshot_lock = asyncio.Lock()
+        # Per-replica sub-pool process PIDs (primary ModelActor pool + per-device
+        # vLLM/SGLang rank pools). Used by report_status to attribute NVML GPU
+        # memory back to the owning replica deterministically, without reading
+        # any process environ.
+        self._model_uid_to_subpool_pids: Dict[str, Set[int]] = {}
 
         if XINFERENCE_DISABLE_METRICS:
             logger.info(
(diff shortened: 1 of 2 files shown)


xinference/core/worker.py
```diff
@@ -2204,6 +2225,26 @@ async def wait_for_load(self, model_uid: str):
...
+        before_pids = self._model_uid_to_gpu_snapshot_before.pop(model_uid, None)
+        if before_pids is not None:
+            try:
+                from ..device_utils import get_per_process_gpu_memory
+
+                async with self._gpu_snapshot_lock:
+                    after_pids = set(
+                        (await asyncio.to_thread(get_per_process_gpu_memory)).keys()
+                    )
+                new_pids = after_pids - before_pids
```
qinxuye: This PID attribution can be wrong when two models launch concurrently on the same worker. The snapshots are node-wide, and `_gpu_snapshot_lock` only protects each NVML query, not the whole load interval. If model A and model B overlap, A's `after_pids - before_pids` can include B's newly spawned vLLM/SGLang worker PID, and then `report_status` will count B's GPU memory under A as well. Please make the attribution model-specific, e.g. filter candidate PIDs by the model's assigned GPU IDs and avoid assigning PIDs already claimed by another model.

qinxuye (commented):
I found one additional issue that does not overlap with the existing review threads.

qinxuye (approved):
LGTM

PR #4980, opened 2026-06-01
bug: Adapt vLLM LoRA request path parameter

amumu96:
## Summary

- Adapt vLLM LoRA request construction for vLLM 0.14.0 and newer by passing `lora_path`.
- Keep `lora_local_path` for older vLLM versions to preserve compatibility.

## Release note

- Fix LoRA module loading with newer vLLM versions that expect `LoRARequest(lora_path=...)`.

## Validation

- `python -m py_compile xinference/model/llm/vllm/core.py`
- `git diff --check -- xinference/model/llm/vllm/core.py`

Fixes #4764

### DIFF
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -476,6 +476,15 @@ class VLLMModel(LLM):
 
         if self.lora_modules is None:
             self.lora_requests = []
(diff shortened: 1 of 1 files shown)


qinxuye (approved):
LGTM

PR #4947, opened 2026-05-27
feat(logging): enhance logging system with JSON format and stdout redirect

m199369309:
## Summary

- Add structured JSON logging support with configurable output format (`text` / `json`)
- Add stdout/stderr redirect to logging system for capturing model process output
- Add environment variables (`XINFERENCE_LOG_LEVEL`, `XINFERENCE_LOG_FORMAT`, `XINFERENCE_LOG_BACKUP_COUNT`) for flexible configuration
- Refactor deploy utils logging initialization to support new features

## Changes

- `xinference/core/log.py` (new): Core logging module with JSON formatter and stream redirect
- `xinference/constants.py`: Add logging-related environment variable constants
- `xinference/deploy/utils.py`: Refactor logging setup to use new log module
- `xinference/deploy/cmdline.py`, `local.py`, `supervisor.py`, `worker.py`: Integrate new logging config
- `xinference/core/model.py`: Use structured logging for model lifecycle
- `xinference/conftest.py`: Test configuration update

## Dependencies

> None — this PR is independent and can be merged standalone.

## Testing

- Verified JSON log output format with `XINFERENCE_LOG_FORMAT=json`
- Verified stdout redirect captures model process output
- Existing test suite passes without regression

PR #4983, opened 2026-06-02
fix(logging): strip all CSI escapes and route flush() through sampling

m199369309:
## Summary

The stdout/stderr redirect added in #4947 (`StreamToLogger` in `xinference/deploy/utils.py`) is meant to sample tqdm progress bars into `xinference.log`, but two defects let raw tqdm output leak through:

- **Junk `[A` lines.** `_ANSI_ESCAPE_RE` only stripped `m`-terminated SGR color codes (`\x1b\[[0-9;]*m`), so CSI cursor-control sequences such as `\x1b[A` (cursor up) and `\x1b[?25l` (hide cursor) survived stripping and landed in the log as junk lines (a lone `[A`). Widened to strip all CSI sequences: `\x1b\[[0-9;?]*[A-Za-z]`.
- **Sampling defeated by `flush()`.** `flush()` dumped the residual buffer raw, bypassing `_handle_progress`. tqdm writes a `\r`-terminated bar (no trailing terminator for `write()`'s split to consume) and calls `flush()` after every frame, so every frame went out un-sampled — a full per-frame storm instead of the intended 25/50/75/100 sampling. The residual is now routed through `_process_line` so sampling applies.

Also guards buffer mutation with a `threading.Lock`: downloads run multi-threaded (`HF_HUB_DOWNLOAD_WORKERS`) and concurrent writes to the shared buffer would interleave.

No public API, CLI, or schema changes. Behavior change is confined to what gets written to `xinference.log` when `XINFERENCE_LOG_CONSOLE=false`.

## Test plan

- [ ] With `XINFERENCE_LOG_CONSOLE=false`, load a model that emits tqdm bars; confirm `xinference.log` shows sampled progress (25/50/75/100) and **no** lone `[A` / cursor-control junk lines.
- [ ] Confirm no full per-frame storm in the log.
- [ ] Multi-threaded download (`HF_HUB_DOWNLOAD_WORKERS>1`): confirm log lines are not interleaved/truncated.
- [ ] `pre-commit run --all-files` (black/flake8/isort/mypy/codespell).

### DIFF
--- a/xinference/deploy/utils.py
+++ b/xinference/deploy/utils.py
@@ -131,7 +131,12 @@ class StreamToLogger:
         if not match:
             return
         pct = int(match.group(1))
-        key = line[:30]
+        # Derive the sampling key from the stable description *before* the
+        # percentage, so every frame of the same bar shares one key. Using
+        # line[:30] folds the changing percent/bar text into the key for
+        # short-description formats (e.g. "Downloading: 26%|"), making each
+        # frame a distinct key and defeating threshold sampling.
+        key = line[: match.start()].strip()
         last = self._last_reported_pct.get(key, 0)
         for threshold in sorted(self._progress_thresholds):
             if last < threshold <= pct:
--- a/xinference/deploy/utils.py
+++ b/xinference/deploy/utils.py
@@ -95,7 +95,10 @@ class StreamToLogger:
         self._last_reported_pct: dict = {}
         # Downloads run multi-threaded (HF_HUB_DOWNLOAD_WORKERS); concurrent
         # writes to the shared buffer would interleave without this lock.
-        self._lock = threading.Lock()
+        # Reentrant: while the lock is held we call logger.info(), and a logging
+        # handler error path (Handler.handleError -> sys.stderr) can re-enter
+        # write() on the same thread; RLock avoids a self-deadlock there.
+        self._lock = threading.RLock()
 
     def write(self, message: str) -> int:
         if not message:
--- a/xinference/deploy/utils.py
+++ b/xinference/deploy/utils.py
@@ -72,7 +72,10 @@ class LoggerNameFilter(logging.Filter):
 
 
 _PROGRESS_RE = re.compile(r"(\d+)%\|")
-_ANSI_ESCAPE_RE = re.compile(r"\x1b\[[0-9;]*m")
+# Strip all CSI sequences (cursor moves, hide-cursor, SGR colors), not just the
+# `m`-terminated color codes, so tqdm artifacts like `\x1b[A` / `\x1b[?25l` do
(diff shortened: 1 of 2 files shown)


xinference/deploy/utils.py
```diff
@@ -90,21 +93,25 @@ def __init__(
         self._buffer = ""
         self._progress_thresholds = {25, 50, 75, 100}
         self._last_reported_pct: dict = {}
+        # Downloads run multi-threaded (HF_HUB_DOWNLOAD_WORKERS); concurrent
+        # writes to the shared buffer would interleave without this lock.
+        self._lock = threading.Lock()
```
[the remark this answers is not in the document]
m199369309: Good catch — adopted `threading.RLock()` in the latest push. The handler is correct: while the lock is held we call `logger.info()`, and the `Handler.handleError -> sys.stderr` path can re-enter `write()` on the same thread, which would self-deadlock on a non-reentrant `Lock`. RLock preserves the cross-thread mutual exclusion while allowing same-thread reentry.

xinference/deploy/utils.py
```diff
@@ -132,11 +142,15 @@ def _handle_progress(self, line: str) -> None:
...
-                self._logger.info(line)
+        # Route residual buffer through the sampling path instead of dumping it
+        # raw. tqdm calls flush() after every frame with a `\r`-terminated bar
+        # still in the buffer; a raw dump here bypasses _handle_progress and
+        # defeats sampling, producing a full per-frame storm in the log.
+        with self._lock:
+            if self._buffer:
+                residual = self._buffer
+                self._buffer = ""
+                self._process_line(residual)
```
qinxuye, line 158: Routing the residual buffer through `_process_line()` is the right direction, but the current sampling key still includes the changing percentage/progress text (`line[:30]`). That means each frame can become a different key, so intermediate frames such as `26%` or `99%` can still be logged as if they were the first 25% threshold for a new progress bar. Please derive the key from a stable part of the progress line, or normalize/remove the dynamic percent/bar fields before tracking `_last_reported_pct`.

qinxuye (commented):
Left one remaining issue.

xinference/deploy/utils.py
m199369309: Good catch. Fixed in aa330622: the sampling key now derives from the stable description before the percentage (`key = line[: match.start()].strip()`) instead of the fixed `line[:30]` slice, so every frame of the same bar shares one key and threshold sampling works for short-description bars too.

Added a regression test (`TestStreamToLoggerProgressSampling`) that feeds a short-description bar through frames 26/30/40/50/60/76/99/100 and asserts only the threshold-crossing frames (26, 50, 76, 100) are logged, plus a test that two distinct bars are tracked independently.

qinxuye (approved):
LGTM

PR #4985, opened 2026-06-02
fix(vllm): read json_schema from `schema_` so guided decoding applies

m199369309:
## Summary

`response_format={"type": "json_schema", ...}` was silently ignored on the vLLM backend — the model returned free-form text instead of schema-constrained JSON.

Root cause is a wrong key lookup in `VLLMModel._sanitize_generate_config` (`xinference/model/llm/vllm/core.py`):

```python
elif response_format.get("type") == "json_schema":
    json_schema = response_format.get("json_schema")
    assert json_schema is not None
    guided_json = json_schema.get("json_schema")   # key does not exist -> None
```

The OpenAI-compatible request body serializes `_compat.JSONSchema` **without** `by_alias` (`restful_api.py` uses `body.dict(exclude_unset=True, ...)`). Since `schema` is a reserved Pydantic name, the model field is `schema_ = Field(alias="schema")`, and without `by_alias` the serialized key is the **field name** `schema_`. So the inner dict only ever has `name` / `schema_` / `strict` — never `json_schema`. `guided_json` therefore stayed `None`, `guided_params` got no `json`, and vLLM skipped guided decoding.

### Fix

Read `schema_` with a `schema` fallback (for raw `extra_body` passthrough):

```python
guided_json = json_schema.get("schema_") or json_schema.get("schema")
```

This matches the already-correct handling elsewhere in the codebase:
- `xinference/model/llm/sglang/core.py` (reads `schema_`, falls back to `schema`)
- `xinference/model/llm/utils.py` (`schema_block.get("schema_") or schema_block.get("schema")`)

vLLM's `core.py` was the lone outlier. Both the chat path and the plain `/v1/completions` path funnel through `_sanitize_generate_config`, so this single change covers both.

### Why fix here, not at serialization

Switching `restful_api.py` to `by_alias=True` would rename the key globally (`schema_` -> `schema`) and break every consumer that currently reads `schema_` (e.g. the chat-config path and the engines listed above). The localized fix is minimal and aligns vLLM with the existing convention.

No public API, request/response schema, or CLI changes. Backward compatible: requests without `response_format`, or with `json_object`, are unaffected.

## Test plan

- [x] Added `TestVLLMSanitizeGenerateConfig` covering: serialized `schema_` key, `schema` alias fallback, `json_object` regression, and no-`response_format` regression.
- [ ] `pytest -vv xinference/model/llm/vllm/tests/test_core_chat_model.py -k SanitizeGenerateConfig` in an env with vLLM installed (these tests are ignored in the non-GPU CI job; I could not run them in my environment).
- [ ] End-to-end: send a `json_schema` `response_format` to `/v1/chat/completions` on a vLLM-served model and confirm the output is schema-valid JSON.
- [ ] `pre-commit run --all-files`.

### DIFF
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -961,7 +961,7 @@ class VLLMModel(LLM):
             elif response_format.get("type") == "json_schema":
                 json_schema = response_format.get("json_schema")
                 assert json_schema is not None
-                guided_json = json_schema.get("json_schema")
+                guided_json = json_schema.get("schema_") or json_schema.get("schema")
 
         sanitized.setdefault("lora_name", generate_config.get("lora_name", None))
         sanitized.setdefault("n", generate_config.get("n", 1))
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -961,7 +961,13 @@ class VLLMModel(LLM):
             elif response_format.get("type") == "json_schema":
                 json_schema = response_format.get("json_schema")
                 assert json_schema is not None
-                guided_json = json_schema.get("schema_") or json_schema.get("schema")
+                # Real serialized key is the field name `schema_` (the model
+                # aliases the reserved `schema`); fall back to `schema` for raw
+                # passthrough. Check `is None` rather than truthiness so a valid
+                # empty schema ({}) is not dropped.
+                guided_json = json_schema.get("schema_")
+                if guided_json is None:
+                    guided_json = json_schema.get("schema")
 
         sanitized.setdefault("lora_name", generate_config.get("lora_name", None))
         sanitized.setdefault("n", generate_config.get("n", 1))
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -1319,7 +1319,9 @@ class VLLMModel(LLM):
             # Extract guided decoding parameters
             guided_params: dict[str, Any] = {}
             guided_json = sanitized_generate_config.pop("guided_json", None)
-            if guided_json:
+            # Check `is not None` rather than truthiness so a valid empty
+            # schema ({}) is forwarded to vLLM instead of being dropped.
+            if guided_json is not None:
                 guided_params["json"] = guided_json
 
             guided_regex = sanitized_generate_config.pop("guided_regex", None)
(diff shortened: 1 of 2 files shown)


xinference/model/llm/vllm/core.py
```diff
@@ -961,7 +961,7 @@ def _sanitize_generate_config(
             elif response_format.get("type") == "json_schema":
                 json_schema = response_format.get("json_schema")
                 assert json_schema is not None
-                guided_json = json_schema.get("json_schema")
+                guided_json = json_schema.get("schema_") or json_schema.get("schema")
```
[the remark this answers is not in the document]
m199369309: Adopted the explicit `is None` check in the latest push, and added a `test_json_schema_empty_schema_preserved` regression test. Note that end-to-end this only matters at the function boundary: downstream `if guided_json:` (core.py:1316) also treats `{}` as falsy, so an empty schema results in no guided decoding either way — but reading `schema_` via `is None` is clearer and more correct.

xinference/model/llm/vllm/core.py
```diff
@@ -961,7 +961,13 @@ def _sanitize_generate_config(
             elif response_format.get("type") == "json_schema":
                 json_schema = response_format.get("json_schema")
                 assert json_schema is not None
-                guided_json = json_schema.get("json_schema")
+                # Real serialized key is the field name `schema_` (the model
+                # aliases the reserved `schema`); fall back to `schema` for raw
+                # passthrough. Check `is None` rather than truthiness so a valid
+                # empty schema ({}) is not dropped.
```
qinxuye, line 967: This preserves `{}` in `sanitized_generate_config`, but the value is still dropped later when building the vLLM guided params because `guided_json` is guarded with `if guided_json:`. Since `{}` is a valid JSON schema, please update that downstream check to `if guided_json is not None:` as well; otherwise the empty-schema test only verifies the intermediate value, not that guided decoding actually receives it.

qinxuye (commented):
The previous issue is only partially fixed; one downstream check still drops empty schemas.

xinference/model/llm/vllm/core.py
m199369309: You're right, thanks. Fixed in 52c7531e: the downstream guard in `async_generate` now uses `if guided_json is not None:` so a preserved empty schema `{}` actually reaches `guided_params["json"]` and vLLM guided decoding, instead of being dropped by the falsy check.

The change is scoped to the `guided_json` guard only and is byte-for-byte equivalent for every non-empty schema and for `None`; the sole behavior change is the `{}` case, which is the intended fix. Updated the `test_json_schema_empty_schema_preserved` comment to note that the async_generate path depends on vLLM and isn't unit-tested locally.

qinxuye (approved):
LGTM

PR #4991, opened 2026-06-04
feat(ui): allow editing API key name and description in edit dialog

m199369309:
## Summary

- Extend the existing "Edit Model Permissions" dialog to also allow editing the API key **name** and **description** fields
- Change the save endpoint from `PUT /v1/admin/keys/{id}/permissions` to the general `PUT /v1/admin/keys/{id}` which already supports all fields
- Update dialog title to "Edit API Key" and add i18n keys for en/zh/ja/ko

## Motivation

Currently, once an API key is assigned, there is no UI to modify its name or description (申请理由). The backend already supports updating these fields via the general update endpoint, but the frontend lacked the entry point.

## Test plan

- [ ] Open the API Key management page
- [ ] Click edit on an existing key → dialog shows name and description fields pre-filled
- [ ] Modify name/description and save → verify changes persist
- [ ] Verify model permissions editing still works as before
- [ ] Test with en/zh/ja/ko locales for correct translations

### DIFF
--- a/xinference/ui/web/ui/src/locales/en.json
+++ b/xinference/ui/web/ui/src/locales/en.json
@@ -454,6 +454,7 @@
     "copied": "Copied to clipboard",
     "editPermTooltip": "Edit permissions",
     "editPermTitle": "Edit Model Permissions —",
+    "editKeyTitle": "Edit API Key —",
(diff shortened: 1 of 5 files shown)


qinxuye (approved):
LGTM

PR #4992, opened 2026-06-04
fix(worker): limit concurrent model launches with semaphore to prevent heartbeat timeouts

m199369309:
## Summary

When deploying 10+ replicas to a single worker node, concurrent fork/IO/GPU operations can overwhelm system resources, causing heartbeat timeouts (`Failed to upload node info: TimeoutError()`) and deployment deadlocks.

This PR adds a per-worker `asyncio.Semaphore` to cap concurrent model launches, configurable via `XINFERENCE_MAX_CONCURRENT_LAUNCHES` (default: 5).

### Changes
- **`xinference/constants.py`**: Add `XINFERENCE_ENV_MAX_CONCURRENT_LAUNCHES` and `XINFERENCE_MAX_CONCURRENT_LAUNCHES` constant
- **`xinference/core/worker.py`**: Wrap heavy launch operations (subpool creation, model download, actor creation, weight loading) in `async with self._launch_semaphore`, with queued/started/finished logging
- **`xinference/core/tests/test_worker.py`**: Add unit tests for semaphore default value, env var override, and concurrency limiting
- **`doc/source/getting_started/environments.rst`**: Document new environment variable

### Design highlights
- Semaphore is per-worker, so multi-worker deployments scale naturally (total parallelism = workers × semaphore value)
- Cancel-aware: checks `cancel_event` immediately after acquiring semaphore (handles cancel during queue wait)
- Uses `_value == 0` instead of `locked()` for more accurate queue detection
- `locals()` capture for auto-recovery happens before semaphore (no behavior change)
- `launching_guard` registration happens before semaphore (prevents duplicate launches)

## Test plan
- [x] `pre-commit run --all-files` passes (black, flake8, isort, mypy, codespell)
- [x] Python syntax check passes for all modified files
- [ ] `pytest -vv xinference/core/tests/test_worker.py` (requires torch in test env)
- [ ] Deploy 12 replicas of jina-embeddings-v3 to a single worker with `XINFERENCE_MAX_CONCURRENT_LAUNCHES=5`
- [ ] Verify `Launch queued` / `Launch started` / `Launch finished` logs appear
- [ ] Verify no consecutive `Failed to upload node info: TimeoutError()` during deployment


m199369309:
Addressed both review comments in 72d03ad7:

1. **`_value == 0` → `locked()`**: Switched to the public API as suggested.
2. **`_launch_waiting` leak fix**: Moved `_was_queued = False` initialization before the `try` block, set `_was_queued = False` after semaphore acquisition, and added a safety decrement in the outer `finally` to handle `CancelledError` during queue wait.

m199369309:
Fixed CI test failures in 3cf18156.

**Root cause**: `xo.create_actor()` returns an `ActorRef`, not the actor instance — direct attribute access like `worker._launch_semaphore` raises `AttributeError`.

**Fix**: Added getter methods to `MockWorkerActor`:
- `get_launch_semaphore_value()`
- `get_launch_active_count()`
- `get_launch_waiting_count()`

Simplified `test_launch_semaphore_concurrency` to not require an actor pool (it was testing Python's `asyncio.Semaphore`, not our code).

xinference/constants.py
```diff
@@ -234,6 +235,10 @@ def get_xinference_home() -> str:
     os.environ.get(XINFERENCE_ENV_STATUS_REPORT_MULTIPLIER, 3)
 )
 
+XINFERENCE_MAX_CONCURRENT_LAUNCHES = int(
```
qinxuye: Please validate this value before constructing the worker semaphore. `asyncio.Semaphore(0)` is legal but starts locked, so `XINFERENCE_MAX_CONCURRENT_LAUNCHES=0` would make every model launch wait forever; negative values fail during worker initialization. Since this is a user-facing environment variable, we should reject values below 1 with a clear configuration error or clamp/fallback to the default.

m199369309:
Fixed XINFERENCE_MAX_CONCURRENT_LAUNCHES validation in b3a00d5e.

**Changes**:
- Added `max(1, ...)` protection to prevent deadlock when value ≤0
- `Semaphore(0)` would cause all launches to wait forever
- `Semaphore(-1)` would raise ValueError during initialization
- Values ≥1 remain unchanged (normal path unaffected)

**Behavior**:
| Input | Before | After |
|-------|--------|-------|
| `5` (default) | `5` | `5` ✅ |
| `2` | `2` | `2` ✅ |
| `0` | Deadlock | `1` (safe fallback) ⚠️ |
| `-1` | ValueError crash | `1` (safe fallback) ⚠️ |

Addresses review from @qinxuye. Gemini Code Assist comments on `locked()` and `_launch_waiting` leak were already addressed in commit `72d03ad7`.

qinxuye (approved):
LGTM

PR #4993, opened 2026-06-05
fix(venv): allow user override

llyycchhee:
Prepare venv expands engine dependencies before user override.


qinxuye (approved):
LGTM. The ordering change is correct: expanding engine dependency placeholders before merging allows user-provided package pins to replace the default engine dependencies as intended. Focused virtualenv merge tests pass locally.

PR #4998, opened 2026-06-06
fix(supervisor): self-heal worker registry after supervisor restart

m199369309:
## Summary

After a supervisor restart, the in-memory worker registry (`_worker_address_to_worker`) is wiped but workers continue sending heartbeats and status reports using cached supervisor refs. The existing `report_worker_status` silently fabricated `_worker_status` entries for unregistered workers, causing:

- `xinference:workers_total` to stay stuck below the real worker count (e.g. 5 workers but only 2 reported)
- The scheduler (which only iterates `_worker_address_to_worker` in `_choose_worker`) to permanently skip those workers for new model launches
- A "green dashboard, broken cluster" scenario: worker-level metrics show all workers alive, but `workers_total` is wrong and new models never get scheduled on the missing workers

## Root cause

`_worker_address_to_worker` (the registry) and `_worker_status` (the heartbeat table) are two separate dicts written at different times:

- **Registry** is only written by `add_worker`, which is called on worker first-connect and on reconnect. Once cached, `get_supervisor_ref` short-circuits and never re-runs `add_worker`.
- **Heartbeat table** is unconditionally fabricated by `report_worker_status` and `receive_heartbeat` for any reporting worker, regardless of whether it is in the registry.

After a supervisor restart, the heartbeat table self-heals (workers keep reporting), but the registry does not self-heal unless every worker happens to go through the `add_worker` path — which is rare.

## Fix

`report_worker_status` now raises `WorkerNotRegisteredError` when a worker reports but is absent from the registry. This triggers the worker's **existing** `report_status` except-branch (already present in `worker.py`), which calls `_clear_supervisor_refs()` and re-runs `add_worker` with `replica_states`, rebuilding the registry and restoring `workers_total`.

`receive_heartbeat` is changed to a no-op for unregistered workers (instead of fabricating a status entry) so dead-node detection is not fooled by ghost heartbeats that would keep the registry permanently stale. Heartbeat cannot trigger re-registration because the worker heartbeat path has no reconnect branch; only `report_status` does.

A new `WorkerNotRegisteredError` exception (subclass of `Exception`, not `RuntimeError`) is introduced: the worker report loop treats some `RuntimeError`s as fatal and would break the recovery loop if we raised `RuntimeError` here.

## Test plan

- [x] Unit tests added in `test_worker.py`:
  - `test_supervisor_report_worker_status_rejects_unregistered_worker`
  - `test_supervisor_report_worker_status_accepts_registered_worker`
- [x] Unit tests added/updated in `test_heartbeat.py`:
  - `test_supervisor_receive_heartbeat_unregistered_worker_is_noop` (new)
  - `test_supervisor_receive_heartbeat_registered_new_worker` (renamed from `test_supervisor_receive_heartbeat_new_worker` to reflect the new registry gate)
  - `test_supervisor_receive_heartbeat_multiple_calls` (updated to register the worker first)
- [x] Existing tests (`test_worker_report_status_reconnects_and_replays_running_models`, `test_worker_report_status_refreshes_supervisor_internal_address_on_reconnect`) already cover the worker-side reconnect/replay path that this fix relies on.
- [x] Verified functional behavior in production cluster (5-worker cluster, supervisor restart scenario)

## Related

- Follow-up: a companion PR will add Prometheus stale-series cleanup and an `unexpected_termination` gauge; that PR is stacked on this one and will be opened shortly.

### DIFF
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -100,7 +100,7 @@ class WorkerNotRegisteredError(Exception):
     branch, which re-runs ``add_worker`` and self-heals the registry (and thus
     ``workers_total``). Subclasses ``Exception`` (not ``RuntimeError``) on
     purpose: the worker report loop treats some ``RuntimeError``s as fatal and
-    breaks, which must never happen here. See optimize/20260603/2026060306.md.
+    breaks, which must never happen here.
     """
 
 
@@ -2670,7 +2670,7 @@ class SupervisorActor(xo.StatelessActor):
             # report_status reconnect branch re-runs add_worker and self-heals
             # the registry, restoring workers_total and replaying running
             # replicas. Do NOT fabricate a _worker_status entry here, otherwise
-            # the registry would stay stale forever. See 2026060306.md.
+            # the registry would stay stale forever.
             raise WorkerNotRegisteredError(worker_address)
 
         if worker_address not in self._worker_status:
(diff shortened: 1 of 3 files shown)


qinxuye (approved):
LGTM

PR #4999, opened 2026-06-06
fix(metrics): drop stale series and add unexpected termination gauge

m199369309:
> **Depends on #4998** — please review and merge #4998 first.
> This branch is stacked on top of #4998's branch. Once #4998 merges into
> `main`, I will rebase this branch onto `main` and the diff will shrink to
> only this PR's changes (2 files, +273/-47).

## Summary

Three related issues with supervisor-side Prometheus metrics when workers die, restart, or are removed:

1. **Stale series accumulate as value=0 zombies** (P1): `update_cluster_metrics` called `gauge.set(stale_labels, 0)` for removed models/workers instead of deleting the series. Since aioprometheus has no `remove()` API, these zombie series accumulated indefinitely, driving cardinality up. Fixed via `MetricDict.values.pop()` through a new `_drop_series()` helper.

2. **`model_status` shows "fake online"** (P0): When a worker crashes, the three death-detection paths (`_check_dead_nodes` heartbeat / reverse-channel, `remove_worker`) only cleaned supervisor's internal dicts but never advanced `StatusGuard` to `TERMINATED`. Since `model_status` is sourced from `get_instance_info()` which only filters `TERMINATED`, dead models lingered as `READY=1` in `/metrics` while already disappearing from the UI.

3. **No signal for "replica down due to worker failure"**: A gap between `model_status` (model-level lifecycle) and `workers_total` (infrastructure). When a worker died, there was no Prometheus series to alert on that said "this specific replica needs redeploy".

## Changes in this PR (2 files, +273/-47)

- `_drop_series()` helper for the only correct way to delete a series in aioprometheus (`collector.values.pop()`); applied to all supervisor-side stale-tracking loops (worker, GPU, model_info, model_gpu_binding, model_status, model_gpu_memory)
- New `xinference:model_unexpected_termination` gauge (replica-level: `model_uid`, `model_name`, `replica_index`), following the existing pull architecture — supervisor maintains `_unexpected_down_replicas` dict, serialized in `get_cluster_metrics_data()`, translated to gauge in `update_cluster_metrics()` with stale-pop cleanup
- `_handle_dead_worker()` consolidates the three death-detection paths: records affected replicas, calls `_remove_worker_from_replica_mappings` (preserving healthy replicas of multi-replica models on surviving workers), advances fully-gone models to `TERMINATED`
- `add_worker` reconciliation via `skip_replica_uids` parameter to `_record_unexpected_down_replicas`: diffs supervisor's previously-mapped replicas against the ones the worker reports this round, so the common "worker restarts with same address" case (which bypasses dead-node detection because the fresh heartbeat resets `failure_remaining_count` before the threshold is hit) is handled correctly
- `_clear_unexpected_down_replicas` in both launch entries so redeploy clears the gauge
- `WARNING` log line in `_record_unexpected_down_replicas` with `model_name` and `replica_index` for searchability
- `_check_dead_nodes` now pops `_worker_model_gpu_memory` for dead workers (aligning with `remove_worker`)

## Commits

- First commit (`fix(supervisor): self-heal worker registry after supervisor restart`) belongs to #4998. Please ignore it in this PR's diff.
- Second commit (`fix(metrics): drop stale series and add unexpected termination gauge`) is this PR's actual scope.

## Design decisions

- **Gauge, not Counter**: the signal must disappear when the same uid is redeployed. Counter cannot express this without manual `pop` which triggers counter-reset semantics in Prometheus and corrupts `rate()`/`increase()`.
- **"Disappear" not "ERROR"**: `model_status` transitions to `TERMINATED` (filtered out) rather than `ERROR`, because ERROR already has its own semantic (launch/runtime failure) and is not filtered by `_drop_terminated_info`, which would create permanent zombies.
- **Pull architecture**: supervisor runs in a separate subprocess (`n_process=0`, `deploy/supervisor.py`) and never exposes `/metrics`. Direct `gauge.set()` in supervisor writes to a private instance that is not the one served by the API process. All new state follows the existing pattern: plain dict in supervisor → `get_cluster_metrics_data()` serialization → API-process gauge translation.
- **Set-difference reconciliation in `add_worker`**: avoids false positives when a worker successfully recovers its models on restart (non-empty `replica_states`). Only replicas that the supervisor still maps to this address but the worker did not report are treated as unexpected terminations.

## Test plan

- [x] Functional verification in production cluster (worker restart, worker crash, redeploy scenarios all validated)
- [ ] Unit tests for `_handle_dead_worker`, `_record_unexpected_down_replicas` (with and without `skip_replica_uids`), `_drop_series`, `_clear_unexpected_down_replicas`, metrics translation, and `add_worker` reconciliation — planned for follow-up commit

## Related

- Depends on: #4998 (registry self-heal after supervisor restart, independent fix)

### DIFF
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -338,17 +338,29 @@ class SupervisorActor(xo.StatelessActor):
             if parsed is None:
                 continue
             base_uid, replica_idx = parsed
-            try:
-                info_list = await self._status_guard_ref.get_instance_info(
-                    model_uid=base_uid
-                )
-                m_name = info_list[0].model_name if info_list else "unknown"
-            except Exception:
-                m_name = "unknown"
+            m_name = "unknown"
+            if self._status_guard_ref is not None:
+                try:
+                    info_list = await self._status_guard_ref.get_instance_info(
+                        model_uid=base_uid
+                    )
+                    m_name = info_list[0].model_name if info_list else "unknown"
+                except Exception:
+                    pass
             collected.append((base_uid, replica_idx, m_name))
-        # Mutate the shared dict without awaiting in between.
+        # Mutate the shared dict without awaiting in between. Re-check that
+        # each replica is still mapped to this dead worker before writing:
+        # during the awaits above, a concurrent redeploy could have called
+        # _clear_unexpected_down_replicas and removed stale entries — blindly
+        # writing collected items back would resurrect them.
         for base_uid, replica_idx, m_name in collected:
-            self._unexpected_down_replicas[(base_uid, replica_idx)] = m_name
+            rep_uid = build_replica_model_uid(base_uid, replica_idx)
+            worker_refs = self._replica_model_uid_to_worker.get(rep_uid)
+            if worker_refs is not None:
+                if not isinstance(worker_refs, (list, tuple)):
+                    worker_refs = [worker_refs]
+                if any(wr.address == worker_address for wr in worker_refs):
+                    self._unexpected_down_replicas[(base_uid, replica_idx)] = m_name
         if collected:
             logger.warning(
                 "Replicas down due to worker failure. worker: %s, replicas: %s",
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -101,7 +101,7 @@ class WorkerNotRegisteredError(Exception):
     branch, which re-runs ``add_worker`` and self-heals the registry (and thus
     ``workers_total``). Subclasses ``Exception`` (not ``RuntimeError``) on
     purpose: the worker report loop treats some ``RuntimeError``s as fatal and
-    breaks, which must never happen here. See optimize/20260603/2026060306.md.
+    breaks, which must never happen here.
     """
 
 
@@ -2449,6 +2449,7 @@ class SupervisorActor(xo.StatelessActor):
         if errors and not suppress_exception:
             raise errors[0]
         self._model_uid_to_replica_info.pop(model_uid, None)
+        self._clear_unexpected_down_replicas(model_uid)
 
         # clear for xavier
         rank0_uid = model_uid + "-rank0"
@@ -2524,9 +2525,11 @@ class SupervisorActor(xo.StatelessActor):
                 model_uid,
                 {"replica": remaining_replica_count, "status": LaunchStatus.READY.name},
             )
+            self._unexpected_down_replicas.pop((model_uid, replica_id), None)
             return remaining_replica_count
 
         self._model_uid_to_replica_info.pop(model_uid, None)
+        self._clear_unexpected_down_replicas(model_uid)
 
         rank0_uid = model_uid + "-rank0"
         if rank0_uid in self._replica_model_uid_to_worker:
@@ -2860,7 +2863,7 @@ class SupervisorActor(xo.StatelessActor):
             # report_status reconnect branch re-runs add_worker and self-heals
             # the registry, restoring workers_total and replaying running
             # replicas. Do NOT fabricate a _worker_status entry here, otherwise
-            # the registry would stay stale forever. See 2026060306.md.
+            # the registry would stay stale forever.
             raise WorkerNotRegisteredError(worker_address)
 
         if worker_address not in self._worker_status:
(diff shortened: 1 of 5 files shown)


m199369309:
## Response to review feedback

Addressed the following review comments in the latest push:

### ✅ Comment #2 (race condition) - Fixed
Added a check in `_record_unexpected_down_replicas` to verify the replica is still mapped to the dead worker before writing to `_unexpected_down_replicas`. This prevents the race condition where a concurrent redeploy could clear the entry during the `await` loop, only to have it re-added at the end.

### ✅ Comment #3 (None check) - Fixed  
Added explicit `None` check for `self._status_guard_ref` before calling `get_instance_info`, avoiding `AttributeError` in edge cases.

### ❌ Comment #1 (MetricDict key format) - Disagree
The bot's suggestion about converting labels to `tuple(sorted(labels.items()))` is incorrect. `aioprometheus.MetricDict` is a `MutableMapping` subclass with a `__keytransform__` method (`metricdict.py:39-58`) that automatically normalizes keys:

```python
collector.values.pop(labels, None)
# ↓ MutableMapping.pop() calls __delitem__(labels)
# ↓ __delitem__ calls __keytransform__(labels)  
# ↓ __keytransform__ converts dict → tuple internally
```

This is the standard Python `MutableMapping` protocol. Manually converting to tuples would bypass the abstraction and couple our code to `MetricDict`'s internal representation. I've verified this works correctly in production — stale series are properly deleted from `/metrics`.

xinference/core/supervisor.py
```diff
@@ -284,6 +303,125 @@ def _remove_worker_from_replica_mappings(self, worker_address: str):
...
+                    await self._status_guard_ref.update_instance_info(
+                        base_uid, {"status": LaunchStatus.TERMINATED.name}
+                    )
+                except Exception:
+                    logger.warning(
+                        "Failed to mark %s TERMINATED in status guard", base_uid
+                    )
+        return affected_replica_uids
+
+    def _clear_unexpected_down_replicas(self, model_uid: str) -> None:
```
qinxuye, line 415: [P1] This only clears unexpected-down markers on redeploy, but explicit termination can make the marker obsolete too. If one replica of a multi-replica model is recorded in `_unexpected_down_replicas` and another replica survives, the model stays in `_model_uid_to_replica_info`; a later `terminate_model()` or final `terminate_model_replica()` removes the model without clearing this dict, so `get_cluster_metrics_data()` can keep exporting a stale `model_unexpected_termination` series. Please clear the model/replica marker on the terminate paths as well.

xinference/core/supervisor.py
```diff
@@ -92,6 +93,18 @@ def callback_for_async_launch(model_uid: str):
...
 
+class WorkerNotRegisteredError(Exception):
+    """Raised when a worker reports status but is absent from the supervisor
+    registry (``_worker_address_to_worker``). Most commonly happens after a
+    supervisor restart clears the in-memory registry while workers keep their
+    cached refs. Raising forces the worker into ``report_status``'s reconnect
+    branch, which re-runs ``add_worker`` and self-heals the registry (and thus
+    ``workers_total``). Subclasses ``Exception`` (not ``RuntimeError``) on
+    purpose: the worker report loop treats some ``RuntimeError``s as fatal and
+    breaks, which must never happen here. See optimize/20260603/2026060306.md.
```
qinxuye: [P2] This stack seems to have brought back local design-doc references that were already removed from #4998. Please rebase/cherry-pick the latest #4998 cleanup, or replace these `2026060306.md` references with comments that are self-contained for the repository.

qinxuye (commented):
I left two inline comments on the current head.

m199369309:
Addressed the two review comments from @qinxuye in the latest push:

**[P1] terminate paths** — `_clear_unexpected_down_replicas` is now called in:
- `terminate_model()` — clears all replica entries for the model_uid
- `terminate_model_replica()` when `remaining_replica_count == 0` — clears all entries for the model_uid
- `terminate_model_replica()` when `remaining_replica_count > 0` — pops only the specific `(model_uid, replica_id)` key

**[P2] local doc refs** — Removed all four `2026060306.md` references from `supervisor.py` and `test_worker.py`. The surrounding comments are already self-contained.

qinxuye (approved):
LGTM

PR #5000, opened 2026-06-08
bld: Replace self-hosted runner with GitHub hosted runner and update actions

qinrui777:
@qinxuye 

**Changes:**
1. Replace self-hosted runner with GitHub hosted runner when build 3 types images;
2. Remove  disk clean up steps ;
3. Reduce timeout threshold to 60 mins ;
4. Upgrade actions/checkout to v6;
5. Upgrade docker/login-action to v4 for remediating workflow reminders

**Validation:**
`docker pull qinruicool/xinference:nightly-main`

### DIFF
--- a/.github/workflows/docker-cd.yaml
+++ b/.github/workflows/docker-cd.yaml
@@ -62,7 +62,7 @@ jobs:
 
   build-gpu:
     name: Build GPU Docker image
-    needs: build-aarch64
+    needs: check-release-permission
     timeout-minutes: 60
     runs-on: ubuntu-24.04 #x86 runner
     permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-cd.yaml
```diff
@@ -125,44 +98,25 @@ jobs:
...
-      - name: Clean docker image cache after build
-        shell: bash
-        if: ${{ always() && github.repository == 'xorbitsai/inference' }}
-        run: |
-          docker system df || true
-          docker system prune -af --volumes || true
-          docker builder prune -af || true
-          docker buildx prune -af || true
-
   build-cpu:
```
qinxuye, line 101: [P1] `build-cpu` no longer depends on `check-release-permission`. On tag pushes, that lets the CPU image job log in and push `${tag}-cpu` even if `check-release-permission` later rejects the release tag, while the other image jobs are still gated through `build-aarch64`. Please keep the parallelism but add `needs: check-release-permission` here so every DockerHub push waits for the release permission check.

qinxuye (commented):
I left one inline comment.

qinxuye (approved):
LGTM

PR #5004, opened 2026-06-08
fix(worker): ensure subpool monitor starts after append_sub_pool

m199369309:
## Summary
- The worker main pool is created with `n_process=0`, so xoscar's `start_monitor()` is a no-op at startup because `sub_processes` is empty at that point. After `append_sub_pool()` adds model subpools, `start_monitor()` is never called again, so `monitor_sub_pools` never runs and subprocess deaths (OOM / crash / kill) go undetected and unrecovered.
- Add `_ensure_subpool_monitor()` that calls `start_monitor()` after every `append_sub_pool` call site (3 total: `_create_subpool`, multi-worker `append_sub_pool`, and `launch_rank0_model`). `start_monitor` has its own `_monitor_task is None` idempotency guard, so at most one monitor task runs per worker process.
- Also fix deprecated `asyncio.get_event_loop()` in `deploy/local.py` (switched to `asyncio.new_event_loop()` + `asyncio.set_event_loop()`).

## Test plan
- [ ] Launch a model and verify `monitor_sub_pools` is running (check worker debug logs for subpool monitoring activity)
- [ ] Kill a model subprocess externally and verify `recover_sub_pool` is triggered within the xoscar monitor interval
- [ ] Verify local mode (`xinference-local`) starts without `DeprecationWarning` for `get_event_loop`

### DIFF
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -905,6 +905,18 @@ class WorkerActor(xo.StatelessActor):
             for model_info in model_infos:
                 self._user_specified_gpu_to_model_uids[dev].remove(model_info)
 
+    async def _ensure_subpool_monitor(self):
+        # The worker main pool is created with n_process=0, so xoscar's
+        # start_monitor() (called once in start()) is a no-op because its guard
+        # `and self.sub_processes` is empty at that point; monitor_sub_pools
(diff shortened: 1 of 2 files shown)


qinxuye (approved):
LGTM

PR #5005, opened 2026-06-08
fix(model): use os._exit for OOM to trigger pool recovery

m199369309:
## Summary
- `sys.exit(1)` in `_handle_oom_error` only raises `SystemExit`, which xoscar's `_ErrorProcessor.__exit__` (`backends/pool.py`) swallows with an unconditional `return True`. The model subprocess stays alive, its `returncode` stays `None`, `monitor_sub_pools` keeps treating it as alive, and `recover_sub_pool` never fires. The model becomes a "zombie" that keeps returning OOM errors without ever recovering.
- Replace `sys.exit(1)` with `os._exit(1)` so the subprocess actually exits at the OS level and xoscar can detect the returncode change and rebuild the sub pool.
- Add a reentrancy gate (`_oom_handling`) to handle concurrent OOM on the sync batch path (where the worker thread does not await the cleanup coroutine and can trigger another OOM before `os._exit` lands).
- Reduce `stop()` timeout from 30s to 5s to bound recovery delay (OOM-site cleanup mostly cannot reclaim already-reserved GPU fragments anyway).
- Add regression tests in both sync (`test_client.py`) and async (`test_async_client.py`) test suites that verify the subprocess actually exits after OOM.

## Test plan
- [ ] Run `test_model_oom_triggers_restart` in both `test_client.py` and `test_async_client.py` — verifies the subprocess exits after OOM
- [ ] Run existing `test_model_error` — verifies OOM error message is still propagated to client
- [ ] Run existing `test_auto_recover` — verifies normal auto-recovery (non-OOM path) still works
- [ ] Manual test: trigger OOM on a GPU model and verify the model recovers within ~5.5s (5s stop timeout + 0.5s xoscar monitor interval)

### DIFF
--- a/xinference/core/model.py
+++ b/xinference/core/model.py
@@ -18,7 +18,6 @@ import inspect
 import json
 import os
 import queue
-import sys
 import time
 import types
 import uuid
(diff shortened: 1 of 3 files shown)


xinference/core/model.py
```diff
@@ -519,45 +518,63 @@ async def stop(self):
...
-        # §4.1: Graceful cleanup instead of os._exit(1).
-        # 1) Stop model to release underlying resources (with timeout guard)
         try:
-            await asyncio.wait_for(self.stop(), timeout=30)
-        except asyncio.TimeoutError:
-            logger.warning(
-                "Model stop() timed out after 30s for %s, proceeding with cleanup",
-                self.model_uid(),
+            worker_ref = await self._get_worker_ref()
+            await asyncio.wait_for(
```
qinxuye: Please include `_get_worker_ref()` in the same bounded xoscar wait. The current timeout only starts after `await self._get_worker_ref()` returns, but resolving an actor ref goes through xoscar actor RPC too and can also hang during an OOM/worker failure path. Since this status report is best-effort before `os._exit(1)`, wrap the whole report operation with `xo.wait_for`/`xoscar.wait_for` (or otherwise bound `_get_worker_ref()` as well) so the process can always reach the hard exit promptly.

qinxuye (commented):
One issue to address before LGTM.

qinxuye (approved):
LGTM

PR #5006, opened 2026-06-08
fix(supervisor): evict dead replica from round-robin when auto-recover exhausted

m199369309:
## Summary
- When worker exhausts `AUTO_RECOVER_LIMIT` and stops recreating a model actor, the dead replica remains in supervisor's round-robin scheduler and `model_unexpected_termination` gauge is never lit. Requests routed to the dead replica fail silently, and ops has no observability signal that N-1 out of N replicas are healthy.
- Add `supervisor.mark_replica_dead()` that evicts the dead replica from the round-robin scheduler, lights up `model_unexpected_termination`, and advances `base_uid` to `TERMINATED` when the single/last replica dies. The method is idempotent and deliberately does not call back `worker.terminate_model` (the worker already terminated the model locally before giving up recreating).
- Worker calls back `supervisor.mark_replica_dead()` after "Stop recreating model actor." with a 5s `xo.wait_for` timeout guard so a stalled supervisor does not hold up the worker's local shutdown. Failure/timeout is non-fatal — the next death detection or redeploy will reconcile.
- Add regression tests in both sync and async client test suites (`test_model_oom_recover_exhausted_evicts_replica`).

**Depends on #5005** (fix/model: use os._exit for OOM to trigger pool recovery) which provides the `os._exit(1)` fix that makes OOM-triggered recovery cycles actually work. This PR includes #5005's commits as its base; the PR3-specific changes are in the supervisor.py `mark_replica_dead` method, the worker callback, and the eviction test.

## Test plan
- [ ] Run `test_model_oom_recover_exhausted_evicts_replica` in both `test_client.py` and `test_async_client.py` — verifies dead replica is evicted after recover limit exhausted
- [ ] Run existing `test_model_oom_triggers_restart` (from #5005) — verifies OOM still triggers subprocess exit
- [ ] Run existing `test_auto_recover` — verifies normal auto-recovery still works
- [ ] Manual test: deploy a model with `XINFERENCE_MODEL_ACTOR_AUTO_RECOVER_LIMIT=1`, trigger repeated OOM, verify `model_unexpected_termination` gauge lights up and `list_models()` drains to empty

### DIFF
--- a/xinference/client/tests/test_async_client.py
+++ b/xinference/client/tests/test_async_client.py
@@ -14,7 +14,6 @@
 
 import asyncio
 import os
-import time
 
 import psutil
 import pytest
@@ -587,7 +586,7 @@ async def test_auto_recover(set_auto_recover_limit, setup_cluster):
             assert "text" in completion["choices"][0]
             break
         except Exception:
-            time.sleep(1)
+            await asyncio.sleep(1)
     else:
         assert False
     await model.close()
@@ -684,7 +683,7 @@ async def test_model_oom_recover_exhausted_evicts_replica(
         if len(await client.list_models()) == 0:
             evicted = True
             break
-        time.sleep(1)
+        await asyncio.sleep(1)
     assert (
         evicted
     ), "dead replica was not evicted from supervisor after recover exhausted"
--- a/xinference/client/tests/test_async_client.py
+++ b/xinference/client/tests/test_async_client.py
@@ -656,6 +656,42 @@ async def test_model_oom_triggers_restart(set_test_oom_error, setup_cluster):
 
 
 @pytest.mark.asyncio
+async def test_model_oom_recover_exhausted_evicts_replica(
+    set_auto_recover_limit, set_test_oom_error, setup_cluster
+):
+    # Once worker exhausts AUTO_RECOVER_LIMIT (=1), it stops recreating and
+    # calls back supervisor.mark_replica_dead, which evicts the dead replica.
+    # For a single-replica model the model is taken fully offline, so
+    # list_models() drains to empty.
+    endpoint, _ = setup_cluster
+    client = AsyncRESTfulClient(endpoint)
+
+    model_uid = await client.launch_model(
+        model_name="qwen1.5-chat",
+        model_engine="llama.cpp",
+        model_size_in_billions="0_5",
+        quantization="q4_0",
+    )
+    assert len(await client.list_models()) == 1
+
+    evicted = False
+    for _ in range(60):
+        try:
+            model = await client.get_model(model_uid=model_uid)
+            await model.generate("Once upon a time, there was a very old computer")
+        except Exception:
+            pass
+        if len(await client.list_models()) == 0:
+            evicted = True
+            break
+        time.sleep(1)
+    assert (
+        evicted
+    ), "dead replica was not evicted from supervisor after recover exhausted"
+    await client.close()
+
+
+@pytest.mark.asyncio
 async def test_async_restful_chat_enable_thinking_injected():
     handle = AsyncRESTfulChatModelHandle.__new__(AsyncRESTfulChatModelHandle)
     handle._model_uid = "test-model"
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -2531,13 +2531,48 @@ class SupervisorActor(xo.StatelessActor):
         self._model_uid_to_replica_info.pop(model_uid, None)
         self._clear_unexpected_down_replicas(model_uid)
 
+        await self._cleanup_distributed_actors(
+            model_uid, terminate_rank0_on_worker=True
+        )
+
+        return 0
+
+    async def _cleanup_distributed_actors(
+        self, model_uid: str, terminate_rank0_on_worker: bool = True
+    ) -> None:
+        """Tear down supervisor-side actors/mappings created for distributed
+        (Xavier) models when the last replica of ``model_uid`` goes away: the
+        rank0 model, the collective manager, and the block tracker.
+
+        Shared by terminate_model_replica (graceful teardown) and
+        mark_replica_dead (auto-recover exhaustion). Without this, exhausting
+        AUTO_RECOVER_LIMIT on a distributed model would leak the rank0 actor and
+        the collective/block-tracker mappings, and a later launch reusing the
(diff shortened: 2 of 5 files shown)


xinference/core/worker.py
```diff
@@ -315,6 +315,30 @@ async def recover_sub_pool(self, address):
...
+                            # light up the failure gauge. add_worker=False:
+                            # only fetch the ref, do not trigger
+                            # re-registration. Wrap in xo.wait_for(5s): this
+                            # runs inside the recover_sub_pool tail path, so a
+                            # stalled supervisor must not hold up the worker's
+                            # local shutdown. A failure/timeout is non-fatal --
+                            # the next death detection / redeploy will
+                            # reconcile.
+                            try:
+                                supervisor_ref = await self.get_supervisor_ref(
```
qinxuye, line 334: Please include `get_supervisor_ref(add_worker=False)` in the same bounded xoscar wait. The comment says this path must not hold up local shutdown, but the timeout currently starts only after the supervisor ref has already been resolved. When `_supervisor_ref` is missing or was cleared, `get_supervisor_ref()` calls `xo.actor_ref()` and other xoscar actor refs, which can block as well. Wrap the whole `get_supervisor_ref + mark_replica_dead` notification in `xo.wait_for(..., timeout=5)` so the recover tail path is actually bounded.

xinference/core/supervisor.py
```diff
@@ -2563,6 +2563,79 @@ async def terminate_model_replica(
...
+        if remaining > 0:
+            # Degraded: healthy replicas remain -> keep READY.
+            await self._status_guard_ref.update_instance_info(
+                base_uid,
+                {"replica": remaining, "status": LaunchStatus.READY.name},
+            )
+            return
+
+        # Single replica / last replica died -> take the model fully offline.
+        self._model_uid_to_replica_info.pop(base_uid, None)
```
qinxuye, line 2677: When the last replica is marked dead, this removes `_model_uid_to_replica_info` and marks the model TERMINATED, but it does not perform the last-model cleanup that `terminate_model()` / `terminate_model_replica()` already do for `rank0_uid`, `_collective_manager_mapping`, and `_block_tracker_mapping`. For Xavier/distributed models, auto-recover exhaustion can leave supervisor-side actors/mappings behind, and a later launch with the same uid may hit stale actors or leaked state. Please reuse/extract the existing final cleanup path here as well.

qinxuye (commented):
Two issues to address before LGTM.

xinference/core/worker.py
m199369309: Fixed in 6c93b8b2. Wrapped the whole notification — `get_supervisor_ref(add_worker=False)` + `mark_replica_dead` — in a single `xo.wait_for(..., timeout=5)` via an inline `_notify_replica_dead()` coroutine, so the blocking `xo.actor_ref` calls inside `get_supervisor_ref` (when the cached ref is missing) are now covered by the timeout. Comment updated to match.

xinference/core/supervisor.py
m199369309: Fixed in 6c93b8b2. Extracted the final distributed-cleanup path (rank0 termination + `_collective_manager_mapping`/`_block_tracker_mapping` destroy) from `terminate_model_replica` into a shared `_cleanup_distributed_actors(model_uid, terminate_rank0_on_worker=...)` helper, and call it from `mark_replica_dead`'s last-replica branch with `terminate_rank0_on_worker=False` (the worker already terminated locally, so only the supervisor-side mapping is dropped). The `_unexpected_down_replicas` marker is deliberately left untouched by the helper so the failure gauge stays lit. `terminate_model` was left as-is to keep this change scoped.

xinference/client/tests/test_async_client.py
```diff
@@ -616,6 +616,81 @@ async def test_model_error(set_test_oom_error, setup_cluster):
...
+    for _ in range(60):
+        try:
+            model = await client.get_model(model_uid=model_uid)
+            await model.generate("Once upon a time, there was a very old computer")
+        except Exception:
+            pass
+        if len(await client.list_models()) == 0:
+            evicted = True
+            break
+        time.sleep(1)
```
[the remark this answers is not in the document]
m199369309: Fixed in 6c93b8b2. Replaced the blocking `time.sleep(1)` with `await asyncio.sleep(1)` in the async retry loops (and removed the now-unused `time` import). The sync `test_client.py` keeps `time.sleep` since those tests are not coroutines.

xinference/core/supervisor.py
qinxuye: This still looks incomplete for rank0. The exhausted worker only calls `terminate_model()` for the replica uid whose subpool died; Xavier rank0 is a separate `model_uid-rank0` actor/subpool created by `launch_rank0_model()`, so a regular replica exhausting recovery does not necessarily terminate rank0 locally. Popping `_replica_model_uid_to_worker[rank0_uid]` with `terminate_rank0_on_worker=False` drops the supervisor mapping, but can leave the rank0 actor/subpool running. Please either keep a bounded best-effort termination for rank0 here, or otherwise prove/detect that rank0 is already gone before only dropping the mapping.

m199369309: Fixed in b6e1a1b7 — took the bounded best-effort termination option.

`mark_replica_dead`'s last-replica branch now calls `_cleanup_distributed_actors(base_uid, terminate_rank0_on_worker=True)` (was `False`), so the supervisor RPCs the worker to terminate the separate `model_uid-rank0` actor/subpool instead of only dropping `_replica_model_uid_to_worker[rank0_uid]`. The rank0 `terminate_model` RPC inside the helper is wrapped in `xo.wait_for(..., timeout=5)` so a stalled worker can't hold up the death-recovery tail path; failure/timeout stays non-fatal and best-effort (logged at debug). `_unexpected_down_replicas` is deliberately left untouched by the helper so the failure gauge stays lit.

Added unit tests in `test_worker.py`: `test_cleanup_distributed_actors_terminates_rank0` (True → rank0 terminated + mapping dropped), `test_cleanup_distributed_actors_skips_rank0_rpc_when_false` (False → mapping-only, no RPC), and `test_mark_replica_dead_last_replica_terminates_rank0` (end-to-end Xavier last-replica path). All 10 supervisor tests pass.

qinxuye (approved):
LGTM

PR #5010, opened 2026-06-09
feat: add MiniCPM5-1B model support

xiaoyesoso:
Register MiniCPM5-1B model in llm_family.json with pytorch, ggufv2 and mlx formats. Add reasoning/hybrid/tool-call abilities, chat template, and virtualenv dependencies for Transformers/MLX/llama.cpp/vLLM/SGLang engines. Fix xoscar compatibility (wait_for, start_python, daemon process), Gradio UI reasoning content display, and MLX tokenizer loading. Add model documentation.

### DIFF
--- a/doc/source/getting_started/installation.rst
+++ b/doc/source/getting_started/installation.rst
@@ -65,7 +65,7 @@ Currently, supported models include:
 
 .. vllm_start
 
-- ``code-llama``, ``code-llama-instruct``, ``code-llama-python``, ``deepseek``, ``deepseek-chat``, ``deepseek-coder``, ``deepseek-coder-instruct``, ``deepseek-r1-distill-llama``, ``gorilla-openfunctions-v2``, ``HuatuoGPT-o1-LLaMA-3.1``, ``llama-2``, ``llama-2-chat``, ``llama-3``, ``llama-3-instruct``, ``llama-3.1``, ``llama-3.1-instruct``, ``llama-3.3-instruct``, ``tiny-llama``, ``wizardcoder-python-v1.0``, ``wizardmath-v1.0``, ``Yi``, ``Yi-1.5``, ``Yi-1.5-chat``, ``Yi-1.5-chat-16k``, ``Yi-200k``, ``Yi-chat``
+- ``code-llama``, ``code-llama-instruct``, ``code-llama-python``, ``deepseek``, ``deepseek-chat``, ``deepseek-coder``, ``deepseek-coder-instruct``, ``deepseek-r1-distill-llama``, ``gorilla-openfunctions-v2``, ``HuatuoGPT-o1-LLaMA-3.1``, ``llama-2``, ``llama-2-chat``, ``llama-3``, ``llama-3-instruct``, ``llama-3.1``, ``llama-3.1-instruct``, ``llama-3.3-instruct``, ``minicpm5-1b``, ``tiny-llama``, ``wizardcoder-python-v1.0``, ``wizardmath-v1.0``, ``Yi``, ``Yi-1.5``, ``Yi-1.5-chat``, ``Yi-1.5-chat-16k``, ``Yi-200k``, ``Yi-chat``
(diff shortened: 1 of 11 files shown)


qinxuye:
Thanks, there is conflict, could you resolve it?

qinxuye:
Actually, you only need to modify llm_family.json, doc file changes will be made by bot automatically.

xiaoyesoso:
> Actually, you only need to modify llm_family.json, doc file changes will be made by bot automatically.

Hi @qinxuye, thanks for the quick review!

I have rebased onto the latest main to resolve the conflict in xinference/core/worker.py , and reverted my manual changes to doc/source/models/builtin/llm/* so that the doc bot can regenerate them automatically.

qinxuye:
pre-commit failed, could you fix it accordingly? thanks.

xiaoyesoso:
> pre-commit failed, could you fix it accordingly? thanks.

Thanks for the patience @qinxuye! I missed running pre-commit run --all-files last time and only ran it on the changed files, so mypy errors slipped through. Just fixed them in chat_interface.py (renamed the reused choices variable and properly narrowed the message type for mypy) and verified pre-commit run --all-files passes locally — all hooks (black / flake8 / isort / mypy / codespell) are green now. PTAL 🙏

xinference/__init__.py
```diff
@@ -37,6 +37,48 @@ def _install():
...
+    # xinference internals (e.g. core/worker.py, core/supervisor.py) call
+    # ``xo.wait_for`` for actor-aware timeouts. Fall back to
+    # ``asyncio.wait_for`` when running against an older xoscar release
+    # (e.g. the latest PyPI build 0.9.6) so the local cluster can boot.
+    import xoscar as _xo
+
+    if not hasattr(_xo, "wait_for"):
+        import asyncio as _asyncio
+
+        _xo.wait_for = _asyncio.wait_for
```
qinxuye: Using `asyncio.wait_for` as the fallback changes the semantics of `xo.wait_for` for actor calls. We already rely on `xo.wait_for` specifically because `asyncio.wait_for` can get stuck when cancellation is routed through a Xoscar actor call, so this shim can reintroduce the timeout hang on older xoscar versions. Please avoid aliasing it to `asyncio.wait_for`; either require a xoscar version that provides `wait_for`, or backport the non-blocking wrapper implementation instead.

xinference/__init__.py
```diff
@@ -37,6 +37,88 @@ def _install():
...
+        _xo.wait_for = _xo_wait_for
+
+    # Compatibility shim: older xoscar (<= 0.9.6) spawns each actor pool as
+    # a daemon process, which makes the worker process unable to spawn the
+    # model subprocess (Python raises ``AssertionError: daemonic processes
+    # are not allowed to have children``). Patch multiprocessing so the
+    # daemon flag is forced off when xoscar tries to start a child pool.
+    import multiprocessing.process as _mp_process
+
+    if not getattr(_mp_process, "_xinf_daemon_patch_applied", False):
```
qinxuye: I do not think we need this monkey patch for older xoscar versions. Patching `multiprocessing.process.BaseProcess.start` at import time changes process-spawning semantics globally for the entire Python process, not just Xinference or xoscar worker subpools, and it also keeps taking effect even when a newer xoscar is installed. Since this code already depends on newer xoscar behavior such as `xo.wait_for` and `append_sub_pool(start_python=...)`, please bump the minimum xoscar version to the first released version that supports the required behavior and remove these compatibility shims instead.

xiaoyesoso: @qinxuye thanks for the review! You're absolutely right — xoscar 0.9.6 already ships with both xo.wait_for and append_sub_pool(start_python=...) , so all the compatibility shims ( __init__.py wait_for backport, daemon patch, and the start_python try/except in worker.py ) were unnecessary. I've removed all of them and rebased onto the latest upstream/main .

qinxuye (approved):
LGTM

PR #5011, opened 2026-06-09
feat: new UI part 3

maoyuehui:
Model Deploy
1. Seven Model Type Forms, Progress Tracking, and Cancel
2. Custom Model Delete and Edit
3. Cache Config
4. Copy Command Line and Parse Command Line

Register Model
1. Refactored Model Registration from a Single Route to Dynamic Routes
2. Added Model Registration Edit Page
3. Added Copy for JSON view

Form Compontens
1. Support for Hidden Form Fields

Running Model Page


frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -0,0 +1,1364 @@
...
+      pollingRef.current = null;
+    }
+  }, []);
+
+  const fetchProgress = useCallback(async () => {
+    const modelUid = form.getFieldValue('model_uid');
+    try {
+      const [progressRes, replicaRes] = await Promise.all([
+        request.get<number | string | { progress?: number | string }>(
+          `/v1/models/${model?.model_name}/progress`
```
qinxuye: This should use the actual launch model UID, not always `model.model_name`. The form allows a custom `model_uid` and `/v1/models/{model_uid}/progress` is keyed by that UID, so progress polling will read the wrong launch when the user customizes it. The same applies to the cancel request below: compute one `launchModelUid = form.getFieldValue('model_uid') || model?.model_name` and use it consistently for progress, replicas, and cancel.

maoyuehui: Keep consistent with previous logic

maoyuehui: Updated: The parameters for the progress, cancel, and replicas APIs now prioritize model_uid; if model_uid is unavailable, they fall back to model_name.

qinxuye (approved):
LGTM

PR #5015, opened 2026-06-11
fix(logs): fix empty node filter dropdown and redesign toolbar layout

m199369309:
## Summary
- Add `GET /v1/cluster/logs/nodes` endpoint using ES `terms` aggregation to return the actual node list from log data (the previous `/v1/cluster/info` endpoint had no `node` field, causing the dropdown to always be hidden)
- Update frontend to fetch nodes from the new endpoint instead of `/v1/cluster/info`
- Redesign toolbar from single-row `flexWrap` into a 3-row layout: node select + search + time controls on row 1, log level chips with label on row 2, node type chips with label on row 3
- Add 2 i18n keys (`logLevel`, `nodeType`) for zh/en/ja/ko

## Test plan
- [ ] Configure `XINFERENCE_ES_URL`, start UI, navigate to Log Center
- [ ] Verify node dropdown appears and lists actual node values from logs
- [ ] Select a node and verify the log table filters correctly
- [ ] Verify toolbar layout matches 3-row structure (search row, level chips row, type chips row)
- [ ] Verify time range picker / refresh / auto-refresh still work as before
- [ ] Test without `XINFERENCE_ES_URL` — dropdown should not appear (graceful degradation)
- [ ] Switch UI language to en/ja/ko and verify chip labels are correct

### DIFF
--- a/xinference/api/routers/admin.py
+++ b/xinference/api/routers/admin.py
@@ -422,6 +422,57 @@ async def search_logs(
     return JSONResponse(content={"hits": hits, "total": total})
 
 
+async def list_log_nodes() -> JSONResponse:
+    es_url = os.environ.get("XINFERENCE_ES_URL", "")
+    if not es_url:
+        raise HTTPException(status_code=503, detail="Elasticsearch is not configured")
+
+    es_index = os.environ.get("XINFERENCE_ES_INDEX", "xinference-logs-*")
+    es_auth = os.environ.get("XINFERENCE_ES_AUTH", "")
+
+    headers = {"Content-Type": "application/json"}
+    auth = None
+    if es_auth:
+        if es_auth.startswith("ApiKey "):
+            headers["Authorization"] = es_auth
+        else:
+            parts = es_auth.split(":", 1)
+            if len(parts) == 2:
+                auth = aiohttp.BasicAuth(parts[0], parts[1])
+
+    url = f"{es_url.rstrip('/')}/{es_index}/_search"
+
+    async def _aggregate(field: str) -> Optional[list[dict[str, Any]]]:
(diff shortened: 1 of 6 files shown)


xinference/api/routers/admin.py
```diff
@@ -422,6 +422,57 @@ async def search_logs(
...
+                if resp.status != 200:
+                    return None
+                data = await resp.json()
+        return data.get("aggregations", {}).get("nodes", {}).get("buckets", [])
+
+    try:
+        # "node" is usually a keyword field; fall back to "node.keyword" if the
+        # mapping is text (terms aggregation requires a keyword/fielddata field).
+        buckets = await _aggregate("node")
+        if buckets is None:
```
qinxuye, line 465: This fallback makes the node list come from `node.keyword` when `node` cannot be aggregated, but the actual `/v1/cluster/logs` query still filters the selected node with `terms` on `node`. In mappings where this fallback is needed, the dropdown can show nodes but selecting one can still return no logs. Please make the node filter use the same resolved field, or otherwise keep aggregation and filtering on a consistent keyword field.

xinference/ui/web/ui/src/locales/en.json
```diff
@@ -397,7 +399,14 @@
     "selectAll": "Select All",
     "editPermissions": "Edit Permissions",
     "editPermissionsTitle": "Edit Permissions",
-    "save": "Save"
+    "save": "Save",
+    "changePassword": "Change Password",
```
qinxuye: These `changePassword*` user-management strings are unrelated to this logs PR, and the matching user-management UI/backend changes are not included here. Please drop this unrelated locale diff from this branch to avoid pulling in part of another PR.

qinxuye (commented):
Two issues found.

qinxuye (approved):
LGTM

PR #5016, opened 2026-06-11
fix(monitor): correct alert rule labels/expressions and add 4 new rules

m199369309:
## Summary
- Fix `$labels.model` → `$labels.model_name` in HighModelErrorRate and TTFTHigh annotations across all 4 locale rule files (`metrics.py` uses `model_name` as the label key)
- Fix TTFT description unit from `ms` to `s` (metric is `time_to_first_token_seconds`)
- Replace RequestQueueBacklog expression: `model_pending_requests` does not exist in `metrics.py`; changed to `model_serve_count / model_request_limit > 0.9` (both gauges exist)
- Fix ModelLoadSlow description: remove "average" (metric is `model_last_load_duration_seconds`, not an average)
- Add 4 new alert rules, all referencing metrics that exist in `xinference/core/metrics.py`:
  - **ReplicaUnexpectedTerminated** (critical): `model_unexpected_termination == 1`
  - **WorkerMemoryHigh** (warning): `worker_memory_used_bytes / worker_memory_total_bytes > 0.9`
  - **RequestLatencyHigh** (warning): `histogram_quantile(0.95, rate(model_request_duration_seconds_bucket[5m])) > 60`
  - **BannedIPsSpike** (warning): `banned_ips_total > 10`
- Update README.md and README_zh-CN.md with node_exporter dependency note and docs for all new rules

## Test plan
- [ ] Verify `promtool check rules monitor/alert/rules.yml` passes
- [ ] Deploy rules to a Prometheus instance with Xinference metrics and confirm:
  - [ ] HighModelErrorRate/TTFTHigh fire with correct `model_name` label in alert description
  - [ ] RequestQueueBacklog fires when concurrency ratio exceeds 90%
  - [ ] New rules (ReplicaUnexpectedTerminated, WorkerMemoryHigh, RequestLatencyHigh, BannedIPsSpike) evaluate without "no metric" errors
- [ ] Check ja/ko/zh-CN locale rule files have identical structure and expressions

### DIFF
--- a/monitor/alert/rules.yml
+++ b/monitor/alert/rules.yml
@@ -47,7 +47,7 @@ groups:
           description: "Node {{ $labels.instance }} mountpoint {{ $labels.mountpoint }} remaining space {{ $value | humanizePercentage }}."
 
       - alert: RequestQueueBacklog
-        expr: xinference:model_serve_count / xinference:model_request_limit > 0.9
+        expr: xinference:model_serve_count / (xinference:model_request_limit > 0) > 0.9
         for: 3m
         labels:
           severity: warning
--- a/monitor/alert/rules.yml
+++ b/monitor/alert/rules.yml
@@ -11,7 +11,7 @@ groups:
           description: "Worker {{ $labels.worker_address }} GPU {{ $labels.gpu_index }} memory usage {{ $value | humanizePercentage }}, lasting 5 minutes."
 
       - alert: ModelRequestErrorRateHigh
-        expr: rate(xinference:model_request_errors_total[5m]) / rate(xinference:model_request_total[5m]) > 0.05
+        expr: sum without (stream) (rate(xinference:model_request_errors_total[5m])) / sum without (stream) (rate(xinference:model_request_total[5m])) > 0.05
         for: 3m
         labels:
           severity: warning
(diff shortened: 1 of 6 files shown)


monitor/alert/rules.yml
```diff
@@ -17,7 +17,7 @@ groups:
           severity: warning
         annotations:
           summary: "Model request error rate exceeds 5%"
-          description: "Model {{ $labels.model }} error rate {{ $value | humanizePercentage }}, lasting 3 minutes."
+          description: "Model {{ $labels.model_name }} error rate {{ $value | humanizePercentage }}, lasting 3 minutes."
```
qinxuye, line 20: This alert's expression above still will not match the actual metric labels. `model_request_total` is recorded with an extra `stream` label, while `model_request_errors_total` is recorded without it, so Prometheus vector matching will not pair the numerator and denominator and this alert returns no series. Please aggregate away `stream` here, and keep the same expression in the other locale rule files and READMEs, for example:

```promql
sum without (stream) (rate(xinference:model_request_errors_total[5m])) / sum without (stream) (rate(xinference:model_request_total[5m])) > 0.05
```

qinxuye (approved):
LGTM

PR #5018, opened 2026-06-11
feat: add jina-embeddings-v5 series support

xiaoyesoso:
Register the four jina-embeddings-v5 models in model_spec.json:

- jina-embeddings-v5-text-nano   (768 dim,  8192 tokens)
- jina-embeddings-v5-text-small  (1024 dim, 32768 tokens)
- jina-embeddings-v5-omni-nano   (768 dim,  8192 tokens, multimodal)
- jina-embeddings-v5-omni-small  (1024 dim, 32768 tokens, multimodal)

Extend the sentence-transformers backend so v5 works end-to-end:

* _resolve_jina_task handles v5's task parameter (retrieval / text-matching / classification / clustering), maps query/document aliases to v5 prompt_name for the retrieval task, and defaults to retrieval when no task is supplied (OpenAI-compatible clients omit it). Unit tests cover all v5 task and alias variants.

* v5-omni models are loaded with model_kwargs={"default_task": "retrieval"} per the model card, and route through SentenceTransformer.encode (instead of the local tokenize+forward helper) so image / video / audio inputs are dispatched to the matching tower.

* The HTTP embedding API accepts dict inputs of the form {"text" | "image" | "video" | "audio": url_or_b64_or_path} for omni models, mirroring jina-clip-v2 / jina-embeddings-v4 multimodal conventions, and translates OpenAI's `dimensions` request into v5-omni's `truncate_dim` for Matryoshka truncation.

### DIFF
--- a/xinference/model/embedding/sentence_transformers/core.py
+++ b/xinference/model/embedding/sentence_transformers/core.py
@@ -52,6 +52,23 @@ JINA_V4_VALID_TASKS = {
     "document",
 }
 
+# jina-embeddings-v5 (text-* and omni-*): custom encode() consumes ``task`` directly.
+# Valid tasks come from the v5 model cards on HuggingFace.
+JINA_V5_VALID_TASKS = {
+    "retrieval",
+    "text-matching",
+    "classification",
+    "clustering",
+}
+# v5 retrieval additionally uses ``prompt_name`` to distinguish query vs document.
+JINA_V5_PROMPT_NAME_ALIASES = {
+    "query": "query",
+    "document": "document",
+    "passage": "document",
+    "retrieval.query": "query",
+    "retrieval.passage": "document",
+}
+
 
 def _resolve_jina_task(
     model_name: str, task: Optional[str]
@@ -62,11 +79,18 @@ def _resolve_jina_task(
         (prompt_name, task_passthrough)
         - v3: (prompt_name, None) — uses prompt_name to look up prompt template
         - v4: (None, task) — passes task directly to model.forward()
+        - v5: (prompt_name_or_None, task) — passes task directly to model.encode();
+              v5 models reject a missing task, so default to ``retrieval`` when the
+              caller does not pass one (OpenAI-compatible clients usually omit it)
         - other models: (None, None)
     """
-    if task is None:
-        return None, None
     model_lower = model_name.lower()
+    if task is None:
+        # v5 mandates a task on every call. Fall back to ``retrieval`` so OpenAI-
+        # style clients that have no notion of a task parameter still work.
+        if "jina-embeddings-v5" in model_lower:
+            return None, "retrieval"
+        return None, None
     if "jina-embeddings-v3" in model_lower:
         prompt_name = JINA_V3_TASK_TO_PROMPT_NAME.get(task)
         if prompt_name is None:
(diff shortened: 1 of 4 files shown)


xinference/model/embedding/sentence_transformers/core.py
```diff
@@ -505,6 +554,89 @@ def base64_to_image(base64_str: str) -> Image.Image:
...
+                )
+
+            # ``encode`` returns ndarray for batches and a 1-D ndarray for a
+            # single item; normalise to a list of per-item vectors.
+            if input_was_string or (
+                hasattr(omni_out, "ndim") and getattr(omni_out, "ndim", 0) == 1
+            ):
+                all_embeddings = [omni_out]
+            else:
+                all_embeddings = list(omni_out)
```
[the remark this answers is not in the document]
xiaoyesoso: Thanks for the catch @gemini-code-assist! Both issues are real and have been fixed in the latest push ( 5f2c6ebf ):

1. Dict iteration bug : Now treats str and dict as single-item inputs via isinstance(sentences, (str, dict)) before iterating, so a bare {"image": "..."} payload is no longer iterated over its keys.
2. Nesting bug : For single-item inputs the (1, dim) ndarray is now unwrapped with omni_out[0] before being returned, so the response embedding is [float, ...] rather than [[float, ...]] .
Verified with the existing 38 unit tests + pre-commit --all-files (all green). PTAL 🙏

qinxuye (commented):
One remaining issue found.

xinference/model/embedding/sentence_transformers/core.py
```diff
@@ -505,6 +554,101 @@ def base64_to_image(base64_str: str) -> Image.Image:
...
+                    **omni_kwargs,
+                )
+
+            # ``encode`` on a single-item list returns either a 2-D ndarray of
+            # shape (1, dim) or a 1-D ndarray of shape (dim,). The downstream
+            # code below iterates ``all_embeddings`` and calls ``.tolist()`` on
+            # each item, expecting a flat 1-D vector per input. Normalise both
+            # shapes accordingly so the response stays ``[float, ...]`` rather
+            # than the accidentally-nested ``[[float, ...]]``.
+            if input_was_single:
```
qinxuye, line 640: This still breaks single-string inputs for the v5-omni branch. The branch normalizes a single item into `all_embeddings = [omni_out[0]]`, but after leaving this branch the shared tail still sees the original `sentences` is a `str` and wraps it again into `[[ndarray]]`. The loop below then receives a Python list and `data.tolist()` raises `AttributeError`, so an OpenAI-compatible `input: "text"` request for `jina-embeddings-v5-omni-*` fails. Please skip that extra wrapping for the omni path, or normalize the omni branch so the shared tail does not re-wrap it.

xiaoyesoso: Great catch @qinxuye! You're right — the shared tail at line 659 re-wraps all_embeddings whenever sentences is a str , which double-wraps our already-normalised omni output and breaks data.tolist() . Fixed in 49506640 by promoting sentences to a list at the end of the omni branch so the tail no longer re-wraps it.

Also added a new test file test_jina_v5_omni_shapes.py with 4 regression tests that mock SentenceTransformer.encode and assert the shape of the embeddings returned by _create_embedding for:

- single str input → flat 1-D vector
- single dict input (e.g. {"image": ...} ) → not iterated as keys
- list of strings → preserves cardinality
- mixed list with text/image dicts → each item resolved correctly
Confirmed that without the fix , the single-string test fails with the exact AttributeError: 'list' object has no attribute 'tolist' you described — so this is now properly guarded against regression. PTAL 🙏

qinxuye:
could you fix lint?

qinxuye (approved):
LGTM

PR #5019, opened 2026-06-12
feat(vllm): support glm5 tool parser

llyycchhee:
Support glm5 tool parser.

### DIFF
--- a/xinference/model/llm/tool_parsers/glm5_tool_parser.py
+++ b/xinference/model/llm/tool_parsers/glm5_tool_parser.py
@@ -1,0 +1,167 @@
+import json
+import logging
+import re
+from typing import Any, Dict, List, Optional, Tuple
+
+from . import register_tool_parser
+from .abstract_tool_parser import ToolParser
+
+logger = logging.getLogger(__name__)
+
+
+@register_tool_parser("glm5")
+class Glm5ToolParser(ToolParser):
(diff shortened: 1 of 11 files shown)


xinference/model/llm/tool_parsers/glm5_tool_parser.py
```diff
@@ -0,0 +1,147 @@
...
+        """
+        Extract tool calls from streaming output.
+
+        Args:
+            previous_text (List[str]): Previous text chunks from the stream.
+            current_text (str): Current accumulated text.
+            delta_text (str): New text delta in this chunk.
+        """
+        if isinstance(current_text, str):
+            if self.tool_call_start_token not in current_text:
```
qinxuye, line 139: Streaming chunks can split the `<tool_call>` start marker, for example the first delta can be `<tool` and the next delta `_call>...`. In that first chunk this branch treats `<tool` as normal content because the full start token is not in `current_text`, so the OpenAI stream can leak partial tool markup before later emitting the tool call. Please suppress suffixes that are prefixes of `self.tool_call_start_token` before returning `delta_text` here, similar to the existing DeepSeek parser, and add a regression test for a split start token.

qinxuye:
Tests failed.

qinxuye (approved):
LGTM

PR #5022, opened 2026-06-12
feat: add Gradio test page for embedding models

xiaoyesoso:
Add a Gradio-based web UI for testing embedding models, mirroring the existing LLM and image/video/audio Gradio interfaces.

* New `EmbeddingInterface` class in `xinference/ui/gradio/embedding_interface.py` with three tabs:
  - Text Encoding: encode text with optional task/dimensions parameters
  - Multimodal Encoding: encode image/video/audio (for omni models)
  - Similarity Comparison: compute cosine similarity between two texts

* New `BuildGradioEmbeddingInterfaceRequest` schema and `build_gradio_embedding_interface` API endpoint at `/v1/ui/embeddings/{model_uid}`.

* Update frontend `running_models/index.js` to show "Launch Web UI" button for embedding models, calling the new endpoint.

* Export `EmbeddingInterface` from `xinference.ui.gradio` and `xinference.ui` modules.

### DIFF
--- a/xinference/ui/web/ui/src/scenes/running_models/index.js
+++ b/xinference/ui/web/ui/src/scenes/running_models/index.js
@@ -581,10 +581,34 @@ const RunningModels = () => {
       disableColumnMenu: true,
       renderCell: ({ row }) => {
         const url = row.url
+        const openUrl = `${endPoint}/` + url
 
         if (url === 'IS_LOADING') {
           return <div></div>
         }
+
+        // ``embeddingModelColumns`` is reused for rerank / flexible tabs
+        // (see ``rerankModelColumns`` / ``flexibleModelColumns`` aliases
+        // below), so the Gradio launcher must only render for embedding
+        // rows. Posting a rerank / flexible ``model_uid`` to
+        // ``/v1/ui/embeddings/{model_uid}`` would otherwise 400 on the
+        // backend ``model_type`` check.
+        if (row.model_type !== 'embedding') {
+          return (
+            <Box
+              style={{
+                width: '100%',
+                display: 'flex',
+                justifyContent: 'left',
+                alignItems: 'left',
+              }}
+            >
+              {renderTerminateButton(row)}
+            </Box>
+          )
+        }
+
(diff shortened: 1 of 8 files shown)


qinxuye:
Thanks, and I will review this later soon.

One tip is that, for planned Xinference 3.0, we decided to add a new frontend system, and for each test/run page, we will have our own page instead of  using gradio.

Code: https://github.com/xorbitsai/inference/tree/main/frontend

@maoyuehui is on this, I don't know if he started the page for embedding (I guess not), if you are interested, welcome to join the 3.0 new UI development.

Before 3.0, we can still have embedding gradio page.

xinference/ui/web/ui/src/scenes/running_models/index.js
```diff
@@ -595,6 +597,79 @@ const RunningModels = () => {
               alignItems: 'left',
             }}
           >
+            <IconButton
```
qinxuye, line 622: This column definition is also reused by the rerank and flexible tabs (`const rerankModelColumns = embeddingModelColumns` and `const flexibleModelColumns = embeddingModelColumns`). Adding the embedding Gradio launcher here makes those tabs show the same button, but clicking it posts a rerank/flexible row to `/v1/ui/embeddings/{model_uid}` and the backend rejects it because `model_type` is not `embedding`. Please keep this launcher only on embedding rows, e.g. by splitting out an embedding-specific columns array or conditionally rendering the button only when `row.model_type === "embedding"`.

xiaoyesoso: Good catch @qinxuye! Fixed in 70f58539 by adding a row.model_type !== "embedding" early-return in the renderCell callback — when the row isn't an embedding, we skip the Gradio launcher and just render renderTerminateButton(row) directly. This keeps the existing rerankModelColumns = embeddingModelColumns / flexibleModelColumns = embeddingModelColumns aliases intact, so no structural changes are needed beyond the one conditional. PTAL 🙏

qinxuye (approved):
LGTM

PR #5025, opened 2026-06-13
feat: add MiniCPM-V-4.6 series support

xiaoyesoso:
Register MiniCPM-V 4.6 (Instruct) and MiniCPM-V 4.6-Thinking, including their pytorch / bnb / awq / gptq / ggufv2 variants, in llm_family.json.

The two models share the new transformers-native architecture `MiniCPMV4_6ForConditionalGeneration` (model_type `minicpmv4_6`), built on SigLIP2-400M + Qwen3.5-0.8B with 256K context and Qwen3.5 vocab (stop_token_ids 248044 / 248046).

Engine adapters:

* `transformers/multimodal/minicpmv46.py`: a new PytorchMultiModalModel subclass that uses `AutoProcessor` + `AutoModelForImageTextToText` and drives the model through `processor.apply_chat_template(..., tokenize= True, return_dict=True, return_tensors="pt")`, forwarding `downsample_mode` / `max_slice_nums` only when the messages actually carry image / video content. OpenAI-style `image_url` / `video_url` payloads are normalized to the v4.6 native `{"type": "image"/"video", "url": "..."}` form.
* `vllm/core.py`: register `MiniCPMV4_6ForConditionalGeneration` in the multimodal architecture list, gated on vLLM >= 0.22.0.
* `sglang/core.py`: register `MiniCPMV4_6ForConditionalGeneration` in `SGLANG_SUPPORTED_VISION_MODEL_LIST` (supported by SGLang main since PR #24998).

Transformers engine requires `transformers>=5.7.0` for the v4.6 architecture, so the family-level virtualenv pins transformers to git main plus `accelerate`, `torchvision`, and `av` for video decoding; GGUF variants go through llama.cpp unchanged.

### DIFF
--- a/xinference/model/llm/transformers/multimodal/minicpmv46.py
+++ b/xinference/model/llm/transformers/multimodal/minicpmv46.py
@@ -152,10 +152,20 @@ class MiniCPMV46Model(PytorchMultiModalModel):
                         continue
                     item_type = item.get("type")
                     if item_type == "image_url":
-                        url = item.get("image_url", {}).get("url", "")
+                        image_url = item.get("image_url")
+                        url = (
+                            image_url.get("url", "")
+                            if isinstance(image_url, dict)
+                            else ""
+                        )
                         new_content.append({"type": "image", "url": url})
                     elif item_type == "video_url":
-                        url = item.get("video_url", {}).get("url", "")
+                        video_url = item.get("video_url")
+                        url = (
+                            video_url.get("url", "")
+                            if isinstance(video_url, dict)
+                            else ""
+                        )
                         new_content.append({"type": "video", "url": url})
                     else:
                         new_content.append(item)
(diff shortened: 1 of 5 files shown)


xinference/model/llm/transformers/multimodal/minicpmv46.py
```diff
@@ -0,0 +1,381 @@
...
+                # Fall back to the first request's tensor and warn loudly,
+                # so mixed-shape batches don't silently corrupt outputs.
+                logger.warning(
+                    "Failed to concatenate multimodal tensor %r across the "
+                    "batch (%s). Falling back to the first request's tensor; "
+                    "batched multimodal inference may be inaccurate.",
+                    key,
+                    e,
+                )
+                merged[key] = tensors[0].to(device)
```
qinxuye: This fallback still lets the batch continue with only the first multimodal tensor when concatenation fails. For example, if two visual requests produce different `pixel_values` shapes, the second request stays in the batch but its visual tensor is dropped, so generation can run against the wrong or missing image data. Please fail fast here, or disable/split incompatible multimodal batching, instead of falling back to `tensors[0]`.

xiaoyesoso: @qinxuye Thanks for the catch — fixed in `e5150cf1` (force-pushed).

`build_prefill_kwargs` now **fails fast** when multimodal tensors from different requests cannot be concatenated, instead of silently falling back to `tensors[0]`. The error message lists the offending shapes and asks the caller to split / retry incompatible visual inputs:

```python
shapes = [tuple(t.shape) for t in tensors]
raise RuntimeError(
    f"Cannot batch MiniCPM-V-4.6 multimodal tensor {key!r} "
    f"across requests: incompatible shapes {shapes} "
    f"({e}). Disable continuous batching for this request "
    f"or split incompatible visual inputs into separate "
    f"batches."
) from e
```

Local `pre-commit run --files xinference/model/llm/transformers/multimodal/minicpmv46.py` is green.
````

qinxuye (approved):
LGTM

PR #5027, opened 2026-06-13
feat: gate model remote-code execution behind a config switch

qxyuan853:
Loading a model could run code shipped inside the model repository (via transformers/sentence-transformers/FlagEmbedding trust_remote_code, the per-engine defaults, and the dynamic script loaders used by a few models), with no way to turn it off. Add XINFERENCE_TRUST_REMOTE_CODE (default off) and route every such loader through it, so model-bundled code only runs when the operator opts in (XINFERENCE_TRUST_REMOTE_CODE=1).


xinference/constants.py
```diff
@@ -109,6 +109,17 @@ def get_xinference_home() -> str:
     os.path.join(XINFERENCE_HOME, "launch_history.db"),
 )
 
+# Whether to allow models to run their own bundled code (transformers /
+# sentence-transformers / FlagEmbedding / engine `trust_remote_code`). Off by
+# default; set XINFERENCE_TRUST_REMOTE_CODE=1 to allow it.
+XINFERENCE_TRUST_REMOTE_CODE = os.environ.get(
```
qinxuye, line 115: This global default change is too broad for existing built-in model integrations. For model implementations that already have dedicated code paths and are known to require remote code (for example MiniCPM-V, InternVL/GLM4V/DeepSeek VL/OCR, and some embedding/rerank modules), we should preserve their current behavior so they do not stop working out of the box. The environment variable is a better fit for the generic loader paths or user-supplied/custom models where remote-code execution is not already an intentional per-model integration decision. Please keep the existing per-model integrations compatible, and apply `XINFERENCE_TRUST_REMOTE_CODE` only to the generic/unknown remote-code paths.

qinxuye (commented):
Found one compatibility blocker around the new default remote-code gate.

qinxuye (commented):
One built-in compatibility case still needs the same treatment.

xinference/model/embedding/sentence_transformers/core.py
```diff
@@ -270,7 +271,7 @@ def to(self, *args, **kwargs):
             self._model = SentenceTransformer(
                 self._model_path,
                 device=self._device,
-                trust_remote_code=True,
+                trust_remote_code=XINFERENCE_TRUST_REMOTE_CODE,
```
qinxuye: This path still includes existing built-in Jina integrations, not only generic/custom sentence-transformers models. `jina-embeddings-v4` and the `jina-embeddings-v5-*` families are registered built-ins and this file already has dedicated Jina handling above, so gating this call on the global env can still make those built-in models fail by default. Please preserve compatibility for the known built-in Jina models, and use `XINFERENCE_TRUST_REMOTE_CODE` only for the generic/unknown sentence-transformers path.

qinxuye: This still is not addressed in the latest commit. The PR now adds `allow_trust_remote_code(model_family)`, which is the right shape, but this `SentenceTransformer(...)` call still uses `XINFERENCE_TRUST_REMOTE_CODE` directly. For built-in Jina embedding families, this should use the same built-in-aware gate, e.g. `trust_remote_code=allow_trust_remote_code(self.model_family)`, so existing registered Jina models keep working by default while custom/generic sentence-transformers models remain gated by the env var.

xinference/model/embedding/sentence_transformers/core.py
```diff
@@ -270,7 +280,7 @@ def to(self, *args, **kwargs):
...
```
qinxuye: This remaining call should use the new built-in-aware gate, not `XINFERENCE_TRUST_REMOTE_CODE` directly. The PR already introduced `allow_trust_remote_code(model_family)`, and `jina-embeddings-v4` / `jina-embeddings-v5-*` are registered built-in integrations handled in this loader. With the current code, those built-in Jina models can still fail by default when the env var is unset. Please switch this to `allow_trust_remote_code(self.model_family)` so built-ins keep compatibility while custom/generic sentence-transformers models remain gated.

qinxuye (commented):
Re-posting the remaining built-in compatibility issue as a fresh inline comment on the latest diff.

qinxuye:
@qxyuan853 could you take a look at the new comments?

qinxuye (approved):
LGTM

qxyuan853:
Hi @qinxuye, I opened an official vulnerability report here: https://github.com/xorbitsai/inference/security/advisories/GHSA-358p-cc53-mf7w

Could you please assign a CVE for this security issue? Thanks!

PR #5029, opened 2026-06-15
feat: add Tencent Hy-MT2 series support

xiaoyesoso:
Register Tencent Hunyuan Hy-MT2 multilingual translation models for deployment via Transformers, vLLM, SGLang, and llama.cpp engines.

Models registered (Hy-MT2-1.8B / 7B / 30B-A3B):
- pytorch (none, fp8) and ggufv2 specs with HF + ModelScope sources
- 33 supported languages, 262144 context length
- Embedded the official chat_template.jinja into llm_family.json
- Correct stop_token_ids per model: 1.8B/7B [120020], 30B [120025,120026]

Engine adapters:
- New transformers/hy_mt2.py adapter forces torch_dtype=bfloat16 to avoid the MPSFloatType embedding error on Apple Silicon and enables trust_remote_code for the custom HunYuanDenseV1/HYV3 modeling code.
- vllm/core.py: add HunYuanDenseV1ForCausalLM and HYV3ForCausalLM to the vLLM (>=0.22.0) supported chat models list.
- sglang/core.py: add the same architectures to SGLANG_SUPPORTED_CHAT_MODELS.

### DIFF
--- a/xinference/model/llm/llm_family.json
+++ b/xinference/model/llm/llm_family.json
@@ -32250,5 +32250,287 @@
     "tool_parser": "deepseek-v4",
     "featured": false,
     "updated_at": 1779522876
+  },
+  {
+    "version": 2,
+    "context_length": 262144,
+    "model_name": "Hy-MT2-1.8B",
+    "model_lang": [
+      "en",
+      "zh"
+    ],
+    "model_ability": [
+      "chat"
+    ],
+    "model_description": "Hy-MT2-1.8B is a dense multilingual translation model from Tencent Hunyuan, supporting 33 languages with instruction-following capabilities for complex translation scenarios.",
+    "model_specs": [
+      {
+        "model_format": "pytorch",
+        "model_size_in_billions": 2,
+        "model_src": {
+          "huggingface": {
+            "quantizations": [
+              "none"
+            ],
+            "model_id": "tencent/Hy-MT2-1.8B",
+            "model_revision": "main"
+          },
+          "modelscope": {
+            "quantizations": [
+              "none"
+            ],
+            "model_id": "Tencent-Hunyuan/Hy-MT2-1.8B",
+            "model_revision": "master"
+          }
+        }
+      },
+      {
+        "model_format": "pytorch",
+        "model_size_in_billions": 2,
+        "quantizations": [
+          "fp8"
+        ],
+        "model_src": {
+          "huggingface": {
+            "quantizations": [
+              "fp8"
+            ],
+            "model_id": "tencent/Hy-MT2-1.8B-FP8",
+            "model_revision": "main"
+          },
+          "modelscope": {
+            "quantizations": [
+              "fp8"
+            ],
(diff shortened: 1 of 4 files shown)


xiaoyesoso:
Addressed Gemini Code Assist's review comments in `f0220038`:

1. **`hy_mt2.py:46`** — Annotated `pytorch_model_config` as `Optional[PytorchModelConfig] = None` in `HyMT2PytorchModel.__init__` to match the parent `PytorchChatModel` signature.
2. **`hy_mt2.py:72`** — `match_json` now returns a descriptive `(False, "Hy-MT2 transformer only supports architectures: …")` tuple when the architecture does not match, consistent with the other transformer adapters in the codebase.

`pre-commit run --files xinference/model/llm/transformers/hy_mt2.py` passes (black/flake8/isort/mypy/codespell).

xinference/model/llm/llm_family.json
```diff
@@ -32250,5 +32250,284 @@
...
+    ],
+    "architectures": [
+      "HunYuanDenseV1ForCausalLM"
+    ],
+    "model_type": "hunyuan_v1_dense",
+    "virtualenv": {
+      "packages": [
+        "transformers>=4.57.0 ; #engine# == \"Transformers\"",
+        "accelerate>=0.28.0 ; #engine# == \"Transformers\"",
+        "#vllm_dependencies# ; #engine# == \"vllm\"",
```
qinxuye, line 32347: The PR registers the Hy-MT2 architectures in `SGLANG_SUPPORTED_CHAT_MODELS`, but these model `virtualenv.packages` only list Transformers/vLLM/llama.cpp dependencies. With virtualenv engine filtering enabled, SGLang is rejected before launch because the model has no `#sglang_dependencies# ; #engine# == "sglang"` marker (`check_engine_by_spec_parameters_with_virtual_env("SGLang", ...)` currently reports that SGLang is not listed for all three Hy-MT2 families). Please add the SGLang dependency marker to the Hy-MT2 virtualenv packages, or drop the SGLang registration if it is not intended to be exposed.

xiaoyesoso: Thanks for catching this — fixed in `0b86b49d`.

I added the `#sglang_dependencies# ; #engine# == "sglang"` marker to the `virtualenv.packages` of all three Hy-MT2 families (`Hy-MT2-1.8B`, `Hy-MT2-7B`, `Hy-MT2-30B-A3B`), so SGLang now passes `check_engine_by_spec_parameters_with_virtual_env` when virtualenv filtering is enabled — consistent with the `SGLANG_SUPPORTED_CHAT_MODELS` registration in `xinference/model/llm/sglang/core.py`.

Per the upstream [Hy-MT2 SGLang deployment doc](https://github.com/Tencent-Hunyuan/Hy-MT2#sglang) the SGLang path is officially supported, so I kept the registration rather than dropping it.

qinxuye (approved):
LGTM

PR #5030, opened 2026-06-15
fix(vllm): force spawn for single-GPU EngineCore and fix health check scheduling

m199369309:
## Summary

- **Force spawn for single-GPU vLLM v1 EngineCore creation** to avoid fork deadlock. When `tensor_parallel_size=1`, vLLM's `MultiprocExecutor` creates EngineCore via `mp.get_context("fork").Process()`, inheriting the parent's multi-thread lock state and causing permanent deadlock in the request processing loop. Setting `VLLM_WORKER_MULTIPROC_METHOD=spawn` creates a clean process. vLLM natively supports this env var (`vllm/envs.py`) and ZMQ IPC uses path strings, so spawn is safe.
- **Move health check creation from `load()` to `wait_for_load()`** — the old code created it when `self._engine` was still `None` for threaded paths (both multi-GPU and single-GPU), silently skipping health check entirely.
- **Fix `AsyncEngineDeadError` import compatibility** — removed from `vllm.engine.async_llm_engine` in vLLM 0.19.0+. Module-level `try/except` in `utils.py` with `RuntimeError` fallback in `_check_healthy()`.
- **Use `call_soon_threadsafe` + `create_task`** instead of `run_coroutine_threadsafe` — the latter wraps coroutines in `concurrent.futures.Future` whose exceptions are silently swallowed; `create_task` produces `asyncio.Task` with visible exceptions.
- **Wait for loading thread in `stop()`** before shutting down the engine, preventing orphan EngineCore subprocesses.

## Test plan

- [x] Single-GPU Qwen3-8B (BF16): 3 consecutive chat requests succeed (previously hung indefinitely)
- [x] Single-GPU Llama-3.1-8B (BF16): chat requests succeed
- [x] `"Begin to check health of vLLM"` log appears for the first time (previously 0 occurrences across all deployments)
- [x] No `Task exception was never retrieved` errors
- [x] Multi-GPU (TP=2) deployment unaffected (uses `XinferenceDistributedExecutorV1`, no fork)
- [ ] `pytest xinference/model/llm/vllm/tests/` passes

## Root cause analysis

Single-GPU vLLM v1 deployments (Qwen3-8B, Llama-3.1-8B, Qwen2.5-Instruct) experienced permanent request hangs:
- Model loaded successfully, CUDA graphs captured
- Any inference request hung indefinitely with no error logs
- EngineCore subprocess stayed alive (stats telemetry continued)
- `serve request count` kept incrementing, no timeout

The fork deadlock occurs because the parent process (ModelActor subpool) has multiple active threads (MainThread event loop, `_loading_thread`, asyncio workers). Fork copies all memory including lock states, but the child only has one thread — it cannot release locks held by the other threads.

Multi-GPU deployments are unaffected because they use `XinferenceDistributedExecutorV1` (xoscar actors, no fork).

### DIFF
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -68,7 +68,7 @@ from ..utils import (
     ChatModelMixin,
     generate_completion_chunk,
 )
-from .utils import vllm_check
+from .utils import AsyncEngineDeadError, vllm_check
 
 logger = logging.getLogger(__name__)
 
@@ -654,12 +654,21 @@ class VLLMModel(LLM):
                 **self._model_config,
             )
             self._enable_v1_if_supported(engine_args)
-            self._engine = AsyncLLMEngine.from_engine_args(engine_args)
 
-        self._check_health_task = None
-        if hasattr(self._engine, "check_health"):
-            # vLLM introduced `check_health` since v0.4.1
-            self._check_health_task = self._loop.create_task(self._check_healthy())
+            def _load():
+                # Force spawn to avoid fork deadlock: vLLM v1 creates
+                # EngineCore via fork, which inherits parent's multi-thread
+                # lock state causing deadlock. spawn creates a clean process.
+                os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
+                try:
+                    self._engine = AsyncLLMEngine.from_engine_args(engine_args)
+                except:  # noqa: E722
+                    logger.exception("Creating vllm engine failed")
+                    self._loading_error = sys.exc_info()
+
+            self._loading_thread = threading.Thread(target=_load)
+            self._loading_thread.start()
+            self._loading_thread.join(1)
 
     def wait_for_load(self):
         if self._loading_thread:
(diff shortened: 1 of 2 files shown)


xinference/model/llm/vllm/utils.py
```diff
@@ -18,18 +18,22 @@
...
-    except Exception:
-        return fn
 
+def vllm_check(fn):
     @functools.wraps(fn)
     async def _async_wrapper(self, *args, **kwargs):
         try:
             return await fn(self, *args, **kwargs)
-        except AsyncEngineDeadError:
+        except (AsyncEngineDeadError, RuntimeError):
```
qinxuye: This makes `vllm_check` catch every `RuntimeError` on the request path. The decorator wraps `async_generate` / `async_chat`, so ordinary request/runtime errors from vLLM or local preprocessing would now call `os._exit(1)` instead of being returned as request failures. Please only fall back to `RuntimeError` when `AsyncEngineDeadError` cannot be imported; when the dedicated exception class is available, keep catching that specific class rather than all `RuntimeError`s.

qinxuye (commented):
Found one issue in the vLLM health-check wrapper.

m199369309 (commented):
Addressed review findings:

1. **`vllm_check` over-catching `RuntimeError`** (qinxuye + Gemini Critical): Fixed. `except (AsyncEngineDeadError, RuntimeError)` narrowed to `except AsyncEngineDeadError`. When the import succeeds (vLLM <0.19), only the specific `AsyncEngineDeadError` is caught — ordinary `RuntimeError` from request paths no longer triggers `os._exit(1)`. When the import fails (vLLM 0.19.0+), the module-level fallback `AsyncEngineDeadError = RuntimeError` is the only degradation path, which is correct since vLLM removed the dedicated exception class.

2. **Bare `except:` in `_check_healthy`** (Gemini Medium): Fixed, replaced with `except Exception:` for PEP 8 compliance. The bare `except:` at the engine creation path (core.py) was already fixed to `except Exception:` in the prior commit.

3. **`_check_health_task` race** (Gemini High): No change needed. The current PR already assigns `self._check_health_task` inside `_start_health_check()` which runs on the event loop — the assignment is correct. The theoretical race window (stop() called before the event loop executes `_start_health_check`) is not practically reachable since `stop()` is only called during model teardown, long after `load()` completes.

xinference/model/llm/vllm/core.py
```diff
@@ -780,6 +816,10 @@ def stop(self):
         logger.info("Stopping vLLM engine")
         if self._check_health_task:
```
qinxuye, line 818: `_check_health_task` is now initialized in `wait_for_load()`, but `stop()` can be called before that point while the background loading thread is still running, for example during cleanup/cancellation. In that window this attribute does not exist, so `stop()` raises `AttributeError` before it can join the loading thread or shut down the engine. Please initialize `self._check_health_task = None` in `__init__` (preferred) or guard this with `getattr`.

m199369309 (commented):
**`_check_health_task` AttributeError** (qinxuye): Fixed. Added `self._check_health_task = None` to `__init__` so `stop()` is safe to call before `wait_for_load()` completes.

qinxuye (approved):
LGTM

PR #5031, opened 2026-06-15
fix(worker): strip test envs before recover and clean up GPU orphans

m199369309:
## Summary

- **Strip `XINFERENCE_TEST_*` envs from cached `launch_args`** before recover and startup recovery. Test-injected env vars (e.g. OOM simulation flags) were being cached alongside model parameters and replayed on every recover, causing infinite recover loops where the model repeatedly crashes with the same test env.
- **NVML-based VRAM polling after terminate** in `recover_sub_pool`. Replaces implicit reliance on process exit timing with explicit VRAM free-ratio check (90% threshold, 30s timeout). Gracefully degrades to psutil if pynvml is unavailable.
- **SIGKILL orphan GPU processes** after `terminate_model` in both recover and normal terminate paths. Spawn-created EngineCore subprocesses can survive `stop()` + `remove_sub_pool` as orphans, holding ~21.5G VRAM and blocking subsequent launches. Uses cmdline verification (`vllm`/`enginecore`/`model_uid`) to safely identify orphans.
- **Per-uid persist dirty tracking** with retry limit (`_PERSIST_RETRY_MAX=3`) prevents infinite retry storms when disk writes fail.
- **Startup recovery path** (`_try_recover_models`) also strips test envs from persisted launch_args.

## Test plan

- [x] Unit tests: `_strip_test_envs` (13 test cases covering prefix matching, reference isolation, non-dict envs, exception safety)
- [x] Unit tests: GPU utility functions (`_parse_gpu_indices`, `_snapshot_gpu_occupying_pids`, `_snapshot_gpu_free_ratio`, `_kill_orphan_gpu_pids`)
- [ ] `pytest xinference/core/tests/test_strip_test_envs.py -v` passes
- [ ] `pytest xinference/core/tests/test_launch_orphan_cleanup.py -v` passes
- [ ] TP=1 OOM self-heal: terminate → VRAM returns to 0 → clean relaunch succeeds
- [ ] TP=2 OOM self-heal: VRAM polling confirms release before recover
- [ ] Multi-model same worker: orphan cleanup doesn't affect other models' processes

## Root cause

When models are recovered (OOM self-heal or worker restart), cached `launch_args` containing test-injected env vars are replayed verbatim, causing the model to crash again with the same test flag. Additionally, `spawn`-created EngineCore subprocesses (introduced for fork deadlock avoidance) survive `stop()` as orphans, holding GPU VRAM and causing `ValueError: Free memory on device ... less than desired` on subsequent launches.

## Design notes

- `pynvml` is an optional dependency — all GPU utility functions gracefully degrade when unavailable
- Test env convention: all test-only env vars must use `XINFERENCE_TEST_` prefix
- Orphan identification uses cmdline matching (not PID diff) to correctly handle the terminate path where the orphan was already running before stop
- Worker PID and other models' known PIDs are excluded from orphan candidates

### DIFF
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -2784,7 +2784,10 @@ class WorkerActor(xo.StatelessActor):
                 # process may disappear, we just ignore it.
                 logger.debug("Fail to get pool addresses, error: %s", e)
 
-        # Resolve GPU indices for terminate-path orphan cleanup
+        # Resolve GPU indices for terminate-path orphan cleanup.
+        # Prefer launch_args["gpu_idx"] (user-specified), but fall back to
+        # model_spec["accelerators"] (allocated devices when gpu_idx=None
+        # and _create_subpool() auto-selected GPUs).
         _gpu_indices_for_terminate: list = []
         if not is_model_die:
             _launch_args = self._model_uid_to_launch_args.get(model_uid)
@@ -2792,6 +2795,12 @@ class WorkerActor(xo.StatelessActor):
                 _gpu_indices_for_terminate = _parse_gpu_indices(
                     _launch_args.get("gpu_idx")
                 )
+            if not _gpu_indices_for_terminate:
+                _model_spec = self._model_uid_to_model_spec.get(model_uid)
+                if _model_spec and _model_spec.get("accelerators"):
+                    _gpu_indices_for_terminate = [
+                        int(dev) for dev in _model_spec["accelerators"]
+                    ]
 
         try:
             logger.debug("Start to destroy model actor: %s", model_ref)
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -305,6 +305,20 @@ async def _wait_pids_dead(pids: set, timeout: float = 5.0):
             await asyncio.sleep(0.2)
 
 
+def _process_or_ancestor_has_uid(proc: "psutil.Process", uid_lower: str) -> bool:
+    """Return True if proc or any of its ancestors has uid_lower in cmdline."""
+    current = proc
+    while current is not None:
+        try:
+            cmd = " ".join(current.cmdline()).lower()
+            if uid_lower in cmd:
+                return True
+            current = current.parent()
+        except Exception:
+            return False
+    return False
+
+
 async def _kill_orphan_gpu_pids(
     device_indices: list,
     pre_pids: set,
@@ -316,8 +330,8 @@ async def _kill_orphan_gpu_pids(
     An orphan is a PID present on the GPU that was not in pre_pids (the set
     of PIDs known to belong to other models or the worker itself).
 
-    Requires model_uid in the process cmdline to avoid killing unrelated
-    processes on shared GPUs.
+    Requires model_uid in the process cmdline (or in an ancestor's cmdline)
+    to avoid killing unrelated processes on shared GPUs.
     """
     post_pids = _snapshot_gpu_occupying_pids(device_indices)
     orphan_pids = [pid for pid in post_pids if pid not in pre_pids]
@@ -334,8 +348,11 @@ async def _kill_orphan_gpu_pids(
             if not p.is_running() or p.status() == psutil.STATUS_ZOMBIE:
                 continue
             cmd = " ".join(p.cmdline()).lower()
-            if uid_lower and uid_lower not in cmd:
-                continue
+            if uid_lower:
+                if uid_lower not in cmd and not _process_or_ancestor_has_uid(
+                    p, uid_lower
+                ):
+                    continue
             if not any(k in cmd for k in ("vllm", "enginecore", "python")):
                 continue
             p.kill()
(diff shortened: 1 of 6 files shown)


xinference/core/worker.py
```diff
@@ -154,6 +155,204 @@ def _exclusive_venv_path_lock(env_path: str):
...
+
+    killed = []
+    uid_lower = model_uid.lower() if model_uid else ""
+    for pid in orphan_pids:
+        try:
+            p = psutil.Process(pid)
+            if not p.is_running() or p.status() == psutil.STATUS_ZOMBIE:
+                continue
+            cmd = " ".join(p.cmdline()).lower()
+            if uid_lower and uid_lower not in cmd:
```
qinxuye: This guard now requires the orphan process command line to contain `model_uid`, but the current code does not make that true for the vLLM child processes this cleanup is trying to reclaim. `ModelActor` sets its process title to `Model: <uid>`, while the vLLM worker title is only `Xinf vLLM worker: <rank>`, and the test also calls `_kill_orphan_gpu_pids` without passing `model_uid`, so it does not cover the production filter. In practice this can make the cleanup skip the leaked EngineCore/worker process entirely. Please add a reliable provenance marker for the launched model, or match from tracked child PIDs, and cover the positive `model_uid` path in the test.

qinxuye (commented):
One new issue found.

m199369309 (commented):
Addressed review finding:

**model_uid provenance for vLLM child processes** (qinxuye): Fixed with a two-pronged approach:

1. **Process title marker**: `VLLMModel.load()` now sets `XINFERENCE_MODEL_UID` env var before engine creation. Both `distributed_executor.py` and `distributed_executor_v1.py` WorkerActors now include `[model_uid]` in their process title: `Xinf vLLM worker: <rank> [<uid>]`.

2. **Ancestry walk for EngineCore**: EngineCore is forked by vLLM itself — we can't control its process title, but it inherits the environment from its parent. Added `_process_or_ancestor_has_uid()` helper that walks up the parent chain checking cmdline. This catches EngineCore whose parent (the vLLM worker) now has model_uid in its title.

3. **Test coverage**: Added `test_kill_orphan_gpu_pids_model_uid_filter` (covers: worker with uid in cmdline → killed, EngineCore with uid only in parent → killed, unrelated process → skipped) and `test_kill_orphan_gpu_pids_model_uid_skips_other_model` (covers: other model's worker → skipped).

**Other findings (already fixed in prior commits, no changes needed):**
- Gemini Critical ×2 (recover/terminate path missing model_uid check): both call sites already pass model_uid
- Gemini High ×2 (missing helper functions): `_kill_orphan_gpu_pids` and `_wait_pids_dead` are defined in worker.py
- Gemini High ×2 (VRAM reclaim loop not breaking on -1): both loops already have `if _free_ratio < 0: break`
- Gemini Medium ×3 (pynvml not shut down): all three call sites already have `try...finally` with `nvmlShutdown()`

xinference/core/worker.py
```diff
@@ -2449,6 +2784,15 @@ async def terminate_model(self, model_uid: str, is_model_die=False):
                 # process may disappear, we just ignore it.
                 logger.debug("Fail to get pool addresses, error: %s", e)
 
+        # Resolve GPU indices for terminate-path orphan cleanup
+        _gpu_indices_for_terminate: list = []
+        if not is_model_die:
+            _launch_args = self._model_uid_to_launch_args.get(model_uid)
+            if _launch_args:
+                _gpu_indices_for_terminate = _parse_gpu_indices(
```
qinxuye, line 2795: This terminate-path cleanup still only derives GPUs from `launch_args["gpu_idx"]`, but the default load-first launch path passes `target_gpu_idx=None` and lets `_create_subpool()` allocate devices internally. That allocated `devices` list is not written back to `launch_args`, so `_gpu_indices_for_terminate` stays empty and this orphan cleanup is skipped for the common default launch path. Please track the actual allocated devices, for example alongside `_model_uid_to_subpool_pids` or by saving them after `_create_subpool()` returns, and use those for terminate cleanup.

m199369309 (commented):
**Terminate path missing GPU indices when `gpu_idx=None`** (qinxuye): Fixed.

When `gpu_idx=None` is passed to launch (the default), `_create_subpool()` auto-selects GPUs via `allocate_devices()` but doesn't write them back to `launch_args`. The terminate path only checked `launch_args["gpu_idx"]`, so `_gpu_indices_for_terminate` stayed empty and orphan cleanup was skipped entirely for the default launch path.

Fix: fall back to `model_spec["accelerators"]` (the allocated devices saved at `model.model_family.accelerators = devices` after `_create_subpool()` returns) when `launch_args["gpu_idx"]` is empty. This covers both user-specified and auto-allocated GPU paths.

**Other findings (already fixed in prior commits, no changes needed):**
- Gemini Critical ×2 (recover/terminate missing model_uid check): both call sites pass model_uid
- Gemini High ×2 (missing helper functions): both functions exist in worker.py
- Gemini High ×2 (VRAM reclaim loop not breaking on -1): both loops have `if _free_ratio < 0: break`
- Gemini Medium ×3 (pynvml not shut down): all three call sites have `try...finally` with `nvmlShutdown()`
- qinxuye (model_uid provenance): fixed with `XINFERENCE_MODEL_UID` env var + ancestry walk

qinxuye (approved):
LGTM

PR #5032, opened 2026-06-15
feat(core): add model loading state machine to prevent routing to loading models

m199369309:
## Summary

- **Introduce explicit model lifecycle state machine**: `registering → loading → ready → stopping → stopped` (plus `error`), replacing implicit state inference from dict membership. This fixes the race condition where `list_models` shows a model but chat requests return `500 AssertionError: self._engine is None` because the engine is still loading in a background thread.
- **New `ModelNotReadyError` exception** (`xinference/core/exceptions.py`) raised by `ModelActor._require_ready()` guard on all inference methods (generate, chat, create_embedding, rerank) when the model is not yet ready.
- **`LaunchStatus.LOADING`** enum value and **`ReplicaStatus.model_state`** field for fine-grained state tracking visible to supervisors and API consumers.
- **Supervisor race condition fix**: swap order in `_launch_one_model` so `wait_for_load` completes *before* the model is added to the route table (both normal and xavier rank0 paths). Previously, the route table was written before `wait_for_load`, allowing requests to hit a still-loading model.
- **HTTP 503 with `Retry-After: 30`** header for loading-state requests in the API layer (`require_model`). Loading is transient so `ModelNotReadyError` is **not** written to negative cache.
- **`WorkerActor._update_model_state()`** helper syncs state transitions to StatusGuard, with auto-creation of `ModelStatus` entries to support registering/loading states (previously only created after `load()` completed).

## Test plan

- [ ] `pytest xinference/core/tests/ -v --timeout=300` passes
- [ ] Fresh launch: `model_state` transitions through registering → loading → ready
- [ ] Chat during loading: returns HTTP 503 (not 500 AssertionError)
- [ ] OOM self-heal recover: model transitions through error → registering → loading → ready
- [ ] `GET /v1/models` includes `model_state` field
- [ ] Non-LLM models (embedding/rerank): loading state is transient (sub-second), no client-visible impact
- [ ] Multi-replica: only `ready` replicas are routed to
- [ ] Terminate: model transitions through stopping → stopped

## Design notes

- **State authority**: `WorkerActor.ModelStatus.model_state` is the authoritative source (routing decisions). `ModelActor._model_state` is a safety net for direct calls bypassing the worker.
- **Backward compatibility**: `model_state` defaults to empty string for existing data, treated as `ready` (current behavior).
- **`wait_ready=False` async launch**: model is visible in `list_models` (with `model_state=loading`) but not routable until ready. Clients can poll `model_state` to know when it's available.

### DIFF
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -2714,8 +2714,13 @@ class SupervisorActor(xo.StatelessActor):
 
         worker_ref = self._replica_model_uid_to_worker.get(replica_model_uid, None)
         if worker_ref is None:
-            raise ValueError(
-                f"Model not found in the model list, uid: {replica_model_uid}"
+            # replica_info exists but worker route not yet registered —
+            # model is being launched (wait_ready=False), not missing.
+            # Raise ModelNotReadyError (503) instead of ValueError (404)
+            # so the negative cache is not poisoned and clients get a
+            # retry-friendly response.
+            raise ModelNotReadyError(
+                f"Model {model_uid} is launching, not yet ready for inference"
             )
         if isinstance(worker_ref, (list, tuple)):
             # get first worker to fetch information if model across workers
@@ -2761,8 +2766,8 @@ class SupervisorActor(xo.StatelessActor):
         )
         worker_ref = self._replica_model_uid_to_worker.get(replica_model_uid, None)
         if worker_ref is None:
-            raise ValueError(
-                f"Model not found in the model list, uid: {replica_model_uid}"
+            raise ModelNotReadyError(
+                f"Model {model_uid} is launching, not yet ready for inference"
             )
         if isinstance(worker_ref, (list, tuple)):
             # get status from first shard if model has multiple shards across workers
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -56,6 +56,7 @@ from ..model.utils import (
     get_engine_params_by_name_with_virtual_env,
 )
 from ..types import PeftModelConfig
+from .exceptions import ModelNotReadyError
 from .launch_strategy import IdleFirstLaunchStrategy
 from .metrics import record_metrics
 from .resource import GPUStatus, ResourceStatus
@@ -1748,7 +1749,6 @@ class SupervisorActor(xo.StatelessActor):
                 rank0_address, _port = await worker_ref.launch_rank0_model(
                     _replica_model_uid, xavier_config
                 )
-                self._replica_model_uid_to_worker[_replica_model_uid] = worker_ref
                 store_address = rank0_address.split(":")[0]
                 store_port = _port
 
@@ -1756,6 +1756,7 @@ class SupervisorActor(xo.StatelessActor):
                 await self._status_guard_ref.update_replica_status(
                     model_uid, replica_id, {"status": LaunchStatus.READY.name}
                 )
+                self._replica_model_uid_to_worker[_replica_model_uid] = worker_ref
                 return rank0_address
 
             replica_gpu_idx = (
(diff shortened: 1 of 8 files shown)


xinference/core/supervisor.py
```diff
@@ -1794,8 +1795,10 @@ async def _launch_one_model(
                     xavier_config=xavier_config,
                     **kwargs,
                 )
-                self._replica_model_uid_to_worker[_replica_model_uid] = worker_ref
+                # Wait for engine to be ready BEFORE adding to route table,
+                # so requests are never routed to a still-loading model.
                 await worker_ref.wait_for_load(_replica_model_uid)
```
qinxuye, line 1800: Delaying `_replica_model_uid_to_worker` until after `wait_for_load` fixes the ready-route race for synchronous launch, but it regresses `wait_ready=false`: after the launch API returns and before the background task reaches this line, `supervisor.get_model()` sees no route entry and raises `ValueError`. `require_model()` then treats that as a real 404 and writes the uid into the negative cache, and the background launch completion path does not invalidate that cache. A request made during loading can therefore keep returning cached 404 for up to the cache TTL even after the model becomes ready. Please keep the loading uid distinguishable from a truly missing uid, for example by making this path return `ModelNotReadyError`/503 while the background launch is still in progress, or by invalidating the negative cache when the async launch task completes successfully.

qinxuye (commented):
One new issue found.

xinference/core/supervisor.py
qinxuye: This still applies on the current head. The launch API invalidates the negative cache only when it returns, but a request made after that and before the background `_launch_model()` reaches the route-table insertion can still cache a 404. The async completion path should either expose the uid as loading and return `ModelNotReadyError`/503, or clear the negative cache again once the background launch succeeds.

m199369309 (commented):
**`wait_ready=False` + negative cache race** (qinxuye): Fixed.

In `supervisor.get_model()` and `describe_model()`, when `replica_info` exists but `_replica_model_uid_to_worker` has no entry yet (model is launching with `wait_ready=False`), now raises `ModelNotReadyError` (HTTP 503, retry-friendly) instead of `ValueError` (HTTP 404, cached as not-found).

This prevents the negative cache from being poisoned during the async launch window. Clients get a 503 with `Retry-After: 30` header, which is the correct semantic for "model exists but not ready yet".

**Other findings (already fixed in prior commits, no changes needed):**
- Gemini High ×3: `_require_ready` already checks all non-ready states; `_update_model_state` already uses `@log_async`; `get_model` in worker.py already guards `error`/`stopping`/`stopped`
- Gemini Medium: error path already calls `_update_model_state(model_uid, "error")`

xinference/core/supervisor.py
```diff
@@ -2753,8 +2766,8 @@ async def describe_model(self, model_uid: str) -> Dict[str, Any]:
         )
         worker_ref = self._replica_model_uid_to_worker.get(replica_model_uid, None)
         if worker_ref is None:
-            raise ValueError(
-                f"Model not found in the model list, uid: {replica_model_uid}"
+            raise ModelNotReadyError(
```
qinxuye, line 2769: This new `ModelNotReadyError` path also reaches the REST `describe_model` endpoint (`GET /v1/models/{model_uid}`), but that handler only maps `ValueError` specially and sends all other exceptions through the generic 500 path. So after a `wait_ready=false` launch, model detail requests can return 500 for a transient loading model while inference requests correctly return 503 via `require_model()`. Please catch `ModelNotReadyError` in the REST describe handler too and return the same retry-friendly 503.

qinxuye (approved):
LGTM

PR #5036, opened 2026-06-15
feat: add random serving benchmark workload

qinxuye:
## Summary

- Add `--dataset-name random` to `benchmark_serving.py` for vLLM-style synthetic request generation without ShareGPT.
- Add random prompt generation helpers with configurable input/output lengths, range ratio, shared prefix length, and seed.
- Thread `ignore_eos` through the benchmark runner so synthetic output lengths are more stable.
- Update benchmark README examples and fix the `--num-prompts` flag spelling.

## Validation

- `python -m black --check benchmark/utils.py benchmark/benchmark_runner.py benchmark/benchmark_serving.py`
- `python -m isort --check-only benchmark/utils.py benchmark/benchmark_runner.py benchmark/benchmark_serving.py`
- `python -m flake8 benchmark/utils.py benchmark/benchmark_runner.py benchmark/benchmark_serving.py`
- `python -m py_compile benchmark/utils.py benchmark/benchmark_runner.py benchmark/benchmark_serving.py`
- `python benchmark/benchmark_serving.py --help`

### DIFF
--- a/benchmark/benchmark_runner.py
+++ b/benchmark/benchmark_runner.py
@@ -12,16 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import aiohttp
 import json
+import logging
 import sys
+import time
 import traceback
 import warnings
-import logging
 from dataclasses import dataclass, field
-import time
 from typing import List, Optional, Tuple
 
+import aiohttp
 import numpy as np
 
 logging.basicConfig(level=logging.INFO)
@@ -56,6 +56,7 @@ class BenchmarkRunner:
         stream: bool,
         api_key: Optional[str] = None,
         print_error: bool = False,
+        ignore_eos: bool = False,
     ):
         self.api_url = api_url
         self.model_uid = model_uid
@@ -65,6 +66,7 @@ class BenchmarkRunner:
         self.stream = stream
         self.api_key = api_key
         self.print_error = print_error
+        self.ignore_eos = ignore_eos
 
     async def run(self):
         await self.warm_up()
@@ -107,6 +109,8 @@ class BenchmarkRunner:
                 "stream": False,
                 "messages": [{"role": "user", "content": prompt}],
             }
+        if self.ignore_eos:
+            pload["ignore_eos"] = True
 
         headers = {"User-Agent": "Benchmark Client"}
         if self.api_key:
@@ -142,7 +146,9 @@ class BenchmarkRunner:
                                 if not chunk_bytes:
                                     continue
 
-                                chunk = remove_prefix(chunk_bytes.decode("utf-8"), "data:")
+                                chunk = remove_prefix(
+                                    chunk_bytes.decode("utf-8"), "data:"
+                                )
 
                                 if chunk == "[DONE]":
                                     latency = time.perf_counter() - st
@@ -157,18 +163,24 @@ class BenchmarkRunner:
 
                                     # Decoding phase
                                     else:
-                                        output.itl.append(timestamp - most_recent_timestamp)
+                                        output.itl.append(
+                                            timestamp - most_recent_timestamp
+                                        )
 
                                     most_recent_timestamp = timestamp
 
                             output.latency = latency
                             output.success = True
-                            output.completion_tokens = data["usage"]["completion_tokens"]
+                            output.completion_tokens = data["usage"][
+                                "completion_tokens"
+                            ]
                         else:
                             resp = await response.json()
                             output.latency = time.perf_counter() - st
                             output.success = True
-                            output.completion_tokens = resp["usage"]["completion_tokens"]
+                            output.completion_tokens = resp["usage"][
+                                "completion_tokens"
+                            ]
             except Exception:
                 output.success = False
                 exc_info = sys.exc_info()
@@ -368,7 +380,9 @@ class BenchmarkRunner:
                 logger.info("Errors encountered during benchmark:")
                 for output in self.outputs:
                     if not output.success:
-                        print(f"Error for prompt with length {output.prompt_len}: {output.error}")
+                        print(
+                            f"Error for prompt with length {output.prompt_len}: {output.error}"
+                        )
             else:
                 logger.info(
                     "Errors were encountered during the benchmark. Run with --print-error to see detailed error messages."
@@ -385,6 +399,7 @@ class ConcurrentBenchmarkRunner(BenchmarkRunner):
         concurrency: int,
         api_key: Optional[str] = None,
         print_error: bool = False,
+        ignore_eos: bool = False,
     ):
         super().__init__(
             api_url,
@@ -393,6 +408,7 @@ class ConcurrentBenchmarkRunner(BenchmarkRunner):
             stream,
             api_key,
             print_error,
+            ignore_eos,
         )
         self.concurrency = concurrency
         self.left = len(input_requests)
--- a/benchmark/benchmark_runner.py
+++ b/benchmark/benchmark_runner.py
@@ -175,7 +175,8 @@ class BenchmarkRunner:
                             output.success = True
                             output.completion_tokens = (
                                 usage["completion_tokens"]
-                                if usage and "completion_tokens" in usage
+                                if isinstance(usage, dict)
+                                and isinstance(usage.get("completion_tokens"), int)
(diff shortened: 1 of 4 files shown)


benchmark/utils.py
```diff
@@ -140,6 +144,260 @@ def sample_requests(
...
+    if vocab_size <= 0:
+        raise ValueError("Tokenizer vocab size must be positive.")
+    return vocab_size
+
+
+def _num_special_tokens_to_add(tokenizer: "PreTrainedTokenizerBase") -> int:
+    num_special_tokens_to_add = getattr(tokenizer, "num_special_tokens_to_add", None)
+    if callable(num_special_tokens_to_add):
+        return int(num_special_tokens_to_add())
+    return 0
```
rogercloud, line 175: `_num_special_tokens_to_add` returns 0 when `num_special_tokens_to_add` is a non-callable attribute (e.g. an int property). While `PreTrainedTokenizerBase` defines it as a method, some tokenizer subclasses may expose it as a plain int. This would silently cause incorrect prompt length calculations.

Consider adding a fallback for the int case:

```python
def _num_special_tokens_to_add(tokenizer: "PreTrainedTokenizerBase") -> int:
    num_special_tokens_to_add = getattr(tokenizer, "num_special_tokens_to_add", None)
    if callable(num_special_tokens_to_add):
        return int(num_special_tokens_to_add())
    if isinstance(num_special_tokens_to_add, int):
        return num_special_tokens_to_add
    return 0
```

benchmark/benchmark_serving.py
```diff
@@ -103,12 +133,30 @@ def main(args: argparse.Namespace):
 
     logger.info("Preparing for benchmark.")
     tokenizer = get_tokenizer(args.tokenizer, trust_remote_code=args.trust_remote_code)
-    input_requests = sample_requests(
-        args.dataset,
-        args.num_prompts,
-        tokenizer,
-        prompt_len_limit=args.prompt_len_limit,
-    )
+    if args.request_rate <= 0.0:
```
rogercloud, line 148: The `--request-rate` validation runs after tokenizer initialization (line 135), which can be expensive for large tokenizers. If the user provides an invalid `request_rate`, they wait for tokenizer loading before getting the error.

Consider moving this check before `get_tokenizer()`:

```python
    logger.info("Preparing for benchmark.")
    if args.request_rate <= 0.0:
        raise ValueError("--request-rate must be positive.")
    tokenizer = get_tokenizer(args.tokenizer, trust_remote_code=args.trust_remote_code)
```

rogercloud (changes requested):
## Summary

Overall the feature is well-implemented — clean separation between CLI parsing (`benchmark_serving.py`), benchmark orchestration (`benchmark_runner.py`), and prompt generation (`utils.py`). The `ignore_eos` threading and `completion_tokens` defensive handling are good improvements. The `--num-prompts` spelling fix in README is also appreciated.

## Issues

1. **Major** — `_num_special_tokens_to_add` returns 0 for non-callable attributes (see inline)
2. **Minor** — `request_rate` validation happens after tokenizer loading (see inline)

## Positive Observations

- `ignore_eos` is cleanly threaded through all three constructor chains (`BenchmarkRunner` → `ConcurrentBenchmarkRunner` → `ServingBenchmarkRunner`)
- `completion_tokens` fallback logic in `send_request` handles missing usage data gracefully in both streaming and non-streaming paths
- `parse_random_range_ratio` provides good error messages for CLI misconfiguration
- `_gen_prompt_decode_to_target_len` retry loop correctly handles tokenizer decode-re-encode mismatches
- Reproducibility is preserved via `np.random.default_rng(seed)` local RNG for prompt generation alongside global `np.random.seed` for request timing

## Validation

- `py_compile` passes on all three changed Python files

benchmark/benchmark_runner.py
```diff
@@ -157,18 +165,31 @@ async def send_request(self, request: tuple, warming_up: bool = False):
...
                             output.latency = time.perf_counter() - st
                             output.success = True
-                            output.completion_tokens = resp["usage"]["completion_tokens"]
+                            usage = resp.get("usage")
+                            output.completion_tokens = (
+                                usage["completion_tokens"]
+                                if isinstance(usage, dict)
+                                and isinstance(usage.get("completion_tokens"), int)
+                                else 0
+                            )
```
rogercloud: **Minor** — non-streaming fallback silently zeroes `completion_tokens`

When `usage` is absent or malformed in a non-streaming response, `completion_tokens` is set to `0` with no log message. This silently zeroes output tokens/s in throughput stats and is hard to diagnose. The streaming path at least approximates with `len(output.itl) + 1`.

Consider adding a `logger.warning` here so operators know when reported stats are unreliable:
```python
else:
    logger.warning(
        "usage missing or malformed in non-streaming response; "
        "completion_tokens set to 0 and throughput stats will be inaccurate."
    )
    output.completion_tokens = 0
```

rogercloud (commented):
## Re-review

All 9 prior findings (Gemini ×7, rogercloud ×2) are fully addressed. Three new minor issues found via inline comments below.

benchmark/benchmark_runner.py
rogercloud: **Minor** — non-streaming fallback silently zeroes `completion_tokens`

When `usage` is absent or malformed in a non-streaming response, `completion_tokens` is set to `0` with no log message. This silently zeroes output tokens/s in throughput stats and is hard to diagnose. The streaming path at least approximates with `len(output.itl) + 1`.

Consider adding a `logger.warning` so operators know when stats are unreliable:
```python
else:
    logger.warning(
        "usage missing or malformed in non-streaming response; "
        "completion_tokens set to 0 and throughput stats will be inaccurate."
    )
    output.completion_tokens = 0
```

benchmark/benchmark_serving.py
```diff
@@ -14,20 +14,48 @@
...
-
+from utils import get_tokenizer, sample_random_requests, sample_requests
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 
+def parse_random_range_ratio(value):
+    try:
+        return float(value)
```
rogercloud: **Minor** — `parse_random_range_ratio` accepts out-of-range floats at parse time

`float(value)` succeeds for inputs like `"1.5"` or `"inf"`. The validation is deferred until runtime inside `_get_sampling_params`, where the error message uses the internal variable name (`"input_range_ratio must be in [0, 1)"`), not the CLI flag name.

Better to validate at parse time:
```python
try:
    f = float(value)
    if not (0.0 <= f < 1.0):
        raise argparse.ArgumentTypeError(
            f"--random-range-ratio must be in [0, 1), got {f}."
        )
    return f
except ValueError:
    pass
```

benchmark/utils.py
```diff
@@ -140,6 +144,262 @@ def sample_requests(
...
+        num_requests,
+        range_ratio,
+        input_len,
+        output_len,
+        tokenizer,
+    )
+    vocab_size = _get_vocab_size(tokenizer)
+    prohibited_tokens = getattr(tokenizer, "all_special_ids", []) or []
+    all_tokens = np.arange(vocab_size)
+    allowed_tokens = np.array(list(set(all_tokens) - set(prohibited_tokens)))
```
rogercloud: **Nit** — use `np.setdiff1d` instead of Python set subtraction

```python
# current
allowed_tokens = np.array(list(set(all_tokens) - set(prohibited_tokens)))
```

This round-trips through two Python sets for an O(vocab_size) allocation. The numpy idiom stays in the C layer:

```python
allowed_tokens = np.setdiff1d(all_tokens, list(prohibited_tokens))
```

`np.setdiff1d` returns sorted unique values from `all_tokens` not in `prohibited_tokens`, same semantics, no intermediate Python objects.

rogercloud (approved):
All three follow-up findings are addressed:

1. `benchmark_runner.py` — `logger.warning` added for non-streaming missing-usage fallback. ✓
2. `benchmark_serving.py` — `_validate_random_range_ratio` helper validates `[0, 1)` for both the scalar and JSON dict paths (also catches invalid `input`/`output` values individually). ✓
3. `utils.py` — `np.setdiff1d` replaces the Python set subtraction. ✓

LGTM.

PR #5039, opened 2026-06-16
test(test_cmdline):add try finally to terminal model

llyycchhee:
Add try finally to terminal model.


qinxuye:
Replaced by #5042

PR #5048, opened 2026-06-18
FEAT: prebuilt pypiserver mirror image as the offline compose package source

qinrui777:
## What this PR does

Ships `xprobe/xinference-pypiserver`: a prebuilt PyPI mirror image carrying every wheel the
xinference runtime may install into per-model virtual environments, and makes it the default
package source of the `offline` Docker Compose profile introduced in #5174. Air-gapped users
get working `vllm` / `sglang` / `transformers` / … model launches with zero wheel preparation.

This is a rework of the original approach in this PR. Testing showed that pointing
`PIP_INDEX_URL` / `PIP_TRUSTED_HOST` at the mirror does **not** reach the model-venv install
chain: per-model venv installs go through `uv` (which ignores `PIP_*` variables), and
xinference's pip-config inheritance (`get_pip_config_args`) only reads `global.*` keys from
pip config files, where `PIP_INDEX_URL` surfaces as `:env:.index-url` and is never inherited.
In a measured launch, `uv` installed packages while the mirror's TX byte counter stayed flat —
all traffic went to public PyPI. The compose/env-var half of this PR is therefore replaced by
the pip.conf + `UV_*` wiring that #5174 already merged (verified to reach `uv`:
`settings(index_url=http://…:8080/simple/, …)` in the worker log, mirror TX counter grows).

## Changes

- **Runtime: resolve direct wheel URLs from the private index** (`rewrite_direct_url_packages_for_index`).
  Direct-URL requirements (the `sgl_kernel` GitHub wheels) bypass any index, breaking sglang
  in air-gapped setups even when the wheel is mirrored. When `settings.index_url` is explicitly
  configured, such requirements are rewritten to `name==version` (keeping local segments, e.g.
  `sgl_kernel==0.3.21+cu130`) and resolved from the mirror. Online deployments are unchanged.
- **Runtime: drop support for CUDA below 13.0** in the virtualenv engine dependencies
  (`PYTORCH_CUDA_WHEEL_URLS` now cu130-only; the `sgl_kernel ; cuda_version < "13.0"` fallbacks
  are removed; the GLM-4.6/4.7 sgl_kernel URLs are aligned to v0.3.21 to match the code).
- **Image build pipeline** (`xinference/deploy/docker/pypiserver/`):
  - `generate_package_lists.py` loads `ENGINE_VIRTUALENV_PACKAGES` and the marker-filter logic
    straight from the xinference sources (no install), so the mirror cannot drift from what the
    runtime installs — this replaces the hand-maintained expansions in `list-packages.sh`.
  - `download_packages.py` locks the torch family to the runtime image's version
    (`TORCH_VERSION=2.11.*`), locks each engine set with `uv pip compile`, and fetches the
    fully-pinned locks with `pip download --no-deps` — one coherent resolution per engine
    instead of a per-line `pip download` loop that baked up to 12 versions of the same package.
    vllm/sglang pin their own torch stacks and are locked unconstrained (their venvs install a
    self-consistent stack). Per-model pins keep their exact versions. Git sources and
    sdist-only downloads are built into wheels so the runtime never compiles.
  - A `selfcheck` build stage re-resolves every engine set, model pin and direct wheel against
    the baked mirror **alone** (`--disable-fallback`); any gap fails the image build instead of
    failing an air-gapped user at model-launch time.
- **CI** (`build-pypiserver-image.yaml`): native amd64/arm64 builds merged into a multi-arch
  manifest. Pushes (including `:latest`) only happen on release tags of `xorbitsai/inference`,
  gated by `check-release-permission`; `workflow_dispatch` builds run the full download +
  selfcheck without publishing.
- **Compose/docs**: the offline profile defaults to the baked image (`--disable-fallback`);
  the previous stock-pypiserver + `./wheels` setup moves to a `docker-compose.byo-wheels.yml`
  override; docs updated accordingly.

## Verification

- `pytest xinference/core/tests/test_virtualenv_package_rewrite.py` — unit tests for the
  direct-URL rewrite (PEP 508 forms, markers, local versions, non-wheel passthrough).
- Generator run for both platforms: sglang set contains `sglang>=0.5.6` + `numpy>=2.4.1` with
  the arch-matching `sgl_kernel` wheel split into the direct-URL list; `httpx==0.24.0`
  correctly classified as a pin, not a URL.
- Full CI build (workflow_dispatch semantics, no publishing) on both architectures:
  amd64 562 files / 11.37 GiB, arm64 561 files / 11.32 GiB, and the offline selfcheck
  passed on both — all 6 engine sets, 150 model pins and 3 direct wheels resolve from the
  mirror alone ("mirror is self-sufficient"). ~20 legacy model pins took the recorded
  unconstrained-fallback path, as designed (old pins conflict with the modern locks;
  per-model venv isolation makes the extra versions legitimate).
  Locked engine stacks: vllm 0.25.1 (torch 2.11.0+cu130, matching the runtime image),
  sglang 0.5.9 + sgl-kernel 0.3.21 (torch 2.9.1+cu130, sglang's own pin),
  transformers/sentence_transformers aligned to torch 2.11.0+cu130.
- End-to-end on a 2×RTX 3090 Ti host (CUDA 13, Docker Compose v5.3), with the image built
  from this branch's pipeline and the runtime patched with this branch's code:
  - `docker compose --profile offline up -d`: worker log shows
    `settings(index_url=http://xinference-pypiserver:8080/simple, extra_index_url=..., trusted_host=...)`
    — pip.conf inheritance reaches the per-model uv installer, and the vLLM engine's
    hardcoded public extra indexes are overridden by the private one.
  - transformers engine (qwen3 0.6b): venv install traffic measured on the mirror
    container's NIC (+92 KiB for the forced extra wheels), inference OK.
  - sglang engine: the requirement list shows the GitHub direct wheel rewritten to
    `sgl_kernel==0.3.21+cu130`, resolved from the mirror; mirror NIC counter grew by
    ~390 MB during the venv install; sgl-kernel 0.3.21 + torch 2.11.0+cu130 landed in the
    venv. (Model load afterwards crashes on nightly-main because that image ships an
    incompatible sglang 0.5.6 / transformers 5.5.0 pair in its parent env — reproducible
    online, unrelated to this PR; will be reported separately.)
  - Air-gap override (`docker-compose.airgap.yml`, internal network, no external routing;
    `curl https://pypi.org` from the container fails): model launch + venv install +
    inference all succeed fully offline through the mirror.

Not run: the readthedocs render of the updated compose docs page.


docker-compose.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+  pypiserver:
+    image: xprobe/pypiserver:latest
+    container_name: pypiserver
+    profiles:
+      - pypiserver
+    ports:
+      - "8080:8080"
+    command: run
+    restart: unless-stopped
```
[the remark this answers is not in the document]
qinrui777: Tested, it looks good without `command`.

xinference/deploy/docker/pypiserver/Dockerfile.pypiserver
```diff
@@ -0,0 +1,75 @@
...
+               && echo "OK  $pkg" || echo "SKIP $pkg" ;; \
+           *) \
+             pip download --dest /data/packages/ --no-cache-dir \
+               --extra-index-url https://download.pytorch.org/whl/cu130 \
+               --extra-index-url https://wheels.vllm.ai/0.19.0/cu130 \
+               --extra-index-url https://xorbitsai.github.io/xllamacpp/whl/cu128 \
+               "$pkg" 2>/dev/null \
+               && echo "OK  $pkg" || echo "SKIP $pkg" ;; \
+         esac; \
+       done < packages.txt \
```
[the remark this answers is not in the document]
qinrui777: The `cu130` index is valid — both `download.pytorch.org/whl/cu130/` and `wheels.vllm.ai/0.19.0/cu130/` return 200 and contain real wheels. Changing to cu124 would actually break vllm (404). The silent skip pattern is intentional: the package list covers multiple platforms and optional deps — a single miss shouldn't fail the build.

xinference/deploy/docker/pypiserver/Dockerfile.pypiserver
```diff
@@ -0,0 +1,75 @@
...
+RUN pip cache purge \
+    && rm -rf /build
+
+# ============================================================================
+# Stage 2 — minimal pypiserver runtime
+# ============================================================================
+FROM pypiserver/pypiserver:latest
+
+COPY --from=downloader /data/packages/ /data/packages/
+RUN chown -R pypiserver:pypiserver /data/packages
```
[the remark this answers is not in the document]
qinrui777: Testing...

xinference/deploy/docker/pypiserver/list-packages.sh
```diff
@@ -0,0 +1,60 @@
...
+ALL=$(jq -r '
+  [ ..
+    | select(.virtualenv? | type == "object")
+    | .virtualenv.packages[]?
+  ] | unique[]
+' $FILES)
+
+# Step 3: filter by platform, exclude opposite arch entries
+# Step 4: strip conditional markers
+PACKAGES=$(echo "$ALL" | grep -v "platform_machine == \"$EXCLUDE_ARCH\"" | sed 's/ ; .*//')
```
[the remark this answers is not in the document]
qinrui777: Testing...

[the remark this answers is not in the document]
qinrui777: Test passed.

xinference/deploy/docker/pypiserver/Dockerfile.pypiserver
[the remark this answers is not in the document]
qinrui777: Test passed.

xinference/deploy/docker/pypiserver/Dockerfile.pypiserver
```diff
@@ -0,0 +1,75 @@
...
+             pip download --dest /data/packages/ --no-cache-dir \
+               "$pkg" 2>/dev/null \
+               && echo "OK  $pkg" || echo "SKIP $pkg" ;; \
+           *) \
+             pip download --dest /data/packages/ --no-cache-dir \
+               --extra-index-url https://download.pytorch.org/whl/cu130 \
+               --extra-index-url https://wheels.vllm.ai/0.19.0/cu130 \
+               --extra-index-url https://xorbitsai.github.io/xllamacpp/whl/cu128 \
+               "$pkg" 2>/dev/null \
+               && echo "OK  $pkg" || echo "SKIP $pkg" ;; \
```
qinxuye: This still makes the pypiserver image a best-effort mirror: every `pip download` failure is hidden by `2>/dev/null` and converted into `SKIP`, so the workflow can publish an image that is missing required wheels and only fails later in offline installs. Since this image is meant to support offline use, please collect failed packages and fail the build for required downloads, or otherwise make the best-effort behavior explicit and avoid publishing it as a complete offline mirror.

qinxuye (commented):
One issue remains around the offline package mirror build.

xinference/deploy/docker/pypiserver/Dockerfile.pypiserver
qinrui777: Fixed in latest commit. Build Result -> https://github.com/qinrui777/inference/actions/runs/28079279864

- **Stopped swallowing stderr** — removed `2>/dev/null` from both pip invocations, so failures are now visible in the build log.
- **Recorded failures** — each SKIP now appends the package name to `/tmp/failed.txt`, then cat the file in console.

.github/workflows/build-pypiserver-image.yaml
```diff
@@ -0,0 +1,142 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # ------------------------------------------------------------------
+  # Build Docker image per architecture on native runners.
+  # Each runner uses list-packages.sh to generate the package list,
+  # then pip download collects all wheels inside the container.
+  # ------------------------------------------------------------------
+  build-and-push-image:
```
rogercloud, line 44: This new publishing job runs on tag pushes and can push Docker Hub artifacts, but unlike `release.yaml` and `docker-cd.yaml` it does not depend on `.github/actions/check-release-permission`. That bypasses the existing guard that prevents non-admin releases from tags outside `main`, while still publishing `pypiserver:<tag>` and `latest`. Please add the same `check-release-permission` job and make the build/manifest jobs depend on it before pushing images.

xinference/deploy/docker/pypiserver/Dockerfile.pypiserver
```diff
@@ -0,0 +1,80 @@
...
+               --extra-index-url https://download.pytorch.org/whl/cu130 \
+               --extra-index-url https://wheels.vllm.ai/0.19.0/cu130 \
+               --extra-index-url https://xorbitsai.github.io/xllamacpp/whl/cu128 "$1" ;; \
+         esac; \
+       } \
+    && while IFS= read -r pkg; do \
+         fetch "$pkg" && echo "OK  $pkg" || { echo "SKIP $pkg"; echo "$pkg" >> /tmp/failed.txt; }; \
+       done < packages.txt \
+    && echo "=== Downloaded $(find /data/packages -name '*.whl' | wc -l) wheels ===" \
+    && if [ -s /tmp/failed.txt ]; then \
```
rogercloud: Recording failed downloads is useful, but this branch still lets the Docker build exit successfully when `/tmp/failed.txt` is non-empty. The workflow would then publish an incomplete offline package mirror, and users only discover the missing dependency later when launching a model without internet access. Please make this fail the build unless every failure is explicitly allowlisted as optional.

docker-compose.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+  pypiserver:
+    image: xprobe/pypiserver:latest
+    container_name: pypiserver
+    profiles:
+      - pypiserver
+    ports:
+      - "8080:8080"
+    command: run
```
rogercloud: The README describes this profile as resolving packages from the local PyPI server instead of the public internet, but pypiserver's default `run` mode can fall back/redirect to PyPI for packages missing locally. That can hide an incomplete mirror on connected machines and fail late in true offline deployments. Please disable fallback here, e.g. `run --disable-fallback /data/packages`.

rogercloud (changes requested):
Requesting changes for the release gating and offline package integrity issues below.

xinference/deploy/docker/pypiserver/Dockerfile.pypiserver
```diff
@@ -0,0 +1,80 @@
...
+#   docker build -f xinference/deploy/docker/pypiserver/Dockerfile.pypiserver \
+#     --build-arg PLATFORM=amd64 -t pypiserver:amd64 .
+#
+#   docker build -f xinference/deploy/docker/pypiserver/Dockerfile.pypiserver \
+#     --build-arg PLATFORM=arm64 -t pypiserver:arm64 .
+
+# ============================================================================
+# Stage 1 — download every wheel needed by xinference model engines
+# ============================================================================
+FROM python:3.12-slim AS downloader
```
rogercloud: **[Updated — downgraded to a minor note.]** Downloader Python should track the runtime image.

I first read the in-repo `Dockerfile` (base `vllm/vllm-openai:v0.17.1`, Python 3.10) and worried this `python:3.12` downloader would produce wheels that can't be installed at runtime. Inspecting the actually-published `xprobe/xinference:latest` config, however, it is built on a newer vLLM base with `PYTHON_VERSION=3.12` and `CUDA_VERSION=13.0.2`. So today the downloader (3.12) matches the runtime interpreter and there is **no ABI mismatch** — and the `cu130` extra indexes are correct for this image.

The only residual concern is forward-looking: this `python:3.12-slim` is hardcoded and decoupled from whatever Python the xinference base image actually ships. Because model venvs are created with `python_path=sys.executable` (`xinference/core/worker.py`), a future base-image Python bump would silently make the mirrored wheels uninstallable, and the failure would only surface in an offline install. Consider deriving/pinning the downloader Python (and CUDA tag) from the target image, or documenting the coupling, so they can't drift. Not a blocker.

docker-compose.yaml
```diff
@@ -0,0 +1,32 @@
+services:
+  xinference:
+    image: xprobe/xinference:latest
+    container_name: xinference
+    ports:
+      - "9997:9997"
+    environment:
+      # - XINFERENCE_MODEL_SRC=modelscope
+      - XINFERENCE_HOME=/data
+      - PIP_INDEX_URL=${PIP_INDEX_URL:-https://pypi.org/simple}
```
rogercloud: **Offline mode is incomplete for vllm/sglang.** Pointing `PIP_INDEX_URL` at the local pypiserver only redirects pip's *primary* index. When xinference bootstraps a vllm/sglang model venv it injects engine-specific extra indexes from `ENGINE_VIRTUALENV_EXTRA_INDEX_URLS` (`apply_engine_virtualenv_settings` in `xinference/core/utils.py`) — `https://wheels.vllm.ai/0.19.0/cu130` and `https://download.pytorch.org/whl/cu130`, with `index_strategy=unsafe-best-match`. These are passed to uv as explicit `--extra-index-url` args, so in a true air-gapped deployment those hosts are unreachable and venv creation for the heavy GPU engines fails despite the local mirror.

**Note:** setting a `PIP_EXTRA_INDEX_URL` *environment variable* does **not** help — `get_pip_config_args()` (`xinference/utils.py`) reads `pip config list` (config files only), not env vars, so the env var never reaches `settings` and the engine injection still happens.

**Suggested fix (no xinference code change).** `inherit_pip_config` (`xinference/core/worker.py`, ~L1707) runs *before* `apply_engine_virtualenv_settings` (~L1714), and the engine injection only fires when `settings.extra_index_url is None`. So provide a pip **config file** in the container (not an env var) that points both indexes at the local mirror:

```ini
# /etc/pip.conf
[global]
index-url = http://pypiserver:8080/simple/
extra-index-url = http://pypiserver:8080/simple/
trusted-host = pypiserver
```

Then `pip config list` -> `get_pip_config_args()` -> `inherit_pip_config` sets `settings.extra_index_url` to the mirror (non-`None`), so `apply_engine_virtualenv_settings` skips injecting the `wheels.vllm.ai` / `download.pytorch.org` internet indexes, and venv creation resolves entirely from pypiserver (the mirror already holds those cu130 wheels by name). Mount this file (e.g. to `/etc/pip.conf`, or via `PIP_CONFIG_FILE`) in the offline profile.

**Two caveats:**
1. This does **not** fix sglang's `sgl_kernel`: it is referenced as a direct GitHub URL at venv-creation time (not an index lookup), so a name-indexed pypiserver cannot serve it and pip will still try to reach `github.com`. That needs xinference core to express such deps as index-resolvable names (or mirror + rewrite). So the config-file fix unblocks **vllm**, not **sglang**.
2. The mechanism is implicit (relies on the inherit-before-apply ordering and the "only inject when `None`" guard). A cleaner long-term fix is an explicit offline switch in xinference that disables engine extra-index injection — out of scope for this PR.

xinference/deploy/docker/pypiserver/list-packages.sh
```diff
@@ -0,0 +1,65 @@
...
+    '#system_torchaudio#') printf 'torchaudio\n' ;;
+    '#system_torchvision#') printf 'torchvision\n' ;;
+    '#system_pandas#')    printf 'pandas\n' ;;
+
+    '#transformers_dependencies#'|transformers_dependencies)
+      printf 'transformers>=4.53.3\naccelerate>=0.28.0\n' ;;
+    '#vllm_dependencies#'|vllm_dependencies)
+      printf 'vllm>=0.11.2\n' ;;
+    '#sglang_dependencies#'|sglang_dependencies)
+      printf 'pybase64\nzmq\npartial_json_parser\nsentencepiece\ndill\nninja\nnumpy>=2.4.1\nsglang>=0.5.6\nsgl_kernel\n' ;;
```
rogercloud: **Duplicated source of truth (already drifting).** This script — and the `--extra-index-url` list in `Dockerfile.pypiserver` — is a hand-maintained copy of `ENGINE_VIRTUALENV_PACKAGES` / `ENGINE_VIRTUALENV_EXTRA_INDEX_URLS` from `xinference/core/virtual_env_manager.py`. Two sources of truth will drift, and the air-gapped install is the worst place to discover it.

**Concrete drift today:**

1. **sglang `sgl_kernel` (this line).** The source defines three entries — two pinned CUDA-13.0 direct-URL wheels (`sgl_kernel-0.3.21+cu130-cp310-abi3-...x86_64.whl` / `...aarch64.whl`, gated by `cuda_version == "13.0" and platform_machine == ...`) plus a `sgl_kernel ; cuda_version < "13.0"` fallback. This line collapses all of that to a bare `sgl_kernel`. Since the published runtime image is CUDA 13.0.2, the runtime resolves to the pinned GitHub URL wheel, not PyPI `sgl_kernel` — so the mirror downloads the wrong artifact. (And see the note below: even mirroring the right wheel doesn't make it servable.)

2. **Extra indexes: global vs per-engine, plus an out-of-band source.** The source applies extra indexes per engine (`vllm → [wheels.vllm.ai/0.19.0/cu130, download.pytorch.org/whl/cu130]`, `sglang → [download.pytorch.org/whl/cu130]`) with `index_strategy = unsafe-best-match`. The Dockerfile instead applies `download.pytorch.org/whl/cu130 + wheels.vllm.ai/0.19.0/cu130 + xorbitsai.github.io/xllamacpp/whl/cu128` to *every* package and passes no index strategy. Note (a) `xorbitsai.github.io/xllamacpp/whl/cu128` is **not** in the source at all and its `cu128` disagrees with the `cu130` used everywhere else, and (b) a transformers-/diffusers-only model can now pull a cu130-tagged torch into the mirror.

3. **Version specs are a manual copy.** `vllm>=0.11.2`, `transformers>=4.53.3`, etc. happen to match the source right now, but they are hand-transcribed; the next bump in `ENGINE_VIRTUALENV_PACKAGES` silently leaves the mirror behind, surfacing only during an offline install.

**Suggested fix — generate from the source instead of re-encoding in bash.** Replace the hardcoded placeholder/index logic with a small Python helper that imports and reuses the real definitions, so the list is exactly what the runtime computes:

```python
from xinference.core.virtual_env_manager import (
    ENGINE_VIRTUALENV_PACKAGES,
    ENGINE_VIRTUALENV_EXTRA_INDEX_URLS,
    expand_engine_dependency_placeholders,
    filter_virtualenv_packages_by_markers,
)
# for each engine: expand placeholders, then
# filter_virtualenv_packages_by_markers(expanded, engine, cuda_version="13.0")
```

This makes the package list (incl. the sgl_kernel URLs) and the per-engine extra indexes track the source automatically, killing all three drifts above. The downloader stage would need the module importable — e.g. `COPY . /src` + `pip install --no-deps -e /src` (please confirm the import surface stays light enough to avoid pulling heavy deps).

**Caveat (related to the offline concern on `docker-compose.yaml`).** Fixing the drift still won't make sglang work offline: the `sgl_kernel` cu130 dependency is referenced by a direct GitHub URL at venv-creation time, and a name-indexed pypiserver cannot serve a direct-URL wheel — the runtime will still try to reach `github.com`. Truly supporting these engines offline needs xinference core to express such deps as index-resolvable names (or to mirror + rewrite them), which is beyond this script.

rogercloud (commented):
Three design-level concerns about whether the offline mirror is actually usable in a true air-gapped deployment. These are independent of the silent-failure / release-gating points already raised.

docker-compose.yaml
qinrui777: Added.

xinference/deploy/docker/pypiserver/Dockerfile.pypiserver
OliverBryant: Addressed in the reworked pipeline: the download loop is gone entirely. Any engine-set lock failure, pin download failure (after a recorded unconstrained retry), direct-wheel or git-wheel failure now exits the build with FATAL — nothing is skipped silently. On top of that, a dedicated selfcheck build stage re-resolves every engine set, model pin and direct wheel against the baked mirror alone (pypiserver started with --disable-fallback) and fails the image build on any gap, so a published image is guaranteed to be a complete offline mirror.

xinference/deploy/docker/pypiserver/Dockerfile.pypiserver
OliverBryant: Made explicit in the rework: PYTHON_VERSION (default 3.12) and TORCH_VERSION (default 2.11.*) are build args documented to track the xprobe/xinference runtime image, and the torch family is locked to that version before the engine sets are resolved, so the mirror carries exactly the interpreter/torch combination the runtime venvs will request.

docker-compose.yaml
OliverBryant: Confirmed and reworked accordingly — measured on a live launch, PIP_INDEX_URL never reached the per-model uv installer (mirror NIC counter stayed flat while uv installed from public PyPI). The env-var approach is dropped; the offline profile now rides the pip.conf + UV_* wiring merged in #5174, whose extra-index-url also overrides the hardcoded vLLM/SGLang public indexes. Verified end-to-end on GPU: worker log shows settings(index_url=http://xinference-pypiserver:8080/simple, extra_index_url=..., trusted_host=...), and an sglang venv install pulled ~390 MB through the mirror NIC, with a fully air-gapped (internal Docker network) launch + inference succeeding.

xinference/deploy/docker/pypiserver/list-packages.sh
OliverBryant: Fixed at the root: list-packages.sh is deleted. generate_package_lists.py loads ENGINE_VIRTUALENV_PACKAGES, the engine extra-index tables and filter_virtualenv_packages_by_markers directly from the xinference sources at build time, so the mirror cannot drift from the runtime definitions. The sgl_kernel direct-URL wheels are downloaded into the mirror, and a new runtime rewrite (rewrite_direct_url_packages_for_index) resolves direct-URL requirements as name==version from the private index when one is configured — verified in the e2e log: sgl_kernel==0.3.21+cu130 installed from the mirror.

xinference/core/worker.py
```diff
@@ -2647,6 +2648,13 @@ def _prepare_virtual_env(
             packages, model_engine, cuda_version
         )
 
+        if settings.index_url:
```
rogercloud: **[MAJOR] Finding A — Rewrite gate breaks online installs for any user with a pip mirror configured**

This gate is a bare `if settings.index_url:`, which cannot distinguish "an offline mirror was explicitly configured for this feature" from "the user merely has a normal pip mirror in `pip.conf`" (Tsinghua/Aliyun/USTC, or a corporate Nexus/Artifactory proxy).

`inherit_pip_config` defaults to `True` (`xinference/model/core.py:136`) and unconditionally copies `pip config list`'s `global.index-url` into `settings.index_url` for every model install (`worker.py:2548-2553` via `get_pip_config_args()` in `xinference/utils.py:42-81`) — online or not.

Concrete failure: a fully-online user with a standard China mirror (or any corporate pip proxy) launches sglang. Pre-PR, the `sgl_kernel` GitHub release wheel (`xinference/core/virtual_env_manager.py:36-37`, carrying a `+cu130` local-version segment) is fetched directly and works. Post-PR, `rewrite_direct_url_packages_for_index` (`xinference/core/utils.py:427`) rewrites it to `sgl_kernel==0.3.21+cu130`, resolved against the user's PyPI-proxy mirror — but PyPI disallows local-version-segment uploads, so no PyPI-compatible mirror can ever serve this. Per the comment at `worker.py:2604-2611`, uv treats an unreachable/unsatisfiable index entry as fatal, not a soft fallback to the original URL. Net: an install that worked before this PR now hard-fails for these users.

**Suggested fix:** gate the rewrite on an explicit offline/mirror-only signal (e.g. a distinct flag set by the docker offline profile) rather than the mere presence of `settings.index_url`.

xinference/deploy/docker/pypiserver/selfcheck.py
```diff
@@ -0,0 +1,155 @@
...
+            for line in in_file.read_text().splitlines()
+            if line.strip() and not line.startswith("#")
+        ]
+        if requirements:
+            gate(f"engine:{in_file.stem}", requirements)
+
+    for entry in json.loads((args.manifest_dir / "pins.json").read_text()):
+        gate(f"pin:{entry['spec']}", [entry["spec"]])
+
+    for url in (args.manifest_dir / "urls.txt").read_text().splitlines():
```
rogercloud, line 143: **[MAJOR] Finding B — Git-sourced package dependencies are excluded from the selfcheck gate**

`selfcheck.py`'s `main()` only re-resolves `engines/*.in` (line 124), `pins.json` (line 133), and `urls.txt` wheel URLs via `wheel_url_to_spec` (line 136) — it never reads or checks any git-sourced requirements list, even though `download_packages.py:304-365` does build git-sourced deps into wheels and bakes them into the mirror (reading `git.txt`).

`rewrite_direct_url_packages_for_index` in `xinference/core/utils.py` intentionally leaves `git+`/`name @ git+` specs unchanged (documented in its own docstring). At runtime, `worker.py:2651-2676` passes the raw git spec straight to uv/pip; an explicit git/URL requirement is always fetched from that URL directly, regardless of a configured index. These specs are live in shipped model specs, not just test fixtures: `xinference/model/image/model_spec.json:1234,1679,1714` and `xinference/model/llm/llm_family.json:1854,2012`.

So an air-gapped launch of such a model passes this build-time selfcheck, then fails at actual model-launch time trying to `git clone` from GitHub with no network — exactly the failure this selfcheck mechanism exists to catch before it reaches a user, silently missed for this one category.

**Suggested fix:** add a git-source check here (realistically: verify the wheel `download_packages.py` built for each git dep exists and installs standalone, since a live git-ref reachability check is impossible offline), or explicitly document git-sourced-dependency models as unsupported by the offline profile.

xinference/core/virtual_env_manager.py
```diff
@@ -35,7 +35,6 @@
         "sglang>=0.5.6",
         'https://github.com/sgl-project/whl/releases/download/v0.3.21/sgl_kernel-0.3.21+cu130-cp310-abi3-manylinux2014_x86_64.whl ; cuda_version == "13.0" and platform_machine == "x86_64"',
         'https://github.com/sgl-project/whl/releases/download/v0.3.21/sgl_kernel-0.3.21+cu130-cp310-abi3-manylinux2014_aarch64.whl ; cuda_version == "13.0" and platform_machine == "aarch64"',
```
rogercloud, line 37: **[MAJOR] Finding C — CUDA <13.0 drop applies unconditionally to the online path, with silent degradation**

Removing the `sgl_kernel ; cuda_version < "13.0"` fallback here (and dropping `cu128`/`cu129` from `PYTORCH_CUDA_WHEEL_URLS` below at line 84) is applied unconditionally in `_prepare_virtual_env` (`xinference/core/worker.py:2574-2650`) — not gated to offline/mirror mode.

Two distinct effects:
1. Torch-index part: a CUDA 12.8/12.9 online user now gets `PYTORCH_CUDA_WHEEL_URLS.get(cuda_version)` → `None` → no auto extra-index-url — same as any previously-unmapped CUDA version (cu121 was already unmapped). This is a reduction in supported versions, not a new failure mode.
2. sgl_kernel part (sharper regression): with the `< "13.0"` fallback gone, a CUDA 12.x online sglang install now resolves **zero** `sgl_kernel` entries (both remaining entries require `cuda_version == "13.0"`), silently, with no install-time error — the failure only surfaces later as a runtime import/missing-kernel error inside sglang.

The commit that made this change frames it as an offline-mirror wheel-matrix concern, but the code affects every deployment including fully online ones, with no breaking-change surfacing beyond a docker-compose docs note.

**Suggested fix:** document this as a product-level breaking change (release notes / CLI warning), and consider raising an explicit, actionable error when sgl_kernel is unavailable for CUDA<13.0 instead of silently installing sglang without its kernel.

xinference/core/tests/test_virtualenv_package_rewrite.py
```diff
@@ -0,0 +1,105 @@
...
+        # too few dash-separated fields to be a valid wheel filename
+        "https://example.com/invalid.whl",
+        "https://example.com/pkg-none-any.whl",
+        # version field does not start with a digit
+        "https://example.com/pkg-vNext-py3-none-any.whl",
+    ]
+    assert rewrite_direct_url_packages_for_index(packages) == packages
+
+
+def test_filter_then_rewrite_sglang_cu130():
```
rogercloud, line 103: **[MINOR] Finding D — No test coverage for the new CUDA-version-marker filtering behavior**

This is the only marker-filtering test (`test_filter_then_rewrite_sglang_cu130`), and it hardcodes `cuda_version="13.0"`. No test anywhere in the repo exercises `filter_virtualenv_packages_by_markers` with a CUDA 12.x or 13.1+ value, so the behavior change from Finding C (sgl_kernel silently dropped for CUDA<13.0) has zero regression coverage.

**Suggested fix:** add a test asserting that for `cuda_version="12.4"` (or similar) the sglang package list contains no `sgl_kernel` entry — documents the intentional change and guards the marker-evaluation logic against future regressions.

xinference/deploy/docker/pypiserver/selfcheck.py
```diff
@@ -0,0 +1,155 @@
...
+    while time.time() < deadline:
+        try:
+            urllib.request.urlopen(f"{base}/health", timeout=3)
+            return
+        except Exception:
+            time.sleep(1)
+    sys.exit(f"FATAL: pypiserver at {index_url} did not become healthy")
+
+
+def wheel_url_to_spec(url: str) -> Optional[str]:
```
rogercloud, line 53: **[MINOR] Finding E — New image-build pipeline scripts have no unit test coverage**

`selfcheck.py` (155 lines), `download_packages.py` (427 lines), and `generate_package_lists.py` (312 lines) are all new with non-trivial branching/parsing logic (lock/resolve/fallback handling, `platform.machine` monkeypatching for cross-arch resolution). No test file anywhere imports or exercises any of these three scripts.

`wheel_url_to_spec` here is independently written from the tested `rewrite_direct_url_packages_for_index` in `xinference/core/utils.py`, so the existing unit tests give only indirect, not-guaranteed confidence in this parser. `build-pypiserver-image.yaml:32` sets a 180-minute timeout on the only job that exercises these scripts, and it triggers only on tag-push/workflow_dispatch, not on ordinary PRs.

**Suggested fix (non-blocking follow-up):** add fast unit tests for `generate_package_lists.py`'s classification logic and this file's `wheel_url_to_spec`, independent of a full image build.

xinference/deploy/docker/pypiserver/download_packages.py
```diff
@@ -0,0 +1,427 @@
...
+            print(f"WARN: keeping sdist {sdist.name}", flush=True)
+            sdist_left.append(sdist.name)
+
+    # ------------------------------------------------------------------
+    # Report.
+    # ------------------------------------------------------------------
+    versions: Dict[str, Set[str]] = defaultdict(set)
+    total_size = 0
+    files = sorted(p for p in dest.iterdir() if p.is_file())
+    wheel_re = re.compile(r"^([A-Za-z0-9_.]+)-([0-9][^-]*)-")
```
rogercloud: **shrink:** hand-rolled `wheel_re` regex reinvents wheel-filename parsing (used only for the `multi_version_packages` report stat). This file already imports and uses `packaging.utils.canonicalize_name` on the same names for the same purpose, so canonicalization is already the expected behavior here — replace with `packaging.utils.parse_wheel_filename`, adding a `.whl`-suffix check plus `try/except InvalidWheelFilename` to preserve the current silent-skip-on-non-wheel behavior (e.g. leftover `.tar.gz` sdist artifacts in the same directory). `net: -3 to -5 lines possible`.

rogercloud (commented):
## Summary

This PR hardens air-gapped deployments against silent egress to public PyPI/GitHub during per-model venv installs. It rewrites direct wheel-URL requirements (e.g. `sgl_kernel` GitHub-release wheels) into `name==version` specs so they resolve from a configured index, ships a prebuilt `xprobe/xinference-pypiserver` mirror image whose package lists are generated directly from xinference source (a solid anti-drift choice), and adds a build-time `selfcheck` gate that fails the image build if the mirror cannot self-sufficiently resolve everything. It also trims the CUDA wheel matrix (drops `cu128`/`cu129` torch index entries and the `sgl_kernel ; cuda_version < "13.0"` fallback) to bound the mirror's size.

## Approach verdict: acceptable-with-reservations

The mirror-image + build-time selfcheck gate is a sound architecture, and generating the package lists straight from xinference source rather than a hand-maintained copy is genuinely good design. However, three changes that are motivated as offline-only concerns are actually wired into the unconditional / already-online install path, and the most important one (Finding A below) is a real regression for a large, common class of online users — recommend fixing that before merge.

Findings B and C are also MAJOR; D and E are MINOR test-coverage gaps. See inline comments for details, and the Simplification note at the bottom.

## Verification performed

- Ran `xinference/core/tests/test_virtualenv_package_rewrite.py`: **7/7 pass**.
- Full editable install with dev deps succeeded in a review worktree; no other test files exercise the changed code paths directly.
- Did not run the Docker image build itself (180-min CI-only step requiring registry access) — relying on the author's reported CI results for that portion.
- Coverage gaps: the three new pypiserver build scripts (`selfcheck.py`, `download_packages.py`, `generate_package_lists.py`) have no unit tests (Finding E); the CUDA-marker behavior change from dropping CUDA<13.0 has no regression test (Finding D).

## Simplification opportunity

`xinference/deploy/docker/pypiserver/download_packages.py:402` — the hand-rolled `wheel_re` regex (used only for the `multi_version_packages` report stat) can be replaced with `packaging.utils.parse_wheel_filename`. This file already imports and uses `packaging.utils.canonicalize_name` on the same names for the same purpose, so canonicalization is already the expected behavior here (no behavior mismatch), and the risk is confined to a reporting statistic. Needs a `.whl`-suffix check plus `try/except InvalidWheelFilename` to preserve the current silent-skip-on-non-wheel behavior (e.g. leftover `.tar.gz` sdist artifacts in the same directory).

`net: -3 to -5 lines possible`

(Two similar suggestions — applying the same swap to `rewrite_direct_url_packages_for_index` in `utils.py:427` and to `wheel_url_to_spec` in `selfcheck.py:50` — were considered and dropped: `parse_wheel_filename` PEP-503-canonicalizes names (`sgl_kernel` → `sgl-kernel`), which would break 5+ existing tests asserting underscore-preserving output, and raises on unparsable filenames rather than returning `None`.)

## Out of scope (verified but not this PR's issue)

CI publishes `:latest` on any tag push to `main`, not only version-format tags (`check-release-permission` only checks who/where, never the tag's format). Confirmed real, but this is a pre-existing repo-wide convention also present in `.github/workflows/docker-cd.yaml` — not introduced by this PR.

xinference/deploy/docker/pypiserver/selfcheck.py
OliverBryant: Addressed using the explicitly-supported alternative from this finding. After supported wheel URLs are rewritten, explicit offline mode now detects any remaining git/non-wheel direct reference and fails before attempting egress. `selfcheck.py` reads `git.txt` and reports those entries as unsupported, and the Compose/README docs list the current affected model paths and custom-image workaround. The mirror claims were narrowed to index-compatible content. Covered by the direct-reference and worker fail-fast tests.

xinference/core/virtual_env_manager.py
OliverBryant: Addressed in a4e5a0d6d by restoring the pre-existing online/non-Docker behavior: the CUDA <13 `sgl_kernel` fallback and the cu128/cu129 PyTorch index mappings are back. Only the prebuilt offline mirror remains cu130-targeted, which is now stated explicitly in the docs; the mirror-size decision no longer changes global runtime support.

xinference/core/worker.py
OliverBryant: Addressed in a4e5a0d6d. Direct-wheel rewriting is now gated by the explicit `XINFERENCE_VIRTUAL_ENV_OFFLINE_INSTALL` flag (default off), and only `offline.env.example` enables it. A normal inherited Tsinghua/corporate pip mirror therefore keeps the original direct URL. Added regression tests for both the normal-mirror and explicit-offline paths.

xinference/core/tests/test_virtualenv_package_rewrite.py
OliverBryant: Covered by `test_filter_sglang_keeps_cuda_12_fallback`. Because Finding C was fixed by restoring CUDA 12 online support instead of accepting its removal, the regression assertion intentionally verifies that CUDA 12.8 retains `sgl_kernel` and excludes the cu130 direct wheel. The existing cu130 test continues to cover the mirror path.

xinference/deploy/docker/pypiserver/selfcheck.py
OliverBryant: Addressed in `xinference/deploy/docker/pypiserver/tests/test_scripts.py`. Fast tests now import and exercise all three new scripts: generator classification/system placeholders, selfcheck wheel URL parsing (including invalid/non-wheel inputs), and downloader wheel filename parsing.

xinference/deploy/docker/pypiserver/download_packages.py
OliverBryant: Addressed in a4e5a0d6d. The reporting loop now uses `packaging.utils.parse_wheel_filename` through `wheel_name_version`, with a `.whl` suffix guard and `InvalidWheelFilename` fallback. Tests cover normalized names plus non-wheel and invalid filenames.

qinxuye (commented):
One remaining release-workflow issue.

.github/workflows/build-pypiserver-image.yaml
```diff
@@ -0,0 +1,150 @@
...
+      - name: Determine image tag
+        shell: bash
+        env:
+          MANUAL_IMAGE_TAG: ${{ inputs.image_tag }}
+        run: |
+          if [[ "$GITHUB_REF" =~ ^"refs/tags/" ]]; then
+            IMAGE_TAG="${GITHUB_REF#refs/tags/}"
+            echo "Will handle tag $IMAGE_TAG"
+          else
+            IMAGE_TAG="$MANUAL_IMAGE_TAG"
```
qinxuye, line 93: Manual dispatch can still overwrite `:latest`: setting `image_tag=latest` and `push_to_dockerhub=true` makes this branch use `IMAGE_TAG=latest`, and `merge-manifest` then pushes `xprobe/xinference-pypiserver:latest`. This contradicts the README statement that manual runs never replace `:latest` and lets a manual build replace the release pointer. Please reject `latest` for `workflow_dispatch` (or otherwise reserve it for tag-triggered releases).

OliverBryant: Fixed in f2e3b543f. Manual dispatch now rejects the exact `latest` tag before the build starts, while tag-triggered releases retain permission to publish `:latest`. The workflow input description also makes this reservation explicit. Validated with actionlint, pre-commit, and git diff --check.

qinxuye (approved):
LGTM

xinference/core/worker.py
rogercloud: Finding A — **FIXED**. The bare `if settings.index_url:` gate is gone; the rewrite is now gated behind `XINFERENCE_VIRTUAL_ENV_OFFLINE_INSTALL` (worker.py:2769-2779), which defaults to `"0"` (constants.py:355-357) and also raises if the flag is on but no `index_url` is configured. A normal online user with an inherited pip mirror keeps the original direct URL. Both paths are covered by new regression tests, and the 9 existing rewrite-function tests still pass. Clean, well-scoped, well-tested fix.

xinference/deploy/docker/pypiserver/selfcheck.py
rogercloud: Finding B — **FIXED**. selfcheck.py now reports `git.txt` entries as UNSUPPORTED (informational), and more importantly a real runtime guard (`find_direct_reference_packages()` in utils.py) makes the offline path raise `ValueError` before install if any git+/direct reference survives the rewrite step (worker.py ~2780). Covered by `test_prepare_virtual_env_offline_mirror_rejects_git_source` using the real `diffusers @ git+...` spec. Replaces silent launch-time failure with an explicit error.

xinference/core/virtual_env_manager.py
rogercloud: Finding C — **FIXED**. `PYTORCH_CUDA_WHEEL_URLS` again carries cu128/cu129/cu130 (virtual_env_manager.py:102-106) and the `sgl_kernel ; cuda_version < "13.0"` fallback is restored (virtual_env_manager.py:38). The mirror image remains intentionally cu130-only, now explicitly documented in using_docker_compose.rst. No conflict between restored online capability and the narrower offline mirror scope.

xinference/core/tests/test_virtualenv_package_rewrite.py
rogercloud: Finding D — **FIXED**. `test_filter_sglang_keeps_cuda_12_fallback` asserts that for `cuda_version="12.8"` the non-cu130 `sgl_kernel` fallback is present and no `+cu130` entry appears — genuinely exercising the previously-untested non-"13.0" marker path.

xinference/deploy/docker/pypiserver/selfcheck.py
rogercloud: Finding E — **PARTIAL**. New `pypiserver/tests/test_scripts.py` covers the riskiest string-parsing helpers across all three scripts (`classify_spec`, `system_placeholder_name`, `wheel_url_to_spec`, `wheel_name_version`) including invalid/non-wheel edge cases — real, meaningful coverage. But the orchestration logic (`download_packages.py`'s 4-pass download/lock/prune/self-pin flow, `selfcheck.py`'s `gate()`/`compile_against_index`, `generate_package_lists.py`'s `main()`/`load_xinference_modules`/`iter_virtualenv_packages`) remains untested; the ~180-min CI image build is still the only safety net for it.

xinference/deploy/docker/pypiserver/download_packages.py
rogercloud: Simplification — **FIXED**. `wheel_name_version()` now uses `packaging.utils.parse_wheel_filename` + `canonicalize_name` with a `.whl` guard and `InvalidWheelFilename` handling — behavior-preserving and tested. Note the net change was +14 lines (not the estimated -3 to -5) because it was extracted into a standalone, testable function with a docstring — a reasonable design choice enabling the new coverage, not a line-count win.

xinference/deploy/docker/pypiserver/download_packages.py
```diff
@@ -0,0 +1,441 @@
...
+        pruned = work / "constraints-pruned.txt"
+        pruned.write_text(
+            "".join(f"{line}\n" for n, line in merged.items() if n != src_name)
+        )
+        proc = pip_wheel(src, pruned)
+        if proc.returncode != 0:
+            print(f"WARN: retrying '{src}' without constraints", flush=True)
+            proc = pip_wheel(src, None)
+            if proc.returncode != 0:
+                sys.exit(f"FATAL: failed to build wheel for '{src}'")
```
rogercloud: **[MINOR] F1 — Git-source dependency handling is inconsistent across build/selfcheck/runtime**

This `sys.exit("FATAL: ...")` hard-aborts the *entire* image build if a git-wheel build fails (even on retry). But per the offline runtime guard added for Finding B, any spec containing a `git+` reference is unconditionally rejected before install in offline mode (`worker.py:2769-2786` + `find_direct_reference_packages` in `utils.py`) — there's no substitution mechanism that lets the runtime pick up this pre-built wheel instead. So git-sourced wheels built here can never actually be used offline, yet a build failure for one of them still aborts the whole mirror image.

Not a functional break (the runtime's rejection is deliberate and correct) — just build-time waste plus a fragile extra failure mode for wheels that were never going to be usable anyway.

**Suggested fix:** either skip building git-sourced wheels for the offline mirror entirely (they're provably unusable), or make git-wheel build failures non-fatal (warn + continue), matching the sdist-build soft-failure pattern already used a few lines above (~386-408) for the same reason.

xinference/core/worker.py
```diff
@@ -2724,7 +2727,11 @@ def _prepare_virtual_env(
             # so a CPU-only host with a CUDA-built torch would otherwise get a
             # GPU wheel that fails to import (missing libcuda.so.1). In that
             # case keep the default CPU index and behavior.
-            if xllamacpp_index_url and cls._is_cuda_device_available():
+            if (
+                not XINFERENCE_VIRTUAL_ENV_OFFLINE_INSTALL
```
rogercloud: **[MINOR] F2 — Offline mode silently disables xllamacpp's GPU-specific wheel selection**

This `not XINFERENCE_VIRTUAL_ENV_OFFLINE_INSTALL` guard (added by this PR) skips the GPU-index switch with no `else` branch logging why. The offline mirror's package generator never references xllamacpp's CUDA index URLs, so the mirror only ever carries the CPU build from PyPI — and none of the docs (`using_docker_compose.rst`, `pypiserver/README.md`, `offline.env.example`) mention llama.cpp/xllamacpp at all.

This is a defensible trade-off (the GPU index would be unreachable offline anyway, and the mirror doesn't carry that wheel), but a GPU host running the llama.cpp engine fully offline silently gets a CPU-only build with degraded performance and zero warning.

**Suggested fix:** add a `logger.info`/`logger.warning` when this block is skipped due to offline mode, plus a doc note alongside the existing "mirror's GPU stack targets CUDA 13.0" callout.

xinference/deploy/docker/pypiserver/tests/test_scripts.py
```diff
@@ -0,0 +1,63 @@
...
+
+def _load_script(name: str) -> ModuleType:
+    spec = importlib.util.spec_from_file_location(name, SCRIPT_DIR / f"{name}.py")
+    assert spec is not None and spec.loader is not None
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+def test_generate_package_list_classification():
```
rogercloud, line 33: **[MINOR] F3 — New tests cover only string-parsing helpers, not orchestration logic**

This file's three tests (`test_generate_package_list_classification`, `test_selfcheck_wheel_url_to_spec`, `test_download_report_wheel_filename_parsing`) exercise the real parsing/classification helpers from all three build scripts, including edge cases — genuinely useful coverage. But none of them reach `download_packages.py`'s actual 4-pass download/lock/prune/self-pin orchestration, `selfcheck.py`'s `gate()`/`compile_against_index`, or `generate_package_lists.py`'s `main()`/`load_xinference_modules`/`iter_virtualenv_packages`. That logic remains covered only by the ~180-minute CI image build.

Not blocking — flagging so the gap is tracked rather than assumed closed by this file's existence.

xinference/core/tests/test_worker.py
```diff
@@ -850,6 +850,59 @@ def test_prepare_virtual_env_inherit_pip_config(monkeypatch):
     assert kwargs["index_url"] == "https://example.invalid/simple"
 
 
+def test_prepare_virtual_env_normal_pip_mirror_keeps_direct_wheel(monkeypatch):
```
rogercloud, line 853: **[MINOR] F4 — Offline-install tests never exercise the real sglang engine-dispatch path**

All three offline-install tests here (`test_prepare_virtual_env_normal_pip_mirror_keeps_direct_wheel`, `test_prepare_virtual_env_offline_mirror_rewrites_direct_wheel`, `test_prepare_virtual_env_offline_mirror_rejects_git_source`) pass `model_engine=None`. That short-circuits `apply_engine_virtualenv_settings`/`expand_engine_dependency_placeholders` — the non-trivial name-normalization + marker-evaluation logic that actually decides whether to inject the cu130 `sgl_kernel` direct-URL wheel or the CUDA<13.0 fallback. That path is only verified at the pure-function level (`test_virtualenv_package_rewrite.py` calling `filter_virtualenv_packages_by_markers`/`rewrite_direct_url_packages_for_index` directly), never through the real `_prepare_virtual_env` entrypoint with `model_engine="sglang"`.

**Suggested fix:** add one offline-install test that passes a real engine name and asserts the resulting package list matches what the pure-function tests expect, closing the gap between the two test suites.

doc/source/getting_started/using_docker_compose.rst
```diff
@@ -158,15 +153,47 @@ CPU override if needed:
...
+   in ``offline.env`` to skip runtime installs entirely and rely on the packages baked into
+   the Xinference image.
+
+.. warning::
+
+   Model specifications containing ``git+`` or other non-wheel direct references cannot be
+   represented faithfully by a Python simple index. In explicit offline-install mode,
+   Xinference rejects them before attempting network egress and reports the offending
+   requirement. The current built-ins in this category include the Transformers path of
+   HunyuanOCR and MiniCPM-V-4.6, plus FLUX.2-klein. Preinstall the required source revision in
```
rogercloud: **[MINOR] F5 — Affected-models list omits `MiniCPM-V-4.6-Thinking`**

This list reads as specific model names ("HunyuanOCR and MiniCPM-V-4.6, plus FLUX.2-klein"). `MiniCPM-V-4.6-Thinking` is a *separate* `model_name` entry in `llm_family.json` with its own `git+` dependency, so a reader would reasonably believe it's unaffected when it isn't — same offline-mode `ValueError` applies to it.

(For the record: FLUX.2-klein is fine as written — there's no model literally named that; only `-4B`/`-9B` variants exist, both have `git+` deps, and the family-level phrasing covers both.)

**Suggested fix:** add `MiniCPM-V-4.6-Thinking` explicitly to this list.

xinference/deploy/docker/docker-compose.byo-wheels.yml
```diff
@@ -0,0 +1,21 @@
...
+#   docker compose --profile offline \
+#     -f docker-compose.yml -f docker-compose.byo-wheels.yml up -d
+#
+# Composable with the air-gap override as well:
+#   docker compose --profile offline -f docker-compose.yml \
+#     -f docker-compose.byo-wheels.yml -f docker-compose.airgap.yml up -d
+
+services:
+  xinference-pypiserver:
+    image: ${XINFERENCE_PYPISERVER_IMAGE:-pypiserver/pypiserver:v2.3.2}
```
rogercloud: **[MINOR] F6 — This override doesn't actually switch away from the baked mirror image if `XINFERENCE_PYPISERVER_IMAGE` is set**

The header comment describes this file as serving a local `./wheels` directory "with the stock pypiserver image instead of the baked mirror," but this line reuses the same env var (`XINFERENCE_PYPISERVER_IMAGE`) that the base `docker-compose.yml` uses for the baked mirror image. Since `offline.env.example`/`.env.example`/the docs all instruct users to set this var to the baked mirror image tag for the normal offline profile, a user who has that set and then applies this override keeps the baked mirror image (just with a `./wheels` mount added on top) rather than switching to stock pypiserver — contradicting this file's own comment.

**Suggested fix:** either pin `image: pypiserver/pypiserver:v2.3.2` literally here (no env var), or document that `XINFERENCE_PYPISERVER_IMAGE` must be unset before applying this override.

xinference/deploy/docker/pypiserver/download_packages.py
```diff
@@ -0,0 +1,441 @@
...
+            sys.exit(f"FATAL: failed to fetch locked engine set '{engine}'")
+
+    # ------------------------------------------------------------------
+    # 2. Per-model pins, constrained to the shared substrate.
+    # ------------------------------------------------------------------
+    pins = json.loads((manifest_dir / "pins.json").read_text())
+    for entry in pins:
+        spec = entry["spec"]
+        pin_name = spec_name(spec)
+        pruned = work / "constraints-pruned.txt"
```
rogercloud, line 295: **shrink:** This block and the one further down (~369) both rebuild `work / "constraints-pruned.txt"` with the identical `"".join(f"{line}\n" for n, line in merged.items() if n != <name>)` pattern, differing only in the excluded name. Both call sites share the same `work`/`merged` types within `main()`, so extracting `write_pruned_constraints(work, merged, exclude_name) -> Path` is a safe, behavior-preserving dedup (edge cases — empty `merged`, name not present — are handled identically by construction). `net: 0 to -2 lines possible` — the value here is removing the literal duplication, not line count.

rogercloud (approved):
## Re-review (verified at `e8d3f467`)

> Note: two further commits (`f7ddb2c6`, `f2e3b543`) landed after this verification pass, touching only `.github/workflows/build-pypiserver-image.yaml` and the pypiserver `README.md` (adding manual `workflow_dispatch` publish controls, with `:latest` explicitly reserved for tag-triggered releases). I read this diff — it's additive, safe, and out of scope of everything below; not independently re-verified with the same rigor as the rest of this review.

### Update summary

Since the last review, the PR added an explicit `XINFERENCE_VIRTUAL_ENV_OFFLINE_INSTALL` flag (default off) that now gates the direct-wheel-URL rewrite, replacing the coarse `if settings.index_url:` heuristic that used to fire for anyone with an ordinary pip mirror configured. A genuine runtime guard was added that rejects any leftover `git+`/non-wheel direct reference before install when offline mode is on, turning a silent launch-time git-clone failure into an immediate, actionable error. Online CUDA 12.8/12.9 support was restored (the `PYTORCH_CUDA_WHEEL_URLS` cu128/cu129 entries and the `sgl_kernel ; cuda_version < "13.0"` fallback are back) while the mirror image itself is now explicitly documented as CUDA-13.0-only. New regression tests cover both rewrite paths, the git-rejection guard, the CUDA-12 fallback, and the three build scripts' string-parsing helpers.

### Prior findings — status (see individual replies on the original threads for detail)

All independently re-verified against current code, not just taken on the author's word:

1. Finding A (rewrite gate too coarse) — **FIXED**. Clean, well-tested flag-based gating.
2. Finding B (git deps escape selfcheck) — **FIXED**. Real runtime guard added, tested.
3. Finding C (CUDA<13.0 online regression) — **FIXED**. Online support restored; mirror's narrower scope documented.
4. Finding D (no CUDA-marker test) — **FIXED**. New test covers the CUDA 12.8 path.
5. Finding E (build scripts untested) — **PARTIAL**. Parsing helpers now tested; orchestration logic still isn't.
6. Simplification (`wheel_re` → `packaging.utils`) — **FIXED**. Net +14 lines due to helper extraction, not a problem.

### New findings this round (all MINOR — no functional regressions found)

- **F1** — Git-source dependency handling is inconsistent across three places: `download_packages.py` builds git-sourced deps into wheels and bakes them into the mirror (with a `sys.exit("FATAL: ...")` that hard-aborts the *entire* image build if a git-wheel build fails, even on retry); `selfcheck.py` reports git entries as UNSUPPORTED but never gates on them; and the runtime (Finding B's fix) unconditionally rejects any `git+` spec before install in offline mode. No substitution mechanism exists anywhere — the git wheels built and baked in can *never* actually be used offline. **Not a functional break** (the runtime rejection is deliberate and correct) — the real cost is build-time waste plus one fragile extra failure mode. See inline comment for suggested fix.
- **F2** — Offline mode silently disables xllamacpp's GPU-specific wheel/index selection with no log or doc. Defensible trade-off (the GPU index is unreachable offline anyway), but zero indication anywhere that a GPU host running llama.cpp fully offline gets a CPU-only build.
- **F3** — New `test_scripts.py` covers only the three build scripts' string-parsing helpers, not their orchestration logic (4-pass download flow, `selfcheck.py`'s `gate()`, `generate_package_lists.py`'s `main()`). Expected scope for a fast unit-test addition, tracked here so the gap isn't lost.
- **F4** — The three new offline-install tests in `test_worker.py` all pass `model_engine=None`, so the actual sglang engine-dispatch path (which decides whether to inject the cu130 `sgl_kernel` URL or the <13.0 fallback) is never exercised end-to-end through the real `_prepare_virtual_env` entrypoint — only at the pure-function level.
- **F5** — The offline docs' affected-models list omits `MiniCPM-V-4.6-Thinking` (a separate model entry with its own `git+` dependency) even though it reads as covering only `MiniCPM-V-4.6`. (For the record, the FLUX.2-klein concern from the initial pass didn't hold up on closer look — there's no model literally named that; only `-4B`/`-9B` exist, both covered by the family-level doc wording.)
- **F6** — `docker-compose.byo-wheels.yml`'s header comment says it switches to the stock pypiserver image, but its `image:` line reuses the same env var (`XINFERENCE_PYPISERVER_IMAGE`) the base compose file uses for the baked mirror — so a user who set that var per the offline setup instructions keeps the baked mirror image even after applying this override, contradicting the file's own comment.

*(Also checked: `download_packages.py`'s kept-failed-sdist behavior vs. `selfcheck.py`'s later `uv pip compile` gate — confirmed working as designed, not raising as a finding.)*

### Simplification opportunities

`download_packages.py`: two identical inline blocks rebuild `constraints-pruned.txt` by excluding one name from the `merged` dict (see inline comment). Extract a small helper — safe, no behavior-change risk. `net: 0 to -2 lines possible`.

### Verdict

All 3 prior MAJOR findings are genuinely fixed with solid engineering — the flag-based gating and runtime git-rejection guard are exactly the right shape and are well tested. The remaining PARTIAL (Finding E) and all new findings (F1–F6) are minor polish/doc/test-gap items with no functional regressions. **Approving** — recommend tracking F1–F6 as follow-ups; F1 and F6 are cheap enough to consider addressing before or shortly after merge, but nothing here blocks it.

**Tests:** `test_virtualenv_package_rewrite.py` (12/12) and `pypiserver/tests/test_scripts.py` (3/3) pass locally. Did not run the full Docker image build (180-min CI-only step).

xinference/deploy/docker/pypiserver/tests/test_scripts.py
OliverBryant: Addressed in 0a21115ca. Added lightweight orchestration tests for all three scripts: generator main walks model JSON and writes the manifest, downloader main covers lock/download/sdist/report flow and proves git sources are skipped, and selfcheck main exercises engine, pin, and URL gates plus unsupported git reporting. These tests use mocked process and network boundaries and complete in milliseconds.

xinference/core/worker.py
OliverBryant: Fixed in 0a21115ca. Offline llama.cpp launches now warn when a usable GPU and CUDA-specific xllamacpp index are detected, explain that the private mirror supplies the CPU build, and point users to a custom image for offline GPU acceleration. The Docker Compose docs and pypiserver README now document the same limitation, with a focused worker test.

xinference/deploy/docker/pypiserver/download_packages.py
OliverBryant: Addressed in 0a21115ca. The second constraints-pruning block belonged to git-wheel construction, which was removed in response to F1 because those wheels cannot be consumed by explicit offline mode. Only the per-model pin pruning remains, so there is no longer duplicate logic to extract.

doc/source/getting_started/using_docker_compose.rst
OliverBryant: Fixed in 0a21115ca. The affected-model list now explicitly includes MiniCPM-V-4.6-Thinking alongside MiniCPM-V-4.6.

PR #5049, opened 2026-06-18
fix(anthropic): honor top-level system prompt and inline system messages (#5037)

### LINKED ISSUE #5037: Claude Code >= v2.1.154版本支持
### Feature request / 功能建议

hello，team! 希望能支持Claude Code 在2.1.154版本后，Claude Code 会在 Anthropic Messages API 的 messages 数组中发送额外的消息角色，包括 ctx、msg 和 system。当前验证消息角色是 user 或 assistant，遇到未知角色就会返回 400 错误，无法在新版本Claude Code中继续使用。是否可以作为一个feature呢？

### Motivation / 动机

https://github.com/vllm-project/vllm/issues/44000

### Your contribution / 您的贡献

vllm/litellm他们也做了一定的支持调整~

Anai-Guo:
## What

Makes the Anthropic-compatible `/anthropic/v1/messages` endpoint honor the **top-level `system` prompt** and **inline `role: system` messages**.

Fixes #5037.

## Why

Today the endpoint silently drops the system prompt:

- `create_message` parses the body but never reads the Anthropic top-level `system` field. Because `system` isn't in the `exclude` set, it ends up in `raw_kwargs` and is passed as `raw_params` to `model.chat(...)`.
- `ModelActor.chat` then does `kwargs.pop("raw_params", None)` (`xinference/core/model.py`), so `raw_params` — and with it the entire system prompt — is discarded before the model is ever called.

So a standard Anthropic request that puts instructions in `system` runs **without any system prompt**. Separately, recent Claude Code clients (>= 2.1.154) also place `role: system` entries *inside* the `messages` array, which the OpenAI-style chat backend expects as a single leading system message.

vLLM hit the same incompatibility and fixed it by normalizing these into a leading system message ([vllm-project/vllm#44283](https://github.com/vllm-project/vllm/pull/44283), merged). This PR takes the same approach for xinference.

## What changed

`create_message` now folds the system content into one leading OpenAI-style `{"role": "system", ...}` message before validation/dispatch, via a small testable helper `_normalize_anthropic_messages`:

- Reads the top-level `system` field (string or list of text blocks).
- Merges any inline `role: system` messages from the array (and removes them from the dispatched list).
- Strips Claude Code's per-request `x-anthropic-billing-header` block (it carries a changing hash that defeats prefix caching — same treatment as upstream vLLM).
- Drops the consumed `system` key from `raw_kwargs`.

Requests with no system content are passed through unchanged, so existing behavior is preserved.

## Tests

Added unit tests in `xinference/core/tests/test_restful_api.py` covering: top-level string system, billing-header stripping, inline-system merging (string + list content), and the no-system pass-through case.

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -1287,7 +1287,7 @@ class RESTfulAPI(CancelMixin):
             normalized.append(msg)
 
         if system_parts:
-            normalized.insert(0, {"role": "system", "content": "".join(system_parts)})
+            normalized.insert(0, {"role": "system", "content": "\n".join(system_parts)})
         return normalized
 
     async def create_message(self, request: Request) -> Response:
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -1242,6 +1242,54 @@ class RESTfulAPI(CancelMixin):
                 self.handle_request_limit_error(e)
                 raise HTTPException(status_code=500, detail=str(e))
 
+    @staticmethod
+    def _normalize_anthropic_messages(
+        raw_system: Any, messages: Optional[List[dict]]
+    ) -> List[dict]:
+        """Fold the Anthropic top-level ``system`` prompt and any inline
+        ``role: system`` messages into a single leading OpenAI-style system
+        message.
+
+        Anthropic's Messages API carries the system prompt in a top-level
+        ``system`` field, and some clients (notably Claude Code >= 2.1.154)
+        additionally place ``role: system`` entries inside the ``messages``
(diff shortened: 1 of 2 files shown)


Anai-Guo:
Thanks @gemini-code-assist — good catch. Switched the merge join from `""` to `"\n"` in `_normalize_anthropic_messages` so adjacent system-prompt blocks (top-level `system` + inline `role: system`) stay separated instead of risking word concatenation, and updated the corresponding test assertion.

qinxuye (approved):
LGTM

PR #5053, opened 2026-06-18
feat: new UI part 5

maoyuehui:
1. MonitorCenter  Pages
2. LogCenter Pages


qinxuye (commented):
Left one non-duplicate inline comment.

frontend/src/components/pages/log-center/pagination.tsx
```diff
@@ -0,0 +1,42 @@
...
+      >
+        {t('logCenter.prevPage')}
+      </Button>
+      <span className="text-sm">
+        {currentPage} / {totalPages || 1}
+      </span>
+      <Button
+        variant="outline"
+        size="sm"
+        disabled={pageFrom + LOG_PAGE_SIZE >= total || pageFrom + LOG_PAGE_SIZE > 10000}
```
qinxuye: This condition still enables Next when `pageFrom` is already at the backend's ES window limit. The API clamps `page_from` to `10000 - size`, so with `LOG_PAGE_SIZE = 200` and `total > 10000`, clicking Next from `pageFrom = 9800` sends `page_from = 10000` but the backend returns the same 9800 page again, while the UI advances its page label. Please cap against the effective maximum, e.g. `pageFrom + LOG_PAGE_SIZE >= Math.min(total, 10000)`.

qinxuye (approved):
LGTM

PR #5055, opened 2026-06-19
fix: avoid blocking the event loop in async update_model_type

kratos0718:
### What

`update_model_type` is an `async` actor method, but it calls the synchronous `requests.get(url, timeout=30)` directly:

```python
async def update_model_type(self, model_type: str):
    ...
    response = requests.get(url, timeout=30)
```

A blocking call inside a coroutine freezes the actor's event loop for the entire duration of the request — up to the 30s timeout while downloading model configurations from `model.xinference.io`. During that window the worker actor cannot service any other request.

### Fix

Run the blocking request in a worker thread with `asyncio.to_thread()`, so the event loop stays responsive:

```python
response = await asyncio.to_thread(requests.get, url, timeout=30)
```

This matches the pattern already used throughout `worker.py` (e.g. `await asyncio.to_thread(...)` appears in several other places in this module). `asyncio` is already imported at the top of the file. No behavior change other than no longer blocking the loop.

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -116,7 +116,7 @@ jobs:
             sudo rm -rf "$AGENT_TOOLSDIRECTORY"
           fi
           pip install -e ".[dev]"
-          pip install "xllamacpp>=0.2.0"
+          pip install "xllamacpp>=0.2.0,<2026.6.9713"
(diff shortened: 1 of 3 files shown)


qinxuye (approved):
LGTM

PR #5060, opened 2026-06-22
FIX: [model] chat template analysis error

llyycchhee:
Fix `Can only get item pairs from a mapping` error.This happens when the arguments field in the message callback is str during a tool_calls callback.


qinxuye (commented):
One remaining issue on the current diff.

xinference/model/llm/utils.py
```diff
@@ -923,11 +923,63 @@ def _transform_messages(
...
+                    arguments = dict(arguments)
+                except (TypeError, ValueError) as exc:
+                    raise TypeError(
+                        "Tool call arguments must be a mapping or JSON object string"
+                    ) from exc
+
+            if arguments is not None and not isinstance(arguments, dict):
+                raise TypeError("Tool call arguments must decode to a JSON object")
+
+            if arguments is not None:
```
qinxuye: This still leaves one edge case unnormalized: when `arguments` is the valid JSON string `"null"`, `json.loads(arguments)` returns `None`, but the current `if arguments is not None` guard skips writing the parsed value back, so `target["arguments"]` remains the raw string. Please update based on key presence instead, e.g. `if "arguments" in target: target["arguments"] = arguments`, and add a small regression case for `"null"`.

qinxuye (approved):
LGTM

PR #5061, opened 2026-06-22
fix(embedding): pin peft/transformers for jina-embeddings-v5 series

m199369309:
## Summary
- Pin `peft==0.18.1` and `transformers==5.7.0` for the four jina-embeddings-v5 models (`text-nano`, `text-small`, `omni-nano`, `omni-small`) in `xinference/model/embedding/model_spec.json`.
- Replaces the previous looser bounds (`peft>=0.15.2`, `transformers>=4.57.0` / `>=5.0.0` / `>=5.1.0`) that were observed to cause runtime regressions for these models.

## Test plan
- [ ] `pytest -vv xinference/model/embedding/tests` passes
- [ ] Manually launch one jina-embeddings-v5 model and verify embedding inference works with `transformers==5.7.0` + `peft==0.18.1`
- [ ] Confirm no other embedding models in the same spec file are affected

### DIFF
--- a/xinference/model/embedding/model_spec.json
+++ b/xinference/model/embedding/model_spec.json
@@ -1579,11 +1579,11 @@
       "packages": [
         "sentence_transformers",
         "einops",
-        "peft>=0.15.2",
+        "peft==0.18.1",
         "accelerate>=0.28.0",
         "#system_torchvision# ; #engine# == \"sentence_transformers\"",
         "#system_torch# ; #engine# == \"sentence_transformers\"",
-        "transformers>=4.57.0"
+        "transformers==5.7.0"
       ]
     },
     "updated_at": 1781240042,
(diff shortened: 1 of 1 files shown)


m199369309:
Re: @gemini-code-assist's critical comments on all four hunks — the pinned versions **do exist on PyPI**:

- `peft==0.18.1` → https://pypi.org/project/peft/0.18.1/ (wheel + sdist; latest is 0.19.1)
- `transformers==5.7.0` → https://pypi.org/project/transformers/5.7.0/ (wheel + sdist; latest is 5.12.1)

The "No matching distribution found" concern appears to stem from the bot's knowledge cutoff — both versions were released after it, so it flagged them as hallucinated. Verified live via `curl https://pypi.org/pypi/peft/json` and `https://pypi.org/pypi/transformers/json`; both return real release artifacts. No change needed to the PR for this.

The substantive thing a human reviewer may still want to check: the same spec file pins other embedding models at `transformers==4.51.3` / `4.52.0` / `4.57.1`, which is mutually exclusive with `5.7.0` if two models were ever co-installed in one virtualenv. In practice each model runs in its own virtualenv so this is fine, but flagging it for transparency.

qinxuye (approved):
LGTM

PR #5065, opened 2026-06-23
FEAT: [UI] add API key management and user management pages

yiboyasss:


frontend/src/components/pages/api-key-management/index.tsx
```diff
@@ -0,0 +1,406 @@
...
+                        onClick={() => handleReveal(key.id)}
+                        disabled={revealingId === key.id}
+                      >
+                        {revealedKeys[key.id] ? <EyeOff className="h-3.5 w-3.5" /> : <Eye className="h-3.5 w-3.5" />}
+                      </button>
+                      <button
+                        type="button"
+                        title={t('common.copySuccess')}
+                        className="text-muted-foreground hover:text-foreground transition-colors"
+                        onClick={() => copyToClipboard(revealedKeys[key.id] || maskKey(key.key_prefix))}
```
qinxuye: This copies the masked fallback when the key has not been revealed yet, so the user gets a success toast but the clipboard contains an unusable value like `sk-...••••`. Please either disable/hide copy until `revealedKeys[key.id]` exists, or reveal the key first and copy the plaintext returned by `/reveal`.

qinxuye (commented):
I found a couple of issues that should be fixed before merging.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -199,24 +199,24 @@ export function Sidebar() {
...
+            Extra: ChevronRight,
+          },
+          {
+            path: '/api-key-management',
+            name: t('menu.apiKeyManagement'),
+            Icon: KeyRound,
+            Extra: ChevronRight,
+          },
+        ],
+        show: clusterUIConfig?.auth_advanced || false,
```
qinxuye: This only gates the system-management menu on `auth_advanced`, but the new pages call endpoints with narrower scopes: `users:manage`, `keys:create`, `keys:manage`, and `admin` for reveal. A non-admin advanced-auth user can see these entries/actions even though they immediately 403. Please derive the current JWT scopes and gate the user/API-key menu entries and privileged actions by the scopes they require.

frontend/src/components/pages/api-key-management/index.tsx
```diff
@@ -0,0 +1,401 @@
...
+    } finally {
+      setRevealingId(null);
+    }
+  };
+
+  const copyToClipboard = (text: string) => {
+    navigator.clipboard.writeText(text).then(() => {
+      toast.success(t('common.copySuccess'));
+    });
+  };
```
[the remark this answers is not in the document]
maoyuehui: There is a public copyText function in lib/utils

frontend/src/components/pages/api-key-management/index.tsx
```diff
@@ -0,0 +1,406 @@
...
+                </button>
+              </div>
+              <DialogFooter>
+                <Button onClick={() => setCreateOpen(false)}>{t('common.confirm')}</Button>
+              </DialogFooter>
+            </div>
+          ) : (
+            <div className="flex flex-col gap-4">
+              <div className="flex flex-col gap-1.5">
+                <Label htmlFor="key-name">
```
maoyuehui: You can consider using
components/ui/form
components/ui/form-field
In the component, there are form and form-field components. The Form component supports initialValues. The fields will be collected through onFinish, as well as mandatory validation, without the need to manually set error
FormField supports configuring properties such as name, label, placeholder, rules, and disabled

frontend/src/components/pages/user-management/index.tsx
```diff
@@ -0,0 +1,603 @@
...
+
+      {/* ── Create Dialog ───────────────────────────────────── */}
+      <Dialog open={createOpen} onOpenChange={setCreateOpen}>
+        <DialogContent className="sm:max-w-md">
+          <DialogHeader>
+            <DialogTitle>{t('userManagement.createUser')}</DialogTitle>
+          </DialogHeader>
+          <div className="flex flex-col gap-4">
+            <div className="flex flex-col gap-1.5">
+              <Label htmlFor="create-username">
```
maoyuehui: You can consider using
components/ui/form
components/ui/form-field
In the component, there are form and form-field components. The Form component supports initialValues. The fields will be collected through onFinish, as well as mandatory validation, without the need to manually set error
FormField supports configuring properties such as name, label, placeholder, rules, and disabled

frontend/src/components/pages/api-key-management/index.tsx
```diff
@@ -0,0 +1,419 @@
...
+                  <TableCell>
+                    <div className="flex items-center gap-1.5 font-mono text-xs">
+                      <span className="text-muted-foreground">
+                        {revealedKeys[key.id] ? revealedKeys[key.id] : maskKey(key.key_prefix)}
+                      </span>
+                      <button
+                        type="button"
+                        title={revealedKeys[key.id] ? t('apiKey.hideKey') : t('apiKey.revealKey')}
+                        className="text-muted-foreground hover:text-foreground transition-colors"
+                        onClick={() => handleReveal(key.id)}
```
qinxuye: The page-level access now allows users with `keys:create`, which is correct for list/create, but these row actions still call endpoints with higher scopes. `reveal` requires `admin`, while toggle/delete require `keys:manage`, so a user with only `keys:create` will see controls that always fail with 403. Please derive the current scopes here as well and hide or disable reveal for non-admin users, and toggle/delete for users without `keys:manage`.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -150,6 +150,11 @@ export function Sidebar() {
     [clusterAuth, token]
   );
 
+  const jwtScopes = useMemo(() => decodeJwtScopes(token), [token]);
+  const isAdmin = jwtScopes.includes('admin');
+  const canManageUsers = isAdmin || jwtScopes.includes('users:manage');
+  const canManageKeys = isAdmin || jwtScopes.includes('keys:create') || jwtScopes.includes('keys:manage');
```
qinxuye: This combines `keys:create` and `keys:manage` for the API-key page entry, but the page still lists and creates keys via `/v1/admin/keys`, and those endpoints require `keys:create`. A user with only `keys:manage` will see this menu item, then immediately hit 403 on page load. Please split the checks, e.g. use `admin || keys:create` for the page/list/create path and keep `admin || keys:manage` only for update/delete actions, or update the backend contract so `keys:manage` also permits listing.

qinxuye (approved):
LGTM

PR #5071, opened 2026-06-24
fix(core): prevent worker metrics loss when cache_tracker init fails

m199369309:
## Summary

- Worker-level metrics (`workers_total`, `_worker_status`-based gauges) were persistently empty because `get_supervisor_ref` coupled `add_worker` registration with `record_model_version` cache sync inside a single try block. When `record_model_version` raised, `_clear_supervisor_refs` wiped `self._supervisor_ref`, so `report_status` / `heartbeat` could never send worker status — even though the supervisor saw `workers_total=4`.
- Decouple the two concerns: registration is tracked by an explicit `_registered` flag, cache sync is isolated in its own try block and treated as an auxiliary feature whose failure must not tear down the worker↔supervisor channel.
- Harden `cache_tracker` so a version-count mismatch between supervisor and worker becomes a logged skip rather than an `AssertionError` that aborts the whole sync path.
- Isolate `record_model_version` in `supervisor.register_model` / `worker.register_model` so an auxiliary-cache failure does not roll back the already-successful `register_fn`. `register_fn` failures **do** roll back via `unregister_fn` (preserving the origin/main safety guarantee), and `_sync_register_model` failures **do** roll back + propagate (preserving cluster-state consistency).

## Changes

- `xinference/core/worker.py`
  - New `self._registered: bool` flag tracking whether `supervisor.add_worker()` has succeeded.
  - `get_supervisor_ref` cache-hit path returns early only when no registration is required, or the worker is already registered; otherwise falls through to run `add_worker`.
  - `record_model_version` moved into its own try block with a defensive `_cache_tracker_ref is None` guard; failures are logged as warnings, not propagated.
  - `_clear_supervisor_refs` resets `_registered` to keep it in sync with the `_supervisor_ref` lifecycle.
  - `register_model`: `register_fn` failure rolls back via `unregister_fn` (restored per maintainer feedback); `record_model_version` failure is warning-only.
- `xinference/core/cache_tracker.py`
  - `record_model_version`: replace `AssertionError` on version-count mismatch with a `logger.warning` + `continue`, so cache sync degrades instead of crashing.
- `xinference/core/supervisor.py`
  - `register_model`: split the single try block into (1) local `register_fn` (rolls back on failure), (2) `record_model_version` cache sync (warning-only, no rollback), (3) `_sync_register_model` worker propagation (rolls back `register_fn` and re-raises on failure).
- `xinference/core/tests/test_worker.py`
  - `set_supervisor_ref_for_test` now also sets `_registered = True` so tests that inject a fake supervisor ref (without going through `add_worker`) still hit the cache-hit path in `get_supervisor_ref(add_worker=True)`.

## Root cause

`get_supervisor_ref` previously did:

```python
try:
    supervisor_ref = await xo.actor_ref(...)
    if add_worker:
        await supervisor_ref.add_worker(...)
    # ... build model_version_infos ...
    await self._cache_tracker_ref.record_model_version(...)  # ← failure here
except Exception:
    self._clear_supervisor_refs()  # ← wiped _supervisor_ref
    raise
```

So any failure in `record_model_version` (e.g. transient `AssertionError` from a version-count mismatch on a re-reporting worker) cleared `_supervisor_ref`, blocking `heartbeat` and `report_status` for the worker. `workers_total` would still count the worker (because `add_worker` had already run), but every gauge sourced from `_worker_status` would stay empty.

## Test plan

- [x] `black --check` / `flake8` / `isort` on modified files — clean.
- [ ] `pytest xinference/core/tests/test_worker.py -k test_report_status` — verifies the injected-ref cache-hit path works with `_registered=True`. (Cannot run locally due to missing `torchvision`; CI will validate.)
- [ ] Unit test: `get_supervisor_ref` with `add_worker=True` calls `supervisor.add_worker` exactly once across retries when `record_model_version` raises.
- [ ] Unit test: `_cache_tracker_ref is None` does not raise `AttributeError` from `get_supervisor_ref`.
- [ ] Unit test: `cache_tracker.record_model_version` with mismatched version counts logs a warning and continues (no `AssertionError`).
- [ ] Unit test: `supervisor.register_model` / `worker.register_model` rolls back local registration (calls `unregister_fn`) and re-raises when `register_fn` fails.
- [ ] Unit test: `supervisor.register_model` rolls back and re-raises when `_sync_register_model` fails.
- [ ] Manual: on a multi-worker cluster, kill+restart `cache_tracker`; verify worker status gauges stay populated and supervisor logs a warning rather than dropping the worker.

## Dependencies

None. This PR is independent of the dashboard-split work tracked separately.

### DIFF
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -1420,15 +1420,36 @@ class SupervisorActor(xo.StatelessActor):
 
             try:
                 register_fn(model_spec, persist)
-                await self._cache_tracker_ref.record_model_version(
-                    generate_fn(model_spec), self.address
+            except ValueError as e:
+                raise e
+            except Exception as e:
+                unregister_fn(model_spec.model_name, raise_error=False)
+                raise e
+
+            # cache sync is an auxiliary feature; failure must not roll back
+            # the already-successful register_fn. Guard against _cache_tracker_ref
+            # being None (defensive).
+            try:
+                if self._cache_tracker_ref is not None:
+                    await self._cache_tracker_ref.record_model_version(
+                        generate_fn(model_spec), self.address
+                    )
+            except Exception:
+                logger.warning(
+                    "Failed to record model version for %s after registration; "
+                    "cache management may be degraded",
+                    model_spec.model_name,
+                    exc_info=True,
                 )
+
+            # _sync_register_model propagates to all workers. If it fails, the
+            # model will not be registered on the workers, so we must roll back
+            # the supervisor-local registration and propagate the error to keep
+            # the cluster state consistent.
+            try:
                 await self._sync_register_model(
                     model_type, model, persist, model_spec.model_name
                 )
-
-            except ValueError as e:
-                raise e
             except Exception as e:
                 unregister_fn(model_spec.model_name, raise_error=False)
                 raise e
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -417,6 +417,12 @@ class WorkerActor(xo.StatelessActor):
             CacheTrackerActor
         ] = None  # type: ignore
         self._progress_tracker_ref: Optional[xo.ActorRefType] = None
+        # Tracks whether this worker has successfully called supervisor.add_worker().
+        # _supervisor_ref being non-None no longer implies registered, because
+        # heartbeat() uses add_worker=False and may populate the ref cache before
+        # registration completes. This flag is the source of truth for registration
+        # state, used by get_supervisor_ref to decide whether add_worker is needed.
+        self._registered: bool = False
 
         # Virtual environment management
         self._virtual_env_manager: XinferenceVirtualEnvManager = None  # type: ignore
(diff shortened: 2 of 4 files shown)


xinference/core/worker.py
```diff
@@ -889,7 +895,12 @@ async def get_supervisor_ref(self, add_worker: bool = True) -> xo.ActorRefType:
         """
         from .supervisor import SupervisorActor
 
-        if self._supervisor_ref is not None:
+        # Cache hit: return immediately only when no registration is required,
+        # or the worker has already been registered. When add_worker=True and
+        # _registered=False, we must fall through to perform add_worker even if
+        # the ref is cached (this happens after heartbeat populated the ref via
+        # add_worker=False before registration completed).
+        if self._supervisor_ref is not None and (not add_worker or self._registered):
```
qinxuye, line 903: This makes `_registered` part of the cached-ref condition, but the tests that inject a fake supervisor ref via `set_supervisor_ref_for_test()` never mark the worker registered. `report_status()` now falls through to `xo.actor_ref(address="test", ...)` instead of using the fake ref, which is why all `test_report_status_*` jobs fail with `not enough values to unpack`. Please either mark the injected test ref as registered or route those tests through the new `add_worker` path.

xinference/core/supervisor.py
```diff
@@ -1420,15 +1420,35 @@ async def register_model(
 
             try:
                 register_fn(model_spec, persist)
-                await self._cache_tracker_ref.record_model_version(
-                    generate_fn(model_spec), self.address
+            except ValueError as e:
+                raise e
+            except Exception as e:
```
qinxuye, line 1425: This split fixes cache-tracker failures, but it also removes the old rollback for non-`ValueError` failures inside `register_fn`. `RegistryManager.register()` adds the custom model to memory before persisting the JSON, and image registration can also do extra engine setup after registration, so an `OSError` or later registration failure can leave a half-registered model. Please keep rollback around `register_fn` failures and only decouple/suppress `_cache_tracker_ref.record_model_version()` failures. The worker-side `register_model` has the same pattern and should be fixed as well.

qinxuye (commented):
Two issues need cleanup before this can land.

qinxuye (approved):
LGTM

PR #5073, opened 2026-06-24
feat(monitor): split Grafana dashboard into 6 sub-dashboards with config store

m199369309:
## Summary

- Splits the monolithic 79-panel Grafana dashboard into 6 focused sub-dashboards: `overview`, `model-load`, `llm-slo`, `gpu-resources`, `host-resources`, `security-audit`. Each is localized in `en` / `ja` / `ko` / `zh-CN` (24 JSONs total).
- Adds a SQLite-backed `MonitorConfigStore` so operators can configure Grafana URL, datasource, alert datasource, cluster name, and per-sub-dashboard UIDs at runtime via `/v1/cluster/monitor_config` (GET/PUT) and `/v1/cluster/monitor_config/check-grafana`, instead of env vars only. Env vars remain as fallback defaults; DB takes priority.
- Web UI monitoring page gains tabs (one per sub-dashboard), a settings dialog wired to the new config endpoints, iframe loading state, and i18n strings for all four locales.
- Includes `generate.py` to regenerate the 24 JSONs from the original monolithic dashboard, plus unit tests (`test_monitor_config_store.py`, 14 cases) and API integration tests (`test_monitor_config_api.py`, 9 cases).

## Motivation

The original `xinference-grafana-dashboard-zh-CN.json` (uid `xinference-overview`) packed cluster overview, supervisor API, worker GPU, DCGM hardware, model status, LLM SLO, host resources, and security audit into a single dashboard. Different audiences (SRE, ML Ops, SecOps) each cared about only a subset, and the monolithic view was hard to navigate. Splitting by audience also lets each dashboard be provisioned, permissioned, and refreshed independently.

The runtime config store removes the need to restart the supervisor to change Grafana settings, and lets different sub-dashboard UIDs be overridden without editing JSON.

### Dashboards (`monitor/dashboard/`)
- New sub-directories `overview/`, `model-load/`, `llm-slo/`, `gpu-resources/`, `host-resources/`, `security-audit/`, each with 4 locale JSONs.
- New `generate.py` — reads the original monolithic `xinference-grafana-dashboard-zh-CN.json`, redistributes panels by row, applies translations, and writes 24 JSONs into the sub-directories. Verifies UID consistency across languages.
- Updated `README.md` / `README_zh-CN.md` to document the new layout, import/provisioning steps, and per-dashboard audience.
- The original 4 monolithic JSONs are kept (they remain the source of truth for `generate.py`).

### Backend
- `xinference/constants.py`: `XINFERENCE_MONITOR_CONFIG_DB_PATH` (defaults to `~/.xinference/monitor_config.db`).
- `xinference/core/monitor_config_store.py` (new): SQLite store with `get_all`, `get_sources`, `get_dashboards`, `update`, `reset`; thread-safe; env-var fallback; DB-priority resolution.
- `xinference/api/routers/admin.py`: `get_ui_config` now reads from the store; new `GET/PUT /v1/cluster/monitor_config`, `POST /v1/cluster/monitor_config/check-grafana`, `POST /v1/cluster/monitor_config/reset` (admin-scoped).
- `xinference/api/restful_api.py`: instantiate `MonitorConfigStore` and attach to `app.state`.

### UI (`xinference/ui/web/ui/`)
- `src/components/grafanaUtils.js`: `buildGrafanaUrl` / `buildGrafanaPanelUrl` accept a `dashboardKey` and resolve UID from `grafana_dashboards` map.
- `src/scenes/monitoring/index.js`: tabbed layout, settings button, iframe loading indicator, persists active tab in `sessionStorage`.
- `src/scenes/monitoring/MonitorConfigDialog.js` (new): form for the config endpoints with health-check feedback.
- `src/locales/{en,zh,ja,ko}.json`: new `monitoring.tab.*` and `monitoring.config.*` strings.

### Tests
- `xinference/core/tests/test_monitor_config_store.py` — 14 cases (defaults, env fallback, DB priority, concurrent writes, reset).
- `xinference/api/tests/test_monitor_config_api.py` — 9 cases (GET/PUT/reset, partial update, check-grafana, full flow).

## Test plan

- [x] `pytest --noconftest xinference/core/tests/test_monitor_config_store.py` — 14 passed.
- [x] `pytest --noconftest xinference/api/tests/test_monitor_config_api.py` — 9 passed.
- [x] `npx eslint src/components/grafanaUtils.js src/scenes/monitoring/index.js src/scenes/monitoring/MonitorConfigDialog.js` — clean.
- [x] `npx prettier --check` on modified UI files — clean.
- [x] `black --check` on new Python files — clean.
- [ ] Manual: import the 6 sub-dashboards into Grafana, verify each renders against a live Prometheus.
- [ ] Manual: configure Grafana URL via the new settings dialog, verify the iframe loads and the health check passes.
- [ ] Manual: switch tabs, refresh, time-range, and auto-refresh still work.

## Dependencies

None. This PR is independent.

Two follow-up PRs will stack on top of this one:
- `fix/security-audit-model-name` — adds the `model_name` label to `_record_audit` counters in `restful_api.py` so the security-audit dashboard's per-model panels populate. Targets the `security-audit/` sub-dashboard introduced here.
- `feat/metrics-label-rename` — renames metric labels `model`→`model_uid`, `type`→`model_type`, `node`→`worker_address` in `core/model.py` and regenerates affected dashboards via `generate.py`. Breaking change; depends on this PR for `generate.py` and the split dashboard layout.


xinference/core/monitor_config_store.py
```diff
@@ -0,0 +1,175 @@
...
+    "cluster_name": "Cluster name (for Grafana var-cluster)",
+    "dashboard_overview": "Overview dashboard UID",
+    "dashboard_model_load": "Model load dashboard UID",
+    "dashboard_llm_slo": "LLM SLO dashboard UID",
+    "dashboard_gpu": "GPU resources dashboard UID",
+    "dashboard_host": "Host resources dashboard UID",
+    "dashboard_security": "Security audit dashboard UID",
+}
+
+ENV_MAPPING = {
```
qinxuye, line 47: This drops the existing `XINFERENCE_GRAFANA_DASHBOARD_UID` fallback. The old `/v1/cluster/ui_config` read that env var for `grafana_dashboard_uid`, but the new store only maps URL/datasource/cluster env vars before falling back to the hard-coded `xinference-overview`. Existing deployments that configured a custom dashboard UID through the env var will silently switch back to the default after upgrading. Please preserve the legacy env mapping at least for `dashboard_overview`, or add an explicit migration path.

monitor/dashboard/model-load/xinference-grafana-dashboard-model-load-en.json
```diff
@@ -0,0 +1,1328 @@
...
+          "refId": "A"
+        }
+      ]
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${datasource}"
+      },
+      "description": "按模型分组的 GPU 显存实时使用量（堆叠），可观测各模型对显存的持续消耗",
```
qinxuye: The English/Korean dashboards still contain Chinese text in non-title fields such as descriptions, legend labels, value mappings, table column renames, and sort/display names. For example this English dashboard still has a Chinese description, and the same file has Chinese `renameByName`/`displayName` values later in the table panel. Please translate these remaining user-visible strings across the non-zh dashboards, not just panel titles and template labels.

qinxuye (commented):
Found two issues on the current head.

xinference/core/monitor_config_store.py
m199369309: Good catch — I initially added `XINFERENCE_GRAFANA_DASHBOARD_UID` back to `ENV_MAPPING`, but after thinking through the data flow I reverted it. Here's the reasoning:

1. **Semantic mismatch**: the old env var configured the UID for the *single* monolithic dashboard. This PR replaces that monolith with 6 fixed-UID sub-dashboards (`xinference-overview`, `xinference-model-load`, ...). The UIDs are baked into the dashboard JSON files, not configurable properties. Mapping the legacy env var to `dashboard_overview` only is arbitrary — why that sub-dashboard and not the other 5?

2. **Would cause 404s**: if a deployment had `XINFERENCE_GRAFANA_DASHBOARD_UID=my-custom-uid`, restoring the mapping would make the UI load `my-custom-uid` for the overview tab. But the actual overview JSON has UID `xinference-overview`, so Grafana would 404. That's strictly worse than falling back to the default.

3. **Config path has changed**: runtime UID configuration is now done via the new `MonitorConfigStore` + the settings dialog (`PUT /v1/cluster/monitor_config`), which persists per-dashboard UIDs to SQLite. The other 4 env vars (URL/datasource/alert-datasource/cluster) remain in `ENV_MAPPING` because they configure runtime *attributes* rather than dashboard identity, so they don't conflict with the fixed-UID JSONs.

So the "silent switch back to default" you flagged is actually the correct behavior here — the custom UID pointed to a dashboard that no longer exists. The latest push (commit `37e4a754`) removes the env var from `ENV_MAPPING`. Let me know if you'd prefer an explicit migration note in the upgrade docs instead.

qinxuye (approved):
LGTM

PR #5074, opened 2026-06-24
fix(monitor): add model_name label to api_key_requests_total counter

m199369309:
## Summary

- The security-audit dashboard's "API Key requests by model" panels group by `model_name`, but `_record_audit` in `xinference/api/restful_api.py` never passed `model_name` to `api_key_requests_total.inc()`. The counter therefore had no `model_name` label, and every per-model panel on the security-audit dashboard showed "No data".
- Add `"model_name": model_name` to the `.inc()` dict so the counter carries the label. The companion histogram `api_key_request_duration_seconds` already received `model_name`, so the two are now consistent.
- Updates the 4 `security-audit` dashboard JSONs (`en`/`ja`/`ko`/`zh-CN`) to query by the now-populated `model_name` label.

## Root cause

`_record_audit` (`restful_api.py:225-292`) builds the audit log entry with `model_name`, but the Prometheus counter `api_key_requests_total.inc()` dict only had `user`, `api_key_name`, `model_id`, `model_type`, `status` — `model_name` was missing. The histogram path (`_duration.observe()`) a few lines below already included `model_name`, so this was an oversight.

`model_name` is resolved from `self._uid_to_model_name.get(model_uid, "")` (populated at model launch/terminate). When the mapping is missing (e.g. request denied before launch completes), `model_name` is `""`, which is the correct fallback for the counter.

## Changes

- `xinference/api/restful_api.py`: one-line addition to the `_requests_total.inc()` dict in `_record_audit`.
- `monitor/dashboard/security-audit/xinference-grafana-dashboard-security-audit-{en,ja,ko,zh-CN}.json`: query panels now use the `model_name` label.

## Test plan

- [ ] Manual: with `XINFERENCE_AUTH_ADVANCED=true`, send API-key-authenticated requests to ≥2 models; verify the security-audit dashboard's "API Key requests by model" panel populates with one series per model.
- [ ] Manual: verify the "API Key requests by user" and "API Key status distribution" panels still work (they don't use `model_name`).
- [ ] `flake8 xinference/api/restful_api.py` — clean.

## Dependencies

**Depends on #5073** (`feat(monitor): split Grafana dashboard into 6 sub-dashboards with config store`). The `security-audit/` dashboard JSONs modified here are introduced by that PR. This branch is stacked on top of #5073; please merge #5073 first, after which this PR can be rebased onto `main` (the only diff that will remain is the one-line `restful_api.py` change plus the 4 `security-audit` JSON updates).

### DIFF
--- a/monitor/dashboard/security-audit/xinference-grafana-dashboard-security-audit-en.json
+++ b/monitor/dashboard/security-audit/xinference-grafana-dashboard-security-audit-en.json
@@ -155,7 +155,7 @@
       },
       "targets": [
         {
-          "expr": "xinference:api_keys_active_total{cluster=~\"$cluster\"}",
+          "expr": "xinference:api_keys_active_total",
           "legendFormat": "Active",
           "refId": "A"
         }
@@ -214,7 +214,7 @@
       },
       "targets": [
         {
-          "expr": "xinference:api_keys_expired_total{cluster=~\"$cluster\"}",
+          "expr": "xinference:api_keys_expired_total",
           "legendFormat": "Expired",
           "refId": "A"
         }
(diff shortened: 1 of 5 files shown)


qinxuye (approved):
LGTM

PR #5076, opened 2026-06-24
feat(models): add model autostart config

bluefish-08:
## Changes

Added model autostart configuration so users can persist launch settings and automatically restore model instances after supervisor/container restart.

This PR includes:

- Added a persistent autostart config file with normalization, validation, sensitive-field redaction, and atomic write semantics.
- Added Supervisor-side autostart scheduling with retry, worker readiness checks, and runtime status tracking.
- Added REST APIs and Python client methods to read, upsert, append, and remove autostart model entries.
- Updated both UIs:
  - `frontend/` (Next UI)
  - `xinference/ui/web/ui` (legacy CRA UI)
- Added an option in the launch model flow: “Set as autostart after successful launch”.
- Added autostart status controls on the running models page and support for removing autostart entries.
- Frontend now uses the final `model_uid` returned by `/v1/models` when persisting autostart, ensuring the saved config matches the actual launched model instance.

## Notes

- Added locking around autostart config read/modify/write to avoid concurrent update races (upsert/remove).


qinxuye:
One design question: now that we already have a SQLite-backed launch history/store, would it be reasonable to use that as the source for autostart instead of maintaining a separate JSON config?

For example, after a model is successfully launched, record its final launch payload in SQLite; when the model is terminated or autostart is disabled, remove that record; on the next supervisor/container startup, reload the saved entries from SQLite and launch them again. This would also cover models that are already loaded without requiring a separate manual "save autostart" path, and avoids duplicating launch payload persistence in two places.

Does this direction make sense here, or is there a reason you prefer the separate autostart config file?

bluefish-08:
> One design question: now that we already have a SQLite-backed launch history/store, would it be reasonable to use that as the source for autostart instead of maintaining a separate JSON config?
> 
> For example, after a model is successfully launched, record its final launch payload in SQLite; when the model is terminated or autostart is disabled, remove that record; on the next supervisor/container startup, reload the saved entries from SQLite and launch them again. This would also cover models that are already loaded without requiring a separate manual "save autostart" path, and avoids duplicating launch payload persistence in two places.
> 
> Does this direction make sense here, or is there a reason you prefer the separate autostart config file?

Thanks for the suggestion. That makes sense.

When I implemented the first version, I did not look closely enough at the fields and behavior of the existing `launch_history` table. From the name and semantics, I treated it more as a user-facing launch record/history, and assumed it was closer to an append-style record of launch actions. So I did not build the autostart state on top of that table initially.

For the first version, I chose a separate JSON config mainly for simplicity and lower coupling: autostart was modeled as supervisor-level desired state, with explicit fields such as `enabled`, `priority`, `max_retries`, and `retry_interval_seconds`, while launch history was treated as UI/user history.

After looking at the current SQLite-backed launch history store more carefully, I agree it is better to reuse it as the persistence backend and avoid duplicating launch payload storage. I will revise the PR to extend `launch_history` with explicit autostart fields and keep the autostart behavior opt-in.

qinxuye:
Tests seemed failed, could you check it?

xinference/core/launch_history_store.py
```diff
@@ -118,29 +153,75 @@ def list(
...
-                        k: v for k, v in data.items() if k in SHAREABLE_KEYS
-                    }
-            # SQLite CURRENT_TIMESTAMP yields naive UTC "YYYY-MM-DD HH:MM:SS";
-            # emit ISO-8601 with explicit Z so clients parse it as UTC.
-            for key in ("created_at", "updated_at"):
-                value = item.get(key)
-                if value:
-                    item[key] = str(value).replace(" ", "T") + "Z"
-            result.append(item)
+            item = self._row_to_item(row, username=None)
```
qinxuye: This full autostart read path is not user-scoped. `list_autostart()` calls `_row_to_item(..., username=None)`, so `/v1/autostart/models` can expose other users' persisted launch payload fields after only the generic autostart redaction. That hides env/token-like values, but not fields that launch history already treats as non-shareable, such as `model_path`, `worker_ip`, or arbitrary custom kwargs. Since autostart is now backed by launch history, please pass the authenticated username through this path and apply the same owner-aware redaction/filtering, or restrict the full config endpoint to a true global/admin-only view.

qinxuye (approved):
LGTM

PR #5075, opened 2026-06-24
feat(metrics): unify worker labels with supervisor and add replica_index

m199369309:
## Summary

Worker-side model metrics used labels `model`, `type`, `node` while supervisor-side used `model_uid`, `model_type`, `worker_address`. This mismatch broke cross-side PromQL joins (the model-load dashboard's per-model tables needed `label_replace` workarounds) and made per-replica breakdown impossible because `model` carried the replica uid suffix.

Rename in `ModelActor._metrics_labels`:
- `model` → `model_uid` (base uid, replica suffix stripped via `parse_replica_model_uid`)
- `type` → `model_type`
- `node` → `worker_address`
- New label: `replica_index` (0-based, `"0"` for single replica)

Also fixes an existing bug where the model-load/llm-slo dashboard PromQL hardcoded `{type="llm"}` (lowercase) which never matched the actual uppercase `LLM` label value; now uses `{model_type="LLM"}`.

## Motivation

1. **Cross-side joins**: the model-load dashboard wants to show per-model QPS (worker-side `model_request_total`) alongside per-model VRAM (supervisor-side `model_gpu_memory_used_bytes`) in the same table. With mismatched label names, this required `label_replace` hacks. After this change, both sides use `model_uid` / `replica_index` / `worker_address`, so a direct `* on (model_uid, replica_index, worker_address) group_left(...)` works.
2. **Per-replica breakdown**: the `每副本明细` (id 406) panel aims to show one row per replica, but `model` was the replica uid (e.g. `qwen1.5-chat-1-0`), so multiple replicas of the same model on the same worker aggregated into one row. Splitting `model` into `model_uid` (base) + `replica_index` lets each replica get its own row.
3. **Label-name consistency**: `type` is a PromQL reserved-ish word and collides with common conventions; `model_type` is unambiguous. `node` is ambiguous (could be hostname); `worker_address` matches the supervisor side.

### Code
- `xinference/core/model.py`: `_metrics_labels` renamed as above; imports `parse_replica_model_uid`.
- `xinference/core/tests/test_metrics.py`: label key and assertion updated from `model` to `model_uid`.

### Dashboards (hand-edited JSONs; no generator script in-tree)
[...]

PR #5080, opened 2026-06-25
fix(metrics): align test assertion with renamed labels

m199369309:
> **Stacked on top of #5075** (`feat/metrics-label-rename`). Please merge #5075 first; this PR only fixes the CI assertion failure on top of it.

## Summary

Fixes the CI failure on PR #5075 (`feat(metrics): unify worker labels with supervisor and add replica_index`).

PR #5075 renamed worker-side metric labels in `ModelActor._metrics_labels` (`model`→`model_uid`, `type`→`model_type`, `node`→`worker_address`, plus newly added `model_name`/`replica_index`). The assertion in `test_metrics_exporter_data` was only partially updated and still expected the old `model=` key, so it failed against the actual exported labels:

```
FAILED xinference/core/tests/test_metrics.py::test_metrics_exporter_data
assert 'format="ggufv2",gpu_index="",model="qwen1.5-chat"' in response.text
```

Note: simply swapping `model=` → `model_uid=` is not enough, because the new `model_name` and `model_type` labels sort alphabetically between `gpu_index` and `model_uid` in the Prometheus output, so the contiguous substring no longer matches. The assertion now includes those intermediate labels.

## Test plan

- [x] Substring now matches the label set exported by `ModelActor._metrics_labels` after PR #5075.
- [ ] CI on this PR confirms the assertion passes across ubuntu/macos/windows × Python 3.10–3.13.


m199369309:
Closing in favor of pushing the fix directly to #5075.

PR #5084, opened 2026-06-27
feat: support VibeThinker with transformers/vLLM and harden list_models

xiaoyesoso:
* Register VibeThinker 1.5B / 3B (pytorch, none) in llm_family.json with both Transformers and vLLM engines.
* Fix get_batch_size_and_seq_len_from_kv_cache: the DynamicCache key_cache branch was nested inside except ImportError, so when HybridCache is importable but kv is a plain DynamicCache, the function fell through to kv[0][0] and raised 'DynamicCache' object is not subscriptable on newer transformers. Move the DynamicCache handling out of the except block so it runs regardless of HybridCache availability.
* Guard supervisor.list_models against stale replicas: a worker may still report a model that supervisor's _model_uid_to_replica_info no longer tracks (e.g. after a failed launch). Drop those entries with a warning instead of raising KeyError.
* Add .inference_home/ to .gitignore so local XINFERENCE_HOME caches are not committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,6 +103,7 @@ celerybeat.pid
 # Environments
 .env
 .venv
+.inference_home
 env/
 venv/
 ENV/
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -2851,8 +2851,23 @@ class SupervisorActor(xo.StatelessActor):
 
         running_model_info = {parse_replica_model_uid(k)[0]: v for k, v in ret.items()}
         # add replica count
+        stale_uids = []
         for k, v in running_model_info.items():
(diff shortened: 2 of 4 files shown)


xiaoyesoso:
Superseded by smaller, single-purpose PRs:

- #5085 — feat: register VibeThinker 1.5B / 3B (transformers + vLLM) + `.gitignore`
- #5086 — fix: handle `DynamicCache` in `get_batch_size_and_seq_len_from_kv_cache` when `HybridCache` is importable (also picks up @gemini-code-assist's `get_seq_length` fallback suggestion from this thread)
- #5087 — fix: drop stale running models without replica info in `supervisor.list_models`

Closing this one in favour of the three above so each change can be reviewed independently. Sorry for the churn.

PR #5085, opened 2026-06-27
feat: register VibeThinker 1.5B / 3B (transformers + vLLM)

xiaoyesoso:
## What

Register [WeiboAI/VibeThinker](https://huggingface.co/WeiboAI) **1.5B** and **3B** in `llm_family.json` under both **Transformers** and **vLLM** engines with `pytorch / none` quantization. Metadata (architecture, context length, chat template, `model_revision`) is taken from the official HF / ModelScope repos.

Also add `.inference_home/` to `.gitignore` so the local `XINFERENCE_HOME` cache used during development is not committed.

## Why

VibeThinker is a small Qwen2-architecture reasoning checkpoint that runs fine on a single consumer GPU; adding it to the built-in registry lets users launch it with one click via the Web UI / CLI without crafting a custom registration file.

## Notes

This PR is split out of #5084 to keep the change scoped to model registration. The two unrelated runtime fixes from that PR are sent separately:

- transformers `DynamicCache` handling: see `fix/transformers-dynamic-cache`
- supervisor `list_models` stale-replica guard: see `fix/supervisor-stale-replica`

## Test

Registered both sizes locally, launched with Transformers and vLLM engines, verified chat completion works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,6 +103,7 @@ celerybeat.pid
 # Environments
 .env
 .venv
+.inference_home
 env/
 venv/
 ENV/
--- a/xinference/model/llm/llm_family.json
+++ b/xinference/model/llm/llm_family.json
@@ -32532,5 +32532,84 @@
     },
     "featured": true,
     "updated_at": 1781520058
+  },
+  {
+    "version": 2,
+    "context_length": 131072,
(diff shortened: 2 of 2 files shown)


qinxuye (approved):
LGTM

PR #5086, opened 2026-06-27
fix: handle DynamicCache in get_batch_size_and_seq_len_from_kv_cache when HybridCache is importable

xiaoyesoso:
## Problem

In `get_batch_size_and_seq_len_from_kv_cache` ([xinference/model/llm/transformers/utils.py](https://github.com/xorbitsai/inference/blob/main/xinference/model/llm/transformers/utils.py)) the `DynamicCache` / `key_cache` handling was nested inside `except ImportError`:

```python
try:
    from transformers import HybridCache
    if isinstance(kv, HybridCache):
        return kv.key_cache[0].shape[bs_idx], kv.get_seq_length()
except ImportError:
    if kv is None:
        return 0, 0
    if hasattr(kv, "key_cache"):
        ...
return kv[0][0].shape[bs_idx], kv[0][0].shape[seq_len_idx] + 1
```

When `transformers.HybridCache` is importable (newer transformers releases) **and** `kv` is a plain `DynamicCache`, the `except` block is skipped and the function falls through to `kv[0][0].shape[...]`. On recent transformers releases `DynamicCache` is no longer subscriptable, so this raises:

```
TypeError: 'DynamicCache' object is not subscriptable
```

## Fix

Move the `None` / `key_cache` branches **out** of the `except` block so they run whenever the cache exposes `key_cache`, regardless of whether `HybridCache` is importable.

Also defensively fall back to `key.shape[seq_len_idx]` when `kv.get_seq_length()` is unavailable (some older transformers / custom cache implementations), matching the existing empty-cache branch — credit @gemini-code-assist for spotting the asymmetry on #5084.

## Notes

Split out of #5084. Only touches `xinference/model/llm/transformers/utils.py`.

### DIFF
--- a/xinference/model/llm/transformers/utils.py
+++ b/xinference/model/llm/transformers/utils.py
@@ -181,17 +181,41 @@ def get_batch_size_and_seq_len_from_kv_cache(kv, xinf_model_obj: "PytorchModel")
         from transformers import HybridCache
 
         if isinstance(kv, HybridCache):
+            if len(kv.key_cache) == 0 or kv.key_cache[0] is None:
+                return 0, 0
             return kv.key_cache[0].shape[bs_idx], kv.get_seq_length()
     except ImportError:
-        if kv is None:
+        pass
+
+    if kv is None:
+        return 0, 0
+
+    if hasattr(kv, "key_cache"):
+        if len(kv.key_cache) == 0 or kv.key_cache[0] is None:
             return 0, 0
 
-        if hasattr(kv, "key_cache"):
-            if len(kv.key_cache) == 0 or kv.key_cache[0] is None:
-                return 0, kv.get_seq_length() if hasattr(kv, "get_seq_length") else 0
+        key = kv.key_cache[0]
+        return (
+            key.shape[bs_idx],
+            (
+                kv.get_seq_length()
+                if hasattr(kv, "get_seq_length")
+                else key.shape[seq_len_idx]
+            ),
+        )
 
-            key = kv.key_cache[0]
-            return key.shape[bs_idx], kv.get_seq_length()
+    # New transformers Cache API (transformers >= 4.57 / 5.x): DynamicCache
+    # exposes `layers` (a list of CacheLayerMixin objects with `.keys` /
+    # `.values` tensors of shape [batch_size, num_heads, seq_len, head_dim])
+    # and `get_seq_length()`, but no longer exposes `key_cache` and (on 5.x)
+    # is not subscriptable. Without this branch the legacy `kv[0][0]` fallback
+    # below raises `TypeError: 'DynamicCache' object is not subscriptable`.
+    if hasattr(kv, "layers") and hasattr(kv, "get_seq_length"):
+        layers = kv.layers
+        first_keys = getattr(layers[0], "keys", None) if len(layers) > 0 else None
+        if first_keys is None or first_keys.numel() == 0:
+            return 0, kv.get_seq_length()
+        return first_keys.shape[bs_idx], kv.get_seq_length()
 
     return kv[0][0].shape[bs_idx], kv[0][0].shape[seq_len_idx] + 1
 


xiaoyesoso:
Pushed `1565e9b` addressing both review comments and the CI black 25.1.0 failure:

1. **Guard the `HybridCache` branch** against an empty / uninitialized cache so `kv.key_cache[0]` no longer risks `IndexError` / `AttributeError` before the cache is populated.
2. **Drop `kv.get_seq_length()` from the empty `key_cache` branch** and return `(0, 0)` directly — the cache is empty so seq_len is guaranteed 0, and some DynamicCache implementations crash inside `get_seq_length` when `key_cache[0]` is `None`.
3. **Wrap the conditional expression in the non-empty return tuple in parentheses** to satisfy black 25.1.0 (this is what the pre-commit CI hook reformatted; verified locally with `black==25.1.0`, `xinference` tree now leaves all 359 files unchanged).

Thanks @gemini-code-assist for catching the empty-cache cases.

xinference/model/llm/transformers/utils.py
```diff
@@ -181,17 +181,28 @@ def get_batch_size_and_seq_len_from_kv_cache(kv, xinf_model_obj: "PytorchModel")
...
 
-        if hasattr(kv, "key_cache"):
-            if len(kv.key_cache) == 0 or kv.key_cache[0] is None:
-                return 0, kv.get_seq_length() if hasattr(kv, "get_seq_length") else 0
+    if kv is None:
+        return 0, 0
 
-            key = kv.key_cache[0]
-            return key.shape[bs_idx], kv.get_seq_length()
+    if hasattr(kv, "key_cache"):
```
qinxuye, line 193: This still misses the newer Transformers `DynamicCache` API. In Transformers 5.x, `DynamicCache` has `get_seq_length()` and is iterable, but it no longer exposes `key_cache` and is not subscriptable. With the current logic it skips this branch and falls through to `kv[0][0]`, which still raises `TypeError: 'DynamicCache' object is not subscriptable`. Please add a branch for `transformers.cache_utils.Cache` / the new cache-layer API, using `get_seq_length()` for the sequence length and the first cache layer's key tensor for batch size, before keeping the legacy tuple fallback.

qinxuye (commented):
I found one remaining cache-shape compatibility issue.

xinference/model/llm/transformers/utils.py
xiaoyesoso: Thanks @qinxuye for catching this! Addressed in the latest force-push (`9bc42b3`).

Added a dedicated branch for the new transformers Cache API before the legacy tuple fallback in `get_batch_size_and_seq_len_from_kv_cache`:

```python
# New transformers Cache API (transformers >= 4.57 / 5.x): DynamicCache
# exposes `layers` (a list of CacheLayerMixin objects with `.keys` /
# `.values` tensors of shape [batch_size, num_heads, seq_len, head_dim])
# and `get_seq_length()`, but no longer exposes `key_cache` and (on 5.x)
# is not subscriptable. Without this branch the legacy `kv[0][0]` fallback
# below raises `TypeError: 'DynamicCache' object is not subscriptable`.
if hasattr(kv, "layers") and hasattr(kv, "get_seq_length"):
    layers = kv.layers
    first_keys = getattr(layers[0], "keys", None) if len(layers) > 0 else None
    if first_keys is None or first_keys.numel() == 0:
        return 0, kv.get_seq_length()
    return first_keys.shape[bs_idx], kv.get_seq_length()
```

It uses the first layer's `keys` tensor for batch size and `get_seq_length()` for the sequence length, exactly as you suggested. The branch is also guarded against empty `layers` and an empty / `None` first `keys` tensor (DynamicLayer initialises `keys` to an empty tensor), returning `(0, get_seq_length())` defensively.

Verified with smoke tests covering: populated 5.x DynamicCache, empty `layers`, empty `keys` tensor, and `None` kv — all return the expected `(batch_size, seq_len)` tuples without raising. `pre-commit` (black / flake8 / isort / mypy / codespell) passes.

qinxuye (approved):
LGTM

Issue #5058, opened 2026-06-21
[Security] Multiple CRITICAL auth/privilege-escalation vulnerabilities in OAuth2 layer (eternal JWT + users:manage to admin)


Galaxync:
### System Info / 系統信息

- Xinference: `main` branch (latest commit at time of report, ~2026-06-21)
- Python: independent (API/OAuth2 layer issue)
- Deployment: applies to both Docker and pip/source deployments that enable the OAuth2 auth mode

### Version info / 版本信息

Xinference `main` branch. The issues are in the OAuth2 authentication layer introduced under `xinference/api/oauth2/` (both `auth_service.py` and `advanced/routes.py`).

### The command used to start Xinference / 用以启动 xinference 的命令

Not deployment-specific. The vulnerabilities are in the OAuth2 auth module activated when Xinference runs with the OAuth2/simple or OAuth2/advanced auth mode. Any invocation that enables these auth modes is affected.

### Reproduction / 复现过程

This report documents **multiple security vulnerabilities** found via logic-vulnerability audit of the OAuth2 auth layer and **verified against the current `main` branch source**. Each is independently exploitable; several chain together.

---

#### Finding 1 — [CRITICAL] JWT expiration verification disabled (eternal tokens)

**Location:** `xinference/api/oauth2/auth_service.py:94-100`

The `AuthService` decodes JWTs with `options={"verify_exp": False}` (line 98, marked `# TODO: supports token expiration`):

```python
payload = jwt.decode(
    token,
    key,
    algorithms=[self.config.algorithm],
    audience=self.config.audience,
    options={"verify_exp": False},   # ← exp claim never checked
)
```

Although `create_access_token` embeds an `exp` claim based on `token_expire_in_minutes`, the decode path never validates it. There is no token blocklist and no logout mechanism in this auth mode, so **every JWT ever issued is valid forever**.

**Impact:** An attacker who briefly obtains any JWT (logs, proxy, XSS, leaked header) retains permanent access — even after the user changes their password, is disabled, or has permissions revoked. The only mitigation is rotating the JWT signing secret, which disrupts every user.

**Steps:**
1. Attacker obtains any user's JWT once (e.g. from a leaked `Authorization` header).
2. Attacker calls any protected endpoint with `Authorization: Bearer <stolen-jwt>` indefinitely.
3. The `exp` claim is present but never checked; the token never expires.

---

#### Finding 2 — [CRITICAL] Privilege escalation via unrestricted permission grant

[...]

PR #5088, opened 2026-06-27
fix(oauth2): verify JWT expiration in AuthService (#5058)

qfmy83:
### What

`AuthService.__call__` decodes JWTs with `options={"verify_exp": False}`
(`xinference/api/oauth2/auth_service.py:98`, marked `# TODO: supports token
expiration`):

```python
payload = jwt.decode(
    token,
    self._config.auth_config.secret_key,
    algorithms=[self._config.auth_config.algorithm],
    options={"verify_exp": False},   # exp claim never checked
)
```

`create_access_token` already embeds an `exp` claim, but the decode path never
checks it, so **every issued JWT is valid forever** — a leaked token keeps
working even after a password change or permission revocation. This is Finding 1
of #5058.

### Fix

Enable expiration verification (`verify_exp: True`). `python-jose` raises
`ExpiredSignatureError` (a subclass of `JWTError`) for expired tokens, which the
existing `except (JWTError, ValidationError)` already turns into a 401, so no
other change is needed.

### Testing

Added `xinference/api/tests/test_oauth2_token_expiration.py`: a token from
`create_access_token` with a negative `expires_delta` is rejected; a valid token
is still accepted.

```
pytest xinference/api/tests/test_oauth2_token_expiration.py -v
```

Scope: this PR fixes Finding 1 (eternal tokens) only. The other findings in
#5058 (permission-grant privilege escalation, etc.) are separate and out of
scope for this change.

Refs #5058

### DIFF
--- a/xinference/api/tests/test_oauth2_token_expiration.py
+++ b/xinference/api/tests/test_oauth2_token_expiration.py
@@ -1,42 +1,61 @@
-"""Regression test for #5058 Finding 1: JWTs issued by create_access_token
-must be rejected once expired (the decode path now uses verify_exp=True).
+"""Regression tests for rejecting expired OAuth2 JWT access tokens."""
 
-    pytest xinference/api/tests/test_oauth2_token_expiration.py -v
-"""
+import json
 from datetime import timedelta
 
 import pytest
-from jose import ExpiredSignatureError, jwt
+from fastapi import HTTPException
+from fastapi.security import SecurityScopes
 
+from xinference.api.oauth2.auth_service import AuthService
+from xinference.api.oauth2.types import AuthConfig, AuthStartupConfig, User
 from xinference.api.oauth2.utils import create_access_token
 
 _SECRET = "unit-test-secret"
 _ALG = "HS256"
 
 
-def _decode(token: str):
-    # same decode contract as AuthService.__call__ (verify_exp enabled by the fix)
-    return jwt.decode(token, _SECRET, algorithms=[_ALG], options={"verify_exp": True})
+@pytest.fixture
+def auth_service(tmp_path):
+    auth_config = AuthConfig(
+        algorithm=_ALG,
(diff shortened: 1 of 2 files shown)


xinference/api/tests/test_oauth2_token_expiration.py
```diff
@@ -0,0 +1,42 @@
...
+import pytest
+from jose import ExpiredSignatureError, jwt
+
+from xinference.api.oauth2.utils import create_access_token
+
+_SECRET = "unit-test-secret"
+_ALG = "HS256"
+
+
+def _decode(token: str):
```
qinxuye: This test helper duplicates the production decode call with `verify_exp=True`, so it does not actually exercise the changed `AuthService.__call__` path. If `auth_service.py` is changed back to `verify_exp=False`, these tests still pass because `_decode` keeps enforcing expiration. Please construct an `AuthService` with a minimal auth config, pass the expired token into `AuthService.__call__`, and assert it raises the 401 credentials error; keep the valid-token case on the same production path as well.

xinference/api/tests/test_oauth2_token_expiration.py
```diff
@@ -0,0 +1,42 @@
+"""Regression test for #5058 Finding 1: JWTs issued by create_access_token
+must be rejected once expired (the decode path now uses verify_exp=True).
+
+    pytest xinference/api/tests/test_oauth2_token_expiration.py -v
+"""
```
qinxuye: CI lint is currently failing because Black reformats this file by inserting a blank line after the module docstring. Please run `pre-commit run black --files xinference/api/tests/test_oauth2_token_expiration.py` or add the blank line here.

qinxuye (commented):
Found two small follow-ups: one test coverage gap and the current Black lint failure.

qinxuye (approved):
LGTM

PR #5089, opened 2026-06-27
fix(oauth2): block privilege escalation in user permission grants (#5058)

qfmy83:
### What

`create_user`, `update_user`, and `update_user_permissions`
(`xinference/api/oauth2/advanced/routes.py`) write the request-supplied
`permissions` list to the database verbatim. All three only require the
`users:manage` scope, with no check that the caller actually holds the
permissions being granted. Since `admin` bypasses every scope check, a
delegated `users:manage` operator can grant themselves (or anyone) the `admin`
superuser scope — a privilege escalation. This is Finding 2 of #5058.

### Fix

Add `_reject_permission_escalation(request, auth, permissions)` and call it in
the three handlers before persisting permissions. The rule is the standard
no-escalation invariant: a caller may only grant permissions they already hold;
the `admin` scope may grant anything. Granting an unheld scope returns 403.

### Testing

Added `xinference/api/tests/test_oauth2_permission_escalation.py`:

- a `users:manage` caller cannot grant `admin` or any unheld scope (403)
- an `admin` caller can still grant anything (no regression)
- granting only already-held permissions is allowed (delegation still works)

```
pytest xinference/api/tests/test_oauth2_permission_escalation.py -v
```

Scope: this is a minimal anti-escalation guard for Finding 2; maintainers may
wish to tighten the policy further (e.g. restrict permission management to
`admin` only). Other findings in #5058 are out of scope for this PR.

Refs #5058

### DIFF
--- a/xinference/api/oauth2/advanced/routes.py
+++ b/xinference/api/oauth2/advanced/routes.py
@@ -89,18 +89,17 @@ def _reject_permission_escalation(
     operator could mint the ``admin`` superuser scope for themselves or others
     (see security report, Finding 2).
     """
-    if requested_permissions is not None:
-        if not isinstance(requested_permissions, list) or not all(
-            isinstance(p, str) for p in requested_permissions
-        ):
-            raise HTTPException(
-                status_code=400, detail="Permissions must be a list of strings"
-            )
+    if not isinstance(requested_permissions, list) or not all(
+        isinstance(p, str) for p in requested_permissions
+    ):
+        raise HTTPException(
+            status_code=400, detail="Permissions must be a list of strings"
+        )
     _, _, caller_scopes = _get_current_user_from_token(request, auth)
     caller_scopes = caller_scopes or []
     if "admin" in caller_scopes:
         return
-    escalated = [p for p in (requested_permissions or []) if p not in caller_scopes]
+    escalated = [p for p in requested_permissions if p not in caller_scopes]
     if escalated:
         raise HTTPException(
             status_code=403,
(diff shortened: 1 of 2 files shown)


xinference/api/oauth2/advanced/routes.py
```diff
@@ -79,6 +79,35 @@ def _get_current_user_from_token(request: Request, auth: AdvancedAuthService):
...
+    request: Request, auth: "AdvancedAuthService", requested_permissions
+) -> None:
+    """Prevent privilege escalation when granting user permissions.
+
+    A caller may only grant permissions they themselves hold; the ``admin``
+    scope may grant anything. Without this, a delegated ``users:manage``
+    operator could mint the ``admin`` superuser scope for themselves or others
+    (see security report, Finding 2).
+    """
+    if requested_permissions is not None:
```
qinxuye: This still allows `requested_permissions=None`. That is fine for the create-user default path when the key is missing, but the update endpoints pass the request value through to `set_user_permissions`, so `{"permissions": null}` reaches `set_user_permissions(..., None)` and raises `TypeError: 'NoneType' object is not iterable` as a 500. Since this helper now owns permissions payload validation, please reject `None` as a bad permissions value here and keep the endpoint defaults passing `[]` when the field is omitted.

xinference/api/tests/test_oauth2_permission_escalation.py
```diff
@@ -0,0 +1,53 @@
+"""Regression test for #5058 Finding 2: a caller may only grant permissions
+they themselves hold (admins may grant anything).
+
+    pytest xinference/api/tests/test_oauth2_permission_escalation.py -v
+"""
```
qinxuye, line 5: CI lint is currently failing because Black reformats this file: it inserts a blank line after the module docstring and wraps a couple of long calls. Please run `pre-commit run black --files xinference/api/tests/test_oauth2_permission_escalation.py`.

qinxuye (commented):
Found two follow-ups: one permissions null validation gap and the current Black lint failure.

qinxuye (approved):
LGTM

PR #5095, opened 2026-06-28
fix(core): harden sub-pool creation and clean GPU orphans on startup

m199369309:
## Summary

Harden worker sub-pool creation against hangs and clean up vLLM GPU orphans left by a previous worker. Five fixes on top of current `main`:

- **H3 — serialize `append_sub_pool`** with an `asyncio.Lock`. xoscar's `MainActorPool.append_sub_pool` has no internal lock; under `XINFERENCE_MAX_CONCURRENT_LAUNCHES>1`, concurrent `create_subprocess_exec` calls deadlock on fork+GIL and block the event loop (which also disables the timeout callback below). The lock covers only `append_sub_pool` (milliseconds), not download/load — launch throughput is unaffected.
- **C — sub-pool creation timeout** (`XINFERENCE_SUBPOOL_LAUNCH_TIMEOUT=60s`, tunable). A single fork that deadlocks in CUDA init would otherwise hang the actor forever. On timeout: psutil-kill leftover `start_sub_pool` children (PPID == this worker), `release_devices`, and re-raise so `launch_builtin_model`'s `try/finally` runs the failure path and the supervisor's `_workers_launching` counter decrements. Scope is a single launch; co-located models are not affected. H3 is what makes C reachable — without serialization, the fork deadlock blocks the event loop and the `xo.wait_for` callback never fires.
- **H2 — pre-launch VRAM recheck**. If free VRAM on the target GPUs is `< 0.90`, SIGKILL PPID==1 vLLM orphans (cmdline-matched) and poll VRAM release. PPID==1 (not the existing diff heuristic) avoids mis-killing another live worker's vLLM processes on shared GPUs (`allow_multi_replica_per_gpu=True`). Best-effort; any error is logged and the launch proceeds.
- **H1 — startup orphan cleanup**. `__post_create__` dispatches an async task that scans all GPUs for PPID==1 vLLM orphans (the previous worker's EngineCore/GPU workers, reparented to init) and SIGKILLs them. `nvmlInit` is bounded by a `SIGALRM` guard (`_nvml_init_with_timeout`) so a stuck GPU driver cannot block startup. Runs after registration; never blocks the worker from serving.
- **G — fix `XINFERENCE_MODEL_UID` injection**. `VLLMModel.load()` previously set `os.environ["XINFERENCE_MODEL_UID"]` from a worker thread — a `putenv` race under concurrent launches, and ineffective anyway because the sub-pool was already forked. The tag is now injected via the `env=` dict in `_create_subpool`, so the sub-pool and its vLLM descendants inherit it.

## Motivation

Production incident: multi-replica vLLM launches (e.g. `qwen3.6` 2 replicas, or `jina-embeddings-v5` 8 replicas with `XINFERENCE_MAX_CONCURRENT_LAUNCHES=5`) hang the worker actor indefinitely. Root causes, in order of severity:

1. **Concurrent fork deadlock (H3)** — 5 launches enter `_create_subpool` concurrently; `append_sub_pool` is unlocked, so 5 `create_subprocess_exec` calls race on fork+GIL, block the event loop, and disable C's timeout callback. Serializing sub-pool creation eliminated the hang in testing (8 replicas, ~13s each).
2. **Cross-restart GPU orphans (H1)** — when a worker dies, its vLLM EngineCore/GPU workers (no `_check_ppid`) are reparented to PID 1 and keep holding VRAM. The next worker's first sub-pool creation deadlocks in CUDA init. "First restart hangs, second restart succeeds" because the second restart benefits from driver/OOM cleanup of the orphans.
3. **Single-fork CUDA-init deadlock (C)** — even without concurrency, a fork-unsafe worker state (vllm/torch already imported, CUDA context initialized) can deadlock a single sub-pool fork. Without a timeout, the actor hangs forever; the supervisor's launching exemption (`_workers_launching`) prevents self-healing.
4. **`os.environ` race (G)** — real `putenv` thread-safety bug; also made the model-uid tag invisible to the sub-pool's vLLM descendants.

## Changes

| File | Change |
|---|---|
| `xinference/constants.py` | Add `XINFERENCE_SUBPOOL_LAUNCH_TIMEOUT` (default 60s, env-tunable). |
| `xinference/core/worker.py` | `_create_subpool`: `asyncio.Lock` (H3) + `xo.wait_for` timeout (C) + VRAM recheck (H2) + `env` injection (G). New `_cleanup_gpu_orphans_on_startup` (H1) dispatched from `__post_create__`. New helpers `_nvml_init_with_timeout`, `_kill_gpu_orphans_by_ppid` (PPID-based, distinct from the existing diff-based `_kill_orphan_gpu_pids`). New `__init__` attr `_subpool_creation_lock`. Internal thresholds `_NVML_INIT_TIMEOUT=10` (module-level, not env-exposed). |
| `xinference/model/llm/vllm/core.py` | Remove `os.environ["XINFERENCE_MODEL_UID"] = self.model_uid` in `load()` (G). **The `XINFERENCE_TRUST_REMOTE_CODE` security gate (PR #5027) is untouched.** |
| `xinference/core/tests/test_subpool_hardening.py` | New mock-based test suite (no GPU required). |

## Backward compatibility

- One new env var `XINFERENCE_SUBPOOL_LAUNCH_TIMEOUT` (default 60s). Old users get the hardening with no config change. Normal sub-pool creation is <1s (fork+exec+import xoscar+bind+write shm; weight download happens later in `create_model_instance`), so 60s is 60x+ headroom.
- `XINFERENCE_MAX_CONCURRENT_LAUNCHES` can stay at its default 5; H3 serializes only sub-pool creation, download/load remain concurrent.
- `_kill_gpu_orphans_by_ppid` only targets PPID==1 + vLLM-cmdline processes, so live workers and their children are never killed, including on shared-GPU multi-worker hosts.
- Supervisor is not modified — `_workers_launching`, reverse-channel detection, and `_handle_dead_worker` are unchanged.
- `XINFERENCE_MODEL_UID` semantics unchanged (still present in the sub-pool child env); only the setting mechanism is corrected.

## Not in scope

The local "B2 [A]/[B]/[C]" orphan-cleanup blocks from the `feather-customized-2-9-0` branch are deliberately not backported here. `main` already has its own GPU helpers (`_snapshot_gpu_occupying_pids`, `_kill_orphan_gpu_pids`, `_snapshot_gpu_free_ratio`, `_wait_pids_dead`) and its own `recover_sub_pool` / `terminate_model` cleanup; re-introducing the 2-9-0 blocks verbatim would duplicate and conflict. This PR adds the missing pieces (startup cleanup, pre-launch recheck, serialization, timeout) on top of what `main` already has.

## Test plan

- [x] `pre-commit run --files <modified>` — black/flake8/isort/mypy/codespell all pass.
- [x] `xinference/core/tests/test_subpool_hardening.py` — mock-based, covers: C timeout (raises + `release_devices` + leftover-child kill), H3 serialization (no overlap across 3 concurrent calls), G env injection (env dict carries the tag; `os.environ` untouched), H2 PPID filter (live-worker child and non-vllm cmdline spared), H1 startup cleanup (nvml failure degrades; orphan SIGKILL path), `_nvml_init_with_timeout` success/failure. **Cannot run locally without `torch`/`torchvision` (same constraint as the existing `test_launch_orphan_cleanup.py`); CI installs the full extras.**
- [ ] CI matrix (Linux/macOS/Windows) — Windows verifies the no-`SIGALRM` degradation path does not crash at import.
- [ ] Manual: multi-replica vLLM launch (e.g. 8 replicas, `MAX_CONCURRENT_LAUNCHES=5`) — no hang, all replicas up.
- [ ] Manual: `kill -9` the worker while a vLLM model is running, restart, launch again — startup orphan cleanup runs and the launch succeeds.

### DIFF
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -62,6 +62,7 @@ from ..constants import (
     XINFERENCE_MODEL_DOWNLOAD_WORKERS,
     XINFERENCE_STATUS_GATHER_TIMEOUT,
     XINFERENCE_STATUS_REPORT_MULTIPLIER,
+    XINFERENCE_SUBPOOL_LAUNCH_TIMEOUT,
     XINFERENCE_TCP_REQUEST_TIMEOUT,
     XINFERENCE_VIRTUAL_ENV_DIR,
     XINFERENCE_VIRTUAL_ENV_SKIP_INSTALLED,
@@ -168,6 +169,9 @@ _PERSIST_RETRY_MAX = 3
 _VRAM_RECLAIM_TIMEOUT = 30
 # VRAM free ratio threshold — ratio >= this means "released".
 _VRAM_READY_RATIO = 0.90
+# nvmlInit timeout (seconds) — bounds a stuck GPU driver so worker startup
+# and pre-launch VRAM checks cannot hang indefinitely.
+_NVML_INIT_TIMEOUT = 10
 
 
 def _strip_test_envs(launch_args: dict) -> Tuple[dict, Set[str]]:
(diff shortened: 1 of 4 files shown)


xinference/core/worker.py
```diff
@@ -1345,9 +1566,127 @@ async def _create_subpool(
...
+        # the actor forever. xo.wait_for raises asyncio.TimeoutError after
+        # XINFERENCE_SUBPOOL_LAUNCH_TIMEOUT; we clean up and re-raise so the
+        # supervisor's _workers_launching counter decrements.
+        logger.debug(
+            "Creating subpool for model %s, start_python=%s, env=%s",
+            model_uid,
+            start_python,
+            env,
         )
+        async with self._subpool_creation_lock:
```
qinxuye: This lock/timeout only wraps the primary `_create_subpool()` call. There are still direct `append_sub_pool()` call sites later in this file: the `need_create_pools()` branch creates additional rank/device pools with `asyncio.gather(...)`, and `launch_rank0_model()` creates the Xavier rank0 pool directly. Those paths can still hit the same unbounded/concurrent `append_sub_pool()` hang this PR is trying to fix. Please factor the protected append logic into a helper and route all `append_sub_pool()` calls through it.

xinference/core/worker.py
```diff
@@ -1345,9 +1566,127 @@ async def _create_subpool(
...
+        # H3: append_sub_pool has no internal lock; concurrent fork+exec
+        # deadlocks on fork+GIL and blocks the event loop, which also
+        # disables the xo.wait_for timeout callback. The asyncio.Lock
+        # serializes only append_sub_pool (milliseconds), not download/load.
+        # C: a single fork that deadlocks in CUDA init would otherwise hang
+        # the actor forever. xo.wait_for raises asyncio.TimeoutError after
+        # XINFERENCE_SUBPOOL_LAUNCH_TIMEOUT; we clean up and re-raise so the
+        # supervisor's _workers_launching counter decrements.
+        logger.debug(
+            "Creating subpool for model %s, start_python=%s, env=%s",
```
qinxuye: Please do not log the full `env` dict here or in the timeout error below. `envs` comes from the launch request and may include secrets such as API keys or tokens; on timeout the current error log would persist those values. Logging the env key names, or a redacted copy, should be enough for diagnostics.

qinxuye (commented):
I found a couple of remaining issues in the current head.

qinxuye (approved):
LGTM

PR #5098, opened 2026-06-29
Feat: (UI) running model add Try to API and add useMenuAuth hook

maoyuehui:
1. running model add Try to API
2. add useMenuAuth hook
3. part style fix


frontend/src/components/pages/api-key-management/index.tsx
```diff
@@ -48,11 +46,7 @@ interface ApiKey {
 
 export default function ApiKeyManagement() {
   const { t } = useI18n();
-
-  const token = Cookies.get('token');
-  const jwtScopes = decodeJwtScopes(token === NO_AUTH ? undefined : token);
-  const isAdmin = jwtScopes.includes('admin');
-  const canManageKeys = isAdmin || jwtScopes.includes('keys:manage');
+  const { isAdmin, keysManageCreate } = useMenuAuth();
```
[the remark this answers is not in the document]
maoyuehui: Code logic changes

frontend/src/components/pages/running-model-detail/components/try-api-drawer.tsx
```diff
@@ -0,0 +1,534 @@
...
+    lines.push('  -H "Content-Type: application/json"');
+    lines.push(`  -d '${JSON.stringify(body, null, 2)}'`);
+  } else {
+    requiredFields.forEach((field) => {
+      const value = resolveValue(field, modelUid);
+      lines.push(
+        `  -F "${field.key}=${field.type === 'file' ? `@${shellValue(value)}` : shellValue(value)}"`
+      );
+    });
+  }
```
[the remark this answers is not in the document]
maoyuehui: Shell show required fields

maoyuehui (commented):
1

frontend/src/components/layout/sidebar.tsx
```diff
@@ -212,17 +210,17 @@ export function Sidebar() {
...
-            show: canManageUsers,
+            show: usersManagePage,
           },
           {
             path: '/api-key-management',
             name: t('menu.apiKeyManagement'),
             Icon: KeyRound,
             Extra: ChevronRight,
-            show: canAccessKeysPage,
+            show: keysManagePage,
```
qinxuye: This now hides the API key page from users that only have `keys:create`, even though the backend list/create routes for `/v1/admin/keys` are protected by `keys:create` and the previous sidebar allowed that scope to enter this page. Please keep page/list/create access tied to `keys:create`, and gate update/delete actions separately with `keys:manage`.

frontend/src/components/pages/api-key-management/index.tsx
```diff
@@ -300,12 +294,12 @@ export default function ApiKeyManagement() {
                   <TableCell>
                     <Switch
                       checked={key.enabled}
-                      disabled={!canManageKeys || togglingId === key.id || isExpired(key.expires_at)}
+                      disabled={!keysManageCreate || togglingId === key.id || isExpired(key.expires_at)}
```
qinxuye: The enabled switch and delete button should not be gated by `keys:create`. The backend update/delete routes use `keys:manage`, so this currently gives manage actions to create-only users and blocks manage-only users. Please use the manage scope for toggle/delete, while keeping create/list access separate.

frontend/src/components/pages/running-model-detail/components/try-api-drawer.tsx
```diff
@@ -0,0 +1,533 @@
...
+
+    config.fields.forEach((field) => {
+      const value = resolveValue(field, modelUid);
+      if (field.type === 'file') {
+        lines.push(
+          `formData.addFormDataPart("${field.key}", "${escapeString(String(value))}", RequestBody.create(new File("${escapeString(String(value))}"), MediaType.parse("application/octet-stream")));${fieldComment(field, '//')}`
+        );
+      } else {
+        lines.push(
+          `formData.addFormDataPart("${field.key}", String.valueOf(${javaValue(value)}));${fieldComment(field, '//')}`
```
qinxuye: For form examples with object fields such as `kwargs`, `javaValue()` emits `objectMapper.writeValueAsString(...)`, but the form branch does not import or instantiate `ObjectMapper`. OCR, document parsing, and image/video form configs all include stringified object kwargs, so the generated Java examples will not compile. Please stringify object form fields in a way that is valid in this branch, or add the required ObjectMapper setup.

frontend/src/components/pages/running-model-detail/components/try-api-drawer.tsx
```diff
@@ -0,0 +1,533 @@
...
+        const key = field.key;
+        lines.push(
+          `  file_${key}, _ := os.Open("${escapeString(String(value))}")${fieldComment(field, '//')}`,
+          `  defer file_${key}.Close()`,
+          `  part_${key}, _ := writer.CreateFormFile("${field.key}", "${escapeString(String(value))}")`,
+          `  io.Copy(part_${key}, file_${key})`
+        );
+      } else {
+        lines.push(
+          `  writer.WriteField("${field.key}", ${goValue(value)})${fieldComment(field, '//')}`
```
qinxuye: Same issue for Go form examples: object values are generated as `json.RawMessage(...)`, but this branch does not import `encoding/json`, and `writer.WriteField` expects a string. For `stringify` form fields like `kwargs`, please emit a JSON string value before passing it to `WriteField`.

qinxuye (commented):
Review comments for PR 5098.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -212,17 +210,15 @@ export function Sidebar() {
...
           {
             path: '/api-key-management',
             name: t('menu.apiKeyManagement'),
             Icon: KeyRound,
             Extra: ChevronRight,
-            show: canAccessKeysPage,
           },
         ],
-        show: (clusterUIConfig?.auth_advanced || false) && (canManageUsers || canAccessKeysPage),
+        show: clusterUIConfig?.auth_advanced || false,
```
qinxuye: This now shows the whole System Management group whenever `auth_advanced` is enabled, while the individual User Management and API Key Management entries no longer have their previous scope gates. Users without `users:manage` or `keys:create` will see pages that only fail with 403 after navigation. Please keep `show: usersManagePage` for `/user-management`, use the `keys:create` page permission for `/api-key-management`, and show the group only when at least one item is visible.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -143,18 +144,15 @@ export function Sidebar() {
   const { t, locale } = useI18n();
   const [collapsed, setCollapsed] = useState(false);
   const branding = getBrandingFromEnv();
+  const { usersManagePage, keysManagePage } = useMenuAuth();
   const { clusterVersion, clusterAuth, clusterUIConfig } = useGlobal();
+
```
qinxuye: `git diff --check` fails here because this added blank line contains trailing whitespace. Please remove it so the patch is clean.

frontend/src/components/pages/api-key-management/index.tsx
```diff
@@ -300,12 +294,12 @@ export default function ApiKeyManagement() {
                   <TableCell>
                     <Switch
                       checked={key.enabled}
-                      disabled={!canManageKeys || togglingId === key.id || isExpired(key.expires_at)}
+                      disabled={togglingId === key.id || isExpired(key.expires_at)}
```
qinxuye: This still lets non-`keys:manage` users toggle keys because the switch is only disabled while loading or expired. The backend protects `PUT /v1/admin/keys/{key_id}` with `keys:manage`, so the UI should also disable this action unless the user has that scope.

frontend/src/components/pages/api-key-management/index.tsx
```diff
@@ -300,12 +294,12 @@ export default function ApiKeyManagement() {
...
                       onChange={() => handleToggleEnabled(key)}
                     />
                   </TableCell>
                   <TableCell className="text-right">
-                    {canManageKeys && (
+                    {isAdmin && (
```
qinxuye: Delete should be gated by the same `keys:manage` permission, not by admin-only. The backend allows `keys:manage` for DELETE, so manage-scope users would incorrectly lose the delete action here.

frontend/src/components/pages/running-model-detail/components/try-api-drawer.tsx
```diff
@@ -0,0 +1,564 @@
...
+              >
+                <X className="size-5" />
+              </Button>
+            </SheetClose>
+            <SheetTitle className="truncate text-xl">Try To API</SheetTitle>
+          </div>
+          <Button
+            type="button"
+            variant="outline"
+            disabled={!keysManagePage}
```
qinxuye: This button links to the API key page, whose list/create routes are protected by `keys:create`, but it is disabled unless the user has `keys:manage`. Users who can create/list keys but not manage existing keys should still be able to get an API key from here. Please gate this with the key page/create permission instead.

frontend/src/components/layout/sidebar.tsx
maoyuehui: There was an error in the previous code

frontend/src/components/pages/api-key-management/index.tsx
maoyuehui: There was an error in the previous code

frontend/src/components/layout/sidebar.tsx
```diff
@@ -212,17 +208,15 @@ export function Sidebar() {
...
           {
             path: '/api-key-management',
             name: t('menu.apiKeyManagement'),
             Icon: KeyRound,
             Extra: ChevronRight,
-            show: canAccessKeysPage,
           },
         ],
-        show: (clusterUIConfig?.auth_advanced || false) && (canManageUsers || canAccessKeysPage),
+        show: clusterUIConfig?.auth_advanced || false,
```
qinxuye: This is still not fixed on the latest head. The latest commit removed the unused `useMenuAuth` import/destructuring, but the System Management entries are still shown solely based on `auth_advanced`. Please restore per-item scope gates: `usersManagePage` for `/user-management`, a `keys:create` page permission for `/api-key-management`, and hide the group when neither entry is visible.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -144,17 +144,13 @@ export function Sidebar() {
   const [collapsed, setCollapsed] = useState(false);
   const branding = getBrandingFromEnv();
   const { clusterVersion, clusterAuth, clusterUIConfig } = useGlobal();
+
```
qinxuye: `git diff --check` still fails on the latest head because this added blank line contains trailing whitespace. Please remove the whitespace.

frontend/src/components/pages/api-key-management/index.tsx
qinxuye: This is still not fixed on the latest head. The switch can still call `PUT /v1/admin/keys/{key_id}` without checking `keys:manage`; the backend requires `keys:manage`, so the UI should disable this action unless the user has that scope.

frontend/src/components/pages/api-key-management/index.tsx
qinxuye: This is still admin-only on the latest head. Delete should be gated by `keys:manage`, matching the backend `DELETE /v1/admin/keys/{key_id}` permission, so users with `keys:manage` but not admin do not lose the action.

frontend/src/components/pages/running-model-detail/components/try-api-drawer.tsx
qinxuye: This is still gated by `keysManagePage` on the latest head. The button only links users to create/list API keys, which is protected by `keys:create`, so users with `keys:create` but not `keys:manage` should not be blocked here.

frontend/src/components/layout/sidebar.tsx
qinxuye: This is still only gated by `auth_advanced`, so any token with an advanced-auth scope can still see both system-management entries. Please use `useMenuAuth()` here as well: gate User Management with `usersManagePage`, API Key Management with `keysManageCreate`, and only show the group when at least one of those entries is visible.

frontend/src/components/pages/api-key-management/index.tsx
qinxuye: This still allows users who can only list/create keys to toggle an existing key, but the backend `PUT /v1/admin/keys/{key_id}` route requires `keys:manage`. Please include `!keysManagePage` in the disabled condition.

frontend/src/components/pages/api-key-management/index.tsx
qinxuye: Delete is still restricted to `isAdmin`, but the backend delete route accepts `keys:manage` as well. Please render this action for `keysManagePage`; keeping reveal admin-only is fine because `/reveal` still requires admin.

frontend/src/components/pages/running-model-detail/components/try-api-drawer.tsx
qinxuye: This button still checks `keysManagePage`, which maps to `keys:manage`, but navigating to the API key page / creating a key should use the `keys:create` permission. Please use the create/page permission (`keysManageCreate`) here so non-admin users with key creation rights can still get an API key.

qinxuye (approved):
LGTM

PR #5099, opened 2026-06-29
FEAT(webui): replace worker IP input with a filtered multi-select selector

leslie2046:
## Summary

This PR improves the Web UI launch model form by replacing the free-text `worker_ip` field with a multi-select worker selector backed by cluster node data.

图片

### Changes
- replace the `worker_ip` text input with a multi-select dropdown in the launch model drawer
- load available worker nodes from cluster info and show them as selectable options
- filter worker options based on the selected device type so GPU-only selections only show GPU-capable workers
- clear incompatible worker selections automatically when the device selection changes
- hide and clear `gpu_idx` when CPU is selected
- normalize selected worker addresses to pure IPs and submit them as comma-separated values like `ip1,ip2,ip3`
- add inline field validation and helper messages for worker loading, invalid selections, and GPU index constraints
- extend the shared select component to support multi-select, checkbox rendering, and helper text
- update i18n strings for the new worker selector UX and add the missing `text2audio_emotion_control` label

From codex(gpt-5.4)


xinference/ui/web/ui/src/scenes/launch_model/components/launchModelDrawer.js
```diff
@@ -582,6 +731,26 @@ const LaunchModelDrawer = ({
     }
   }
 
+  const fetchWorkers = async () => {
+    if (!open) return
+
+    setIsLoadingWorkers(true)
+    setHasWorkerLoadFailed(false)
+    try {
+      const data = await fetchWrapper.get('/v1/cluster/info?detailed=true')
```
qinxuye, line 743: This makes the worker selector depend on `/v1/cluster/info?detailed=true`, but that route requires the `admin` scope when auth is enabled, while launching a model only requires `models:start`. A non-admin user who can launch models will get 403 here, the worker selector becomes disabled, and any restored `worker_ip` from history/command line will be rejected during validation. Please keep a manual-input fallback when the worker list cannot be loaded, or expose a worker-list endpoint usable by the model launch scope and only enforce filtered selection when that list is available.

leslie2046: what a pity

leslie2046: Good catch. `/v1/cluster/info` is admin-scoped while model launch only requires `models:start`. I kept a frontend fallback for `worker_ip`: when the worker list cannot be loaded, the field falls back to manual comma-separated input, and filtered worker validation is only enforced when the worker list is available. Restored `worker_ip` values from history/command line remain usable in that fallback path.

qinxuye (approved):
LGTM

PR #5103, opened 2026-06-29
feat: add Unlimited-OCR support

xiaoyesoso:
### What does this PR do?

Registers [Baidu Unlimited-OCR](https://github.com/baidu/Unlimited-OCR) — the
follow-up to DeepSeek-OCR for "one-shot long-horizon document parsing" — as a
first-class image/OCR model in Xinference, runnable via the `transformers`
engine.

- Adds `UnlimitedOCRModel` in `xinference/model/image/ocr/unlimited_ocr.py`
  that wraps the bundled `model.infer` / `model.infer_multi` APIs.
- Registers the new class in `SUPPORTED_ENGINES["transformers"]` in
  `xinference/model/image/ocr/__init__.py`.
- Adds the model spec entry in `xinference/model/image/model_spec.json`
  (HuggingFace: `baidu/Unlimited-OCR`, ModelScope: `PaddlePaddle/Unlimited-OCR`).

### How is it tested?

End-to-end on a local 2×GPU host:

1. `POST /v1/models` → `HTTP 200`, model loaded into an isolated virtualenv
   pinned at `transformers==4.57.1` (matches the upstream README versions).
2. `GET /v1/models` lists the running replica.
3. `POST /v1/images/ocr` with a sample chart image returns

   ```json
   {
     "text": "Kid's Favourite Fruits\n\n![](images/0.jpg)\n\n![](images/1.jpg)\n...",
     "model": "unlimited-ocr",
     "success": true,
     "model_size": "gundam",
     "base_size": 1024,
     "image_size": 640,
     "crop_mode": true
   }

### DIFF
--- a/xinference/model/image/ocr/unlimited_ocr.py
+++ b/xinference/model/image/ocr/unlimited_ocr.py
@@ -14,6 +14,7 @@
 
 import logging
 import os
+import shutil
 import tempfile
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
@@ -144,7 +145,12 @@ class UnlimitedOCRModel(OCRModel):
                         setattr(config, key, value)
             if getattr(config, "pad_token_id", None) is None:
                 config.pad_token_id = self._tokenizer.eos_token_id
-            if self._device != "cpu":
+            # Resolve the target device once so the model lands on the GPU
+            # picked by Xinference's scheduler (``self._device``) instead of
+            # always falling back to ``cuda:0``. Default to ``cuda`` when no
+            # device is configured, mirroring the previous behavior.
+            device = self._device or "cuda"
+            if device != "cpu":
                 model = AutoModel.from_pretrained(
                     self._model_path,
                     config=config,
@@ -152,7 +158,7 @@ class UnlimitedOCRModel(OCRModel):
                     use_safetensors=True,
                     torch_dtype=torch.bfloat16,
                 )
-                self._model = model.eval().cuda()
+                self._model = model.eval().to(device)
             else:
                 model = AutoModel.from_pretrained(
                     self._model_path,
@@ -272,6 +278,7 @@ class UnlimitedOCRModel(OCRModel):
         ngram_window = kwargs.pop("ngram_window", 128)
 
         temp_image_path = None
+        output_path = None
         try:
             with tempfile.NamedTemporaryFile(suffix=".jpg", delete=False) as f:
                 image.save(f.name, "JPEG")
@@ -312,6 +319,12 @@ class UnlimitedOCRModel(OCRModel):
                     os.remove(temp_image_path)
                 except OSError:
                     pass
+            # ``model.infer`` always writes ``result.md`` (and possibly
+            # intermediate debug artifacts) under ``output_path`` when
+            # ``save_results=True``. Remove the whole temp dir to avoid
+            # leaking one directory per OCR call.
+            if output_path and os.path.exists(output_path):
+                shutil.rmtree(output_path, ignore_errors=True)
 
     def _ocr_multi(
         self,
@@ -331,6 +344,7 @@ class UnlimitedOCRModel(OCRModel):
         ngram_window = kwargs.pop("ngram_window", 1024)
 
         temp_paths: List[str] = []
+        output_path = None
         try:
             for img in images:
                 if img.mode in ["RGBA", "CMYK"]:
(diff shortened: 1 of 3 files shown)


xiaoyesoso:
Thanks for the review! Both issues fixed in commit ab2d672:

**1. Hardcoded `.cuda()` → `.to(device)`**

The model now lands on the GPU picked by Xinference's scheduler (`self._device`) instead of always `cuda:0`:

```python
device = self._device or "cuda"
if device != "cpu":
    ...
    self._model = model.eval().to(device)
```

This avoids the multi-GPU OOM / resource conflict when the scheduler places two replicas on different GPUs.

**2. `tempfile.mkdtemp()` cleanup**

Both `_ocr_single` and `_ocr_multi` now initialize `output_path = None` before the try block and `shutil.rmtree(output_path, ignore_errors=True)` in the `finally` block, so the temp dir (containing `result.md` and any intermediate debug artifacts) is removed after every OCR call instead of accumulating one directory per inference.

xinference/model/image/ocr/unlimited_ocr.py
```diff
@@ -0,0 +1,391 @@
...
+        logger.info("Unlimited-OCR kwargs: %s", kwargs)
+
+        if self._model is None or self._tokenizer is None:
+            raise RuntimeError("Model not loaded. Please call load() first.")
+
+        prompt = kwargs.pop("prompt", "<image>document parsing.")
+        model_size = kwargs.pop("model_size", "gundam")
+        max_length = kwargs.pop("max_length", 32768)
+        no_repeat_ngram_size = kwargs.pop("no_repeat_ngram_size", 35)
+        save_results = kwargs.pop("save_results", False)
```
qinxuye, line 215: Can we either honor `save_results` (with a user-provided `save_dir`, consistent with DeepSeek-OCR) or stop exposing it here? Right now this value is popped and passed into `_ocr_single` / `_ocr_multi`, but both helpers always create a temporary output directory, call the upstream model with `save_results=True`, and then delete that directory in `finally`. A caller setting `save_results=True` therefore gets no persisted files even though the parameter is documented, so the option is silently ignored.

xiaoyesoso:
Thanks for the review! Fixed in commit 1876165:

**Honor `save_results` with `save_dir`**

The previous implementation silently ignored `save_results` — always creating a temp directory and deleting it in `finally`, leaving callers with no persisted files.

Now consistent with DeepSeek-OCR:

```python
# When save_results=True and save_dir is provided:
- Use save_dir as output path (create if needed)
- Include saved_files in response with paths
- Do NOT clean up user-provided directory

# When save_results=False or save_dir not provided:
- Create temp directory via mkdtemp()
- Clean it up in finally (existing behavior)

# Validation:
if save_results and not save_dir:
    raise ValueError("save_dir must be provided when save_results=True")
```

The response now includes `saved_files` when results are persisted:
```json
{
  "text": "...",
  "saved_files": {
    "output_dir": "/path/to/dir",
    "result_file": "/path/to/dir/result.md"
  }
}
```

xinference/model/image/ocr/unlimited_ocr.py
```diff
@@ -0,0 +1,426 @@
...
+                for key, value in language_config.items():
+                    if not hasattr(config, key) or getattr(config, key, None) is None:
+                        setattr(config, key, value)
+            if getattr(config, "pad_token_id", None) is None:
+                config.pad_token_id = self._tokenizer.eos_token_id
+            # Resolve the target device once so the model lands on the GPU
+            # picked by Xinference's scheduler (``self._device``) instead of
+            # always falling back to ``cuda:0``. Default to ``cuda`` when no
+            # device is configured, mirroring the previous behavior.
+            device = self._device or "cuda"
```
qinxuye: Can we avoid defaulting to CUDA when Xinference did not pass an explicit device? In CPU-only launches or `n_gpu=None`, the worker disables GPUs via the subpool environment but does not pass `device="cpu"` into the model constructor, so `self._device` remains `None` here and this still tries to load onto CUDA, bypassing the CPU branch below. Please fall back based on availability, e.g. `device = self._device or ("cuda" if torch.cuda.is_available() else "cpu")`.

xiaoyesoso: Thanks, fixed. The fallback now uses CUDA only when torch.cuda.is_available(), otherwise CPU:

device = self._device or ("cuda" if torch.cuda.is_available() else "cpu")

This should avoid bypassing the CPU branch for CPU-only launches / n_gpu=None. pre-commit passed for the changed file.

qinxuye (approved):
LGTM

PR #5111, opened 2026-07-01
FEAT: Switch default frontend to Next.js app

OliverBryant:
## What & why

Switch the default web UI to the Next.js app under `frontend/`, shipped as a **static export bundled into the `xinference` package and served by the backend itself** — single process, no Node runtime needed at runtime. `pip install xinference` gives the full product again, same as the old CRA UI. This follows the approach used in xagent#732.

## How it works

**Frontend (static export)**
- `next.config.ts`: output mode is env-driven — builds default to `export`; `NEXT_OUTPUT=standalone` keeps the Node server; `next dev` sets none (export would break dynamic routes in dev). Rewrites (the dev API proxy) only apply outside export mode; in production the UI is same-origin with the API.
- Dynamic routes (`launch-model/[modelType]`, `register-model/[modelType]`, `register-model/[modelType]/[modelName]`, `running-model/[modelUid]`) are split into a server wrapper (`generateStaticParams` emits a `__shell__` placeholder page) + a client page that reads real params from the URL (`useParams`/`useSearchParams` in Suspense).
- `layout.tsx` drops request-time `cookies()`/`headers()` and the server-side cluster-auth fetch; `AppInit` now resolves `/v1/cluster/auth` client-side, and `I18nProvider` falls back to the browser language on first visit.
- Root `/` redirects to `/launch-model` client-side (server `redirect()`/config redirects are unsupported in export).

**Backend (`xinference/api/frontend_static.py`)**
- `mount_frontend()` serves `/_next` assets plus a SPA catch-all that reconstructs the route table from the export's `__shell__` files, serves RSC `.txt` prefetch payloads, guards path traversal, returns JSON 404 for unmatched API prefixes (derived from the live route table), and redirects legacy `/ui/` to `/`. No-op when the export is absent (API-only).
- `ensure_spa_fallback_last()` keeps the catch-all at the end of the route table so runtime Gradio mounts at `/{model_uid}` are not shadowed.
- Replaces the previous `XINFERENCE_FRONTEND_ENDPOINT` redirect; override the export location with `XINFERENCE_FRONTEND_DIST_DIR` if needed.

**Packaging**
- `npm run build` stages the export at `xinference/ui/web/dist` via a postbuild hook; `setup.py build_web` (hooked on install/develop/sdist, as before) runs npm and asserts the staged output. MANIFEST grafts the staged dist, so wheels/sdists are self-contained. Docker flows are unchanged (`python setup.py build_web` still works; the staged dist is untracked so `git restore .` keeps it).

## CI & tests

- Lint job: ESLint → `next build` (static export) → verify `out/` shape (`index.html`, `404.html`, `__shell__` pages) → serve the real export through FastAPI and run the routing tests.
- `xinference/api/tests/test_frontend_static.py`: unit tests on a synthetic export (no Node needed) — shell pattern building, backend-prefix 404s, RSC payloads, traversal rejection, Gradio-mount reordering.
- `xinference/api/tests/test_frontend_static_real_export.py`: integration tests against the real `frontend/out` (skipped when not built).

## Verification

- Static export build: 19/19 routes, all 4 dynamic routes emit `__shell__` shells; postbuild staging works.
- 37/37 backend tests pass (also with `--noconftest`, as the CI lint job runs them).
- Built a wheel locally: all 90 export files included under `xinference/ui/web/dist`.
- `pre-commit` (black/flake8/isort/mypy/codespell) passes; ESLint 0 errors.

## Notes

- The old CRA app under `xinference/ui/web/ui` is untouched here; it can be removed in a follow-up.
- `next@15.2.3` has a published advisory (CVE-2025-66478); worth bumping in a follow-up.

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -42,8 +42,20 @@ jobs:
         run: cd frontend && npm ci
       - name: ESLint Check
         run: cd frontend && npx eslint .
-      - name: Build frontend
+      - name: Build frontend (static export)
         run: cd frontend && npm run build
+      - name: Verify static export output
+        run: |
+          test -f frontend/out/index.html
+          test -f frontend/out/404.html
+          # Dynamic routes must emit __shell__ placeholder pages.
+          find frontend/out -name '__shell__*' | grep -q .
+      - name: Serve static export from backend
+        run: |
+          pip install fastapi httpx pytest xoscar
+          pytest --noconftest -vv \
+            xinference/api/tests/test_frontend_static.py \
+            xinference/api/tests/test_frontend_static_real_export.py
 
   build_test_job:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -36,14 +36,14 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: 16
+          node-version: 20.19.0
       # ESLint and Prettier must be in `package.json`
       - name: Install Node.js dependencies
-        run: cd xinference/ui/web/ui && npm ci
+        run: cd frontend && npm ci
       - name: ESLint Check
-        run: cd xinference/ui/web/ui && npx eslint .
-      - name: Prettier Check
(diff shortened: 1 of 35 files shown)


qinxuye (approved):
LGTM

PR #5120, opened 2026-07-02
fix(vllm): silence repeated GuidedDecodingParams info log per request

m199369309:
## Summary

- Resolve `GuidedDecodingParams` / `StructuredOutputsParams` once at module load instead of on every `async_generate()` call.
- Demote the noisy "GuidedDecodingParams not found" log from INFO to DEBUG; keep the WARNING for the no-support-at-all case.
- Re-run detection from `VLLMModel.load` after `VLLM_VERSION` is reassigned during lazy vLLM initialization.

## Root cause

vLLM renamed `GuidedDecodingParams` to `StructuredOutputsParams` somewhere in the 0.x series (confirmed in 0.22.0). The existing version gate `supports_guided = VLLM_VERSION < 1.12.0` still passes for 0.22, but `hasattr(_sampling_params, "GuidedDecodingParams")` returns `False`, so the `elif supports_guided` branch fired an INFO log on **every inference request** because the whole detection block lived inside `async_generate()`.

The `StructuredOutputsParams` fallback works correctly — the message is pure log noise with no functional impact.

## Approach

- Move the detection into a module-level `_init_guided_decoding_classes()` helper that runs once at import.
- `GuidedDecodingParams` / `StructuredOutputsParams` become module globals, read directly by `async_generate()`.
- Hook a re-detection call into `VLLMModel.load` (where `VLLM_VERSION` gets reassigned for the lazy-load case where vLLM was not importable at module import time).
- Side benefit: removes a per-request `import vllm.sampling_params` from the hot path.

## Test plan

- [x] `pre-commit run --files xinference/model/llm/vllm/core.py` (black / flake8 / isort / mypy / codespell all green)
- [x] Module imports cleanly without vLLM installed (`VLLM_INSTALLED=False` path → globals stay `None`)
- [x] Branch-coverage tests for `_init_guided_decoding_classes` (4 cases):
  - vLLM 0.22.0 with only `StructuredOutputsParams` → fallback set, `GuidedDecodingParams=None`
  - vLLM 0.10.0 with only `GuidedDecodingParams` → primary set
  - vLLM 0.5.0 (too old) → both `None`, early return
  - vLLM 1.13.0 (`supports_guided=False`) → `GuidedDecodingParams=None`, `StructuredOutputsParams` still detected
- [ ] End-to-end on a real vLLM 0.22 deployment (Ornith-1.0-35B): confirm no new INFO log per request and `guided_json` still constrains output

### DIFF
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -191,9 +191,9 @@ StructuredOutputsParams: Optional[Type[Any]] = None
 
 def _init_guided_decoding_classes() -> None:
     # Also re-invoked from VLLMModel after VLLM_VERSION is reassigned at runtime.
+    # Detect into locals and publish at the end so a concurrent async_generate
+    # coroutine (paused at an await) never observes a transient None.
     global GuidedDecodingParams, StructuredOutputsParams
-    GuidedDecodingParams = None
-    StructuredOutputsParams = None
     if not (
         VLLM_INSTALLED
         and VLLM_VERSION is not None
@@ -211,8 +211,12 @@ def _init_guided_decoding_classes() -> None:
                 VLLM_VERSION,
             )
         return
+
+    local_guided: Optional[Type[Any]] = None
+    local_structured: Optional[Type[Any]] = None
+
     if supports_guided and hasattr(_sampling_params, "GuidedDecodingParams"):
-        GuidedDecodingParams = _sampling_params.GuidedDecodingParams
+        local_guided = _sampling_params.GuidedDecodingParams
     elif supports_guided:
         logger.debug(
             "GuidedDecodingParams not found in vLLM %s, "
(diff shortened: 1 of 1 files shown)


m199369309:
Thanks @gemini-code-assist for catching this — the race is real.

In the xoscar runtime, multiple `VLLMModel` actors in the same subpool share process state, and a coroutine in `async_generate()` paused at an `await` (e.g. `_get_tokenizer`) could observe the transient `None` while `_init_guided_decoding_classes()` was resetting globals before recomputing them.

Fixed in 2cd76624: detection now resolves into locals and publishes the globals only at the end. The publish window shrinks from "entire detection pass" to two bytecode-level `STORE_GLOBAL` ops, and the early-return paths (vllm too old / `ImportError`) no longer touch the globals at all.

The residual "two assignments aren't a single atomic op" theoretical window isn't worth a lock for this code path: the read side in `async_generate()` is `if guided_params and GuidedDecodingParams: ... elif guided_params and StructuredOutputsParams:` — losing either branch mid-publish only degrades one request to "no guided decoding applied", with no corruption or crash. A `threading.Lock` would be overkill given asyncio + GIL semantics here.

qinxuye (approved):
LGTM

PR #5121, opened 2026-07-02
feat(engine): [image, audio] sglang image vllm asr engines

OliverBryant:
#4896

### DIFF
--- a/xinference/model/image/sglang/core.py
+++ b/xinference/model/image/sglang/core.py
@@ -93,12 +93,13 @@ class SGLangDiffusionModel:
         # family first to keep the caller's (possibly shared builtin) spec
         # untouched
         if model_spec is not None:
-            model_spec = model_spec.copy()
-            model_spec.model_ability = [
+            restricted_spec = model_spec.copy()
+            restricted_spec.model_ability = [
                 ability
-                for ability in (model_spec.model_ability or [])
+                for ability in (restricted_spec.model_ability or [])
                 if ability in SGLANG_SUPPORTED_ABILITIES
             ]
+            model_spec = restricted_spec
         self.model_family = model_spec
         self._model_uid = model_uid
         self._model_path = model_path
--- a/xinference/model/image/sglang/core.py
+++ b/xinference/model/image/sglang/core.py
@@ -1,0 +1,220 @@
+# Copyright 2022-2026 XProbe Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+import logging
+import random
+import re
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
+
+from ....types import LoRA
+from ..utils import handle_image_result
+
+if TYPE_CHECKING:
+    from ..core import ImageModelFamilyV2
+
+logger = logging.getLogger(__name__)
+
+# Image models verified against the sglang-diffusion compatibility matrix:
+# https://docs.sglang.io/docs/sglang-diffusion/compatibility_matrix
+SGLANG_SUPPORTED_IMAGE_MODELS = (
+    "Qwen-Image",
+    "Qwen-Image-2512",
+    "Z-Image",
+    "Z-Image-Turbo",
+    "FLUX.1-dev",
+)
+
+
+def _filter_kwargs_by_dataclass_fields(
+    kwargs: Dict[str, Any], dataclass_type: Any, purpose: str
+) -> Dict[str, Any]:
+    import dataclasses
+
+    valid_keys = {f.name for f in dataclasses.fields(dataclass_type)}
+    dropped = sorted(set(kwargs) - valid_keys)
+    if dropped:
+        logger.info("Dropping args unsupported by SGLang %s: %s", purpose, dropped)
+    return {k: v for k, v in kwargs.items() if k in valid_keys}
+
+
+class SGLangDiffusionModel:
+    def __init__(
+        self,
+        model_uid: str,
+        model_path: Optional[str] = None,
+        device: Optional[str] = None,
+        lora_model: Optional[List[LoRA]] = None,
+        lora_load_kwargs: Optional[Dict] = None,
+        lora_fuse_kwargs: Optional[Dict] = None,
+        model_spec: Optional["ImageModelFamilyV2"] = None,
+        gguf_model_path: Optional[str] = None,
+        lightning_model_path: Optional[str] = None,
+        **kwargs,
+    ):
+        if gguf_model_path:
+            raise ValueError(
+                "GGUF quantization is not supported by the SGLang image engine, "
+                "please use the diffusers engine instead"
+            )
+        if lightning_model_path:
+            raise ValueError(
+                "Lightning LoRA acceleration is not supported by the SGLang image "
+                "engine, please use the diffusers engine instead"
+            )
+        if lora_model:
+            raise ValueError(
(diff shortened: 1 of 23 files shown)


xinference/model/image/sglang/core.py
```diff
@@ -0,0 +1,220 @@
...
+                "Controlnet is not supported by the SGLang image engine, "
+                "please use the diffusers engine instead"
+            )
+        self.model_family = model_spec
+        self._model_uid = model_uid
+        self._model_path = model_path
+        self._device = device
+        self._model = None
+        self._model_spec = model_spec
+        self._abilities = model_spec.model_ability or []  # type: ignore
```
qinxuye, line 112: The SGLang model copies the full ability list from the diffusers model spec, but this implementation only provides `text_to_image`. All five registered SGLang models advertise `image2image`, and three also advertise `inpainting`, so the running instance is reported as supporting endpoints that will fail with `AttributeError` (and may even be auto-selected by the image-edit API). Please either implement those methods for SGLang or expose only `text2image` for this engine.

OliverBryant: Fixed in c00117c07: both the SGLang and vLLM image engines now restrict the advertised abilities to `text2image` on an engine-local copy of the model family, so `to_description()` / `list_models` no longer report image2image or inpainting for these engines and the image-edit API will not auto-select them. The shared builtin spec stays untouched, with tests covering both engines.

OliverBryant:
## End-to-end benchmark: Z-Image on diffusers / vLLM / SGLang

Ran all three image engines for real on a GPU box against this PR's HEAD, same prompt and generation config, measuring client-side wall-clock of `POST /v1/images/generations` (3 timed runs per engine after warmup, one engine loaded at a time).

**Setup**
- GPU: 1x NVIDIA GeForce RTX 3090 Ti (24 GB), model pinned to a single GPU via `gpu_idx`
- Host env: Python 3.12, torch 2.11.0, vllm 0.24.0, sglang 0.5.7, transformers 5.14.1
- Model: `Z-Image`, defaults from the builtin spec (`num_inference_steps=28`, `guidance_scale=4`), `1024*1024`, `n=1`, `response_format=b64_json`
- Prompt: `A red bicycle leaning against a weathered stone wall at golden hour, cinematic lighting`
- Each engine runs in its own model virtualenv (vLLM venv resolves the `vllm-omni==0.24.*` + `vllm==0.24.*` pair from the spec; SGLang venv is self-contained with `sglang[diffusion]`'s own torch 2.9.1 stack)

**Single-request latency**

| Engine | Run 1 | Run 2 | Run 3 | Avg | vs diffusers |
|---|---|---|---|---|---|
| diffusers | 48.9 s | 49.7 s | 51.2 s | **49.9 s** | 1.00x |
| vLLM (vllm-omni 0.24.0) | 40.3 s | 41.3 s | 42.6 s | **41.4 s** | **1.21x** |
| SGLang (sglang[diffusion] 0.5.7) | 43.9 s | 44.9 s | 46.6 s | **45.2 s** | **1.10x** |

**Concurrency: 4 simultaneous requests** (fired at once after warmup)

| Engine | Request latencies (sorted) | Makespan | Per-image under load | vs its own serial |
|---|---|---|---|---|
| diffusers | 53.3 / 108.4 / 164.9 / 223.4 s | **223.4 s** | 55.9 s | +12% |
| vLLM | 40.2 / 81.3 / 123.7 / 167.4 s | **167.4 s** | 41.9 s | +1% |
| SGLang | 44.2 / 89.6 / 136.9 / 185.9 s | **185.9 s** | 46.5 s | +3% |

Reading the numbers:
- **vLLM** is the only engine that accepts requests concurrently: the actor no longer serializes calls (`allow_batch=True`) and every request is submitted to vLLM-Omni's engine queue immediately, with a single dispatcher routing outputs back by request id. Z-Image's pipeline does not support request batching in vllm-omni, so GPU execution is still one-at-a-time — the win is zero added overhead under load (41.9 s vs 41.4 s serial), immediate admission, and per-request error isolation rather than throughput multiplication.
- **diffusers** and **SGLang** serialize at the model actor: concurrent callers queue outside the engine. They still complete correctly, with diffusers paying ~12% per-image overhead under load.
- Net effect for 4 requests: vLLM finishes the batch 1.33x faster than diffusers and 1.11x faster than SGLang.

**Cross-request GPU batching: sd3.5-medium** (vllm-omni pipeline with `supports_request_batch=True`)

Z-Image's vllm-omni pipeline declares `supports_request_batch = False` ([pipeline_z_image.py](https://github.com/vllm-project/vllm-omni/blob/main/vllm_omni/diffusion/models/z_image/pipeline_z_image.py)), and the engine refuses `max_num_seqs > 1` for such pipelines, so its GPU execution is serial upstream. To verify the dispatcher composes with true request batching, sd3.5-medium (SD3 pipeline, `supports_request_batch = True`) was launched with `max_num_seqs=4` and hit with 4 simultaneous requests:

| | Latencies | Makespan |
|---|---|---|
| single request | 5.5 s | -- |
| 4 concurrent (steady state) | 5.6 / 21.46 / 21.48 / 21.48 s | **21.5 s** |

Requests in the same wave complete with identical timestamps (21.478/21.460/21.478 s) -- they went through the DiT as one batched tensor -- and the engine logs `[RequestBatch] engine init max_num_seqs=4`. On this compute-saturated 3090 Ti the batched throughput matches serial FLOPs, so the makespan equals serial time; on GPUs where a single request underutilizes compute, the same path yields real throughput gains. sd3.5-medium is now registered for the vLLM engine.

All three engines produced valid, prompt-faithful 1024x1024 images (visually verified).

**Bugs found and fixed while testing** (pushed to this branch):
- f0c0293c1 — vLLM engine failed to launch: the concurrent-dispatch locks made the model instance unpicklable for the actor-subprocess handoff (`cannot pickle '_thread.lock' object`)
- 0e75c686a — SGLang generation crashed with `Object of type UUID is not JSON serializable`: xinference injects `request_id` as `uuid.UUID` and SGLang's `SamplingParams` accepts the field but its runtime JSON-serializes it; now stringified
- 44495568e — SGLang generation crashed with `No such file or directory: 'ninja'`: the ninja wheel installs the binary into the venv's bin directory, which was not on PATH in the model subprocess; now prepended at load time

**Environment notes (outside this PR's scope)**
- The SGLang venv install currently needs `XINFERENCE_VIRTUAL_ENV_SKIP_INSTALLED=0` until xorbitsai/xoscar#193 lands: with skip-installed on, a bare host `sglang` is treated as satisfying `sglang[diffusion]` and the venv comes up without the diffusion runtime
- On hosts with transformers >= 5 the `huggingface-hub<1.0` pin in `ENGINE_VIRTUALENV_PACKAGES["diffusers"]` cannot co-resolve; the box was tested with it relaxed to `<2.0` (pre-existing on main, not touched here)

xinference/model/image/vllm/core.py
```diff
@@ -0,0 +1,474 @@
...
+                continue
+            self._route_message(msg)
+
+    def _route_message(self, msg: Any) -> None:
+        from vllm_omni.engine.messages import ErrorMessage, OutputMessage
+
+        if isinstance(msg, ErrorMessage):
+            error = RuntimeError(f"vLLM-Omni engine error: {msg.error}")
+            request_id = getattr(msg, "request_id", None)
+            if getattr(msg, "fatal", False) or request_id is None:
```
qinxuye: A fatal vLLM-Omni error currently only fails the existing waiters; it does not mark the model as closed or shut down the unusable engine. The next request therefore passes the `_closed` check, gets submitted to a dead backend, and may wait forever for output. Please transition the model to a terminal state on fatal `ErrorMessage`s and dispatcher failures (atomically with failing the waiters), so subsequent requests fail immediately, and add a regression test that submits another request after a fatal engine error.

OliverBryant: Fixed in 10ff57948: fatal `ErrorMessage`s and dispatcher-loop failures now go through a `_fail_and_close` terminal transition that atomically marks the model closed, records the fatal error, drains and fails the in-flight waiters, and shuts the engine down. Subsequent `_submit_and_wait`/`text_to_image` calls fail immediately with the recorded cause instead of being submitted to the dead backend. Non-fatal unroutable errors still only fail the in-flight waiters. Added regression tests that submit another request after a fatal engine error and after a dispatcher failure.

qinxuye (approved):
LGTM

PR #5125, opened 2026-07-03
ci: fix Metal CI failures with latest MLX on macOS 26 runners

OliverBryant:
## Summary

`macos-latest` runners migrated to macOS 26, and the Metal CI job started failing across every suite. The root cause is not an OS or dependency-resolution issue: mlx >= 0.31 binds lazy arrays and streams to the OS thread that creates them, and several of our MLX call paths touch the same model from different threads. This PR fixes those call paths so the Metal job runs green on `macos-latest` with the **latest unpinned** mlx / mlx-lm / mlx-vlm.

### Fixes

- **Audio models (whisper / f5-tts / kokoro)**: `ModelActor` runs synchronous model methods via `asyncio.to_thread`, so `load()` and later inference calls could land on different threads and crash the worker with `RuntimeError: There is no Stream(gpu, N) in current thread`. Added `MLXModelThreadMixin` to route every call for a model instance through one persistent single-thread executor.
- **Distributed pipeline parallelism**: shard-to-shard tensor exchange constructed `mx.array` payloads on the actor event-loop thread, so the receiving compute thread's `mx.eval` failed with the same stream error. The failing shard's exception then sat in an unawaited task while its peer blocked forever on the exchange — on CI this surfaced as a silent 50-minute hang in `test_distributed_model` with no traceback. Tensors now cross shards as numpy: the sender converts on its compute thread (forcing evaluation on the stream that owns the graph) and the receiver converts back on its own compute thread.
- **`test_distributed_model` robustness**: await the two shard generations with `asyncio.gather` so a single-shard failure surfaces immediately instead of deadlocking the peer, and wrap `xo.create_actor_pool` / `xo.create_actor` with timeouts so a stuck fork fails fast instead of eating the whole pytest timeout.

### CI workflow

- Keep the runner on `macos-latest` and the MLX stack unpinned (latest mlx / mlx-lm / mlx-vlm).
- Drop the stale `transformers<4.54` pin, which conflicts with mlx-lm >= 0.31's `transformers>=5.0.0` requirement.
- Chain the four Metal suites explicitly with `&&`.

### Verification

The distributed hang was reproduced and the fix verified locally on macOS 26 / Apple Silicon with mlx 0.32.0 + mlx-lm 0.31.3 (the exact versions the runner resolves): a standalone equivalent of `test_distributed_model` hangs before the change and completes after. The full Metal job (`build_test_job (macos-latest, metal, 3.10)`) is green on CI with all four suites passing.

### DIFF
--- a/xinference/model/llm/mlx/distributed_models/core.py
+++ b/xinference/model/llm/mlx/distributed_models/core.py
@@ -30,6 +30,30 @@ logger = logging.getLogger(__name__)
 DEBUG_DISTRIBUTED_MLX = bool(int(os.getenv("XINFERENCE_DEBUG_DISTRIBUTED_MLX", "0")))
 
 
+def _to_wire(result: "mx.array"):
+    """Encode an mx.array for the inter-shard exchange.
+
+    Must run on the compute thread: np.array() evaluates and copies there,
+    on the stream that owns the pending graph. numpy has no bfloat16 and
+    np.array() on a bfloat16 mx.array raises, so ship its raw bits as
+    uint16 and let _from_wire restore the dtype.
+    """
+    if result.dtype == mx.bfloat16:
+        return np.array(result.view(mx.uint16)), True
+    return np.array(result), False
+
+
+def _from_wire(payload) -> "mx.array":
+    """Decode an inter-shard payload back into an mx.array.
+
+    Must run on the compute thread so the new mx.array belongs to the
+    stream that will evaluate it.
+    """
+    data, is_bf16 = payload
+    result = mx.array(data)
+    return result.view(mx.bfloat16) if is_bf16 else result
+
+
 class ReceiverActor(xo.StatelessActor):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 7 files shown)


xinference/model/llm/mlx/distributed_models/core.py
```diff
@@ -90,12 +94,17 @@ def _send_stage_result(self, result: "mx.array"):
                 "Start to send %s partial result to rank %d", self.model_uid, last_rank
             )
 
+        # np.array() evaluates and copies on this (compute) thread, whose
+        # stream owns the pending graph. mx.array payloads must never cross
+        # threads on mlx>=0.31.
+        payload = np.array(result)
```
qinxuye: This conversion does not support MLX `bfloat16` arrays. With MLX 0.32.0, `np.array(mx.array(..., dtype=mx.bfloat16))` raises a `RuntimeError`, and the built-in Qwen3 MLX specs explicitly support the `bf16` quantization. A multi-worker Qwen3 bf16 launch will therefore fail here. Please preserve dtype metadata and encode bfloat16 through a supported representation (for example, its `uint16` bits), restore it on the receiver, and add a regression test.

qinxuye (approved):
LGTM

PR #5129, opened 2026-07-03
fix(llm): normalize tool_call arguments to dict for Qwen3 family templates

m199369309:
## Summary

The builtin chat templates for `Qwen3-Coder`, `qwen3.5`, and `qwen3.6`
iterate over `tool_call.arguments|items`, which crashes when `arguments`
is a JSON-encoded string (the OpenAI spec format) rather than a dict.
Affected requests fail with `Jinja2 error: Can only get item pairs from a
mapping` on any turn that includes a previous tool call.

This PR normalizes string-typed `arguments` to dict at the message layer
in `ChatModelMixin.get_full_context`, before template rendering. The
detection is content-driven (matches `tool_call.arguments|items` in the
template) rather than name-driven, so future templates copied from this
family automatically benefit.

## Background

OpenAI requires `tool_calls[].function.arguments` to be a JSON-encoded
string. The Qwen3-Coder style templates iterate over it with `|items`,
which only works on mappings. The HF Jinja sandbox does not register a
`from_json` filter, so the fix cannot live in the templates themselves.

## Approach

- `_chat_template_needs_dict_arguments(chat_template)` — detects the
  `tool_call.arguments|items` pattern via substring match (lru_cached).
- `_normalize_tool_call_arguments_to_dict(messages)` — for each assistant
  message with `tool_calls`, `json.loads` string-typed `arguments` into a
  dict. Malformed JSON is left as-is so downstream surfaces a clear error.
- Both are invoked at the top of `get_full_context`, before the
  vision/audio branch (so vision models in the same family are covered).

## Edge cases

| Input | Behavior |
|---|---|
| `arguments` is valid JSON string | parsed to dict |
| `arguments` is dict | passthrough (no change) |
| `arguments` is empty string | skipped (preserved as \`\"\"\`) |
| `arguments` is malformed JSON | skipped (preserved; downstream surfaces clear error) |
| `arguments` is None / missing | skipped |
| Non-assistant message with `tool_calls` | skipped |

## Test plan

- [x] \`pytest -vv xinference/model/llm/tests/test_utils.py::test_normalize_tool_call_arguments_to_dict\`
- [x] \`pytest -vv xinference/model/llm/tests/test_utils.py::test_qwen3_family_get_full_context_handles_string_arguments\`
- [x] \`pre-commit run --files xinference/model/llm/utils.py xinference/model/llm/tests/test_utils.py\`
- [ ] Manual: chat completion with a tool-call turn against a Qwen3-Coder model

## Out of scope

- Modifying the builtin templates themselves (HF sandbox lacks \`from_json\`)
- Backfilling \`from_json\` filter via monkey-patch (high coupling risk)
- Non-builtin custom templates (user-supplied templates must still handle
  string arguments themselves)

### DIFF
--- a/xinference/model/llm/utils.py
+++ b/xinference/model/llm/utils.py
@@ -147,6 +147,44 @@ class ChatModelMixin:
         jinja_env.globals["raise_exception"] = raise_exception
         return jinja_env.from_string(chat_template)
 
+    @staticmethod
+    @functools.lru_cache(maxsize=64)
+    def _chat_template_needs_dict_arguments(chat_template: Optional[str]) -> bool:
+        # Detect Coder-style templates that iterate `tool_call.arguments|items`.
+        # Content-driven (not name-driven) so future models copying this
+        # template style are covered automatically.
+        return chat_template is not None and (
+            "tool_call.arguments|items" in chat_template
+        )
+
+    @staticmethod
+    def _normalize_tool_call_arguments_to_dict(messages: List[Dict]) -> List[Dict]:
+        # OpenAI spec sends tool_calls.function.arguments as a JSON-encoded
+        # string, but Coder-style templates (Qwen3-Coder / qwen3.5 / qwen3.6)
+        # require a dict to iterate via `|items`. The HF Jinja sandbox does
+        # not register a `from_json` filter, so we normalize at the message
+        # layer before template rendering.
+        for message in messages:
+            if not isinstance(message, dict) or message.get("role") != "assistant":
+                continue
+            tool_calls = message.get("tool_calls")
+            if not isinstance(tool_calls, list):
+                continue
+            for tc in tool_calls:
+                if not isinstance(tc, dict):
+                    continue
+                fn = tc.get("function")
+                if not isinstance(fn, dict):
+                    continue
+                args = fn.get("arguments")
+                if isinstance(args, str) and args:
+                    try:
+                        fn["arguments"] = json.loads(args)
+                    except json.JSONDecodeError:
+                        # leave as-is so downstream surfaces the malformed JSON
+                        pass
+        return messages
+
     def _build_from_raw_template(
         self, messages: List, chat_template: str, **kwargs
     ) -> str:
@@ -164,6 +202,8 @@ class ChatModelMixin:
         tokenize=False,
         **kwargs,
     ):
+        if self._chat_template_needs_dict_arguments(chat_template):
+            messages = self._normalize_tool_call_arguments_to_dict(messages)
         if (
             "vision" not in self.model_family.model_ability
             and "audio" not in self.model_family.model_ability
--- a/xinference/model/llm/utils.py
+++ b/xinference/model/llm/utils.py
@@ -180,8 +180,14 @@ class ChatModelMixin:
                 normalized.append(message)
                 continue
             # Identify indices that need rewriting (string arguments that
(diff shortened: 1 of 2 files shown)


m199369309:
Thanks @gemini-code-assist for the review — the in-place mutation concern is valid.

Pushed fix in bf6b437c: `_normalize_tool_call_arguments_to_dict` is now non-mutating. It deep-copies only the affected message + tool_call + function dict when a string argument is successfully parsed; messages without string arguments are returned by reference (no copy in the common case).

Tests updated:
- `test_normalize_tool_call_arguments_to_dict` now also asserts the input messages are unchanged post-call.
- `test_qwen3_family_get_full_context_handles_string_arguments` drops the `[dict(m) for m in base_messages]` shallow-copy workaround and instead asserts the input is unchanged.

Re: the 3 failing CI jobs (macos-3.10 `test_log_rotation`, windows-3.13 `test_worker` socket bind, macos-metal-3.10 `test_whisper_mlx`) — these are pre-existing flaky/environmental tests unrelated to the chat_template fix. The PR's own tests (`test_normalize_tool_call_arguments_to_dict`, `test_qwen3_family_get_full_context_handles_string_arguments`) pass on all platforms.

xinference/model/llm/utils.py
```diff
@@ -147,6 +147,67 @@ def raise_exception(message):
...
+                if not isinstance(tc, dict):
+                    continue
+                fn = tc.get("function")
+                if not isinstance(fn, dict):
+                    continue
+                args = fn.get("arguments")
+                if not (isinstance(args, str) and args):
+                    continue
+                try:
+                    parsed = json.loads(args)
```
qinxuye, line 202: After `json.loads(args)` succeeds, this should still verify that the parsed value is a dict before rewriting `arguments`. JSON strings such as `"[]"`, `"null"`, or `"1"` parse successfully, but the Qwen templates still call `tool_call.arguments|items` and will fail with the same `Can only get item pairs from a mapping` error. This should match the existing `_normalize_tool_calls` contract and only accept JSON objects, either by raising a clear error or by leaving non-object parsed values unmodified.

m199369309:
Thanks @qinxuye for the catch — valid bug. Pushed fix in b3959182.

**Root cause:** `json.loads(args)` succeeds for `"[]"`, `"null"`, `"1"`, etc., producing list / None / int. My code wrote these directly into `arguments`, but the Qwen templates iterate `tool_call.arguments|items` which only works on mappings — same crash as the original string-arguments bug.

**Fix:** After `json.loads`, check `isinstance(parsed, dict)`. Non-object parsed values are left as-is (original string preserved), letting the template surface the error naturally. This matches the contract of the existing `_normalize_tool_calls` helper.

**Test coverage:** `test_normalize_tool_call_arguments_to_dict` now covers 3 new edge cases — JSON list (`"[1, 2, 3]"`), JSON null (`"null"`), JSON number (`"42"`) — all expected to be preserved as-is. Both unit and integration tests pass locally; pre-commit clean.

Re: the 4 failing CI jobs (macOS 3.10/3.13 `test_cancel` model download timeout, macOS metal `test_whisper_mlx` ConnectionRefused, Windows 3.13 `test_metrics_exporter_server` supervisor unavailable) — all pre-existing flaky/environmental tests unrelated to this PR. The PR's own tests pass on all platforms; Ubuntu + Windows 3.10 + lint all green.

qinxuye (approved):
LGTM

PR #5130, opened 2026-07-03
feat(vllm): serialize vLLM engine construction to avoid flashinfer JIT deadlock

m199369309:
## Summary

When deploying MoE architectures (e.g. \`Qwen3_5MoeForConditionalGeneration\`)
on GPU architectures without prebuilt flashinfer wheels (e.g. Blackwell
sm_120), multiple worker processes concurrently trigger JIT compilation
of fused_moe kernels during \`AsyncLLMEngine.from_engine_args\` (cuda graph
capture phase). flashinfer serializes JIT via \`filelock\`; if the lock
holder enters a CUDA/cutlass resource deadlock, the remaining workers
retry forever, resulting in a fleet-wide \`futex_wait_queue\` deadlock
within minutes of launch.

This PR adds a \`_serialize_flashinfer_jit()\` context manager on
\`VLLMModel\` that wraps the three \`from_engine_args\` call sites so
engine construction is serialized per-architecture.

## Approach

1. **Architecture-gated**: only MoE arches in \`_FLASHINFER_WARMUP_ARCHES\`
   frozenset (currently \`Qwen3_5MoeForConditionalGeneration\`)
2. **Cross-process lock**: \`fcntl.flock\` on
   \`{XINFERENCE_CACHE_DIR}/flashinfer_warmup/{arch}.lock\`
3. **Thread-safe timeout**: uses \`LOCK_NB\` + \`time.sleep\` polling
   (NOT \`signal.SIGALRM\`) so it works inside \`_loading_thread\` which
   runs in a non-main thread
4. **Failure mode**: on timeout, logs warning and proceeds without the
   lock (engine construction MUST proceed regardless)
5. **Kill switch**: \`XINFERENCE_DISABLE_FLASHINFER_WARMUP=1\`
6. **Windows guard**: \`if os.name == \"nt\": yield; return\`

The first process to acquire the lock compiles and fills the flashinfer
cache; subsequent processes hit the cache and pass through quickly
without racing on flashinfer's internal \`fused_moe_120.lock\`.

## Design decisions

| Decision | Choice | Rationale |
|---|---|---|
| Lock impl | \`fcntl.flock(LOCK_EX \\| LOCK_NB)\` (stdlib) | No new dep; thread-safe polling |
| Lock path | \`{cache_dir}/flashinfer_warmup/{arch}.lock\` | Per-arch, not per-model (flashinfer cache is keyed by GPU arch + kernel variant) |
| Timeout impl | \`LOCK_NB\` + \`sleep\` polling | \`signal.alarm()\` does not work in non-main threads (\`_loading_thread\`) |
| Failure mode | log warning + proceed | Engine construction MUST proceed regardless of lock acquisition |
| Kill switch | env var | Allows quick disable for debugging |
| Windows | early return | fcntl unavailable on Windows |

## Test cases

| Test | Verifies |
|---|---|
| \`test_serialize_skips_unsupported_arch\` | Non-MoE arch passthrough; no lock dir created |
| \`test_serialize_acquires_lock_for_qwen3_5_moe\` | MoE arch creates & releases lock file |
| \`test_serialize_disabled_via_env\` | Env var disables lock creation |
| \`test_serialize_releases_lock_on_exception\` | Lock released even if engine construction raises |
| \`test_serialize_yields_without_engine\` | Context manager yields regardless of \`self._engine\` state |

## Test plan

- [x] \`pytest -vv xinference/model/llm/vllm/tests/test_core_chat_model.py::TestFlashinferWarmup\` (5/5 passed)
- [x] \`pre-commit run --files xinference/model/llm/vllm/core.py xinference/model/llm/vllm/tests/test_core_chat_model.py\`
- [ ] Multi-instance PoC on Blackwell sm_120 with \`Qwen3_5MoeForConditionalGeneration\` (tracked separately; see follow-up issue)

## Out of scope

- Generalizing to other MoE architectures (extensible via the frozenset)
- Making warmup prompts configurable (YAGNI for now)
- Integrating with external precompile tools (follow-up)
- supervisor hang-detection tuning (only if construction > 600s in practice)

### DIFF
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -24,6 +24,7 @@ import sys
 import threading
 import time
 import uuid
+from contextlib import contextmanager
 from functools import partial
 from typing import (
     TYPE_CHECKING,
@@ -163,6 +164,27 @@ except ImportError:
     VLLM_VERSION = None
 
 DEFAULT_VLLM_VERSION = version.parse("0.21.0")
+
+# Architectures known to trigger flashinfer fused_moe JIT deadlock on
+# Blackwell sm_120 when multiple replicas launch concurrently. The lock
+# serializes engine construction (which includes cuda graph capture +
+# JIT) so the first process fills the flashinfer cache and subsequent
+# processes hit the cache without racing on flashinfer's internal
+# `fused_moe_120.lock`.
+_FLASHINFER_WARMUP_ARCHES = frozenset(
+    {
+        "Qwen3_5MoeForConditionalGeneration",
+    }
+)
+
+# Env var to disable the serialization for debugging / fallback.
+_DISABLE_WARMUP_ENV = "XINFERENCE_DISABLE_FLASHINFER_WARMUP"
+
+# Per-process lock-acquisition timeout (seconds). Cold compile on
+# Blackwell sm_120 + MoE-256 may take 30+ min; cache hits finish in <1s.
+_FLASHINFER_WARMUP_TIMEOUT = int(
+    os.getenv("XINFERENCE_FLASHINFER_WARMUP_TIMEOUT", "1800")
+)
 
 
 def _get_effective_vllm_version() -> version.Version:
@@ -483,6 +505,115 @@ class VLLMModel(LLM):
         # For older versions, check the environment variable
         return envs.is_set("VLLM_USE_V1") and envs.VLLM_USE_V1
 
+    @contextmanager
+    def _serialize_flashinfer_jit(self):
+        """Cross-process lock to serialize vLLM engine construction.
+
+        Why: Qwen3_5MoeForConditionalGeneration (qwen3.5 / qwen3.6) on
+        Blackwell sm_120 triggers flashinfer fused_moe JIT during
+        ``AsyncLLMEngine.from_engine_args`` (cuda graph capture phase),
+        not on first inference. With multi-instance launches (e.g.
+        2 replicas x TP=4 = 8 worker processes), all processes race on
+        flashinfer's internal ``fused_moe_120.lock`` and deadlock. This
+        wrapper forces serialization around the engine-construction call
+        so the first process compiles and fills the cache, and subsequent
+        processes get cache hits without lock contention.
+
+        Non-target architectures / disabled via env: yield without locking.
+
+        Thread-safety: uses ``fcntl.LOCK_NB`` + ``time.sleep`` polling
+        (NOT ``signal.SIGALRM``) so it works inside ``_loading_thread``
+        which runs in a non-main thread.
+
+        Failure mode: if the lock cannot be acquired within timeout, the
+        engine is constructed without the lock (falls back to original
+        racing behavior). Engine construction MUST proceed regardless.
+        """
+        if os.name == "nt":
+            yield
+            return
+
+        if os.getenv(_DISABLE_WARMUP_ENV) == "1":
+            logger.debug(
+                "flashinfer jit serialize skipped via %s for model %s",
+                _DISABLE_WARMUP_ENV,
+                self.model_uid,
+            )
+            yield
+            return
+
+        architectures = getattr(self.model_family, "architectures", []) or []
+        if not any(a in _FLASHINFER_WARMUP_ARCHES for a in architectures):
+            yield
+            return
+
+        import fcntl
+
+        from ....constants import XINFERENCE_CACHE_DIR
+
+        matched_arch = next(a for a in architectures if a in _FLASHINFER_WARMUP_ARCHES)
+        lock_dir = os.path.join(XINFERENCE_CACHE_DIR, "flashinfer_warmup")
+        os.makedirs(lock_dir, exist_ok=True)
+        # Keyed by architecture name (not model_uid): flashinfer cache is
+        # keyed by GPU arch + kernel variant, so same arch -> same cache ->
+        # same lock is correct.
+        lock_path = os.path.join(lock_dir, f"{matched_arch}.lock")
+
+        logger.info(
+            "flashinfer jit serialize START model=%s arch=%s lock=%s",
+            self.model_uid,
+            matched_arch,
+            lock_path,
+        )
+        t0 = time.time()
+        fd = os.open(lock_path, os.O_CREAT | os.O_RDWR, 0o644)
+        lock_acquired = False
+        try:
+            # Poll with LOCK_NB instead of blocking LOCK_EX: fcntl.flock has
+            # no native timeout, and signal.alarm() does not work in non-main
+            # threads (ValueError: signal only works in main thread of the
+            # main interpreter). LOCK_NB + sleep is fully thread-safe.
+            deadline = t0 + _FLASHINFER_WARMUP_TIMEOUT
+            while True:
+                try:
+                    fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                    lock_acquired = True
+                    break
+                except (BlockingIOError, OSError):
+                    if time.time() >= deadline:
+                        logger.warning(
+                            "flashinfer jit serialize TIMEOUT model=%s "
+                            "after %.1fs - continuing without lock "
+                            "(may race on flashinfer cache)",
+                            self.model_uid,
+                            time.time() - t0,
+                        )
+                        break
+                    time.sleep(0.5)
+
+            if lock_acquired:
+                logger.info(
+                    "flashinfer jit serialize LOCK ACQUIRED model=%s waited=%.1fs",
+                    self.model_uid,
+                    time.time() - t0,
+                )
+
+            try:
+                yield
+            finally:
+                if lock_acquired:
+                    try:
+                        fcntl.flock(fd, fcntl.LOCK_UN)
+                    except Exception:
+                        pass
+        finally:
+            os.close(fd)
+            logger.info(
+                "flashinfer jit serialize DONE model=%s elapsed=%.1fs",
+                self.model_uid,
+                time.time() - t0,
+            )
+
     def load(self):
         try:
             import vllm
@@ -670,9 +801,12 @@ class VLLMModel(LLM):
                                         loop=loop,
                                     )
 
-                            self._engine = XinferenceAsyncLLMEngine.from_engine_args(
-                                engine_args
-                            )
+                            with self._serialize_flashinfer_jit():
+                                self._engine = (
+                                    XinferenceAsyncLLMEngine.from_engine_args(
+                                        engine_args
+                                    )
+                                )
                         else:
                             from vllm.v1.executor.abstract import Executor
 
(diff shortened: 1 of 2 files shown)


m199369309:
Thanks @gemini-code-assist for the review — all 4 suggestions applied in 9ff27ce6.

**High priority fixes:**

1. **`errno` import** (core.py): Added `import errno` alongside `import fcntl` so we can distinguish error codes.

2. **Filesystem-not-supported detection** (core.py): `fcntl.flock` raising `OSError` with errno NOT in `(EAGAIN, EWOULDBLOCK, EACCES)` is now treated as a non-retryable "filesystem doesn't support locking" error (e.g. NFS returning `ENOTSUP` / `ENOSYS`). We log a warning and proceed immediately without the lock instead of hanging for the full `_FLASHINFER_WARMUP_TIMEOUT` (default 1800s) — a serious operational hazard on network filesystems.

**Medium priority test improvements:**

3. **`test_serialize_releases_lock_on_exception`** now asserts re-entry takes <1s. Previously, a leaked lock would still pass the test because the 2s short timeout would eventually fire — the new timing assertion catches this case directly.

4. **New `test_serialize_unsupported_filesystem`** mocks `fcntl.flock` to raise `OSError(ENOTSUP, ...)` and asserts the context manager proceeds within 1s instead of hanging for 1800s.

All 6 `TestFlashinferWarmup` tests pass locally; pre-commit clean.

Re: the single failing CI job (`macos-latest, metal, 3.10` — `test_whisper_mlx.py::test_restful_api_for_whisper` ConnectionRefused): this is a pre-existing flaky audio-transcription test on macOS metal, unrelated to the vLLM flashinfer warmup hook (which lives in `xinference/model/llm/vllm/`). All other 9 CI jobs pass on Ubuntu / Windows / macOS non-metal.

xinference/model/llm/vllm/core.py
```diff
@@ -670,9 +823,12 @@ def _get_executor_cls(
                                         loop=loop,
                                     )
 
-                            self._engine = XinferenceAsyncLLMEngine.from_engine_args(
-                                engine_args
-                            )
+                            with self._serialize_flashinfer_jit():
```
qinxuye, line 826: This outer lock serializes concurrent `from_engine_args` calls, but it does not appear to serialize the vLLM worker processes that actually run `load_model` inside a single engine construction. In the custom distributed executor, `from_engine_args` creates all workers and `_run_workers("load_model")` dispatches the worker calls concurrently, so a single TP=4 Qwen3.5/Qwen3.6 launch can still have several worker processes racing on flashinfer JIT while this parent-side lock is held. If the deadlock is caused by worker-level concurrent JIT, this only reduces concurrency between replicas and may not eliminate the race. The lock (or another serialization mechanism) likely needs to sit on the worker `load_model`/JIT path for the target architecture, or worker loading needs to be serialized for that path.

m199369309:
## Update: Root Cause Confirmed — v1 Lock Position Is Wrong, Real Fix Is AOT Wheel

Following @qinxuye's feedback that the supervisor-side lock doesn't serialize worker-process JIT, I ran a full investigation on the production worker (Blackwell RTX 6000D sm_120f, CUDA 13.0, flashinfer 0.6.11.post2). **The v1 approach in this PR is confirmed ineffective, and the real root cause is different from what we assumed.**

### Real Root Cause: flashinfer JIT Compilation Failure on sm_120 (Not Lock Contention)

**Investigation steps:**

1. **flashinfer source analysis** (`flashinfer/jit/core.py:415-451`, tag `v0.6.11.post2`):
   - `FLASHINFER_JIT_VERBOSE=1` silently enables debug flags (`-O0 --device-debug -DCUTLASS_DEBUG_TRACE_LEVEL=2`) due to PR [#1946](https://github.com/flashinfer-ai/flashinfer/pull/1946)'s verbose→debug fallback. PR [#3135](https://github.com/flashinfer-ai/flashinfer/pull/3135) (open) fixes this.
   - With debug flags, `ptxas` segfaults (SIGSEGV, exit 139) on CUTLASS sm_120 MoE templates at file 7/97.
   - Without debug flags (`FLASHINFER_JIT_DEBUG=0`), JIT still fails at file 18/97 with 55 namespace resolution errors (`tensorrt_llm::CutlassMoeFCRunner` etc. not found in binding file, despite symbols existing in `moe_kernels.h`).

2. **Single-process JIT test** (no concurrency, no lock contention):
   - Still fails → the original "deadlock" was NOT lock contention, but a **crash-retry loop** (22000+ lock acquisitions = 8 processes each failing fast and retrying).
   - v1's serialization cannot help because the first worker alone can't compile.

3. **Why v1's lock is in the wrong place:**
   - `_serialize_flashinfer_jit()` runs in the **xinference worker process** (uses `/data/inference/.venv`).
   - flashinfer JIT actually happens in **vLLM EngineCore subprocess** (uses the model's dedicated venv `/data/models/virtualenv/v4/qwen3.6/vllm/3.11.15`).
   - Different venvs → v1 warms the wrong flashinfer. Even if v1's lock worked, it wouldn't prevent the EngineCore subprocess from JIT-ing.

### Working Fix: AOT Prebuilt Wheel (Zero xinference Code Change)

`flashinfer-jit-cache==0.6.11.post3+cu130` ships sm_120f-compiled `.so` files. `build_and_load()` loads the AOT `.so` in **0.0s** (vs 10-30min JIT), completely bypassing the JIT path:

```python
# verify_single_jit.py — single-process test on dedicated venv
spec = gen_cutlass_fused_moe_module(nvcc_flags=[], device_arch="120")
spec.build_and_load()  # elapsed=0.0s, RESULT=SUCCESS
```

**Production deployment:** qwen3.6 (TP=4, 2 replicas) launched successfully, chat returns normal responses, no JIT processes, no `fused_moe_120.lock` contention. The original "deadlock" is eliminated.

### One Remaining Issue: vllm 0.21.0 Hard-Pins flashinfer-cubin==0.6.8.post1

vllm 0.21.0's metadata pins `flashinfer-cubin==0.6.8.post1`. xinference's venv install (with `skip_installed=True` → uv `--no-deps`) doesn't resolve deps, but flashinfer-cubin 0.6.8.post1 ends up in the venv (from residual or vllm wheel), causing version mismatch with `flashinfer-jit-cache==0.6.11.post3`.

**Current workaround:** `FLASHINFER_DISABLE_VERSION_CHECK=1` env var (officially supported by flashinfer, bypasses the version check). Verified working — flashinfer 0.6.8.post1 Python binding can load jit-cache 0.6.11.post3 `.so` (ABI compatible for fused_moe path).

### Proposed Patch: Post-Install Force-Upgrade (Seeking Maintainer Feedback)

To eliminate the `FLASHINFER_DISABLE_VERSION_CHECK=1` dependency, I'm considering a post-install hook in `xinference/core/worker.py`:

```python
# After main install_packages call
if model_engine == "vllm" and "Qwen3_5MoeForConditionalGeneration" in architectures:
    subprocess.run([
        uv_path, "pip", "install", "-p", env_path,
        "--no-deps", "--upgrade",
        "--extra-index-url", "https://flashinfer.ai/whl/cu130",
        "flashinfer-python==0.6.11.post3",
        "flashinfer-cubin==0.6.11.post3",
        "flashinfer-jit-cache==0.6.11.post3+cu130",
    ], check=True)
```

`--no-deps --upgrade` bypasses vllm 0.21.0's hard pin and force-upgrades the 3 flashinfer packages.

### Questions for Maintainers

1. **Direction check:** Should we (a) close this PR and implement the post-install hook, (b) keep v1 as a defense-in-depth layer (even though it's ineffective for the dedicated-venv case), or (c) wait for flashinfer PR #3135 to merge and upgrade flashinfer?
2. **Patch approach:** Is the post-install `subprocess.run` acceptable, or should we extend xoscar's `install_packages` to support `--upgrade`/`--no-deps` kwargs for cleaner abstraction?
3. **Scope:** Should the flashinfer force-upgrade be gated on `Qwen3_5MoeForConditionalGeneration` architecture (current proposal), or applied to all vllm models on sm_120?

Happy to open a separate PR for the post-install hook if that's the preferred direction. Full investigation details documented in `optimize/20260702/2026070208.md` (on branch `feather-customized-2-10-0`).

cc @qinxuye

qinxuye:
Thanks for the deeper investigation. Given the new findings, I do not think we should merge the current parent-side lock implementation.

If the JIT happens in the vLLM EngineCore subprocess / dedicated model virtualenv, and a single-process JIT still fails, then this lock is not fixing the real problem. Keeping it as a defense-in-depth layer would add complexity and may make the failure mode look fixed when it is not.

My preference is to close or abandon this PR and open a new PR for the actual fix: make sure the target vLLM model virtualenv gets the working flashinfer AOT packages for this architecture. Please gate it narrowly, reuse the existing virtualenv install / post-install patch path where possible, and keep the fallback behavior clear if those wheels are unavailable. We can review the exact package versions, CUDA gating, and install mechanism in that new PR.

m199369309:
Closing in favor of a new PR with the v3 approach (AOT wheel + post-install force-upgrade).

## Why closing

@qinxuye's feedback on this PR was:
> "My preference is to close or abandon this PR and open a new PR for the actual fix: make sure the target vLLM model virtualenv gets the working flashinfer AOT packages for this architecture. Please gate it narrowly, reuse the existing virtualenv install / post-install patch path where possible, and keep the fallback behavior clear if those wheels are unavailable."

The v1 approach here (supervisor-side `_serialize_flashinfer_jit` context manager) was proven wrong on two fronts:

1. **Wrong venv** — the hook ran in the xinference worker process (`/data/inference/.venv`), but the actual JIT happens in the vLLM EngineCore subprocess which uses a **different** model-dedicated venv. Pre-warming the wrong venv can't prevent JIT in the EngineCore.
2. **Single-process JIT still fails** — even without concurrency, flashinfer 0.6.8.post1's JIT fails on sm_120 due to (a) ptxas segfault from debug flags injected by `gen_jit_spec()` (PR #1946 verbose→debug fallback bug), and (b) namespace resolution failure for `tensorrt_llm::*` symbols. Serializing concurrent JIT can't fix a single-process compile failure.

## The v3 approach (new PR)

Root cause analysis: see [`dist/2026070209.md`](https://github.com/m199369309/inference/blob/main/dist/2026070209.md) (full Step 1-9 investigation) — summarized:

- **AOT pre-compiled wheel bypasses JIT entirely**: `flashinfer-jit-cache==0.6.11.post3+cu130` ships sm_120f-compiled `fused_moe_120.so`; `build_and_load()` loads the `.so` in 0.0s, no ninja/nvcc/filelock — eliminates the deadlock possibility at the root.
- **Post-install force-upgrade solves vllm's hard pin**: vllm 0.21.0 metadata hard-pins `flashinfer-cubin==0.6.8.post1`. Using `uv pip install --no-deps --upgrade` after the main install bypasses this pin.
- **Narrow gating**: `needs_flashinfer_aot()` only triggers for `model_engine == "vllm"` + `Qwen3_5MoeForConditionalGeneration` architecture.
- **Clear fallback**: if the upgrade fails (e.g., wheel unavailable), logs a warning and sets `FLASHINFER_DISABLE_VERSION_CHECK=1` — the model still starts (verified ABI-compatible: 0.6.8.post1 Python binding can load 0.6.11.post3 AOT `.so` for the `fused_moe` path).

## What's removed

The new PR deletes all of v1's code:
- `_serialize_flashinfer_jit()` context manager
- 3× `with self._serialize_flashinfer_jit():` wrappers in `vllm/core.py`
- 3 constants (`_FLASHINFER_WARMUP_ARCHES`, `_DISABLE_WARMUP_ENV`, `_FLASHINFER_WARMUP_TIMEOUT`)
- `TestFlashinferWarmup` test class (5 tests)

## End-to-end verification (production worker node)

- 2 replicas READY, chat returns normally
- No JIT processes, no `fused_moe_120.lock` contention
- `flashinfer-python`/`cubin`/`jit-cache` all at `0.6.11.post3` (version match, no env var bypass needed)
- `Post-install: flashinfer AOT upgrade SUCCEEDED — fused_moe JIT bypassed` log appears

New PR link will be added as a follow-up comment once opened.

m199369309:
Follow-up: the v3 PR is now open at https://github.com/xorbitsai/inference/pull/5137.

PR #5133, opened 2026-07-03
feat: permission/scope alignment (core — no live-read)

m199369309:
## Summary

Aligns the frontend permission list, backend route scopes, and menu/route gating so the three match 1:1. Adds an admin-target-takeover guard to prevent non-admin users from modifying admin accounts. Adds `owner_username` to the API key list response so non-admin callers can render the owner column. Fixes the `reveal_api_key` scope to match the frontend button guard.

## What changed vs the original mega-PR

Based on maintainer @qinxuye's feedback, this PR is a **scoped-down version** of the original 7-commit series. The live-read changes (C5 `/v1/users/me` + `PermissionsContext`, C7 `_get_caller_live_perms`) are **dropped** because the live-read design only worked at the UI menu layer, not at the API route layer (`Security(scopes=[...])` still checks JWT snapshot). Live-read will be revisited in a separate follow-up — see the open issue linked in the comments.

## Commits

| # | Commit | Summary |
|---|---|---|
| 1 | \`feat(ui): add JWT utils for permission checks\` | New \`utils/jwt.js\` with \`parseTokenFromSession\` / \`hasPermission\` / \`isAdmin\` helpers (base64url-safe) |
| 2 | \`feat(ui): expand ALL_PERMISSIONS list and add i18n keys\` | Add \`models:register\`, \`logs:list\`, \`monitor:view\` to user management UI + 4 locales |
| 3 | \`feat(ui): gate B-class menu items and routes by permission scope\` | Menu visibility by scope + \`PermissionGate\` route guard + monitoring config buttons admin-only |
| 4 | \`feat(api): rename route scopes to align with UI permissions\` | \`models:start/stop\` → \`models:write\`, \`models:add/unregister\` → \`models:register\`, \`admin\` → \`logs:list\` for log routes. \`SCOPE_ALIASES\` bridge keeps legacy tokens working. **BREAKING** for new-scope tokens on old deployments. |
| 5 | \`fix: update auth test fixtures for merged models:write scope + JWT base64url\` | conftest user3 \`models:start\` → \`models:write\`; remove terminate-rejected assertions (merge makes them pass); jwt.js base64url normalization |
| 6 | \`feat(api): extend admin-target-takeover guard to all sensitive write handlers\` | New \`_reject_admin_target_takeover\` prevents non-admin \`users:manage\` users from deleting / changing password / disabling / changing permissions of admin users |
| 7 | \`feat(api,ui): owner_username in list_api_keys + reveal scope fix\` | (G1) \`owner_username\` field in list_api_keys response + frontend renderCell preference; (J1) \`reveal_api_key\` scope \`admin\` → \`keys:manage\` |
| 8 | \`fix(api): align API key read route scopes with frontend contract\` | New \`require_keys_read\` dependency accepts \`keys:create\` OR \`keys:manage\`; applied to GET \`/v1/admin/keys\` and GET \`/v1/admin/keys/{id}\` |

## Background

Xinference's permission system had three structural issues:

1. **UI permissions vs backend scopes drifted**: frontend \`models:write\` was a dead permission (no route used it); backend \`models:start\` / \`models:stop\` / \`models:add\` / \`models:unregister\` were not assignable via the UI.
2. **Menu visibility was unconditional**: zero-permission users saw dead menus that 403'd on click; monitoring / logs pages were exposed to all logged-in users.
3. **Admin-target-takeover gap**: a delegated \`users:manage\` operator could delete an admin account, change an admin's password, or disable the only admin — locking the deployment out.

This PR addresses all three. The alias bridge (commit 4) ensures existing tokens carrying legacy scope names keep working during the transition period; the bridge is scheduled for removal in a future release.

## ⚠️ Breaking change (commit 4)

Tokens carrying only the new scope names (\`models:write\`, \`models:register\`, \`logs:list\`) issued by newer deployments will NOT work on older deployments that still expect the legacy names. The alias bridge is one-directional (legacy → new).

Admin default permissions now include \`models:register\`, \`logs:list\`, \`monitor:view\`.

## Test plan

- [x] \`pytest -vv xinference/api/tests/test_oauth2_scope_alias_migration.py\` (11/11 passed)
- [x] \`pre-commit run --files <all modified files>\`
- [ ] \`pytest -vv xinference/api/tests/test_oauth2_admin_target_takeover.py\` (requires \`jose\` dep; CI will run)
- [ ] \`pytest -vv xinference/client/tests/test_client_with_auth.py\` (regression — verified locally that user3 with \`models:write\` can now terminate, matching the merged-scope semantics)
- [ ] \`cd xinference/ui/web/ui && npx eslint . && npx prettier --check .\`
- [ ] Manual: full permission matrix (admin / zero-perm / \`monitor:view\` / \`logs:list\` / \`users:manage\` / \`keys:create\` / \`keys:manage\`)

## Out of scope (follow-up)

- **Live-read permission changes**: the original mega-PR included \`/v1/users/me\` + \`PermissionsContext\` + \`_get_caller_live_perms\` to make admin permission changes take effect without re-login. Maintainer feedback revealed this only works at the UI layer, not the API route layer (\`Security\` still checks JWT snapshot). Dropped from this PR; tracked in a separate follow-up issue for proper design discussion.
- **Force-logout / JWT active invalidation**: separate follow-up.
- **SCOPE_ALIASES removal**: scheduled for a future release after deprecation warnings confirm zero alias hits in production.

### DIFF
--- a/xinference/api/oauth2/advanced/auth_service.py
+++ b/xinference/api/oauth2/advanced/auth_service.py
@@ -41,6 +41,9 @@ logger = logging.getLogger(__name__)
 _TRUSTED_PROXY_SET: Optional[set] = None
 
 
+from ..scope_aliases import _normalize_scopes  # noqa: E402
+
+
 def _get_trusted_proxies() -> set:
     global _TRUSTED_PROXY_SET
     if _TRUSTED_PROXY_SET is None:
@@ -103,6 +106,7 @@ class AdvancedAuthService:
                 "models:list",
                 "models:read",
                 "models:write",
+                "models:register",
                 "keys:create",
                 "keys:manage",
                 "users:manage",
@@ -110,6 +114,8 @@ class AdvancedAuthService:
                 "cache:delete",
                 "virtualenv:list",
                 "virtualenv:delete",
+                "logs:list",
+                "monitor:view",
             ]
             self._db.create_user(
                 username="admin",
@@ -546,6 +552,7 @@ class AdvancedAuthService:
                 _audit("success", user=username or "", auth_type="jwt")
             return user
 
+        token_scopes = _normalize_scopes(token_scopes)
         for scope in security_scopes.scopes:
             if scope not in token_scopes:
                 _audit("insufficient_scope", user=username or "", auth_type="jwt")
--- a/xinference/api/oauth2/advanced/routes.py
+++ b/xinference/api/oauth2/advanced/routes.py
@@ -68,27 +68,6 @@ def get_advanced_auth(request: Request) -> AdvancedAuthService:
     return request.app.state.advanced_auth
 
 
-def require_keys_read(request: Request):
-    """Dependency for API key read routes (list / get).
-
-    Accepts ``keys:create`` OR ``keys:manage`` (or ``admin`` wildcard).
-    FastAPI's ``Security(scopes=[...])`` is AND-semantics, so a custom
-    dependency is needed for OR. This aligns the backend read path with
-    the frontend ``PermissionGate`` contract (which shows the API Key
-    Management page to users holding either scope) and with the handler
-    ``is_admin`` check (which treats both as admin-capable).
-    """
-    auth = get_advanced_auth(request)
-    _, _, scopes = _get_current_user_from_token(request, auth)
-    scopes = scopes or []
-    if "admin" in scopes or "keys:create" in scopes or "keys:manage" in scopes:
-        return True
-    raise HTTPException(
-        status_code=403,
-        detail="Not enough permissions: requires keys:create or keys:manage",
-    )
-
-
 def _get_current_user_from_token(request: Request, auth: AdvancedAuthService):
     """Extract current user info from the Authorization header."""
     token = request.headers.get("Authorization", "").replace("Bearer ", "")
@@ -561,6 +540,32 @@ def register_advanced_auth_routes(api: "RESTfulAPI") -> None:
     _rl = getattr(auth_service, "_rate_limiter", None)
     if _rl is not None:
         api._app.state.rate_limiter = _rl
+
+    async def require_keys_read(
+        request: Request,
+        _user=Security(auth_service),  # validates JWT + user exists + enabled + audit
+    ):
+        """Dependency for API key read routes (list / get).
+
+        Accepts ``keys:create`` OR ``keys:manage`` (or ``admin``
+        wildcard). FastAPI's ``Security(scopes=[...])`` is
+        AND-semantics, so a custom dependency is needed for OR. This
+        reuses the standard ``auth_service`` dependency (which validates
+        JWT, user existence, enabled status, and audit logs) with no
+        required scopes, then performs the OR check on the JWT scopes —
+        consistent with how ``Security(scopes=[...])`` checks JWT
+        scopes (not DB permissions).
+        """
+        # Security(auth_service) already ran full validation. Now check
+        # OR scopes on the JWT payload (same source as Security(scopes=...)).
+        _, _, scopes = _get_current_user_from_token(request, auth_service)
+        scopes = scopes or []
+        if "admin" in scopes or "keys:create" in scopes or "keys:manage" in scopes:
+            return True
+        raise HTTPException(
+            status_code=403,
+            detail="Not enough permissions: requires keys:create or keys:manage",
+        )
 
     router.add_api_route("/token", advanced_login, methods=["POST"])
     router.add_api_route("/v1/auth/refresh", advanced_refresh, methods=["POST"])
--- a/xinference/api/oauth2/advanced/routes.py
+++ b/xinference/api/oauth2/advanced/routes.py
@@ -107,6 +107,35 @@ def _reject_permission_escalation(
         )
 
 
+def _reject_admin_target_takeover(
+    request: Request, auth: "AdvancedAuthService", target_user_id: int
+) -> None:
+    """Prevent non-admin callers from performing sensitive write operations
+    (delete / change password / disable / change permissions) on admin users.
+
+    Without this guard, a delegated ``users:manage`` operator could delete
+    an admin account, change an admin's password and log in as them, or
+    disable the only admin and lock the deployment out. Admin callers
+    bypass this check entirely.
+    """
+    _, _, caller_scopes = _get_current_user_from_token(request, auth)
+    caller_scopes = caller_scopes or []
+    if "admin" in caller_scopes:
+        return
+    target = auth.db.get_user_by_id(target_user_id)
+    if target is None:
+        # Caller already validated existence upstream; this is a defensive
+        # double-check. Treat missing user as not-admin to avoid leaking
+        # existence via 403-vs-404 distinction.
+        return
+    target_perms = target.get("permissions") or []
+    if "admin" in target_perms:
+        raise HTTPException(
+            status_code=403,
+            detail="Non-admin cannot perform this action on an admin user",
+        )
+
+
 # --- Auth endpoints ---
 
 
@@ -251,6 +280,8 @@ async def update_user(user_id: int, request: Request) -> JSONResponse:
     body = await request.json()
 
     if "enabled" in body:
+        # Layer A: non-admin cannot disable/enable an admin user.
+        _reject_admin_target_takeover(request, auth, user_id)
         enabled = int(body["enabled"])
         if enabled:
             auth.enable_user(user_id)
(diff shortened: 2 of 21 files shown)


m199369309:
Thanks @gemini-code-assist for the review — all 3 suggestions applied in db66e143, plus a real CI regression fix:

**1. C4 test regression (4 failing Ubuntu CI jobs — real breakage, not flaky):**

`conftest.py` user3 had \`permissions=["models:list", "models:read", "models:start"]\`. After C4 merged \`models:start\` + \`models:stop\` into a single \`models:write\` scope (with alias bridge), user3's \`models:start\` expands to \`models:write\` which covers BOTH launch AND terminate. The tests \`test_client_auth\` / \`test_async_client_auth\` expected user3's \`terminate_model\` to raise \`RuntimeError\`; after C4 it succeeds → "DID NOT RAISE" failure.

Fix: conftest user3 now uses \`models:write\` directly, and the \`with pytest.raises(RuntimeError): terminate\` assertions for user3 are removed (both JWT and API key paths). The "can launch but can't terminate" distinction is intentionally gone per C4's merge design (see PR description breaking change). Admin (user1) still terminates downstream, preserving the \`list_models() == 1\` → \`== 0\` assertions.

**2. N+1 query in list_api_keys (routes.py):** admin path now uses a single \`list_users()\` call instead of per-user \`get_user_by_id\`. Non-admin path keeps N+1 (user set is typically 1).

**3. JWT base64url decoding (utils/jwt.js):** \`atob()\` throws DOMException on base64url. Now normalizes \`-\`→\`+\` and \`_\`→\`/\` before decoding.

**4. Null-safety (apikey_management/index.js):** \`permissions.includes('admin')\` → \`permissions?.includes('admin')\` to handle the initial-load case where PermissionsContext hasn't resolved yet.

Re: macOS / Windows CI — those jobs were still pending when the fix pushed; the Ubuntu jobs should now pass. The pre-existing flaky tests (test_log_rotation on macOS, test_whisper_mlx on metal, test_worker socket bind on Windows) are unrelated to this PR.

m199369309:
Follow-up fix in 181d4936: the N+1 fix from db66e143 introduced a mypy error that broke the lint CI job.

**Root cause:** The local variable `u` in the admin branch (`for u in auth.db.list_users()`) was inferred as `Dict[str, Any]`. The else branch originally reused the variable name `u` for `auth.db.get_user_by_id(uid)` (which returns `Optional[Dict]`) — type collision. The first rename to `owner` then collided with the `owner: Optional[int]` function parameter, causing mypy to infer `int` and report `Value of type int is not indexable` on `owner["username"]`.

**Fix:** Renamed the else-branch variable to `owner_user`, avoiding both the loop-variable type collision and the parameter shadowing.

Lint CI should now pass.

xinference/api/oauth2/advanced/auth_service.py
```diff
@@ -546,6 +552,7 @@ def _audit(
                 _audit("success", user=username or "", auth_type="jwt")
             return user
 
+        token_scopes = _normalize_scopes(token_scopes)
```
qinxuye, line 555: The new live-read helpers do not take effect for routes protected by `Security(auth_service, scopes=[...])`, because this dependency still authorizes only the scopes embedded in the JWT. For example, if an admin grants `keys:create` in the DB after the user has logged in, `/v1/users/me` and the UI see the new permission, but a request to `/v1/admin/keys` is rejected here before `list_api_keys()` can call `_get_caller_live_perms()`. If this PR wants mid-session permission changes to work, the dependency needs to authorize against DB-current user permissions for JWT users, or the live-read claim needs to be dropped.

xinference/ui/web/ui/src/router/index.js
```diff
@@ -69,31 +71,62 @@ const routes = [
...
+          </PermissionGate>
+        ),
       },
       {
         path: 'apikey_management',
-        element: <ApiKeyManagement />,
+        element: (
+          <PermissionGate
+            requiredScope="keys:create"
+            alternateScope="keys:manage"
```
qinxuye, line 108: This frontend contract does not match the backend route dependencies. `keys:manage` passes the page gate here, and `list_api_keys()`/`get_api_key()` also treat `keys:manage` as admin-capable inside the handler, but the GET routes are still protected by `Security(..., scopes=["keys:create"])`. A user with only `keys:manage` can reach the page but the initial `/v1/admin/keys` request is rejected before the handler runs. Please either make the backend read routes accept the same `keys:create || keys:manage` contract, or remove `keys:manage` from the frontend/handler read path.

m199369309:
@qinxuye Thanks for both reviews — valid points. This PR has been re-scoped to address them:

**Feedback 1 (live-read doesn't work at route layer):** You're right — the \`_get_caller_live_perms\` helper and \`/v1/users/me\` + \`PermissionsContext\` only worked at the UI menu layer, not for actual API calls (\`Security(scopes=[...])\` still checks JWT snapshot). This created a worse UX than pre-PR (user sees a menu item that 403s on click).

**Action taken:** Dropped C5 (\`/v1/users/me\` + \`PermissionsContext\`) and C7's live-read portion (\`_get_caller_live_perms\` + 3 handler switches) from this PR. The remaining C7 pieces (G1 \`owner_username\` + J1 \`reveal\` scope) are kept as independent improvements — they don't depend on live-read.

Live-read design discussion moved to #5134 with four options (A: Security queries DB per request / B: drop live-read / C: shorten JWT + re-read on refresh / D: WebSocket push). Recommendation is C. Your input on direction would be appreciated.

**Feedback 2 (frontend/backend scope contract mismatch on read routes):** Fixed in commit \`a40fbe9e\`. New \`require_keys_read\` dependency accepts \`keys:create\` OR \`keys:manage\` (FastAPI's \`Security(scopes=[...])\` is AND-only, so a custom dependency is needed for OR). Applied to GET \`/v1/admin/keys\` and GET \`/v1/admin/keys/{id}\`. Write routes unchanged (each maps to exactly one scope).

**PR now contains 8 commits** (force-pushed to replace the original 7-commit mega-PR + 3 fix commits):

1. feat(ui): add JWT utils for permission checks
2. feat(ui): expand ALL_PERMISSIONS list and add i18n keys
3. feat(ui): gate B-class menu items and routes by permission scope
4. feat(api): rename route scopes to align with UI permissions
5. fix: update auth test fixtures for merged models:write scope + JWT base64url
6. feat(api): extend admin-target-takeover guard to all sensitive write handlers
7. feat(api,ui): owner_username in list_api_keys + reveal scope fix
8. fix(api): align API key read route scopes with frontend contract

Note: force-push invalidated the previous inline comments, but the design feedback they raised is fully addressed above. CI is re-running on the new head.

xinference/api/oauth2/advanced/routes.py
```diff
@@ -68,6 +68,27 @@ def get_advanced_auth(request: Request) -> AdvancedAuthService:
...
+
+    Accepts ``keys:create`` OR ``keys:manage`` (or ``admin`` wildcard).
+    FastAPI's ``Security(scopes=[...])`` is AND-semantics, so a custom
+    dependency is needed for OR. This aligns the backend read path with
+    the frontend ``PermissionGate`` contract (which shows the API Key
+    Management page to users holding either scope) and with the handler
+    ``is_admin`` check (which treats both as admin-capable).
+    """
+    auth = get_advanced_auth(request)
+    _, _, scopes = _get_current_user_from_token(request, auth)
```
qinxuye: This custom dependency fixes the OR-scope contract, but it bypasses the validation that `AdvancedAuthService.__call__` normally performs for JWT users. `_get_current_user_from_token()` only decodes the JWT payload; it does not verify that the user still exists or is still enabled in the DB. As a result, a user with a `keys:manage` token can be disabled after login and still pass this dependency for `GET /v1/admin/keys` and `GET /v1/admin/keys/{id}` until the JWT expires. The OR-scope read dependency should keep the same authentication checks as the normal Security path (at least user lookup + enabled check, ideally by factoring/reusing the existing auth dependency) before accepting `keys:create || keys:manage`.

m199369309:
@qinxuye Good catch — valid security regression. Pushed fix in d9f44df3.

**Root cause:** `require_keys_read` only called `_get_current_user_from_token` (JWT decode) — it skipped the user DB lookup + enabled check + audit log that `AdvancedAuthService.__call__` performs for standard `Security(scopes=[...])` routes. A disabled/deleted user with a still-valid JWT could pass the read dependency.

**Fix:** `require_keys_read` now uses `Security(auth_service)` as a sub-dependency (no required scopes = enforces authentication + user validity + enabled + audit, but not any specific scope). The OR check (`keys:create || keys:manage || admin`) is then performed manually on the JWT scopes — consistent with how `Security(scopes=[...])` checks JWT scopes (not DB permissions).

Implementation note: `require_keys_read` is now a closure inside `register_advanced_auth_routes` so it can reference the `auth_service` instance (which is only available after app initialization, not at module definition time).

Pre-commit clean (black/flake8/isort/mypy/codespell). CI re-running on the new head.

qinxuye (approved):
LGTM

PR #5081, opened 2026-06-25
ci: prepare GPU T4 runner migration

qinrui777:
**Background**

Replace AWS GPU-CI runner (self-hosted type) with new GPU power larger runner

```bash
 ### new GPU power larger runner softer version
 -------Check software version-------

pip 24.0 from /usr/lib/python3/dist-packages/pip (python 3.12)
conda 26.3.2
/usr/bin/python3
/usr/bin/python3.12
/opt/miniforge/condabin/conda

/usr/bin/python3.12 --version
Python 3.12.3

Name: setuptools
Version: 68.1.2
```

**Changes:**
- Runner changed to gpu-t4
- Add `Set up conda` back for new gpu runner as newly-provisioned everytime
- Remove `SELF_HOST_PYTHON` and `SELF_HOST_CONDA`
- Fix  shell script interpret issue 

**Test Result:**
Pytest can be executed, its potential errors will be fix in the near future.

[build logs](https://github.com/xorbitsai/inference/actions/runs/28153268395/job/83376014864#logs)

PR #5141, opened 2026-07-07
CI: re-enable GPU CI on gpu-t4 runners with faster setup

OliverBryant:
## What do these changes do?

Re-enables the GPU CI that was commented out in #5081 (migration from the self-hosted machine to GitHub-hosted `gpu-t4` runners), and speeds it up **without dropping any test**. The suite runs on a single GPU runner, serially, as before.

### Changes

- Move the GPU suite out of `build_test_job` into a dedicated `gpu_test_job` running on `gpu-t4`.
- Replace ~50 serial `pip install` calls with a few batched `uv pip install` calls (relative order preserved so version pins still win), and cache the uv wheel cache via `actions/cache`.
- Pin `torch==2.7.0` first so dependencies resolve against it instead of downloading the latest torch and then downgrading.
- Install `hf_transfer` and set `HF_HUB_ENABLE_HF_TRANSFER=1` to speed up model downloads — runners are ephemeral, so every run downloads models from scratch.
- Skip the web UI npm build (`NO_WEB_UI=1`) — GPU tests never serve the UI; the API server only warns when the bundle is absent.
- Drop `--cov` from GPU runs: there is no codecov config or upload step anywhere in the repo, so the coverage.xml was write-only overhead.
- Deduplicate the accidental double run of `test_f5tts.py`.
- Fix the broken `pip install -U onnxruntime-gpu==1.16.0; sys_platform == 'linux'` line — the unquoted environment marker was executed as a shell command and failed silently.
- Each test file still runs in its own pytest process so GPU memory is fully released between files, and the run stops at the first failure, matching the previous behavior.

### Cost guardrails (runners are billed per-minute)

- The GPU job only runs for pull requests and pushes to `main`, so a push to a PR branch no longer double-bills push + PR runs.
- `timeout-minutes: 90` bounds billing if a test hangs (pytest's own per-test timeout is 50min).

### Notes for reviewers

- The `gpu-t4` runner group must allow this repository; note GitHub larger runners do not serve fork PRs by default (jobs queue forever) — the runner group settings need to allow it if fork PRs should run GPU CI.
- If uv's stricter joint resolution surfaces a dependency conflict that serial pip papered over, splitting the offending install batch restores the old behavior.


OliverBryant:
Superseded: the GPU CI work will be consolidated into #5107.

PR #5143, opened 2026-07-07
feat: (UI) API Key, User Management, refreshToken, change password after first login

maoyuehui:
Pages：
1. Added username display to the sidebar and adjusted the bottom section layout.
2. Added a change password dialog after the first login.
3. Updated the API Key Management page; added User and Permission Mode fields to the create/edit forms.
4. Optimized the User Management page content.
5. Added auth-storage for managing token and refresh_token.
6. Added refreshToken.

Components：
1. date-time-picker now supports toggling time (hour/minute) display.
2. Added a password visibility toggle icon to Input when type="password".
3. Updated Sonner success toast styling to use a green background.
4. Optimized the MultiSelect dropdown style.


frontend/src/components/pages/api-key-management/api-key-dialog.tsx
```diff
@@ -0,0 +1,451 @@
...
+
+  const buildCreatePayload = (values: KeyFormValues) => {
+    const body: Record<string, unknown> = {
+      model_permissions: buildPermissionPayload(values),
+    };
+
+    if (values.name) body.name = values.name;
+    if (values.description) body.description = values.description;
+    if (values.user_id) {
+      body.user_id = /^\d+$/.test(values.user_id) ? Number(values.user_id) : values.user_id;
```
qinxuye: Please send this field as `owner`, not `user_id`. The backend `create_api_key` reads `body.get("owner")`, so the selected owner is currently ignored and admin-created keys end up assigned to the current user.

frontend/src/components/pages/api-key-management/index.tsx
```diff
@@ -1,112 +1,128 @@
...
     } catch {
       setKeys([]);
     } finally {
       setLoading(false);
     }
   }, []);
 
+  const fetchUsers = useCallback(async () => {
+    try {
+      const data = await request.get<ApiKeyUser[]>('/v1/admin/users');
```
qinxuye, line 101: This page always fetches `/v1/admin/users`, but users with only `keys:create` or `keys:manage` can access the API key page and do not have `users:manage`, so this will show a 403 on load. The key list already returns `owner_username`; please only fetch users when the current token can manage users, and use `owner_username` for display otherwise.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -209,7 +236,7 @@ export function Sidebar() {
             name: t('menu.userManagement'),
             Icon: Users,
             Extra: ChevronRight,
-            show: usersManagePage
+            show: usersManagePage,
```
qinxuye, line 239: In this same group, please show the API Key menu when either `keys:create` or `keys:manage` is present. A user with only `keys:manage` can manage existing keys through the backend, but the sidebar currently hides the page because the API Key item checks only `keys:create`. The group-level `show` condition should use the same combined permission check too.

frontend/src/components/pages/api-key-management/index.tsx
```diff
@@ -182,233 +228,297 @@ export default function ApiKeyManagement() {
...
+
+  const renderKeyValue = (key: ApiKey) => {
+    const revealedValue = revealedKeys[key.id];
+    const displayValue = revealedValue || `${key.key_prefix}${'*'.repeat(44)}`;
+    return (
+      <div className="flex min-w-0 items-center gap-2">
+        <span className="min-w-0 truncate font-mono text-xs text-muted-foreground">
+          {displayValue}
+        </span>
+        {isAdmin && (
```
qinxuye: This should be gated by `canManageKeys`, not `isAdmin`. The backend reveal endpoint requires `keys:manage`, so a non-admin user with `keys:manage` can edit/delete/toggle keys on this page but cannot reveal/copy them because these buttons are hidden.

frontend/src/components/pages/api-key-management/index.tsx
```diff
@@ -182,233 +228,297 @@ export default function ApiKeyManagement() {
...
+                              onChange={() => handleToggleEnabled(key)}
+                            />
+                          )}
+                        </div>
+                      )}
+                      {renderField(
+                        t('apiKey.owner'),
+                        key.user_id == null
+                          ? '-'
+                          : userNameMap.get(String(key.user_id)) || key.user_id
```
qinxuye: Please prefer the backend-provided `owner_username` here before falling back to the local `users` map or `user_id`. Non-admin callers with `keys:create` / `keys:manage` cannot fetch `/v1/admin/users`, and `list_api_keys` already includes `owner_username` specifically so this column can still show the owner name for them. The `ApiKey` type should include `owner_username?: string | null` as well.

qinxuye (approved):
LGTM

PR #5144, opened 2026-07-08
feat: enable advanced database-backed authentication by default

OliverBryant:
## Summary
- Xinference has three built-in SQLite databases (launch history, monitor config, advanced auth). Previously, only "advanced auth" (user accounts / API keys / permissions, SQLite-backed) was disabled by default, requiring `XINFERENCE_AUTH_ADVANCED=true` to opt in.
- This PR flips `XINFERENCE_AUTH_ADVANCED` to be enabled by default, so Xinference now requires authentication to access the API out of the box:
  - On first startup, an `admin` account is auto-created with a one-time password printed to the server log (reusing the existing `_init_admin` logic), and the user is forced to change the password on first login.
  - The JWT secret and encryption key are auto-generated and persisted under `XINFERENCE_HOME/auth/` when not explicitly set via environment variables, so restarts and multi-process deployments keep using the same keys instead of invalidating previously issued tokens.
  - Setting `XINFERENCE_AUTH_ADVANCED=false` still restores the previous no-auth or `--auth-config` file-based mode (the existing mutual-exclusivity check between the two modes is preserved).
- This is a **breaking change**: after upgrading, unauthenticated requests will be rejected, and users need to retrieve the initial admin password from the server log to log in.
- Updated test fixtures accordingly: several internal tests start the REST API without credentials via fixtures like `setup`/`setup_cluster`, which would otherwise fail once advanced auth defaults to on. Updated `conftest.py`, `test_metrics.py`, `test_client.py`, `test_async_client.py`, `test_llm_model.py`, and `test_restful_api.py` to explicitly disable advanced auth in those fixtures, keeping them focused on their original (non-auth) test purpose.
- Updated `doc/source/user_guide/auth_system.rst` to document the new default behavior and how to disable it.

## Test plan
- [x] `pytest xinference/core/tests/test_restful_api.py`: 26 passed, 1 pre-existing failure unrelated to this change (`test_lang_chain`, missing `langchain` dependency, fails identically on `main`)
- [x] `pytest xinference/api/tests/test_oauth2_permission_escalation.py xinference/api/tests/test_oauth2_token_expiration.py xinference/core/tests/test_monitor_config_store.py xinference/api/tests/test_launch_history.py`: 61 passed
- [x] Manually verified: with no environment variables set, `XINFERENCE_AUTH_ADVANCED` defaults to `True`, JWT/encryption keys are auto-generated and persisted, and reused across restarts; explicitly setting it to `0/false/no` (case-insensitive) correctly disables advanced auth
- [x] `pre-commit run --files <changed files>` passes (black/flake8/isort/mypy/codespell)
- [ ] Did not run the full distributed/GPU CI matrix; recommend running the full test suite in CI before merging to catch any tests that implicitly relied on the previous no-auth-by-default behavior


xinference/constants.py
```diff
@@ -90,13 +90,96 @@ def get_xinference_home() -> str:
...
+    import time
+
+    env_val = os.environ.get(env_name, "")
+    if env_val:
+        return env_val
+
+    secret_path = os.path.join(XINFERENCE_AUTH_DIR, file_name)
+    os.makedirs(XINFERENCE_AUTH_DIR, exist_ok=True)
+
+    stale_after_seconds = 10
```
qinxuye: The recovery window cannot reach the stale threshold: this loop waits at most `50 * 0.1 = 5` seconds, while an empty file is not removable until it is older than 10 seconds. An immediate restart after a writer crashes therefore exhausts the loop and raises `RuntimeError` before this process can recover (I reproduced this with a fresh empty secret file). Please use a deadline/retry window that extends beyond the stale threshold and add a regression test for a recently-created empty file.

doc/source/user_guide/auth_system.rst
```diff
@@ -7,7 +7,18 @@ Simple OAuth2 System (experimental)
 Xinference builds an In-memory OAuth2 authentication and authorization system using the account-password mode.
 
 .. note::
-   If you don't have authentication and authorization requirements, you can use Xinference as before, without any changes.
+   Since **v3.0**, Xinference enables the :ref:`advanced, database-backed
+   authentication system <user_guide_advanced_auth_system>` by default
+   (``XINFERENCE_AUTH_ADVANCED``), replacing the in-memory system described
+   below as the default behavior. On first startup, an ``admin`` account is
```
qinxuye: This documentation now contradicts the implementation: the service no longer auto-creates an `admin` account or prints a random password. The same obsolete flow is repeated in the Initial admin account section below and in the locale catalogs. Please replace it with the new `/setup` / `POST /v1/admin/setup` first-run flow so operators are not told to look for credentials that will never be emitted.

xinference/api/oauth2/advanced/routes.py
```diff
@@ -229,6 +234,53 @@ async def advanced_logout(request: Request) -> JSONResponse:
...
+
+async def setup_status(request: Request) -> JSONResponse:
+    auth: AdvancedAuthService = get_advanced_auth(request)
+    needs_setup = auth.needs_setup()
+    return JSONResponse(
+        content={"needs_setup": needs_setup, "initialized": not needs_setup}
+    )
+
+
+async def setup_admin(request: Request) -> JSONResponse:
```
qinxuye, line 268: This public endpoint allows whichever network client arrives first to create the full-privilege administrator. `BEGIN IMMEDIATE` guarantees only one winner, but it does not establish that the winner is the deployment owner; a fresh or upgraded instance exposed on the network can therefore be taken over before the operator opens the UI. Please require a trusted bootstrap credential, restrict setup to a trusted local channel, or provide an equivalent trust mechanism. Also reject immediately when `needs_setup()` is false before computing the bcrypt hash; otherwise this permanently remains an unauthenticated CPU-expensive endpoint even after setup is complete.

qinxuye: Documenting the exposure does not resolve the takeover risk: default-on authentication should not require the operator to win a race before the service port becomes reachable. A practical fix is a one-time setup token. Support `XINFERENCE_AUTH_SETUP_TOKEN` for containers and Kubernetes; when it is unset, generate a random token, persist it under the auth directory with mode `0600`, and print it once at startup. Require that token in a header or request field for `POST /v1/admin/setup`, verify it with `secrets.compare_digest`, and remove the persisted token after the first admin is created. The setup UI can prompt for it while `create_first_user` remains the atomic final guard. Please add missing-token, wrong-token, valid-token, and concurrent-setup tests.

xinference/api/restful_api.py
```diff
@@ -137,6 +138,18 @@ def __init__(
...
+            if self._advanced_auth_service.needs_setup():
+                setup_token = get_or_create_setup_token()
+                logger.warning(
+                    "\n"
+                    + "=" * 60
+                    + "\n"
+                    + "  FIRST-RUN SETUP REQUIRED (token shown only until used)\n"
+                    + "  Create the initial admin account at POST /v1/admin/setup\n"
+                    + "  (or via the web UI's setup page), supplying this token:\n"
+                    + f"  Setup token: {setup_token}\n"
```
qinxuye: Please do not log an operator-provided `XINFERENCE_AUTH_SETUP_TOKEN` verbatim. For containers/Kubernetes this value may come from a Secret specifically to keep the bootstrap credential out of logs; exposing it here lets any log reader create the first full-privilege admin before setup completes. Only print the token when it was auto-generated. When the environment variable is set, log a generic instruction to use the configured setup token.

frontend/src/components/pages/setup/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+              <Input />
+            </FormField>
+            <FormField
+              name="password"
+              label={t('login.password')}
+              placeholder={t('userManagement.passwordPlaceholder')}
+              rules={[
+                { required: true, message: t('userManagement.passwordRequired') },
+                {
+                  validator: (val: unknown) => typeof val === 'string' && val.length >= 8,
```
qinxuye: This hard-codes 8 while the backend now accepts a configurable `XINFERENCE_PASSWORD_MIN_LENGTH`. With a different deployment policy the setup form and API disagree, and values below 8 can never be submitted through the UI. Please expose the effective minimum from `/v1/admin/setup/status` and use it here, or keep the policy fixed consistently on both sides.

qinxuye (approved):
LGTM

PR #5145, opened 2026-07-08
chore: remove legacy React web UI now that Next.js frontend is the default

OliverBryant:
## Summary
- Remove the legacy `xinference/ui/web/ui` React/Material-UI frontend, which was fully superseded by the Next.js app under `frontend/` in #5111 (build, packaging, and backend serving are already wired to `frontend/`).
- Update `AGENTS.md` (and its `CLAUDE.md` symlink) to describe the current `frontend/` directory and its npm/eslint commands instead of the removed directory.

## Test plan
- [ ] Confirm no remaining functional references to `xinference/ui/web/ui` in `setup.py`, `MANIFEST.in`, `xinference/api/restful_api.py`, or CI workflows
- [ ] `pip install -e .` builds the Next.js frontend and the backend serves it correctly
- [ ] `pre-commit run --files AGENTS.md`

PR #5146, opened 2026-07-09
fix(frontend): route params via usePathname(); remove Gradio, add LaTeX rendering

OliverBryant:
## Summary

**Part 1 — route param fix (original scope):**
- The static-exported Next.js frontend prerenders each dynamic route (`running-model/[modelUid]`, `launch-model/[modelType]`, `register-model/[modelType]/[modelName]`) with a single `__shell__` placeholder param via `generateStaticParams()`. The backend (`xinference/api/frontend_static.py`) maps any real URL under that route to this shell's HTML, and also to its prebuilt RSC flight payload for prefetch/navigation requests.
- `useParams()` reads from the flight router's param tree, which stays pinned to the baked-in `"__shell__"` literal on these shell-backed routes forever. So pages relying on it to read the real model uid / model type never see the real value: `describe_model` gets called with `uid: __shell__` forever, and the page spins forever with its title stuck on `__shell__`.
- `usePathname()` instead reflects the actual browser URL (Next seeds `canonicalUrl` from `location.href` on the client per `create-initial-router-state.js`, and updates it on every client-side navigation), so it is not subject to this problem.
- Added `getPathSegmentsAfter(pathname, prefix)` in `route-params.ts`, replaced `useParams()` with it in all four affected `page-client.tsx` files.

**Part 2 — remove Gradio, fix OCR formula rendering (follow-up to #5145):**
- Gradio was only used to mount a per-model chat/image/embedding demo UI at `/{model_uid}`, driven by the old React web UI (removed in #5145). The Next.js frontend never used those `/v1/ui/*` mount endpoints, so the mounting handlers/routes/schemas and the `xinference/ui/gradio/*` interface builders were dead. Removed them, along with the `gradio` dependency (`setup.cfg`, Docker requirements) and a stale doc section.
- Kept and relocated `xinference/ui/gradio/utils/latex.py` → `xinference/model/image/ocr/latex.py`: `DeepSeekOCRModel` still uses `process_ocr_result_with_latex` to post-process LaTeX in OCR output. Updated its 3 call sites in `deepseek_ocr.py`.
- The new frontend's OCR result panel renders text through the shared markdown renderer, which had no LaTeX support — formulas showed up as raw `$...$` text. Wired `remark-math` + `rehype-katex` (already installed, unused) into `markdown-renderer.tsx`, extending the `rehype-sanitize` schema so KaTeX's HTML survives sanitization.
- `visualize_ocr`'s bounding-box output stays Python-client-only (no REST endpoint) — out of scope here.

## Test plan
- [x] `npx tsc --noEmit`, `npx eslint .`, `npm run build` all pass (0 errors; same pre-existing `any` warnings as before)
- [x] Verified via Next.js router source that `usePathname()` is seeded from `window.location.href` on both hard loads and soft navigation, so it can't get stuck on the shell placeholder
- [x] Repo-wide grep sweep: no remaining references to any removed Gradio symbol outside vendored `thirdparty/` code and the standalone `examples/gradio_chatinterface.py` tutorial (intentionally kept, unrelated to the removed mounting code)
- [x] `python3 -m py_compile` on every changed Python file
- [x] Manually traced the KaTeX + rehype-sanitize pipeline with a standalone script matching the exact plugin order used in `markdown-renderer.tsx`: formulas render as proper `<span class="katex">` markup, an XSS `<script>` probe is still stripped
- [x] User confirmed in a live deployment that the `usePathname()` fix resolves the original spinner/`__shell__` bug
- [ ] Not run: `pre-commit` / `pytest` (no Python env with project deps in this sandbox) — please run before merging, especially `xinference/model/image/ocr/tests/test_deepseek_ocr_safe_eval.py` and `xinference/api/tests/test_frontend_static.py`


frontend/src/components/ui/markdown-renderer.tsx
```diff
@@ -1,10 +1,55 @@
...
+// rehype-sanitize's GitHub-derived default schema doesn't know about the
+// classes/attributes rehype-katex's HTML output needs (span/math/semantics/...
+// with "katex*" classes and inline "style"), so it would strip them back out.
+// Extend the schema rather than replace it, so GFM sanitation elsewhere is
+// unaffected.
+const katexSchema = {
+  ...defaultSchema,
+  attributes: {
+    ...defaultSchema.attributes,
+    span: [...(defaultSchema.attributes?.span || []), 'className', 'style', 'ariaHidden'],
```
qinxuye: This opens the sanitizer to any raw HTML `span style=...` whenever `parseHtml` is enabled, not just to KaTeX output. OCR/document results are model-produced text and are rendered with `parseHtml`, so a result can inject arbitrary inline styles into the UI. Please avoid allowing `style` on user/model raw HTML globally; for example sanitize raw HTML before running KaTeX, or only allow the attributes/classes needed on KaTeX-generated nodes.

setup.cfg
```diff
@@ -35,6 +34,7 @@ install_requires =
     aiohttp
     pydantic
     fastapi>=0.110.3,<0.137
+    python-multipart>=0.0.18  # required by FastAPI's Form/File/UploadFile params; used to come in transitively via gradio
```
qinxuye, line 37: Removing Gradio from the default dependencies leaves the standalone Gradio example documented in `examples/gradio_chatinterface.py` broken for a normal Xinference install. Please update the example instructions and/or docs to tell users to install Gradio explicitly, e.g. `pip install gradio`, before running that script.

frontend/src/components/ui/markdown-renderer.tsx
OliverBryant: Fixed in 582105d18 (well, actually f9928148c) — reordered the rehype pipeline so `rehypeSanitize` runs with the plain `defaultSchema` before `rehypeKatex` generates its HTML, instead of extending the schema to allow `style`/`className` on every `span`. Raw HTML from the model/user is sanitized first (no style/className carve-out), and KaTeX's own output is generated afterwards so it's never re-sanitized. Verified with a standalone unified/remark/rehype test that a malicious `<span style="position:fixed;...">` gets stripped to a bare `<span>` while $inline$ and $$block$$ formulas still render correctly.

setup.cfg
OliverBryant: Fixed in 582105d18 — removed the whole examples/ directory instead (podcast demo, translate, CLI chatbot, LangChain/Streamlit doc chat, gradio_chatinterface.py, and several tutorial notebooks), since none of it is exercised by CI or referenced from product code, and it had drifted out of sync with the current API in ways beyond just the Gradio dependency. Also cleaned up every doc page and README (all 9 language variants) that linked into it.

setup.cfg
```diff
@@ -202,6 +202,7 @@ audio =
     vocos  # For F5-TTS
     librosa  # For F5-TTS
     jieba  # For F5-TTS
+    pydub  # For F5-TTS & MegaTTS3; used to come in transitively via gradio
```
qinxuye, line 205: Adding `pydub` and `ffmpy` only to the `audio` extra does not restore them in the official Docker images. Both Dockerfiles install the standalone requirements files and then install Xinference with `pip install --no-deps "."`, while neither the GPU nor CPU `requirements-models.txt` contains these packages. After removing Gradio, F5-TTS/MegaTTS3 will therefore fail to import `pydub`, and IndexTTS2 will fail to import `ffmpy` in those images. Please add both dependencies to the applicable Docker model requirements as well.

qinxuye (commented):
One Docker dependency gap still needs to be addressed.

setup.cfg
OliverBryant: Good catch, fixed in f64ece473. Added pydub/ffmpy to requirements/requirements-models.txt (GPU image) and requirements_cpu/requirements_cpu-models.txt (CPU image) directly, since both Dockerfiles install those standalone requirements files and then `pip install --no-deps "."`, bypassing setup.cfg entirely. Also discovered while checking this that python-multipart and orjson were already present in requirements-ml.txt/requirements-models.txt for the GPU image, so that one was never actually affected - removed the redundant duplicate entries I'd added there. Left requirements_aarch64.txt alone since it doesn't package F5-TTS/MegaTTS3/IndexTTS2 at all.

qinxuye (approved):

PR #5149, opened 2026-07-09
feat(api): validate system message ordering for strict Qwen3-family chat templates

m199369309:
## Summary

Add early validation at the API layer (before stream/non-stream split) to reject requests with non-first `system` messages on models whose chat template enforces system-first ordering.

## Background

The Qwen3-family chat template (used by Ornith-1.0-35B, qwen3.5, qwen3.6, Nex-N2) contains a strict guard:

```jinja
{%- if message.role == "system" and not loop.first %}
    {{- raise_exception("System message must be at the beginning.") }}
{%- endif %}
```

When a client sends multiple `system` messages (e.g. an agent framework appending a mid-conversation reminder as a system message), the template raises during Jinja render. Currently this results in:
- **Non-stream:** HTTP 500 with an opaque error
- **Stream:** HTTP 200 followed by SSE error events

Both make debugging unnecessarily difficult.

## Changes

| File | Change |
|------|--------|
| `model/llm/llm_family.py` | Add `STRICT_SYSTEM_FIRST_MARKERS` + `is_strict_system_first_template()`; expose `strict_system_first: bool` in `to_description()` |
| `model/llm/utils.py` | Add `MessageRoleOrderError` (ValueError subclass) + `check_system_role_order()` validator |
| `api/restful_api.py` | Gate the check on `desc["strict_system_first"]` before stream/non-stream fork → clean 400 on both paths |
| `model/llm/tests/test_system_role_order.py` | 9 unit tests for template detection + message validation |
| `api/tests/test_system_role_order.py` | 6 integration tests for gating + HTTPException conversion |

## Key design decisions

1. **API layer before stream split** — the check runs before `EventSourceResponse` is constructed, so both stream and non-stream get a clean HTTP 400 (not 200+SSE). This is better than checking in the worker render path where the stream header has already been sent.

2. **Gated on a registration-time bool** — `strict_system_first` is computed once per model registration (in `to_description()`) and cached by `describe_model`. The check itself is a simple `O(n)` scan of the messages list. **Zero per-request overhead** from template scanning.

3. **Zero regression for lenient templates** — models without the guard (`strict_system_first=False`) skip the gate entirely. No behavioral change.

4. **Backward compatible** — `strict_system_first` is a new key in the existing `describe_model` response dict; unknown keys are ignored by existing consumers.

## Testing

- 15 new tests all pass
- Manually verified against a production Ornith-1.0-35B deployment (results in design doc)
- Test matrix: strict model (non-stream + stream) → 400; lenient model → 200; strict model + leading system → 200

There is a known boundary: models with an empty `chat_template` that fall back to a tokenizer-provided template containing the guard are not detected (the check only looks at the registered `chat_template`). This does not affect any currently registered model and is left for future work.

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -2558,6 +2558,19 @@ class RESTfulAPI(CancelMixin):
                     detail=f"Only {total_call_family} support tool messages",
                 )
 
+        # Reject misplaced ``system`` messages before entering the worker for
+        # models whose chat template requires system-first ordering (Qwen3
+        # family: Ornith-1.0-35B / qwen3.5 / qwen3.6 / Nex-N2). Placed before
+        # the stream/non-stream split so BOTH paths return a clean 400 instead
+        # of the worker raising mid-render (non-stream 500 / stream 200+SSE).
+        if desc.get("strict_system_first"):
+            from ..model.llm.utils import MessageRoleOrderError, check_system_role_order
+
+            try:
+                check_system_role_order(messages)
+            except MessageRoleOrderError as ve:
+                raise HTTPException(status_code=400, detail=str(ve))
+
         if "skip_special_tokens" in raw_kwargs and await model.is_vllm_backend():
             kwargs["skip_special_tokens"] = raw_kwargs["skip_special_tokens"]
         if body.stream:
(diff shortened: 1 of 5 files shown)


qinxuye (approved):
LGTM

PR #5151, opened 2026-07-10
fix(embedding): implement truncate_prompt_tokens, OOM eviction, and wait_for_load after recovery

m199369309:
## Summary

Three coupled fixes addressing jina-embeddings-v3 OOM crashes cascading into persistent replica zombies. Combines B1 (truncation), B2 (eviction on failure, both branches), and B3a (wait_for_load after recovery).

## B1: Implement truncate_prompt_tokens for embedding models

`truncate_prompt_tokens` was previously only implemented for vLLM LLMs. For embeddings the parameter was silently accepted and ignored — long documents were processed at full length, triggering O(L²) attention OOM.

**Changes:**
- `embedding/core.py`: Add `_truncate_sentences()` — token-based truncation with character fallback. Never raises. Applied at both choke points.
- `schemas/requests.py`: Add `truncate_prompt_tokens: Optional[int]` field.

## B2: Evict zombie replicas on recovery failure (both branches)

When OOM kills a replica subprocess and auto-recovery itself fails:
- **Unbounded branch** (default `recover_count=None`): silently continued without notifying supervisor → zombie replica stays in routing table.
- **Bounded branch** (`AUTO_RECOVER_LIMIT>=1`): same gap — exception escaped `recover_sub_pool`, leaving "stopping" zombie causing persistent 500s.

**Fix:** Wrap `recover_model` in `try/except` in both branches → call `_evict_replica_from_supervisor` (idempotent `mark_replica_dead`).

## B3a: Mark recreated/recovered replicas "ready" via wait_for_load

`recover_model` and `_try_recover_models` called `launch_builtin_model` (sets `model_state="loading"`) but never called `wait_for_load` to transition to `"ready"` — leaving recreated replicas permanently "loading".

**Fix:** Add `await self.wait_for_load(...)` after `launch_builtin_model` in both methods, mirroring the normal launch path.

## Files changed

| File | Changes |
|------|---------|
| `model/embedding/core.py` | B1: `_truncate_sentences()` + constants + choke points |
| `api/schemas/requests.py` | B1: `truncate_prompt_tokens` field |
| `core/worker.py` | B2: try/except in both branches + B3a: wait_for_load in recover_model and _try_recover_models |
| `model/.../test_truncate_prompt_tokens.py` | 8 B1 unit tests |
| `api/tests/test_embedding_truncate.py` | 7 schema tests |
| `core/tests/test_worker.py` | 10 tests (B2 unbounded/bounded + B3a recover_model/_try_recover_models) |

## Backward compatibility

- `truncate_prompt_tokens` defaults to `None` — existing clients unaffected.
- Unbounded recovery still retries infinitely on success. Behavior only changes on failure.
- `wait_for_load` is idempotent; on an already-loaded model it is near-instant.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/xinference/model/embedding/core.py
+++ b/xinference/model/embedding/core.py
@@ -19,7 +19,7 @@ import logging
 import os
 from abc import abstractmethod
 from collections import defaultdict
-from typing import Annotated, Dict, List, Literal, Optional, Tuple, Union
+from typing import Annotated, Any, Dict, List, Literal, Optional, Tuple, Union
 
 from xoscar import extensible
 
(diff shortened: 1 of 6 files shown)


xinference/model/embedding/core.py
```diff
@@ -363,6 +382,65 @@ def _extract_sentences_kwargs(self, args, kwargs):
...
+                    "truncate_prompt_tokens<0 but max_tokens unknown for %s; skip",
+                    self._model_uid,
+                )
+                return sentences
+
+        is_str = isinstance(sentences, str)
+        texts = [sentences] if is_str else list(sentences)
+        truncated: List[str] = []
+        for text in texts:
+            s = text if isinstance(text, str) else str(text)
```
qinxuye: This corrupts valid structured embedding inputs when `truncate_prompt_tokens` is set. The request schema accepts `List[List[int]]`, `Dict[str, str]`, and `List[Dict[str, str]]`, but token arrays are stringified before `_fix_langchain_openai_inputs()` can decode them, a single dict is iterated as its keys, and multimodal dict items are converted to literal dict strings. For example, `[[1, 2, 3]]` becomes a string input and `{"text": "hello"}` becomes `["text"]`, so the model embeds the wrong content. Please normalize token arrays before truncation and only truncate text values while preserving non-text/multimodal structure.

xinference/model/embedding/core.py
```diff
@@ -363,6 +382,65 @@ def _extract_sentences_kwargs(self, args, kwargs):
...
+
+        This method must never raise: a tokenizer/decode failure degrades to a
+        character-based cut so embedding serving is not interrupted.
+        """
+        if truncate_prompt_tokens is None:
+            return sentences
+
+        # Resolve the effective token limit.
+        n_tokens: Optional[int]
+        if truncate_prompt_tokens > 0:
```
qinxuye, line 433: The `else` branch also catches `truncate_prompt_tokens == 0`, although the new schema test explicitly treats zero as valid and the vLLM implementation handles zero as an explicit `max_length=0`. This currently turns zero into the model's full `max_tokens` limit. Please either handle zero consistently with vLLM or reject it in schema validation instead of silently changing its meaning.

qinxuye (commented):
Two input-handling issues need to be addressed before approval.

xinference/model/embedding/core.py
qinxuye: The token-array part is fixed, but multimodal string payloads are still corrupted here. `_truncate_value()` applies `_truncate_text()` to every string value in a dict, while the existing Jina embedding paths use string-valued `image`, `video`, and `audio` fields for URLs, file paths, and base64 data. A long `{"image": "data:image/..."}` or media URL is therefore truncated and becomes unusable (I reproduced a 222-character base64 payload being cut to 32 characters with `n_tokens=8`). Please truncate only the `text` field and leave `image`/`video`/`audio` and other non-text fields unchanged; add a regression test with a media payload longer than the token limit.

qinxuye: Rechecked the current head and this is still reproducible. A 222-character `data:image/...;base64,...` payload is reduced to 32 characters with `truncate_prompt_tokens=8`, because every string value in the dict is passed through `_truncate_text()`. Please truncate only the `text` field and preserve `image`, `video`, `audio`, URLs, paths, and base64 payloads unchanged.

qinxuye (approved):
LGTM

PR #5154, opened 2026-07-10
fix: replace Korean text with Japanese in ja locale launchModel keys

kejhz653-stack:
The launchModel section in ja.ts incorrectly contained Korean translations for modelName through noPackagesInstalled; restore proper Japanese strings.


qinxuye (approved):
LGTM

PR #5162, opened 2026-07-12
fix(worker): use os.replace for atomic launch_args persist on Windows (#5161)

### LINKED ISSUE #5161: [Bug] Failed to persist launch_args on Windows: os.rename raises FileExistsError [WinError 183], silently breaks auto-recovery
## Summary

On Windows, `_persist_launch_args` in `xinference/core/worker.py` (line 1295, v2.12.0) uses `os.rename(tmp_path, filepath)`. On Windows, `os.rename` raises `FileExistsError: [WinError 183]` when the target file already exists. As a result, the `worker_recovery/<worker>/models.json` file silently fails to update after each new model launch — only the **first** model launch is persisted; subsequent launches leave `models.json` reflecting only the first model, and `models.json.tmp` retains the newer (failed-to-commit) content. Auto-recovery on next worker restart only restores the first model.

## Environment

- **Xinference version**: 2.12.0
- **OS**: Windows 11 (10.0.26100)
- **Python**: 3.12.10 (virtualenv at `C:\Users\<user>\AppData\Local\xinference\venv`)
- **Install method**: `pip install "xinference[transformers]"`
- **Launch command**: `xinference-local -H 127.0.0.1 -p 9997`
- **Models launched**: `bge-m3` (embedding), `bge-reranker-v2-m3` (rerank), both via `xinference launch` CLI

## Steps to reproduce

1. Start Xinference on Windows:
   ```
   xinference-local -H 127.0.0.1 -p 9997
   ```
2. Launch first model (e.g. `bge-m3`):
   ```
   xinference launch -e http://127.0.0.1:9997 -n bge-m3 -t embedding -u bge-m3
   ```
3. Launch second model (e.g. `bge-reranker-v2-m3`):
   ```
   xinference launch -e http://127.0.0.1:9997 -n bge-reranker-v2-m3 -t rerank -u bge-reranker-v2-m3
   ```

## Expected behavior

Both launches succeed and `worker_recovery/<worker>/models.json` correctly persists metadata for both models, enabling auto-recovery of both on next worker restart.

## Actual behavior

The second launch succeeds for inference, but logs the following warning:

```
2026-07-12T09:48:51.557Z WARNING xinference.core.worker pid:44860 role:local address:127.0.0.1:44279 node:BigBox Failed to persist launch_args, auto-recovery may not work
  File "...\venv\Lib\site-packages\xinference\core\worker.py", line 1295, in _persist_launch_args
    os.rename(tmp_path, filepath)
FileExistsError: [WinError 183] 当文件已存在时，无法创建该文件。:
  'C:\\Users\\<user>\\.xinference\\worker_recovery\\127.0.0.1_44279\\models.json.tmp'
  -> 'C:\\Users\\<user>\\.xinference\\worker_recovery\\127.0.0.1_44279\\models.json'
2026-07-12T09:48:51.559Z INFO xinference.core.worker pid:44860 ... Launch finished: model_name=bge-reranker-v2-m3, model_uid=bge-reranker-v2-m3-0 (active: 0/5, queued: 0)
```

After the second launch, inspecting `worker_recovery/127.0.0.1_44279/` shows:

```
Name             Length  LastWriteTime
----             ------  -------------
models.json         535  2026/7/12 17:47:05   <- contains ONLY bge-m3-0 entry
models.json.tmp    1103  2026/7/12 17:48:51   <- contains BOTH bge-m3-0 AND bge-reranker-v2-m3-0 (failed commit)
```

`models.json` content (only first model):

```json
{
  "bge-m3-0": {
    "model_uid": "bge-m3-0",
    "model_name": "bge-m3",
    "model_type": "embedding",
    "gpu_idx": [0],
    "trust_remote_code": true,
    "launch_ts": 1783849600
  }
}
```

So if Xinference restarts, **only `bge-m3-0` would auto-recover**; `bge-reranker-v2-m3-0` would be silently lost (user would need to manually re-launch it).

## Root cause

`os.rename(src, dst)` behavior:

- **Linux / macOS**: atomic overwrite if `dst` exists — works as intended
- **Windows**: raises `FileExistsError: [WinError 183]` if `dst` already exists

Python documentation explicitly recommends `os.replace` for cross-platform atomic replacement: https://docs.python.org/3/library/os.html#os.replace

> "If dst exists and is a file, it will be replaced silently if the user has permission. The operation is atomic if and only if src and dst are on the same filesystem."

## Suggested fix

In `xinference/core/worker.py` line 1295:

```python
# Before (raises FileExistsError on Windows when dst already exists)
os.rename(tmp_path, filepath)

# After (cross-platform atomic replace — same behavior on Linux/macOS, fixes Windows)
os.replace(tmp_path, filepath)
```

Same fix likely applies to `_remove_persisted_launch_args` if it uses `os.rename` analogously (line ~1309+ in the same file).

## Additional context

- This bug appears on the `v2.x` branch; `_persist_launch_args` was introduced by PR #4859 ("graceful OOM handling, worker model auto-recovery, and heartbeat log enhancement").
- The warning is non-fatal for inference (models still serve correctly), but auto-recovery is silently broken on Windows.
- Workaround for affected users: manually overwrite `models.json` with the contents of `models.json.tmp` after each new `xinference launch`.

## Reproducibility

Reproduces 100% on Windows 11 with two consecutive `xinference launch` calls against the same worker. Does not reproduce on Linux/macOS due to `os.rename` semantics difference.

Anai-Guo:
## Summary

On Windows, `os.rename(src, dst)` raises `FileExistsError: [WinError 183]` when `dst` already exists — unlike POSIX, it is **not** an atomic overwrite. Both `_persist_launch_args` and `_remove_persisted_launch_args` in `xinference/core/worker.py` rely on the temp-file + rename pattern to atomically update `worker_recovery/<worker>/models.json`.

As a result, on Windows every persist **after the first** silently fails:
- `models.json` keeps reflecting only the first launched model,
- `models.json.tmp` is left holding the newer (uncommitted) content,
- the failure is swallowed by the surrounding `except Exception` (only a debug/warning log), so it is silent,
- auto-recovery on the next worker restart only restores the first model.

## Fix

Replace `os.rename(tmp_path, filepath)` with `os.replace(tmp_path, filepath)` at both call sites. `os.replace` overwrites the destination atomically on **all** platforms (Windows and POSIX), which is exactly the intended semantics — the docstring already says *"Atomically persist current launch_args to disk."*

This also matches the existing atomic-write in the same file (the `.pth` write already uses `os.replace(str(tmp_file), str(pth_file))`), so the fix is consistent with established style in the codebase.

No behavior change on Linux/macOS; fixes the silent breakage on Windows.

Fixes #5161

### DIFF
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -1041,7 +1041,7 @@ class WorkerActor(xo.StatelessActor):
             tmp_path = filepath + ".tmp"
             with open(tmp_path, "w") as f:
                 json.dump(data, f, ensure_ascii=False)
-            os.rename(tmp_path, filepath)
+            os.replace(tmp_path, filepath)
             logger.debug(
                 "Persisted launch_args for %d models to %s", len(data), filepath
             )
@@ -1064,7 +1064,7 @@ class WorkerActor(xo.StatelessActor):
                 tmp_path = filepath + ".tmp"
                 with open(tmp_path, "w") as f:
                     json.dump(data, f, ensure_ascii=False)
-                os.rename(tmp_path, filepath)
+                os.replace(tmp_path, filepath)
         except Exception:
             logger.warning("Failed to update persisted launch_args", exc_info=True)
 


qinxuye (approved):
LGTM

PR #5163, opened 2026-07-12
Polish docs and comments in inference (#5161)

bglglzd:
Small scoped patch based on the reported behavior.

Related to #5161.


qinxuye:
This PR is unrelated to #5162 and does not address #5161. PR #5162 contains the actual Windows `os.replace` fix, while this PR only removes trailing whitespace from documentation files. Please remove the issue reference and clarify the purpose of this PR, or close it if the whitespace-only cleanup was unintended.

PR #5170, opened 2026-07-13
fix(llama.cpp): resolve mmproj against the model directory so image input works

### LINKED ISSUE #5169: 加载自定义多模态模型报错：Exception: [address=127.0.0.1:16632, pid=39104] image input is not supported - hint: if this is unexpected, you may need to provide the mmproj
### System Info / 系統信息

Python 3.10.20
xinference 3.12.0
winds11

- [ ] docker / docker

### Version info / 版本信息

xinference 3.12.0

xinference-local --host 127.0.0.1 --port 9997

### Reproduction / 复现过程

本地下载好的模型：
mmproj-BF16.gguf
Qwen3.5-0.8B-BF16.gguf

启动命令：
xinference launch --model-name Qwen3.5-0.8B-BF16 --model-type LLM --n-gpu auto --replica 1 --reasoning_content false --model-engine llama.cpp --model-format ggufv2 --size-in-billions 0_8 --quantization default --model-uid Qwen3.5-0.8B-BF16 --multimodal-projector mmproj-BF16.gguf --model-path D:\xinference_dify\xinference_models\Qwen3.5-0.8B-BF16.gguf

### Expected behavior / 期待表现

上传图片后能正常chat

Anai-Guo:
### What
When launching a custom multimodal GGUF model with `--multimodal-projector`, the projector was never actually loaded, so llama.cpp rejected image input with:

> image input is not supported - hint: if this is unexpected, you may need to provide the mmproj

### Why
In `llama_cpp/core.py::load()`, when `--model-path` points at a single `.gguf` **file**, `self.model_path` is that file. The projector was then resolved with:

```python
mmproj = os.path.join(self.model_path, multimodal_projector)
```

which yields a nonexistent path such as `.../Qwen3.5-0.8B-BF16.gguf/mmproj-BF16.gguf`. `params.mmproj.path` is set to that bogus path, llama.cpp silently starts without a projector, and every image request fails.

### Fix
Resolve the projector against the model **directory** (`os.path.dirname(model_path)`) instead of the file path, and honor absolute projector paths as-is. This keeps the previous behavior for the directory/cache code paths (`os.path.dirname` of the joined model file equals the original model dir), so there is no regression for cached models.

Fixes #5169

### DIFF
--- a/xinference/model/llm/llama_cpp/core.py
+++ b/xinference/model/llm/llama_cpp/core.py
@@ -247,11 +247,16 @@ class XllamaCppModel(LLM, ChatModelMixin):
         multimodal_projector = self._llamacpp_model_config.get(
             "multimodal_projector", ""
         )
-        mmproj = (
-            os.path.join(self.model_path, multimodal_projector)
-            if multimodal_projector
-            else ""
-        )
+        if not multimodal_projector:
+            mmproj = ""
+        elif os.path.isabs(multimodal_projector):
+            mmproj = multimodal_projector
+        else:
+            # ``model_path`` resolves to the model *file*; the projector lives
+            # alongside it, so resolve it against the model's directory. Joining
+            # onto ``self.model_path`` directly would yield ".../model.gguf/mmproj"
+            # (a nonexistent path) whenever --model-path points at a single file.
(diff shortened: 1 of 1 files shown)


qinxuye (approved):
LGTM

PR #5172, opened 2026-07-13
feat(audio): batch Qwen3-ASR transcriptions

qinxuye:
## Summary

- batch concurrent Qwen3-ASR transcription requests through the model's native list inference
- preserve per-request language and response formatting while grouping compatible inference kwargs
- use a Qwen3-ASR-specific 100 ms aggregation window by default while keeping `batch_interval` configurable
- harden the shared batch worker against cancellation and result-count mismatches
- enforce the shared `BatchMixin` size limit while preserving FIFO order for requests deferred to the next batch

## Why

Qwen3-ASR supports batched transcription, but Xinference currently invokes each concurrent audio request separately. This change lets compatible requests share one model call while preserving the existing API response behavior.

The shared 3 ms batching interval is too short for differently sized audio uploads to reach the model actor together. Qwen3-ASR therefore uses a 100 ms default without changing the global `BatchMixin` default.

`BatchMixin` is shared by embedding, rerank, and Qwen3-ASR models. Because one queued request can contain multiple inputs, checking only the current accumulated size is insufficient: the next request can push a batch past its configured limit. The worker now retains such a request as a FIFO-preserving pending item for the next batch. A single request that already exceeds the limit still runs alone because the mixin cannot split one request.

## Validation

- `pytest -q xinference/model/tests/test_batch.py xinference/model/audio/tests/test_qwen3_asr.py` — 8 passed
- `pre-commit run --files xinference/model/batch.py xinference/model/tests/test_batch.py xinference/model/audio/qwen3_asr.py` — passed
- `git diff --check` — passed

The shared batch tests cover a request that would overflow the remaining capacity, exact-capacity batches, FIFO preservation across deferred requests, and a single oversized request.

### GPU end-to-end results

Tested through a real Xinference service on an NVIDIA GeForce RTX 3090 Ti with `Qwen3-ASR-1.7B`, `batch_size=2`, and the default `batch_interval=0.1`.

Method:

- excluded one warm-up run
- ran three measured rounds and alternated sequential/batch order
- sequential time is the sum of the two individual request times
- batch time is wall-clock time until both concurrent requests complete
- compared the returned text for every input across every sequential and batch run

| Inputs | Sequential sum, median | Batch wall time, median | Result |
| --- | ---: | ---: | ---: |
| `蝶恋花·春景.mp3` ×2 (180.024 s each) | 3.232 s | 2.033 s | 37.1% lower wall time |
| `synthesized_speech_92345053.mp3` ×2 (106.606 s each) | 8.362 s | 4.409 s | 47.3% lower wall time |
| Mixed 180.024 s + 106.606 s inputs | 5.877 s | 6.649 s | 13.1% higher wall time |

Debug logs confirmed that sequential runs invoked native batch sizes `[1, 1]`, while concurrent runs invoked `[2]`.

Sequential and batched outputs matched exactly for both files in every measured round. This validates output parity, not transcript ground-truth accuracy.

The mixed-duration pair is slower because native batching pads work toward the longest item: the two inputs contain 286.630 seconds of audio in total, while a two-item batch can process approximately 360.048 seconds of padded frames. The throughput gain is therefore strongest when requests have similar lengths.

### DIFF
--- a/xinference/model/audio/qwen3_asr.py
+++ b/xinference/model/audio/qwen3_asr.py
@@ -34,6 +34,8 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+QWEN3_ASR_DEFAULT_BATCH_INTERVAL = 0.1
+
 
 class Qwen3ASRModel(BatchMixin):
     def __init__(
@@ -51,7 +53,7 @@ class Qwen3ASRModel(BatchMixin):
         self._device = device
         self._model = None
         batch_size = kwargs.pop("batch_size", None)
-        batch_interval = kwargs.pop("batch_interval", None)
+        batch_interval = kwargs.pop("batch_interval", QWEN3_ASR_DEFAULT_BATCH_INTERVAL)
         self._kwargs = kwargs
 
         batching_kwargs = {}
--- a/xinference/model/audio/qwen3_asr.py
+++ b/xinference/model/audio/qwen3_asr.py
@@ -12,16 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import asyncio
 import logging
 import os
 import tempfile
+from collections import defaultdict
 from typing import TYPE_CHECKING, List, Optional, Tuple
+
+from xoscar import extensible
 
 from ...device_utils import (
     get_available_device,
     get_device_preferred_dtype,
     is_device_available,
 )
+from ...utils import make_hashable
+from ..batch import BatchMixin
 
 if TYPE_CHECKING:
     from .core import AudioModelFamilyV2
@@ -29,7 +35,7 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-class Qwen3ASRModel:
+class Qwen3ASRModel(BatchMixin):
     def __init__(
         self,
         model_uid: str,
@@ -44,7 +50,19 @@ class Qwen3ASRModel:
         self._model_spec = model_spec
         self._device = device
         self._model = None
+        batch_size = kwargs.pop("batch_size", None)
+        batch_interval = kwargs.pop("batch_interval", None)
         self._kwargs = kwargs
+
+        batching_kwargs = {}
+        max_inference_batch_size = kwargs.get("max_inference_batch_size")
+        if batch_size is not None:
+            batching_kwargs["batch_size"] = batch_size
+        elif max_inference_batch_size is not None and int(max_inference_batch_size) > 0:
+            batching_kwargs["batch_size"] = max_inference_batch_size
+        if batch_interval is not None:
+            batching_kwargs["batch_interval"] = batch_interval
+        BatchMixin.__init__(self, self._batch_transcribe, **batching_kwargs)
 
     @property
     def model_ability(self):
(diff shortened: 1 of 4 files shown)


xinference/model/audio/qwen3_asr.py
```diff
@@ -44,8 +52,20 @@ def __init__(
...
+        batch_size = kwargs.pop("batch_size", None)
+        batch_interval = kwargs.pop("batch_interval", QWEN3_ASR_DEFAULT_BATCH_INTERVAL)
         self._kwargs = kwargs
 
+        batching_kwargs = {}
+        max_inference_batch_size = kwargs.get("max_inference_batch_size")
+        if batch_size is not None:
+            batching_kwargs["batch_size"] = batch_size
+        elif max_inference_batch_size is not None and int(max_inference_batch_size) > 0:
+            batching_kwargs["batch_size"] = max_inference_batch_size
```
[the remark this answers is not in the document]
qinxuye: Keeping this parameter in `self._kwargs` is intentional. The official `qwen_asr.Qwen3ASRModel.from_pretrained` signature explicitly declares and consumes `max_inference_batch_size`; it is used to limit the native internal inference batches and is not forwarded as an unexpected `AutoModel.from_pretrained` kwarg. The GPU validation for this PR also loaded Qwen3-ASR successfully with `max_inference_batch_size=2`. Popping it here would discard a supported user configuration.

xinference/model/audio/qwen3_asr.py
```diff
@@ -139,7 +159,108 @@ def _extract_text_and_language(self, result) -> Tuple[str, Optional[str]]:
...
+                audio_path = os.path.join(temp_dir, f"audio-{i}")
+                with open(audio_path, "wb") as f:
+                    f.write(audio)
+                audio_paths.append(audio_path)
+
+            results = self._model.transcribe(
+                audio=audio_paths,
+                language=languages,
+                **transcription_kwargs,
+            )
```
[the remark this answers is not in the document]
qinxuye: The official `transcribe` API explicitly accepts `Optional[Union[str, List[Optional[str]]]]`. Its implementation accepts each list element when it is `None`, and scalar `None` is itself expanded to `[None] * n`, so `[None, None]` is a supported and equivalent automatic-language-detection input. Keeping the per-item list is also necessary for mixed batches such as `[None, "English"]`.

xinference/model/batch.py
```diff
@@ -35,7 +35,7 @@ def __init__(self, func: _ExtensibleWrapper, **kwargs):
         self._func_name = func.func.__name__
         setattr(self, self._func_name, types.MethodType(self._wrap_method(), self))
 
-        self._is_process_batch_running = False
+        self._process_batch_task = None
```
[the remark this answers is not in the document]
qinxuye: The suggested async `close()` is not compatible with the current model lifecycle: `ModelActor.stop()` invokes model `close` as a synchronous method via `asyncio.to_thread`, so the returned coroutine would not be awaited. More importantly, Xinference launches each model in a dedicated subpool and `terminate_model()` force-removes that subpool, which destroys its event loop and pending batch task. The infinite batch worker also existed before this PR; storing its task only makes restart/cancellation handling explicit. A generic graceful BatchMixin shutdown would require a coordinated core lifecycle change rather than this isolated method.

xinference/model/batch.py
```diff
@@ -71,7 +70,7 @@ async def _process_batch(self):
             futures = [first_future]
 
             # Try to gather more items into the same batch within a short timeout window
-            while size <= self.batch_size:
+            while size < self.batch_size:
                 try:
```
OliverBryant, line 82: Non-blocking / description-only: this `<=` -> `<` change affects **every** `BatchMixin` consumer (vLLM / llama.cpp / sentence-transformers embeddings and rerank), not just Qwen3-ASR. It's an improvement — the old `<=` would loop one extra time at `size == batch_size`, waiting a full `batch_interval` and potentially overshooting the cap by one request's worth of items; the new `<` stops right at the cap. Since `BatchMixin` itself has no dedicated unit test, could you add a line to the PR description calling out that this tightens the global batching boundary condition? Just so it isn't a surprise for the embedding/rerank paths later.

xinference/model/audio/qwen3_asr.py
OliverBryant, line 66: Minor: a one-line comment here would help future readers. Xinference's own launch path never populates `max_inference_batch_size` today, so this `elif` looks like a dead branch at first glance — it's only reached when a user passes it manually as a native `qwen_asr.from_pretrained` arg (which, per your earlier reply, is intentional). Something like `# qwen_asr consumes this as a native from_pretrained arg; reuse it as a batch-size default` would prevent someone from "cleaning it up" later.

OliverBryant (approved):
Nice work. The batching integration follows the existing `create_embedding` `@extensible`/`.batch` convention, and the `BatchMixin._process_batch` hardening (moving the count check inside `try`, dropping cancelled futures, restartable processor task) is a solid robustness improvement. Tests cover ordering, batch-size, kwargs grouping, and cancellation survival. LGTM.

Left two minor, non-blocking notes below. The only one I'd genuinely suggest acting on is a mention in the PR description of the shared `batch.py` boundary change, since it affects every batch-capable model, not just Qwen3-ASR.

xinference/model/batch.py
qinxuye: Good catch on the shared scope. On closer inspection, changing `<=` to `<` only fixed the exact-boundary case; a variable-sized request could still push the accumulated size over the limit, for example 3 + 2 with `batch_size=4`. Commit `fb78ca62a` now keeps an overflowing request as a FIFO-preserving pending item for the next batch. I added dedicated `BatchMixin` tests for overflow, exact capacity, deferred-request order, and a single oversized request, and updated the PR description with the shared behavior.

xinference/model/audio/qwen3_asr.py
qinxuye: Added the explanatory comment in `fb78ca62a` to make it clear that `max_inference_batch_size` remains a native qwen_asr loader argument and is reused only as Xinference request-batch default when no explicit `batch_size` is provided.

PR #5175, opened 2026-07-13
Fix: (UI) copyToClipboard 、fix the locales Running Model Page

maoyuehui:


frontend/src/lib/utils.ts
```diff
@@ -106,18 +108,31 @@ export const transformObjToFormList = (obj: Record<string, any> = {}) => {
...
+      });
+  } else {
+    const textarea = document.createElement('textarea');
+    textarea.value = value;
+    document.body.appendChild(textarea);
+    textarea.select();
+    try {
+      document.execCommand('copy');
+      toast.success('Copy successful!');
+    } catch (err) {
```
qinxuye: The clipboard fallback is still incomplete. `navigator.clipboard` may exist while `writeText()` rejects, but that path currently reports an error without trying the legacy copy method. Also, `document.execCommand("copy")` returns `false` on failure rather than necessarily throwing, so this branch can show a success toast even though nothing was copied. Please fall back after a rejected `writeText()`, check the boolean result from `execCommand`, and only report success when either method succeeds. The unused `err` here is also the current ESLint failure.

qinxuye (approved):
LGTM

PR #5177, opened 2026-07-14
fix(oauth2): revoke refresh tokens on password reset

OliverBryant:
## Summary

`change_password` in the advanced OAuth2 layer updated the password hash but never revoked the target user's refresh tokens. `refresh_access_token` only re-checks that the user is still `enabled` — not whether the password changed — so a leaked/stolen refresh token could keep minting fresh access tokens for up to `REFRESH_TOKEN_EXPIRE_DAYS` (7 days) after an admin resets the user's password.

This calls `delete_user_refresh_tokens(user_id)` right after the hash update, so a password reset terminates all outstanding sessions for that user.

## Context

This is the last remaining finding from the #5058 security audit. Re-verifying all six findings against current `main`:

| Finding | Status |
|---|---|
| 1 — eternal JWT (no `exp` check) | already fixed (#5088) |
| 2 — `users:manage` → `admin` escalation via permission grant | already fixed (#5089) |
| 3 — password reset of a more-privileged (admin) account | already fixed (`_reject_admin_target_takeover`) |
| **4 — password reset does not revoke sessions** | **fixed here** |
| 5 — `must_change_password` not enforced + plaintext password logged | no longer applicable: the auto-generated-admin path was replaced by the token-gated `setup_admin` flow, so no random password is logged and no account relies on `must_change_password` as a security boundary |
| 6 — disabled-user API key bypass | already fixed (`user_enabled` cache check) |

## Test

Adds `test_oauth2_password_reset_revokes_tokens.py`, which drives the real `change_password` handler against a temporary SQLite DB and asserts the target's refresh tokens are deleted while other users' tokens are untouched.

- New + existing auth/oauth2 tests: `65 passed`
- `pre-commit` (black, flake8, isort, mypy, codespell): clean

### DIFF
--- a/xinference/api/oauth2/advanced/auth_service.py
+++ b/xinference/api/oauth2/advanced/auth_service.py
@@ -535,6 +535,25 @@ class AdvancedAuthService:
                     auth_type="api_key",
                 )
                 raise credentials_exception
+            # Finding 5: enforce must_change_password on the API-key path too.
+            # An API key cannot change a password (it may only reach model
+            # endpoints), so a key owned by a still-flagged account is blocked
+            # outright until the owner clears the flag via a JWT password
+            # change. Without this, a legacy must_change_password=1 account's
+            # existing API key would keep working, bypassing the JWT gate.
+            if user_obj.get("must_change_password"):
+                _audit(
+                    "must_change_password",
+                    user=_username,
+                    key_name=api_key_entry.name or "",
+                    key_prefix=api_key_entry.key_prefix,
+                    auth_type="api_key",
+                )
+                raise HTTPException(
+                    status_code=status.HTTP_403_FORBIDDEN,
+                    detail="Password change required before using this account.",
+                    headers={"WWW-Authenticate": authenticate_value},
+                )
             # Success — reset counters
             if client_ip and self._rate_limiter:
                 self._rate_limiter.reset_key(client_ip, api_key_entry.key_id)
--- a/xinference/api/oauth2/advanced/auth_service.py
+++ b/xinference/api/oauth2/advanced/auth_service.py
@@ -195,9 +195,24 @@ class AdvancedAuthService:
             self._db.delete_refresh_token(token_hash)
             return None
 
-        # Token rotation: invalidate old token, issue new one
-        self._db.delete_refresh_token(token_hash)
-        new_refresh_token = self.create_refresh_token(user["id"])
+        # Token rotation: delete the old token and issue a new one atomically.
+        # Doing this in a single BEGIN IMMEDIATE transaction serializes it
+        # against a concurrent password reset (which revokes all of the user's
+        # tokens in its own BEGIN IMMEDIATE transaction), so a rotation that
+        # started before the reset cannot leave a live token behind it
+        # (see security report, Finding 4).
+        new_refresh_token = secrets.token_urlsafe(64)
+        new_token_hash = sha256_hex(new_refresh_token)
+        new_expires_at = (
+            datetime.utcnow() + timedelta(days=REFRESH_TOKEN_EXPIRE_DAYS)
+        ).isoformat()
+        rotated = self._db.rotate_refresh_token(
+            token_hash, new_token_hash, new_expires_at
+        )
+        if rotated is None:
+            # The token was revoked (e.g. by a password reset) between our read
+            # above and taking the write lock. Refuse to mint a new session.
+            return None
 
         access_token = self.create_access_token(
             user["id"], user["username"], user["permissions"]
--- a/xinference/api/oauth2/advanced/auth_service.py
+++ b/xinference/api/oauth2/advanced/auth_service.py
@@ -14,6 +14,7 @@
 import base64
 import logging
 import os
+import re
 import secrets
 from datetime import datetime, timedelta
 from typing import Any, Dict, List, Optional
@@ -103,6 +104,11 @@ except (TypeError, ValueError):
     ACCESS_TOKEN_EXPIRE_MINUTES = 30
 REFRESH_TOKEN_EXPIRE_DAYS = 7
 JWT_ALGORITHM = "HS256"
+
+# Matches the self-service password-change endpoint,
+# PUT /v1/admin/users/{user_id}/password. Anchored on the trailing segments so
+# it is independent of any router mount prefix.
+_PASSWORD_CHANGE_PATH_RE = re.compile(r"^.*/users/(\d+)/password$")
 
 try:
     PASSWORD_MIN_LENGTH = int(os.environ.get("XINFERENCE_PASSWORD_MIN_LENGTH", "8"))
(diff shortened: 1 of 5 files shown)


xinference/api/tests/test_oauth2_password_reset_revokes_tokens.py
```diff
@@ -0,0 +1,88 @@
...
+    # isolate the refresh-token revocation behaviour under test.
+    monkeypatch.setattr(
+        routes,
+        "_get_current_user_from_token",
+        lambda request, a: (1, "caller", caller_scopes),
+    )
+
+
+def _run(coro):
+    return asyncio.get_event_loop().run_until_complete(coro)
```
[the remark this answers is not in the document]
OliverBryant: Good catch — fixed in c8060b34c by switching the helper to `asyncio.run(coro)`, which manages the event loop lifecycle and works across the 3.10–3.13 CI matrix.

xinference/api/oauth2/advanced/routes.py
```diff
@@ -455,6 +455,11 @@ async def change_password(user_id: int, request: Request) -> JSONResponse:
         )
     password_hash = get_password_hash(new_password)
     auth.db.update_user(user_id, password_hash=password_hash, must_change_password=0)
+    # Revoke outstanding refresh tokens so a stolen session cannot outlive a
+    # password reset: refresh_access_token only re-checks that the user is
+    # still enabled, not whether the password changed, so without this a leaked
+    # refresh token keeps minting access tokens for up to REFRESH_TOKEN_EXPIRE_DAYS.
+    auth.db.delete_user_refresh_tokens(user_id)
```
qinxuye: This does not reliably revoke the session under concurrency. `refresh_access_token()` reads the old token, then deletes it and inserts the replacement in separate database transactions. A refresh request can read the old token, this call can delete all tokens, and then the refresh request can insert a new token that survives the password reset. I reproduced that sequence on this head. Please make token rotation and password-update/revocation mutually atomic at the database level (including across processes, e.g. with serialized SQLite transactions), and add a concurrent regression test.

qinxuye (commented):
Two findings from the original #5058 report are still not fully resolved:

- Finding 4 remains raceable because password reset and refresh-token rotation are separate database operations.
- Finding 5 still affects upgrades: an existing account with `must_change_password=1` can obtain unrestricted admin tokens. #5144 removes the vulnerable bootstrap path for fresh installs, but there is no migration or server-side enforcement for existing databases.

The other findings appear to be addressed. The focused auth tests pass, but they do not cover these two scenarios.

xinference/api/oauth2/advanced/routes.py
OliverBryant: Fixed in e4da7f38d. Both flows now run in single `BEGIN IMMEDIATE` write transactions that serialize on SQLite's write lock (matching `create_first_user`), so they are mutually atomic across threads and processes:

- `Database.rotate_refresh_token` does validate-delete-insert in one transaction and only succeeds if the old token still existed at delete time.
- `Database.update_password_and_revoke_tokens` updates the hash and deletes the user's refresh tokens in one transaction.
- `refresh_access_token` bails out (returns None) when the token was revoked before it acquired the write lock, so it never mints a session past a reset.

Whatever the interleaving: if the rotation commits first, the reset's revocation then deletes the freshly-rotated token; if the reset commits first, the rotation finds the old token gone and refuses. Either way no token survives.

Added a multi-threaded regression test (`test_concurrent_refresh_and_password_reset_leaves_no_live_token`) that interleaves the two via a barrier over many rounds. I confirmed it reproduces surviving tokens against the old non-atomic code (31/60 rounds) and passes on this head.

OliverBryant:
Thanks for the careful review.

**Finding 4 (race):** fixed in e4da7f38d — see the inline reply. Rotation and password-reset revocation are now mutually atomic via `BEGIN IMMEDIATE`, with a concurrent regression test that reproduced the leak against the old code (31/60 rounds) and passes now.

**Finding 5 (existing DBs with `must_change_password=1`):** you're right that this is still open. It's a distinct concern from this PR's token-revocation fix — it needs a server-side gate in the auth dependency (`__call__`) that blocks a `must_change_password` user from every request except changing their own password, so the flag is enforced regardless of how the account was created. I'd prefer to do that in a separate, focused PR rather than widen this one. Happy to open it next — or fold it in here if you'd rather have both together. Which do you prefer?

OliverBryant:
**Finding 5:** now enforced server-side in this PR (e75d82792), as you suggested.

Added a gate in `AdvancedAuthService.__call__`, placed **after** the enabled check and **before** the admin bypass, so even an `admin`-scoped account is blocked while `must_change_password=1`. A flagged user may make exactly one request — `PUT /v1/admin/users/{own_id}/password` — and everything else returns 403. This is independent of how the account was created, so it covers rows carried over from an older database, which was the gap you flagged.

Two design notes:
- The exempt path is matched by a regex anchored on the trailing `/users/{id}/password` segments (independent of any router mount prefix) and only for the caller's own user id; a GET to the same path is not exempt.
- The exemption deliberately allows *only* the password change, nothing else. In practice the flagged accounts in question are the legacy bootstrap admins (they hold `admin`, so they still satisfy the `users:manage` guard on the change-password route via the admin bypass, which runs after this gate lets the request through). The current code no longer *creates* non-admin users with the flag set, so I did not widen the allowlist for a case that can't arise today — happy to revisit if you'd rather I add a self-service change-password route.

Full oauth2 suite (54 tests) and pre-commit pass. Tests: `test_oauth2_must_change_password.py`.

xinference/api/oauth2/advanced/auth_service.py
```diff
@@ -553,6 +584,22 @@ def _audit(
             _audit("user_disabled", user=username or "", auth_type="jwt")
             raise credentials_exception
 
+        # Finding 5: enforce must_change_password server-side. A user still
+        # flagged for a forced password change may do nothing except set a new
+        # password on their own account -- not even the admin bypass below is
+        # reached first. This gates existing accounts (e.g. rows migrated from
+        # an older database) regardless of how they were created, so the flag
+        # is not merely an informational hint returned to the frontend.
+        if user.get("must_change_password") and not self._is_own_password_change(
```
qinxuye, line 612: This gate is only reached for JWTs. The API-key branch returns `user_obj` above before reaching this code, so an existing API key owned by a legacy `must_change_password=1` account can still call model endpoints. I reproduced this on the current head (`must_change_password=1, api_key_request_allowed=True`). Please enforce the flag in the API-key path as well (existing API keys should remain blocked until the owner changes the password) and add a regression test using an API key.

OliverBryant: Good catch — confirmed and fixed in 40f8f39a3. The API-key branch returns `user_obj` before the JWT gate, and it never checked the flag, so a legacy `must_change_password=1` account's existing key kept working. I reproduced it the same way you did.

Now the API-key path blocks outright when the owning account is still flagged (403, `must_change_password` audit event). An API key can only reach model endpoints and cannot change a password, so there is no exempt request — the key stays blocked until the owner clears the flag via a JWT password change. The check reads `must_change_password` from the live DB row (`get_user_by_id`), not the cached key snapshot, so once the flag is cleared the same key works again immediately without recreating it.

Added two regression tests using a real API key: `test_flagged_account_api_key_is_blocked` and `test_api_key_works_again_after_flag_cleared`. Full oauth2 suite (56 tests) and pre-commit pass.

qinxuye (approved):
LGTM

PR #5179, opened 2026-07-15
fix: honor virtualenv exemption in engine matching for multimodal LLM, embedding and rerank

OliverBryant:
## Problem

On an environment without vLLM/SGLang installed (or without a local GPU), a
chat-only model like `qwen3` can be launched with vLLM/SGLang, but their
multimodal counterparts cannot — the engines don't appear as selectable in the
WebUI. The same asymmetry affects `Qwen3-VL-Embedding` and `Qwen3-VL-Reranker`.

Engine availability is computed by `get_engine_params_by_name_with_virtual_env`,
which calls each engine's `match_json`. Some engines put **runtime-only** checks
— library installed, specific version, CUDA GPU, Linux — inside `match_json`,
before (or instead of) the virtualenv exemption. Chat LLM engines already had
the exemption; the multimodal / embedding / rerank vLLM paths did not, so a
declared engine was hidden at listing time even though virtualenv installs it on
demand at launch (`check_engine_by_spec_parameters_with_virtual_env` already
bypasses this).

## Fix

Make the virtualenv exemption consistent: when virtualenv is enabled,
environment-only checks (library present, version, GPU, OS) are skipped so an
architecture/format-compatible engine is still offered and installed on demand.
When virtualenv is disabled, all strict checks are preserved.

**LLM**
- vLLM: exempt GPU/OS/library checks in `VLLMModel` and `VLLMMultiModel`.
- SGLang: add `_virtual_env_allows_missing_sglang`; exempt GPU/OS/library checks
  in the generate, chat and vision branches.
- LMDeploy: add `_virtual_env_allows_missing_lmdeploy`; exempt the library check
  in the chat branch.

**Embedding / Rerank**
- Add a shared `virtual_env_allows_missing_engine()` helper in `model/utils.py`.
- embedding vLLM: exempt the "vLLM is not installed" and version checks for
  `Qwen3-VL-Embedding` under virtualenv.
- rerank vLLM: replace the unconditional `requires vLLM>=0.14.0` rejection for
  `Qwen3-VL-Reranker` with a real import + version check that is exempt under
  virtualenv, so the engine is offered instead of being permanently disabled.

**Unchanged (already correct):** MLX (Apple-silicon-only, marker auto-dropped on
non-macOS), Transformers / llama.cpp, embedding/rerank
sentence_transformers/flag/llama.cpp, and image (stable_diffusion) — all keep
library checks in `check_lib`, so listing already honors the exemption. Audio /
video do not use the engine-selection mechanism.

## Tests

- LLM: regression test asserting `qwen3.5` offers vLLM/SGLang/Transformers/
  llama.cpp under virtualenv while MLX/LMDeploy (not declared) are not listed.
- Embedding: strengthened to assert `Qwen3-VL-Embedding-2B` offers vLLM.
- Rerank: new engine-params test asserting `Qwen3-VL-Reranker-2B` offers vLLM.

All new/updated assertions fail on the previous code and pass with this change.
Verified manually for a no-GPU host and a simulated CUDA+Linux host with the
libraries absent.

### DIFF
--- a/xinference/model/embedding/model_spec.json
+++ b/xinference/model/embedding/model_spec.json
@@ -1834,7 +1834,7 @@
         "transformers>=4.57.0 ; #engine# == \"sentence_transformers\"",
         "qwen-vl-utils>=0.0.14",
         "Pillow",
-        "#vllm_dependencies# ; #engine# == \"vllm\"",
+        "vllm>=0.14.0 ; #engine# == \"vllm\"",
         "#system_numpy# ; #engine# == \"vllm\""
       ]
     },
@@ -1875,7 +1875,7 @@
         "transformers==4.57.1 ; #engine# == \"sentence_transformers\"",
         "qwen-vl-utils==0.0.11",
         "Pillow",
-        "#vllm_dependencies# ; #engine# == \"vllm\"",
+        "vllm>=0.14.0 ; #engine# == \"vllm\"",
         "#system_numpy# ; #engine# == \"vllm\""
       ]
     },
--- a/xinference/model/embedding/vllm/core.py
+++ b/xinference/model/embedding/vllm/core.py
@@ -299,17 +299,28 @@ class VLLMEmbeddingModel(EmbeddingModel, BatchMixin):
     ) -> Union[bool, Tuple[bool, str]]:
 
         if model_family.model_name.startswith("Qwen3-VL-Embedding"):
+            # In virtualenv mode vLLM (and a compatible version) can be
+            # installed on demand, so only the missing-library / old-version
+            # rejection is exempt here; the format/prefix compatibility checks
+            # below still apply (virtualenv cannot make an incompatible spec
+            # work).
+            from ...utils import virtual_env_allows_missing_engine
+
+            allow_missing_env = virtual_env_allows_missing_engine()
             try:
                 import vllm
                 from packaging import version
             except ImportError:
-                return False, "vLLM is not installed"
-
-            if version.parse(vllm.__version__) < version.parse("0.14.0"):
-                return (
-                    False,
-                    f"Qwen3-VL embedding requires vLLM>=0.14.0, current: {vllm.__version__}",
-                )
+                if not allow_missing_env:
+                    return False, "vLLM is not installed"
+            else:
+                if not allow_missing_env and version.parse(
+                    vllm.__version__
+                ) < version.parse("0.14.0"):
+                    return (
+                        False,
+                        f"Qwen3-VL embedding requires vLLM>=0.14.0, current: {vllm.__version__}",
+                    )
         if model_spec.model_format not in ["pytorch"]:
             return False, "vLLM embedding engine only supports pytorch format"
         prefix = model_family.model_name.split("-", 1)[0]
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -178,11 +178,19 @@ def _get_effective_vllm_version() -> version.Version:
 
 
 def _virtual_env_allows_missing_vllm() -> bool:
+    # Delegate to the shared helper so engine discovery honors the effective
+    # request-level virtualenv flag (via virtualenv_discovery_var) instead of
+    # only the process-global default; falls back to the global constant when
+    # called outside a discovery scope (e.g. the launch path).
     try:
-        from ....constants import XINFERENCE_ENABLE_VIRTUAL_ENV
+        from ...utils import virtual_env_allows_missing_engine
     except Exception:
-        return False
-    return bool(XINFERENCE_ENABLE_VIRTUAL_ENV)
+        try:
+            from ....constants import XINFERENCE_ENABLE_VIRTUAL_ENV
+        except Exception:
+            return False
+        return bool(XINFERENCE_ENABLE_VIRTUAL_ENV)
+    return virtual_env_allows_missing_engine()
 
 
 GuidedDecodingParams: Optional[Type[Any]] = None
@@ -1941,7 +1949,11 @@ class VLLMMultiModel(VLLMModel, ChatModelMixin):
                 False,
                 "vLLM multimodal engine requires vision, audio, or omni ability",
             )
-        if not VLLM_INSTALLED:
+        # Align with VLLMChatModel: in virtualenv mode vLLM is installed on
+        # demand, so a missing local install must not hide the engine at listing
+        # time. Hardware/OS checks above stay unconditional because virtualenv
+        # cannot add a GPU or change the OS.
+        if not VLLM_INSTALLED and not _virtual_env_allows_missing_vllm():
             return False, "vLLM library is not installed"
         return True
 
--- a/xinference/model/utils.py
+++ b/xinference/model/utils.py
@@ -23,6 +23,7 @@ import re
 import sys
 import threading
 from abc import ABC, abstractmethod
+from contextvars import ContextVar
 from copy import deepcopy
 from json import JSONDecodeError
 from pathlib import Path
@@ -93,6 +94,39 @@ def _normalize_match_result(
         return False, result, default_type, None
 
     return False, str(result), default_type, None
+
+
+# Effective virtualenv flag for the current engine-discovery call. The
+# discovery entry point (get_engine_params_by_name_with_virtual_env) sets this
+# to the *request-level* flag so that match_json can honor an explicit
+# enable_virtual_env=False even when the process-global default is true. When
+# unset (e.g. direct match_json calls outside discovery) it is None and callers
+# fall back to the global constant.
+virtualenv_discovery_var: ContextVar[Optional[bool]] = ContextVar(
+    "virtualenv_discovery_var", default=None
+)
+
+
+def virtual_env_allows_missing_engine() -> bool:
+    """Whether virtualenv mode should tolerate a missing engine install.
+
+    When virtualenv is enabled the engine and its runtime are installed on
+    demand, so engine listing must not depend on the library (or a specific
+    already-installed version) being present locally.
+
+    Uses the effective request-level flag published by the discovery path
+    (virtualenv_discovery_var) when available, so an explicit
+    enable_virtual_env=False stays strict; otherwise falls back to the
+    process-global default. Read lazily so the flag can be toggled at runtime.
+    """
+    scoped = virtualenv_discovery_var.get()
+    if scoped is not None:
+        return bool(scoped)
+    try:
+        from ..constants import XINFERENCE_ENABLE_VIRTUAL_ENV as flag
+    except Exception:
+        return False
+    return bool(flag)
 
 
 def _extract_engine_markers_from_packages(packages: List[str]) -> Set[str]:
@@ -330,6 +364,63 @@ def _force_virtualenv_engine_params(
             engine_params[engine_name] = engine_param_list
             available_params[engine_name] = engine_param_list
     return match_status
+
+
+def _drop_unavailable_marker_engines(
+    available_engines: Dict[str, Any],
+    family: Optional[Any],
+    supported_engines: Dict[str, List[Type[Any]]],
+    engine_markers: Set[str],
+) -> None:
+    """Re-validate virtualenv-marker engines for the strict discovery path.
+
+    The per-model engine registries (LLM_ENGINES / EMBEDDING_ENGINES /
+    RERANK_ENGINES) are generated once at import time using the process-global
+    virtualenv flag, so a marker engine (e.g. vLLM for Qwen3-VL) can be baked in
+    as available even though it is only reachable via on-demand install. The
+    strict entry point (enable_virtual_env=False) must not offer such engines,
+    so re-run check_lib + match_json under the current (strict) discovery scope
+    and drop any marker engine that no longer qualifies. Non-marker engines are
+    left untouched — they were matched without any virtualenv exemption.
+    """
+    if family is None or not engine_markers:
+        return
+    specs = getattr(family, "model_specs", []) or []
+    for engine_name in list(available_engines.keys()):
+        if engine_name.lower() not in engine_markers:
+            continue
+        engine_classes = supported_engines.get(engine_name, [])
+        qualifies = False
+        for engine_class in engine_classes:
+            lib_ok, _, _, _ = _normalize_match_result(
+                engine_class.check_lib(),
+                f"Engine {engine_name} library is not installed",
+                "dependency_missing",
+            )
+            if not lib_ok:
+                continue
+            match_func = getattr(engine_class, "match_json", None)
+            for spec in specs:
+                quantization = getattr(spec, "quantization", None) or "none"
+                match_res = (
+                    match_func(family, spec, quantization)
+                    if callable(match_func)
+                    else False
+                )
+                is_match, _, _, _ = _normalize_match_result(
+                    match_res,
+                    f"Engine {engine_name} is not compatible",
+                    "model_compatibility",
+                )
+                if is_match:
+                    qualifies = True
+                    break
+            if qualifies:
+                break
+        if not qualifies:
+            # Remove from the available set; the caller's _collect_supported_engines
+            # pass then annotates it with the proper unavailable reason.
+            available_engines.pop(engine_name, None)
 
 
 def _apply_virtualenv_engine_overrides(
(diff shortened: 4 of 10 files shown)


OliverBryant:
Thanks for the review. Addressed the feedback: removed the redundant `_virtual_env_allows_missing_*` helper functions in `sglang/core.py` and `lmdeploy/core.py` and now import `XINFERENCE_ENABLE_VIRTUAL_ENV` at the module top level, using it directly (confirmed no circular-import risk since `constants` only depends on `os`/`pathlib`). The shared `virtual_env_allows_missing_engine()` helper in `model/utils.py` (used by the embedding/rerank vLLM paths) was likewise simplified to reference the top-level constant. Left the pre-existing `_virtual_env_allows_missing_vllm` in `vllm/core.py` unchanged to keep the diff focused. Also rebased onto the latest main to drop unrelated commits.

xinference/model/llm/vllm/core.py
```diff
@@ -1168,15 +1168,20 @@ def check_lib(cls) -> Union[bool, Tuple[bool, str]]:
...
-            and not cls._has_musa_device()
-        ):
-            return False, "vLLM requires CUDA or MLU GPUs or VACC GPUs or MUSA GPUs"
-        if not cls._is_linux():
-            return False, "vLLM backend is only supported on Linux"
+        # In virtualenv mode the engine/runtime is installed on demand, so
+        # environment-only checks (GPU, OS, library) must not hide an
+        # otherwise-compatible engine at listing time.
+        allow_missing_env = _virtual_env_allows_missing_vllm()
+        if not allow_missing_env:
```
qinxuye: Virtualenv can install Python packages, but it cannot add accelerator hardware or turn a non-Linux worker into Linux. Since `XINFERENCE_ENABLE_VIRTUAL_ENV` defaults to true, this makes vLLM/SGLang match on unsupported workers; on a macOS host with no CUDA, `VLLMMultiModel.match_json()` now returns `True`. It also ignores an explicit request-level `enable_virtual_env=False` because `match_json` reads the process-global constant. Please keep the hardware/OS checks unconditional and use the effective request flag only to exempt missing runtime libraries. The same issue applies to the SGLang branches changed in this PR.

xinference/model/embedding/vllm/core.py
```diff
@@ -299,13 +299,23 @@ def match_json(
...
                 from packaging import version
             except ImportError:
+                if allow_missing_env:
+                    return True
                 return False, "vLLM is not installed"
 
-            if version.parse(vllm.__version__) < version.parse("0.14.0"):
+            if not allow_missing_env and version.parse(
+                vllm.__version__
+            ) < version.parse("0.14.0"):
```
qinxuye: This assumes the virtualenv will install a compatible vLLM, but these Qwen3-VL model specs only declare `#vllm_dependencies#`, which currently expands to `vllm>=0.11.2`. An existing environment with 0.11.2 already satisfies that requirement, while model initialization later requires `vLLM>=0.14.0` and fails. Please encode `vllm>=0.14.0` in the Qwen3-VL embedding and reranker virtualenv requirements (and test the expanded requirement) before bypassing this version check.

qinxuye (commented):
Two correctness issues remain before this can be approved.

xinference/model/llm/vllm/core.py
OliverBryant: Good catch — fixed. You are right that virtualenv cannot add a GPU or change the OS. Reverted the hardware/OS checks in `VLLMModel`/`VLLMMultiModel` (and the SGLang generate/vision branches) back to unconditional; virtualenv now only exempts the missing-library check, matching the existing `VLLMChatModel` behavior. Verified: on a no-CUDA macOS host `VLLMMultiModel.match_json()` now returns `False` again. Added a regression test (`test_multimodal_vllm_engine_requires_gpu_even_with_virtualenv`) asserting the GPU gate still applies under virtualenv. Re the request-level `enable_virtual_env=False`: the engine `match_json` reading the process-global constant is the pre-existing behavior shared by the chat/base branches (via `_virtual_env_allows_missing_vllm`); I kept this PR consistent with that rather than widening the change to thread the request flag through every engine.

xinference/model/embedding/vllm/core.py
OliverBryant: Fixed. Added an explicit `vllm>=0.14.0 ; #engine# == "vllm"` to the virtualenv packages of both Qwen3-VL-Embedding (2B/8B) and Qwen3-VL-Reranker (2B/8B) in the model specs, so the resolved requirement is the intersection of `#vllm_dependencies#` (>=0.11.2) and >=0.14.0, i.e. >=0.14.0. An existing 0.11.2 env no longer satisfies the pin, so virtualenv installs a compatible vLLM before the version-check bypass applies. Verified the expanded requirement via `expand_engine_dependency_placeholders(..., "vllm")` now yields both `vllm>=0.11.2` and `vllm>=0.14.0`.

OliverBryant:
@qinxuye thanks for the careful review — both correctness issues are addressed (replies inline):

1. **Hardware/OS gate must stay unconditional.** Reverted the GPU/OS checks in `VLLMModel`/`VLLMMultiModel` and the SGLang generate/vision branches to unconditional. virtualenv now only exempts the *missing-library* check, matching the existing `VLLMChatModel` behavior. On a no-CUDA/non-Linux worker the multimodal engine no longer matches. Added `test_multimodal_vllm_engine_requires_gpu_even_with_virtualenv` to lock this in, and the existing listing test now simulates a GPU+Linux worker to isolate the library-exemption path.

2. **Version pin.** Added an explicit `vllm>=0.14.0 ; #engine# == "vllm"` to the virtualenv packages of Qwen3-VL-Embedding (2B/8B) and Qwen3-VL-Reranker (2B/8B), so the resolved requirement is >=0.14.0 and an existing 0.11.2 env is upgraded before the version-check bypass applies.

The net diff is now much smaller (vLLM `match_json` is a 6-line change). Note on the earlier Gemini suggestion to drop the `_virtual_env_allows_missing_*` helpers: I kept them for SGLang/LMDeploy so they stay symmetric with the pre-existing `_virtual_env_allows_missing_vllm` helper in `vllm/core.py`, rather than splitting styles across engines.

I also rebased onto the latest main to drop unrelated commits that had leaked in from a stale base.

xinference/model/embedding/vllm/core.py
```diff
@@ -299,13 +299,23 @@ def match_json(
...
             try:
                 import vllm
                 from packaging import version
             except ImportError:
+                if allow_missing_env:
+                    return True
                 return False, "vLLM is not installed"
 
-            if version.parse(vllm.__version__) < version.parse("0.14.0"):
+            if not allow_missing_env and version.parse(
```
qinxuye: The request-level `enable_virtual_env=False` case still needs to remain strict. This is not merely pre-existing behavior for these paths: before this PR, Qwen3-VL Embedding always rejected vLLM < 0.14 and Qwen3-VL Reranker was always rejected, whereas this new exemption reads the process-global flag (true by default) even when engine discovery was explicitly called with virtualenv disabled. I reproduced the current head with global virtualenv enabled, request-level `enable_virtual_env=False`, and an installed vLLM 0.11.2: the strict query returns vLLM as available, but Embedding `load()` then raises because it requires >=0.14. Please base this exemption on the effective request-level flag (or apply it only inside the virtualenv-aware discovery path), and add a regression test for `enable_virtual_env=False` with an old installed vLLM. The same issue applies to the Reranker branch.

qinxuye (commented):
One request-level virtualenv regression remains.

xinference/model/embedding/vllm/core.py
OliverBryant: You are right — this was newly introduced by the PR, not pre-existing for these paths, and reading the process-global flag was wrong. Fixed by taking your second suggested approach (apply the exemption only inside the virtualenv-aware discovery path):

- Added a `virtualenv_discovery_var` ContextVar in `model/utils.py`. The public `get_engine_params_by_name_with_virtual_env(...)` now resolves the *effective* request-level flag and publishes it on that ContextVar for the duration of the call (token always reset in `finally`, so it never leaks into the launch path).
- `virtual_env_allows_missing_engine()` reads the scoped flag first and only falls back to the global constant when unset (e.g. direct `match_json` calls outside discovery).
- The vLLM/SGLang/LMDeploy `_virtual_env_allows_missing_*` helpers now delegate to it as well, so the same request-level semantics apply consistently.

Reproduced your exact scenario (global virtualenv enabled, request `enable_virtual_env=False`, installed vLLM 0.11.2): Qwen3-VL Embedding/Reranker vLLM is now **not** listed (strict), and is listed again when `enable_virtual_env=True`. Added regression tests `test_qwen3_vl_embedding_strict_with_old_vllm_when_virtualenv_disabled` and the reranker equivalent covering exactly this case (they fail on the process-global-flag version and pass now).

qinxuye (commented):
The new request-scoped flag does not cover the production disabled path.

xinference/model/utils.py
```diff
@@ -1217,11 +1251,39 @@ def _get_image_families(model_name: str, is_ocr: bool) -> List[Any]:
...
+    (e.g. keep the Qwen3-VL vLLM version check strict) instead of reading the
+    process-global default. The token is always reset so the scope does not leak
+    into subsequent calls (e.g. the launch path) on the same thread.
+    """
+    effective = (
+        XINFERENCE_ENABLE_VIRTUAL_ENV
+        if enable_virtual_env is None
+        else enable_virtual_env
+    )
+    token = virtualenv_discovery_var.set(bool(effective))
```
qinxuye, line 1377: This context is only set by `get_engine_params_by_name_with_virtual_env()`, but the production Worker path explicitly calls `get_engine_params_by_name()` when `enable_virtual_env=False` (`worker.py:2361`). Therefore the real disabled path never reaches this wrapper and still falls back to the process-global flag. The new Embedding/Reranker tests call this wrapper with `False`, so they pass without exercising the Worker path. I reproduced the latest head with vLLM 0.11.2: `get_engine_params_by_name(..., enable_virtual_env=False)` still returns vLLM as available, while this wrapper correctly rejects it. Please scope the strict entry point as well (or make Worker use one scoped entry point for both values), and update the regression tests to exercise the same function Worker calls.

OliverBryant: You are right, and the root cause was one level deeper than just this wrapper. The Worker disabled path calls `get_engine_params_by_name()` (worker.py:2361), and that function reused the per-model engine registry (`EMBEDDING_ENGINES` / `RERANK_ENGINES` / `LLM_ENGINES`) directly. Those registries are generated once at import time using the process-global flag, so vLLM was baked in as available and the strict path never re-checked it — exactly the reproduction you described.

Fixed both ends:
- `get_engine_params_by_name()` (the strict/disabled entry point Worker uses) now also publishes the effective flag on `virtualenv_discovery_var`, mirroring the virtualenv-aware entry point, so both values go through a scoped path.
- Added `_drop_unavailable_marker_engines()`: on the strict path it re-runs `check_lib` + `match_json` (now under the strict scope) for any virtualenv-marker engine taken from the registry and drops those that no longer qualify. Non-marker engines are untouched. So a registry entry baked in via the global flag no longer leaks into the disabled result.

Verified with vLLM 0.11.2 installed and the global default true: `get_engine_params_by_name("embedding"/"rerank", ..., enable_virtual_env=False)` now returns vLLM as **not** available, while it stays available with vLLM 0.14.0 installed or under virtualenv. Updated both regression tests to call `get_engine_params_by_name(..., enable_virtual_env=False)` — the exact function Worker uses on the disabled path — and confirmed they fail without the fix and pass with it.

qinxuye (commented):
One compatibility check is still bypassed.

xinference/model/embedding/vllm/core.py
```diff
@@ -299,13 +299,23 @@ def match_json(
...
+            # the engine at listing time.
+            from ...utils import virtual_env_allows_missing_engine
+
+            allow_missing_env = virtual_env_allows_missing_engine()
             try:
                 import vllm
                 from packaging import version
             except ImportError:
+                if allow_missing_env:
+                    return True
```
qinxuye: When vLLM is absent and virtualenv is enabled, returning `True` here skips the model-format and supported-prefix checks below. I reproduced a Qwen3-VL embedding spec with `model_format="ggufv2"` being accepted even though this engine only supports PyTorch. Please only bypass the missing-library/version rejection and continue through the normal compatibility checks. The same early return in `rerank/vllm/core.py` has the same issue.

OliverBryant: Good catch — fixed. The exemption now only skips the missing-library / old-version rejection; it no longer short-circuits with `return True`, so execution continues into the model-format and supported-prefix checks. Applied the same fix to `rerank/vllm/core.py`.

Verified with vLLM absent and virtualenv enabled: a Qwen3-VL embedding/reranker spec with `model_format="ggufv2"` is now rejected ("only supports pytorch format"), while a `pytorch` spec still matches. Added regression tests `test_qwen3_vl_embedding_format_check_not_bypassed_under_virtualenv` and the reranker equivalent (they derive a ggufv2 spec and assert it is rejected under the virtualenv scope).

xinference/model/embedding/model_spec.json
```diff
@@ -1835,6 +1835,7 @@
         "qwen-vl-utils>=0.0.14",
         "Pillow",
         "#vllm_dependencies# ; #engine# == \"vllm\"",
+        "vllm>=0.14.0 ; #engine# == \"vllm\"",
```
qinxuye, line 1837: Appending this beside `#vllm_dependencies#` does not produce a safe intersection in the actual install path. Placeholder expansion and marker filtering leave both `vllm>=0.11.2` and `vllm>=0.14.0`. With `skip_installed` and a parent vLLM 0.11.2–0.13.x, xoscar treats the first requirement as satisfied and pins that installed version, then tries to resolve the second requirement, resulting in a constraint such as `vllm==0.13.0` together with `vllm>=0.14.0` and resolution failure. Please ensure only one vLLM requirement reaches the installer, e.g. replace the generic `#vllm_dependencies#` entry for these specs with `vllm>=0.14.0`, and cover the prepared install list (not only engine discovery). The reranker specs have the same issue.

qinxuye (commented):
The virtualenv install path still has a conflicting vLLM constraint.

xinference/model/embedding/model_spec.json
OliverBryant: Fixed — you are right that appending produced two conflicting vLLM requirements in the prepared install list. Replaced the generic `#vllm_dependencies#` entry with `vllm>=0.14.0` for the Qwen3-VL embedding (2B/8B) and reranker (2B/8B) specs, so only a single vLLM requirement reaches the installer instead of both `vllm>=0.11.2` and `vllm>=0.14.0`.

Verified on the actual prepared install path (placeholder expansion + marker filtering, the same functions the Worker uses): the vLLM requirement list is now exactly `["vllm>=0.14.0"]`. Added regression tests `test_qwen3_vl_embedding_prepared_install_list_has_single_vllm_req` and the reranker equivalent that assert this over the prepared install list, not just discovery.

xinference/model/llm/lmdeploy/core.py
```diff
@@ -195,7 +215,7 @@ def match_json(
                 False,
                 f"Model architectures {llm_family.architectures} are not in LMDeploy supported chat list",
             )
-        if not LMDEPLOY_INSTALLED:
+        if not LMDEPLOY_INSTALLED and not _virtual_env_allows_missing_lmdeploy():
```
qinxuye: This exemption is not connected to a working virtualenv install path. `LMDEPLOY` is absent from both `ENGINE_VIRTUALENV_PACKAGES` and the model virtualenv engine markers, so `get_engine_params_by_name_with_virtual_env("LLM", "InternVL3", enable_virtual_env=True)` still reports `Engine LMDEPLOY is not listed in model virtualenv packages`, and launch cannot install it on demand. LMDeploy also needs the same CUDA/Linux gating discussed above before it can safely be exposed. Please either remove this exemption from this PR, or add the package/marker/platform handling and a regression test covering the actual discovery and install path.

OliverBryant: You are right — the LMDeploy exemption was dead code: LMDEPLOY is not in ENGINE_VIRTUALENV_PACKAGES and no model declares it as a virtualenv marker, so it could never be installed on demand, and it also lacked the CUDA/Linux gating. Rather than add full LMDeploy virtualenv support (packages + markers + platform gating), which is out of scope for this PR, I removed the LMDeploy change entirely. `lmdeploy/core.py` is now identical to main, and the PR is focused on the vLLM/SGLang paths for Qwen3.5 / Qwen3-VL that actually have a working virtualenv install path. Updated the LLM commit message accordingly.

qinxuye (approved):
LGTM

PR #5181, opened 2026-07-16
BUG: sglang venv inherits incompatible parent transformers under skip_installed

OliverBryant:
## What do these changes do?

Fixes sglang engine launches failing inside `xprobe/xinference:nightly-main` (and any environment whose parent site-packages ship transformers 5.x next to sglang 0.5.x).

### Root cause

- The Docker image installs `transformers==5.5.0` and `--no-deps sglang==0.5.6` into the same environment. All sglang 0.5.x releases (0.5.6 through 0.5.9) declare `transformers==4.57.1`.
- Under transformers 5.x, importing `sglang.srt.entrypoints.engine` crashes while defining `DeepseekVL2Config(PretrainedConfig)`: transformers 5's `PretrainedConfig.__init_subclass__` wraps subclasses with `dataclass(...)`, which raises `TypeError: non-default argument 'vision_config' follows default argument` (sglang `srt/configs/deepseekvl2.py`).
- With `XINFERENCE_VIRTUAL_ENV_SKIP_INSTALLED=1` (the default), the parent's sglang 0.5.6 satisfies the `sglang>=0.5.6` spec in `ENGINE_VIRTUALENV_PACKAGES`, so per-model venvs skip installing sglang entirely — and nothing pulls a compatible transformers into the venv. Every sglang model launch then fails at engine import (reproduced with qwen3 0.6B, `--model-engine sglang`, 2×RTX 3090 Ti, CUDA 13).

### Fix

Rather than hard-pinning a transformers version (which would go stale as sglang evolves), derive the required spec from the inherited engine's own dependency metadata at launch time:

- New `ENGINE_CRITICAL_DEPENDENCIES` map and `get_engine_critical_dependency_specs()` in `virtual_env_manager.py`: when the parent env's engine copy will be inherited by the venv (its version satisfies the requested spec) and the parent copy of a critical dependency (transformers, for sglang) violates the engine's declared requirement (`Requires-Dist`), the engine's declared spec is added to the venv install list. The venv-installed copy shadows the parent's because child site-packages precede the parent `.pth` entry.
- Wired into `WorkerActor._prepare_virtual_env` after marker filtering.

The check is a no-op when:
- the parent environment is consistent (dependency satisfies the engine's requirement),
- the engine is absent from the parent (the resolver then installs the engine with its full dependency closure),
- the requested spec forces a fresh engine install into the venv,
- the user supplies an explicit spec for the dependency (user specs win).

Because the spec is read from the installed engine's metadata, it automatically follows whatever the installed sglang version declares — no maintenance needed when sglang bumps its transformers requirement.

### Verification

- Reproduced the incompatibility in an isolated venv with sglang 0.5.6: under `transformers==5.5.0`, importing `sglang/srt/configs/deepseekvl2.py` raises the exact `TypeError` seen in the nightly image; under `transformers==4.57.1` the same import succeeds.
- Added `TestGetEngineCriticalDependencySpecs` (10 cases) to `xinference/core/tests/test_virtual_env_manager.py`; the full file passes (30 passed).

Not run: an end-to-end sglang model launch on GPU (requires the CUDA Docker environment).

### DIFF
--- a/xinference/core/tests/test_virtual_env_manager.py
+++ b/xinference/core/tests/test_virtual_env_manager.py
@@ -17,6 +17,7 @@
 See optimize/20260702/2026070209.md for root cause analysis.
 """
 
+import importlib.metadata
 import os
 from unittest import mock
 
@@ -27,6 +28,7 @@ from ..virtual_env_manager import (
     FLASHINFER_AOT_PACKAGES,
     FLASHINFER_AOT_WHEEL_URL,
     apply_flashinfer_aot_post_install,
+    get_engine_critical_dependency_specs,
     needs_flashinfer_aot,
 )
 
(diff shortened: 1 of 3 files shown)


qinxuye (approved):
LGTM

PR #5183, opened 2026-07-16
fix(audio): add WeTextProcessing virtualenv dependency for MegaTTS3

m199369309:
MegaTTS3 requires WeTextProcessing for Chinese/English text normalization during speech synthesis. This was verified locally — without this dependency, the model fails at inference time when processing input text.

This change adds the dependency as a `virtualenv` block in `model_spec.json`, matching the pattern used by other audio models (CosyVoice2, Qwen3-TTS, VoxCPM2, etc.).


m199369309:
Closing this PR as requested.

PR #5185, opened 2026-07-16
feat(llama.cpp): install GPU xllamacpp wheel from CUDA-matched index

OliverBryant:
## Problem

When the `llama.cpp` engine creates its virtual environment, `xllamacpp` is installed from the default PyPI index (or an inherited PyPI mirror), which only publishes the **CPU** build. On GPU hosts this silently yields a CPU-only runtime, and importing the CPU wheel in a GPU process can crash the model subprocess. Users otherwise have to know to manually reinstall from the GPU wheel index:

```
pip install xllamacpp --force-reinstall --index-url https://xorbitsai.github.io/xllamacpp/whl/cu132
```

## Change

During `_prepare_virtual_env`, for the `llama.cpp` engine, detect the system CUDA version (via `xoscar.virtualenv.platform.get_cuda_version()`) and, when a compatible prebuilt GPU wheel exists, install `xllamacpp` from the matching GPU index:

- CUDA `13.x` (and newer) → `https://xorbitsai.github.io/xllamacpp/whl/cu132`
- CUDA `12.8+` → `https://xorbitsai.github.io/xllamacpp/whl/cu128`

The GPU index becomes the primary `index_url` so it wins over any inherited index (e.g. a Tencent/Tsinghua PyPI mirror pulled in via `inherit_pip_config`). The previously-inherited `index_url` is demoted to `extra_index_url`, so it stays available as a resolution fallback and keeps the resolver responsive when the GPU index is slow to reach.

Hosts without a compatible GPU (no CUDA detected, or an older CUDA line with no prebuilt wheel) keep their inherited index and install the CPU build, as before.

## Notes

- Only `cu128` and `cu132` GPU wheel indexes are published by xllamacpp today; the mapping is confined to those.
- `xllamacpp` GPU wheels are self-contained `abi3` wheels with no required runtime dependencies.
- New unit tests cover the CUDA-version-to-index mapping and the `_prepare_virtual_env` behavior (GPU host → GPU index primary with the inherited mirror demoted to extra-index; CPU host → inherited mirror kept).

## Testing

- `black`, `isort`, `flake8` clean on the changed files.
- `pytest xinference/core/tests/test_utils.py` — 7 passed.
- Verified end-to-end on a real GPU host (RTX 3090 Ti, torch cu128) that the engine now installs `xllamacpp` from the `cu128` GPU index as the primary index with the inherited mirror demoted to extra-index (`index_url=.../whl/cu128, extra_index_url=[<mirror>]`), and that the launch no longer errors on resolution.
- Not yet verified: actual GPU-offloaded inference on that host, because it currently cannot reach `github.com` (where the GPU wheels are hosted), so the resolver falls back to the mirror's CPU wheel. The index-selection logic is confirmed; GPU-runtime confirmation is pending a host with network access to the wheel index.


xinference/core/worker.py
```diff
@@ -2593,6 +2593,39 @@ def _prepare_virtual_env(
...
+
+            xllamacpp_index_url = get_xllamacpp_cuda_index_url(cuda_version)
+            if xllamacpp_index_url:
+                logger.info(
+                    "Detected CUDA %s, installing GPU build of xllamacpp "
+                    "exclusively from %s",
+                    cuda_version,
+                    xllamacpp_index_url,
+                )
+                settings.index_url = xllamacpp_index_url
```
qinxuye, line 2683: This does not guarantee that the GPU wheel is installed when the parent environment already has a CPU build of `xllamacpp` satisfying `xllamacpp>=0.2.6`. `XINFERENCE_VIRTUAL_ENV_SKIP_INSTALLED` defaults to true, and xoscar filters such a requirement out before invoking `uv`, so this index URL is never used. The xllamacpp installation instructions explicitly require `--force-reinstall` when replacing the CPU wheel. Please force the GPU-wheel installation (or otherwise bypass the satisfied-package filter for `xllamacpp`) and add a test that exercises the real skip-installed path; the current fake VEM only verifies that the index setting is passed through.

OliverBryant: Good catch — this was exactly the failure I hit in practice. Fixed in the latest revision.

For the `llama.cpp` engine on a CUDA host, `_prepare_virtual_env` now:
1. uninstalls any existing `xllamacpp` from the venv first (`_uninstall_venv_package`), and
2. forces this install (`skip_installed=False` for this call only),

so the GPU wheel actually replaces a same-version CPU build instead of being filtered out as already-satisfied.

Verified end to end on a real GPU host (RTX 3090 Ti, 2×), reproducing the bug and confirming the fix:
- Install CPU wheel from a PyPI mirror → `get_device_info()` = `['CPU']`, 0 CUDA symbols in the `.so`.
- With skip-installed on, re-running the install against the GPU index is a no-op: `Audited 1 package ... Would make no changes` (the bug).
- After the fix (uninstall + forced install from the GPU index) → `get_device_info()` = `['CUDA0', 'CUDA1', 'CPU']`, 4552 CUDA symbols in the `.so`.

Added `test_prepare_virtual_env_llama_cpp_gpu_force_reinstalls_over_skip_installed`, which drives the real skip-installed path and asserts that `xllamacpp` is uninstalled first and `skip_installed=False` is passed to `install_packages` (rather than only checking the index setting is forwarded).

xinference/core/worker.py
```diff
@@ -2593,6 +2623,48 @@ def _prepare_virtual_env(
...
+        # user then gets a CPU runtime while believing the GPU build was
+        # installed. Restricting to the GPU index guarantees the GPU wheel (or a
+        # visible failure). xllamacpp GPU wheels are self-contained abi3 wheels
+        # with no required runtime dependencies, so an exclusive index does not
+        # break resolution.
+        force_reinstall_xllamacpp = False
+        if model_engine and model_engine.lower() == "llama.cpp":
+            from .virtual_env_manager import get_xllamacpp_cuda_index_url
+
+            xllamacpp_index_url = get_xllamacpp_cuda_index_url(cuda_version)
```
qinxuye, line 2670: This currently breaks all Ubuntu test jobs. `get_cuda_version()` returns `torch.version.cuda` (13.0 on those runners) even when no GPU/driver is available, so the code installs the cu132 wheel and model import then fails with `ImportError: libcuda.so.1: cannot open shared object file`. The CUDA version describes the PyTorch build, not actual device availability. Please gate this path with `xoscar.virtualenv.platform.check_cuda_available()` (or `torch.cuda.is_available()`), keep the CPU index/no-uninstall behavior when it is false, and add a regression test for `cuda_version="13.0"` with CUDA availability false.

OliverBryant: Fixed. The GPU path is now gated on real device availability, not just the PyTorch-build CUDA version.

`_prepare_virtual_env` now checks `WorkerActor._is_cuda_device_available()`, which wraps `xoscar.virtualenv.platform.check_cuda_available()` (falling back to `torch.cuda.is_available()`). The GPU index / uninstall / force-reinstall path only runs when it returns true; otherwise the inherited CPU index and default behavior are kept, so the CPU-only runners install the CPU wheel as before.

Confirmed on the host that this distinguishes the two: with the GPU visible, `get_cuda_version()=12.8` and `check_cuda_available()=True`; with `CUDA_VISIBLE_DEVICES=` set (simulating a CPU-only host with a CUDA-built torch), `get_cuda_version()` still reports `12.8` but `check_cuda_available()=False` — exactly the runner situation you described.

Added `test_prepare_virtual_env_llama_cpp_no_gpu_device_keeps_cpu_index` (`cuda_version="13.0"`, CUDA availability false) asserting the inherited index is kept and no uninstall/force-reinstall happens.

qinxuye (approved):
LGTM

PR #5186, opened 2026-07-16
FEAT(webui): show per-model GPU memory in Running Models

OliverBryant:
## Motivation

The backend already collects **per-process GPU memory** via `nvmlDeviceGetComputeRunningProcesses` and attributes it to each model replica (see `get_per_process_gpu_memory` in `device_utils.py` and the attribution logic in `WorkerActor.report_status`). However, this data was only exposed through the Prometheus metric `model_gpu_memory_used_bytes` (for Grafana) and never surfaced in the REST API, so the Web UI had no way to display it. The Running Models page only showed GPU indexes, not actual memory usage.

## Changes

**Backend** (`xinference/core/supervisor.py`)
- In `SupervisorActor.list_models()`, aggregate `_worker_model_gpu_memory` (keyed by `worker_address -> replica_model_uid -> {gpu_idx -> bytes}`) into base-model granularity, summing across all workers and replicas of the same base model.
- Attach the result as a `gpu_memory` field (`{gpu_idx: bytes}`) on each model entry. Since the REST layer spreads the full spec (`**model_info`), it flows through `/v1/models` automatically without any API-layer change.
- Grouping reuses the existing `parse_replica_model_uid` logic already used for the `running_model_info` keys, so base uids line up exactly. A malformed replica uid is skipped (logged) instead of breaking the whole listing.

**Frontend** (`frontend/`, Next.js UI)
- Add a **GPU Memory** entry next to GPU Indexes in the Resources section of the Running Models detail view.
- Single GPU shows e.g. `3.20 GiB`; multi-GPU shows one line per device (`GPU 0: … / GPU 1: …`); auto GiB/MiB; `-` when no data.
- i18n added for en / zh / ja / ko.

## Notes

- Per-process collection is **NVIDIA-only** (via pynvml), consistent with the existing backend behavior. On other accelerators (NPU / ROCm, etc.) it shows `-`.
- No breaking changes: the new field is additive and the REST schema is otherwise unchanged.

## Verification

- `black` on the Python file; project-pinned `prettier@3.8.3` reports the frontend files unchanged (already conformant). `git diff --stat` confirms additions only, no incidental reformatting.
- Full pytest, tsc type-check, and on-device GPU verification were not run in this environment (no runtime deps / no NVIDIA GPU). Suggested manual check: launch a model on a GPU host, confirm `/v1/models` returns `gpu_memory`, and verify it renders on the Running Models page.

### DIFF
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -3195,6 +3195,35 @@ class SupervisorActor(xo.StatelessActor):
         self._replica_gpu_cache = replica_gpu_cache
 
         running_model_info = {parse_replica_model_uid(k)[0]: v for k, v in ret.items()}
+
+        # Aggregate per-process GPU memory (real-time, from
+        # nvmlDeviceGetComputeRunningProcesses) into base-model granularity so
+        # the REST /v1/models response can surface it. _worker_model_gpu_memory
+        # is keyed by worker_address -> replica_model_uid -> {gpu_idx -> bytes}.
+        # GPU indices are local to each worker, so the worker dimension must be
+        # preserved: aggregating by gpu_idx alone would merge same-numbered GPUs
+        # on different physical hosts. Result shape:
+        # {base_uid: {worker_address: {gpu_idx: bytes}}}, summed across replicas
+        # that live on the same worker.
+        base_gpu_memory: Dict[str, Dict[str, Dict[int, int]]] = defaultdict(
+            lambda: defaultdict(lambda: defaultdict(int))
+        )
+        for worker_address, per_model in self._worker_model_gpu_memory.items():
+            for replica_uid, per_gpu in per_model.items():
+                try:
+                    base_uid = parse_replica_model_uid(replica_uid)[0]
+                except Exception:
+                    # A malformed replica uid must not break model listing;
+                    # skip this entry instead of propagating the error.
+                    logger.warning(
+                        "list_models: skip malformed replica uid %s in "
+                        "per-model GPU memory",
+                        replica_uid,
+                    )
+                    continue
+                for gpu_idx, mem in per_gpu.items():
+                    base_gpu_memory[base_uid][worker_address][int(gpu_idx)] += int(mem)
+
         # add replica count
         stale_uids = []
         for k, v in running_model_info.items():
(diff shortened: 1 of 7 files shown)


xinference/core/supervisor.py
```diff
@@ -3170,6 +3170,21 @@ async def _fetch_one(
...
+        # is keyed by worker_address -> replica_model_uid -> {gpu_idx -> bytes};
+        # sum across all workers and replicas of the same base model.
+        base_gpu_memory: Dict[str, Dict[int, int]] = defaultdict(
+            lambda: defaultdict(int)
+        )
+        for per_model in self._worker_model_gpu_memory.values():
+            for replica_uid, per_gpu in per_model.items():
+                base_uid = parse_replica_model_uid(replica_uid)[0]
+                for gpu_idx, mem in per_gpu.items():
+                    base_gpu_memory[base_uid][int(gpu_idx)] += int(mem)
```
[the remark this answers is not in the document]
OliverBryant: Good catch — confirmed. `parse_replica_model_uid` does `int(parts.pop())`, so a replica uid whose last segment is non-numeric (e.g. a custom `-rank0` suffix) would raise `ValueError` and crash this critical listing path. Fixed in the latest push: the call is now wrapped in a try/except that skips (and logs a warning for) the malformed uid instead of propagating. I used a broad `except Exception` plus a warning log so any future parsing change stays contained.

xinference/core/supervisor.py
```diff
@@ -3195,6 +3195,31 @@ async def _fetch_one(
...
+                    # A malformed replica uid must not break model listing;
+                    # skip this entry instead of propagating the error.
+                    logger.warning(
+                        "list_models: skip malformed replica uid %s in "
+                        "per-model GPU memory",
+                        replica_uid,
+                    )
+                    continue
+                for gpu_idx, mem in per_gpu.items():
+                    base_gpu_memory[base_uid][int(gpu_idx)] += int(mem)
```
qinxuye: GPU indices are local to each worker, so aggregating only by `gpu_idx` merges different physical devices when replicas run on multiple workers. For example, `worker-a/demo-0/GPU 0 = 2 GiB` plus `worker-b/demo-1/GPU 0 = 3 GiB` is returned as a single `GPU 0 = 5 GiB` (and the collapsed model entry may show only one worker address). Please preserve the worker dimension, e.g. `{worker_address: {gpu_idx: bytes}}`; alternatively, if the intended value is model-wide memory, expose one total instead of presenting it as per-GPU usage. Please add a multi-worker regression test for this case.

qinxuye (commented):
One distributed aggregation issue remains.

xinference/core/supervisor.py
OliverBryant: You're right, thanks. GPU indices are worker-local, so summing by `gpu_idx` alone would merge same-numbered GPUs on different hosts. Fixed in the latest push: the worker dimension is now preserved. The field shape is `{worker_address: {gpu_idx: bytes}}`, summed only across replicas that live on the same worker. The Running Models page renders per-worker rows and prefixes each with the worker address when a model spans multiple workers.

qinxuye (approved):
LGTM

PR #5188, opened 2026-07-17
FEAT: slim GPU Docker image without pre-installed inference engines

OliverBryant:
## What this PR does

Rewrites the GPU image (`xinference/deploy/docker/Dockerfile`) from a
`vllm/vllm-openai:v0.21.0`-based build that pre-installs every engine and model
dependency, to a slim image based on `nvidia/cuda:13.0.2-devel-ubuntu22.04`
(the same base / system layout the vLLM image itself is built on: deadsnakes
Python 3.12, apt ffmpeg).

| | before (`xprobe/xinference:latest`) | after (this PR) |
|---|---|---|
| Base image | `vllm/vllm-openai:v0.21.0` | `nvidia/cuda:13.0.2-devel-ubuntu22.04` |
| Image size (uncompressed) | **35.2 GB** | **13.9 GB** (-60%) |
| Engines in image | vllm, sglang, transformers, llama.cpp + full audio/image/video stacks | transformers only |

The image now ships only:

* Python 3.12 (deadsnakes, same as the vLLM base) + pip/setuptools/wheel
* a shared CUDA torch stack: `torch==2.11.*+cu130`, torchvision, torchaudio,
  `torchcodec==0.14.0` (all from the cu130 index)
* `transformers==5.13.1` + `accelerate==1.14.0` (the default Transformers engine)
* xinference with its core dependencies (`.[otel]`) and the prebuilt Web UI
  (built in a separate `node:20-slim` stage; the final image carries no Node runtime)
* runtime system tools: git (VCS installs in venvs), ffmpeg/libsndfile1 (audio
  decode), libgl1/libglib2.0-0 (opencv), curl, procps, build-essential +
  python3.12-dev (sdist compile fallback)

All pins are exposed as build args (`TORCH_VERSION`, `TORCHCODEC_VERSION`,
`TRANSFORMERS_VERSION`, `ACCELERATE_VERSION`, `PIP_INDEX`, `TORCH_INDEX`).

## Why engines can be dropped from the image

`XINFERENCE_ENABLE_VIRTUAL_ENV` defaults on: at launch time the worker creates a
per-model virtual environment (`worker._prepare_virtual_env` → xoscar's uv
manager) and installs exactly the engine packages the model spec declares
(`ENGINE_VIRTUALENV_PACKAGES` / `#<engine>_dependencies#` placeholders). Online
deployments install from PyPI; offline deployments install from the private
mirror wired up by the `offline` compose profile (#5174).

The venvs are created with `--system-site-packages`, and
`XINFERENCE_VIRTUAL_ENV_SKIP_INSTALLED` (default on) skips anything the parent
environment already satisfies — so what the image *does* bake in stays shared
across all model venvs.

`xinference/model/llm/vllm/` only imports vllm lazily (TYPE_CHECKING guards and
function-local imports; the `distributed_executor*` modules with top-level vllm
imports are themselves imported lazily), so an engine-less parent environment is
safe — the CPU image has always run without vllm.

## Design decisions

**torch stays in the image.**

1. Model specs reference the parent environment's torch family via
   `#system_torch#` / `#system_torchvision#` / `#system_torchaudio#` /
   `#system_torchcodec#` placeholders (92 / 56 / 18 / 1 occurrences across the
   built-in spec JSONs). xoscar's `resolve_system_requirement` raises
   `RuntimeError` when the referenced package is missing, so a torch-less image
   would break every one of those specs at launch.
2. `torch` is a core `install_requires` of xinference itself — a torch-less
   image is not achievable with a plain `pip install .` anyway.
3. With skip-installed, one shared ~5 GB CUDA torch is reused by every venv
   whose resolution lands on the same version, instead of being downloaded per
   model.

The `+cu130` build is installed explicitly from `download.pytorch.org/whl/cu130`:
the worker parses the `+cu130` local suffix of the installed torch to
auto-configure the matching PyTorch wheel index for venv installs, and
`2.11.*` matches the torch-family lock of the prebuilt pypiserver mirror
image (#5048).

**Baked packages are pinned, not floating.** Found the hard way during offline
validation, twice: (1) with skip-installed, the parent environment's installed
versions become uv resolution *constraints* for venv installs — a floating
`transformers>=4.53.3` resolved to a release the offline mirror did not carry
and every sentence_transformers-engine launch failed with `No solution found`;
(2) the `#system_numpy#`/`#system_pandas#` spec placeholders pin the parent
versions into venv install requests outright — a floating numpy landed on a
version absent from the mirror and every vllm-engine launch failed the same
way. Every version baked into the image must therefore also exist on the
offline mirror — hence explicit pins (`TRANSFORMERS_VERSION`,
`ACCELERATE_VERSION`, `NUMPY_VERSION`, `PANDAS_VERSION`), aligned with the
mirror's locks, with a comment in the Dockerfile explaining the invariant.

**torchcodec pinned to 0.14.0 (was 0.10.0).** The previous image's
`torchcodec==0.10.0` pairs with torch 2.10 per the
[upstream compatibility table](https://github.com/pytorch/torchcodec#installing-torchcodec)
and cannot load against torch 2.11 (`undefined symbol` in every
`libtorchcodec_core*.so`) — latent breakage that surfaced because newer
sentence-transformers imports torchcodec unconditionally. 0.14.0 supports
torch >= 2.11, ships an FFmpeg-4 core that works with Ubuntu 22.04's apt
ffmpeg, and is carried by the #5048 mirror. Installed from the cu130 index for
CUDA decode support.

**transformers/accelerate stay in the image.** They satisfy
`ENGINE_VIRTUALENV_PACKAGES["transformers"]`, so the default engine works out
of the box and its venv install is a no-op (measured: the qwen3 venv is 128 KB).
It also keeps `XINFERENCE_ENABLE_VIRTUAL_ENV=0` usable as a minimal offline
escape hatch (docs updated: it now only covers the Transformers engine).

**devel base, not runtime.** Kept `-devel` so nvcc + CUDA headers (plus
build-essential/python3.12-dev) remain available as a fallback for venv
packages that only ship sdists and compile CUDA extensions on the fly. This
costs ~3-4 GB over `-runtime`; if we later verify all supported model venvs
resolve to binary wheels (the #5048 mirror already builds sdists into wheels
at mirror build time), switching to `-runtime` is a one-line follow-up.

**pandas added.** Seven audio specs reference `#system_pandas#`, which
hard-fails when pandas is absent from the parent environment.

**The final layer installs with uv.** uv is a runtime dependency anyway (it
powers the per-model venv installer); its parallel downloads with per-request
timeouts also make image builds robust against throttled index connections
that stall a single serial pip session.

**ninja-build added.** vLLM's engine startup shells out to `ninja` for
runtime JIT compilation and dies with `FileNotFoundError: 'ninja'` without it
(the vllm-openai base image shipped it; a slim base must provide it
explicitly). numpy is additionally kept below 2.4 because every vllm release
on the mirror requires `numpy<2.4` via mistral-common, and `#system_numpy#`
pins the image's numpy into the vllm venv request.

## Relation to #5048 and CUDA support

* This slim image + the prebuilt `xprobe/xinference-pypiserver` mirror image
  (#5048) form the complete offline story: the slim image provides the
  Python 3.12 + torch 2.11+cu130 parent environment, the mirror provides every
  wheel the per-model venv installer may request. The pin-alignment invariant
  above is the contract between the two.
* Consistent with #5048, CUDA below 13.0 is not supported by this image
  (host driver >= 580). Docs updated accordingly.

## Other changes

* The GPU `requirements/` files (`requirements-base/ml/models.txt`) are
  removed — the rewritten Dockerfile was their only consumer.
* `.dockerignore` excludes `frontend/` build artifacts so a local
  `node_modules` cannot leak into the image via `COPY . /opt/inference`.
* Docs: `using_docker_image.rst` (new base, CUDA >= 13.0, "what is inside the
  GPU image"), `using_docker_compose.rst` + `offline.env.example`
  (`XINFERENCE_ENABLE_VIRTUAL_ENV=0` scope).
* `.github/workflows/docker-cd.yaml` needs no changes: the workflow passes no
  build args and `PIP_INDEX` is kept. The unused `LLAMA_CPP_USE_CUDA` arg is
  gone.

## Verification

On a 2×RTX 3090 Ti host (driver 580.x, CUDA 13.0, Docker 29.6.1):

* **Build**: `docker build -f xinference/deploy/docker/Dockerfile .` → 13.9 GB
  (vs 35.2 GB for `xprobe/xinference:latest` on the same host). Web UI served
  by the backend, `torchcodec`/`torch`/`transformers` import cleanly,
  `/status` healthy in ~20 s.
* **Online launch (Transformers engine)**: `xinference launch -n qwen3 -s 0_6
  -f pytorch --model-engine transformers` → venv created with
  `['transformers>=4.53.3', 'accelerate>=0.28.0']`, skip-installed satisfies
  everything from the parent env (venv weighs 128 KB), chat completion OK.
* **Offline launch (private mirror)**: containers wired per the `offline`
  compose profile documentation (pip.conf + `UV_*` env) against a mirror image
  built from the #5048 pipeline. Worker log shows
  `settings(index_url=http://xinference-pypiserver:8080/simple, ...,
  skip_installed=True)`; `bge-small-zh-v1.5` (sentence_transformers engine)
  resolves and installs its venv from the mirror alone and serves
  `/v1/embeddings` correctly.
* **Offline vLLM launch (private mirror)**: qwen3 0.6B with `--model-engine
  vllm` — the venv resolves and installs vllm 0.25.1 (+ transitive deps,
  ~4.4 GB venv, shared torch reused from the image) entirely from the mirror,
  the engine JIT-compiles through the image's nvcc + ninja, and
  `/v1/chat/completions` returns correct responses.
* Not covered: `Dockerfile.cpu` / `Dockerfile.aarch64` (untouched), zh_CN doc
  locale regeneration, and vllm/sglang launches (their venv install chain is
  identical to the sentence_transformers path exercised above, just larger
  downloads for sglang; the engines are no longer part of the image
  contract).


OliverBryant:
Pushed a follow-up: the final image layer is now installed with `uv pip install --system` instead of serial pip. uv is already a runtime dependency (it powers the per-model venv installer), and its parallel downloads with per-request timeouts recover from throttled index connections that stall a single serial pip session. No dependency or content changes.

xinference/deploy/docker/Dockerfile
```diff
@@ -1,81 +1,111 @@
...
+# significant: the worker reads it to auto-configure the matching PyTorch
+# wheel index for per-model venv installs. pandas backs the #system_pandas#
+# placeholder used by several audio model specs.
+RUN pip install --no-cache-dir -i "$PIP_INDEX" --upgrade pip "setuptools<82" wheel && \
+    pip install --no-cache-dir "torch==${TORCH_VERSION}" torchvision torchaudio \
+      "torchcodec==${TORCHCODEC_VERSION}" --index-url "$TORCH_INDEX" && \
+    pip install --no-cache-dir -i "$PIP_INDEX" pandas
+
+COPY . /opt/inference
+COPY --from=web-builder /opt/inference/xinference/ui/web/dist /opt/inference/xinference/ui/web/dist
```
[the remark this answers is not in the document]
OliverBryant: The path is correct as written: the repository convention is `xinference/ui/web/dist` throughout — `MANIFEST.in` (`graft xinference/ui/web/dist`), `setup.py` (`_web_dest_path = "xinference/ui/web/dist/index.html"`), and `frontend/scripts/stage-export.mjs` all use it, and the Python backend serves the UI from that location. Verified at runtime: the Web UI loads from a container built from this Dockerfile.

xinference/deploy/docker/Dockerfile
```diff
@@ -1,81 +1,111 @@
...
+# ---------------------------------------------------------------------------
+FROM node:20-slim AS web-builder
 
-ENV NVM_DIR=/usr/local/nvm
-ENV NODE_VERSION=20.19.0
+WORKDIR /opt/inference/frontend
+COPY frontend ./
+RUN npm ci && npm run build
+# `npm run build` stages the static export at ../xinference/ui/web/dist via
+# its postbuild hook (frontend/scripts/stage-export.mjs).
```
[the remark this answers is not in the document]
OliverBryant: Same as above — `xinference/ui/web/dist` matches `frontend/scripts/stage-export.mjs`, which stages the export to `../xinference/ui/web/dist`. The comment is accurate as written.

xinference/deploy/docker/Dockerfile
```diff
@@ -1,81 +1,111 @@
...
-    # Clear the Conda cache
-    /opt/conda/bin/conda clean --all -y
+# Shared CUDA torch stack (see header). The +cu130 local version is
+# significant: the worker reads it to auto-configure the matching PyTorch
+# wheel index for per-model venv installs. pandas backs the #system_pandas#
+# placeholder used by several audio model specs.
+RUN pip install --no-cache-dir -i "$PIP_INDEX" --upgrade pip "setuptools<82" wheel && \
+    pip install --no-cache-dir "torch==${TORCH_VERSION}" torchvision torchaudio \
+      "torchcodec==${TORCHCODEC_VERSION}" --index-url "$TORCH_INDEX" && \
+    pip install --no-cache-dir -i "$PIP_INDEX" pandas
```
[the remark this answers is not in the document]
OliverBryant: Partially addressed: pandas was already pinned via `PANDAS_VERSION`. torchvision/torchaudio are now pinned too (`TORCHVISION_VERSION=0.26.0`, `TORCHAUDIO_VERSION=2.11.0` — the exact versions the previous floating resolution produced against torch 2.11.*+cu130, so no behavior change from the validated image).

xinference/deploy/docker/Dockerfile
```diff
@@ -1,81 +1,129 @@
...
+# is packaged via MANIFEST.in; skip the npm build hook of setup.py.
+# transformers/accelerate satisfy ENGINE_VIRTUALENV_PACKAGES["transformers"]
+# so the default engine's venv install is satisfied by the parent env.
+# uv (a runtime dependency anyway, it powers the per-model venv installer)
+# also does the image install: parallel downloads with per-request timeouts
+# recover from throttled/stalled index connections that serial pip rides
+# to a crawl.
+RUN pip install --no-cache-dir -i "$PIP_INDEX" uv && \
+    NO_WEB_UI=1 uv pip install --system --no-cache --index-url "$PIP_INDEX" \
+      ".[otel]" "transformers==${TRANSFORMERS_VERSION}" "accelerate==${ACCELERATE_VERSION}"
```
qinxuye: [P1] Preserve the dependencies required by supported Transformers models

This installs only the core/OTel dependencies plus `transformers` and `accelerate`, but the existing per-model virtualenv manifests are not a complete replacement for the deleted image requirements. For example, `qwen2-vl-instruct` expands only `#transformers_dependencies#`, while `qwen2_vl.py` imports `qwen_vl_utils` when processing a request; that package was supplied by the old image and is only listed in the `transformers` extra, so the new official GPU image will raise `ModuleNotFoundError` for a model it still advertises as supported. DeepSeek-VL2 has the same problem with `attrdict`/`einops`. Please either keep the relevant extras in the parent image (for example `.[otel,transformers]`, plus the supported quantization dependencies) or complete and validate the per-model virtualenv package declarations before removing the shared packages.

xinference/deploy/docker/Dockerfile
```diff
@@ -1,81 +1,129 @@
...
+# https://github.com/pytorch/torchcodec#installing-torchcodec (0.12+ ~ torch
+# >=2.11). Its FFmpeg 4-8 runtime libraries are provided by apt ffmpeg below.
+ARG TORCHCODEC_VERSION=0.14.0
+ARG TORCH_INDEX=https://download.pytorch.org/whl/cu130
+# Pinned (not floating) on purpose: with skip_installed, per-model venv
+# resolution pins the parent environment's installed versions as constraints,
+# and the #system_numpy#/#system_pandas# spec placeholders pin the parent
+# versions into venv install requests outright — so every version baked here
+# must also exist on the offline mirror (xprobe/xinference-pypiserver). Keep
+# these aligned with the mirror's locks when bumping either side.
```
qinxuye: [P1] Enforce the runtime/mirror version contract from one source

These exact runtime pins are not propagated to the pypiserver build: `Dockerfile.pypiserver` passes only the floating `TORCH_VERSION=2.11.*`, and `generate_package_lists.py` converts the other `#system_*#` placeholders to unpinned package names. Once a newer torchvision/torchcodec/transformers/numpy/pandas release appears, a same-tag mirror rebuild can omit the exact versions baked into this image, and the parent-version constraints used by `skip_installed` will make offline venv resolution fail. Please feed the runtime pins into the mirror lock/self-check (or move both images to a shared constraints file) rather than relying on manual alignment.

xinference/core/virtual_env_manager.py
```diff
@@ -43,6 +43,32 @@
     "transformers": [
         "transformers>=4.53.3",
         "accelerate>=0.28.0",
+        # Model-side dependencies formerly baked into the GPU image. Keep
```
qinxuye: [P1] Keep optional model packages out of the global Transformers engine defaults

`ENGINE_VIRTUALENV_PACKAGES` is used for every Transformers model on every platform, not just by this Linux Docker image. Moving the former image-wide extras here now makes even a plain text model resolve the full quantization/multimodal stack (`autoawq`, `gptqmodel`, `datasets`, `eva-decord`, etc.). I reproduced the generated `transformers.in` against the supported target platforms: `uv pip compile --python-platform x86_64-apple-darwin` is unsatisfiable because `eva-decord` has no matching macOS wheel, and the Windows target fails while building `autoawq` metadata. On Linux it resolves 141 packages, so the previously 128 KB Qwen3 venv now downloads and duplicates a large unrelated stack. Please keep the engine defaults minimal and declare Qwen2-VL, DeepSeek-VL2, and quantization dependencies in the relevant model specs with engine/format/platform markers, or preserve those compatibility extras only in the Docker parent environment; do not change the cross-platform defaults for every Transformers launch.

xinference/core/worker.py
```diff
@@ -3261,6 +3265,7 @@ def check_cancel():
                                     "_resolve_architectures",
                                     lambda: None,
                                 )(),
+                                model_format=model_format,
```
qinxuye: [P1] Use the resolved model format for dependency selection. This passes the original request value, which may be null even though match_llm has selected a concrete quantized spec. For example, launching qwen2-instruct with size 7 and quantization Int4 while omitting model_format resolves to the gptq spec, but this call still passes null and installs only transformers/accelerate, not gptqmodel or datasets. Please derive the format from model.model_family.model_specs[0].model_format (with an appropriate fallback) and cover the auto-selected path in the test.

xinference/deploy/docker/Dockerfile
```diff
@@ -1,81 +1,113 @@
...
+# by the offline pypiserver generator.
+# uv (a runtime dependency anyway, it powers the per-model venv installer)
+# also does the image install: parallel downloads with per-request timeouts
+# recover from throttled/stalled index connections that serial pip rides
+# to a crawl.
+RUN pip install --no-cache-dir -i "$PIP_INDEX" uv && \
+    NO_WEB_UI=1 uv pip install --system --no-cache --index-url "$PIP_INDEX" \
+      --extra-index-url "$TORCH_INDEX" --index-strategy unsafe-best-match \
+      --constraint /opt/requirements-runtime.txt \
+      ".[otel]" transformers accelerate
```
qinxuye, line 108: [P1] The compatibility-extra migration is still incomplete. The slim image and the Transformers engine defaults no longer provide several packages required by advertised models: the pinned qwen-chat remote tokenizer imports tiktoken, while baichuan-2-chat imports sentencepiece and transformers_stream_generator, but both families still declare only #transformers_dependencies# and that placeholder expands to transformers/accelerate. These launches will fail with ModuleNotFoundError in the new image. Please add the missing packages to the relevant model specs (and audit the remaining packages removed from the old Transformers image requirements), or retain the necessary compatibility packages in the parent image.

doc/source/getting_started/using_docker_image.rst
```diff
@@ -10,11 +10,22 @@ Xinference provides official images for use on Dockerhub.
 
    Starting from **Xinference v2.0**, to use the CUDA version of the image, the minimum CUDA version must be **CUDA 12.9**.
 
+.. versionchanged:: v2.13.0
+
+   Starting from **Xinference v2.13.0**, the GPU image is a *slim* image based on
```
qinxuye: [P2] This image change is part of the Xinference 3.0 breaking release, not v2.13.0. Please change both the versionchanged directive and the prose to v3.0.0 so this page agrees with the 3.0 migration guide in #5194.

xinference/deploy/docker/Dockerfile
qinxuye: The audit is still incomplete. qwen2.5-vl-instruct and Qwen3-VL-Instruct/Thinking use Qwen2VLChatModel, whose inference paths import qwen_vl_utils, but those family specs still declare only the core Transformers placeholder. MiniCPM-V-2.6, MiniCPM-V-4.5, and InternVL3 likewise import decord for video input, while their specs do not declare eva-decord. These paths were previously satisfied by the image-wide Transformers requirements and will now fail with ModuleNotFoundError. Please add the packages to the corresponding Transformers model specs and include them in the mirror-generation test before resolving this thread.

qinxuye, line 108: [P1] qwen3.5 and qwen3.6 are still missing qwen-vl-utils for the Transformers engine. Both families advertise the vision ability and are registered to Qwen2VLChatModel, whose inference paths import process_vision_info from qwen_vl_utils, but their virtualenv declarations contain only transformers>=5.2.0 for the Transformers branch. The previous dependency thread was resolved after adding this package to qwen2.5-vl and Qwen3-VL, but these two families will still fail on multimodal requests in the slim image. Please add qwen-vl-utils!=0.0.9 with the Transformers marker for both families and include them in the model-pin test.

qinxuye (approved):
LGTM

Issue #5189, opened 2026-07-17
First launch after a virtualenv (re)build always fails: model subprocess is spawned BEFORE venv packages are installed, caching host-version base libraries in sys.modules


kejhz653-stack:
### System Info / 系統信息

- OS: Ubuntu Linux (x86_64)
- Python: 3.12.13 (conda)
- Host site-packages: torch 2.13.0+cu130, numpy 2.5.1
- Model virtualenv (IndexTTS2, used to expose the bug): torch==2.11.0+cu130, numpy 2.2.6, numba 0.66, torchvision 0.26.0+cu130
- CUDA: 13.0, single NVIDIA GPU
- Virtual env feature: enabled (default), venv created by uv with --system-site-packages

### Running Xinference with Docker? / 是否使用 Docker 运行 Xinfernece？

- [ ] docker / docker
- [ ] pip install / 通过 pip install 安装
- [x] installation from source / 从源码安装

### Version info / 版本信息

v2.12.0 (source install, editable)

### The command used to start Xinference / 用以启动 xinference 的命令

```bash
HF_ENDPOINT=https://hf-mirror.com XINFERENCE_MODEL_SRC=modelscope xinference-local --host 0.0.0.0 --port 9991
```

### Reproduction / 复现过程

1. Use any model whose `virtualenv.packages` pins base libraries to versions **different from the host** (I used IndexTTS2 with `numpy>=2.0,<2.3` and `torch==2.11.0+cu130`, while the host has numpy 2.5.1 / torch 2.13.0+cu130).
2. Delete the model's virtualenv (`DELETE /v1/virtualenvs?model_name=...` from the Web UI, or `rm -rf ~/.xinference/virtualenv/v4/<model>`), then launch the model.
3. The **first** launch always fails with a version-mismatch error. The exact symptom varies depending on which venv package hits a version check first — two observed variants of the same root cause:

```
ImportError: Numba needs NumPy 2.4 or less. Got NumPy 2.5.
```
(numba loaded from the venv, but `numpy` in the process is the HOST 2.5.1)

```
RuntimeError: operator torchvision::nms does not exist
→ DeserializeMessageFailed → xoscar.errors.ServerClosed: Remote server unixsocket:///... closed
```
(torchvision loaded from the venv at `.../virtualenv/v4/IndexTTS2/.../torchvision/__init__.py`, but the stack shows it registering ops against the HOST torch at `.../miniconda3/envs/.../torch/library.py` — two different installations in one process)

[...]

PR #5191, opened 2026-07-17
fix(image/ocr): import torchvision lazily in deepseek_ocr (#5189)

Anai-Guo:
## Problem

`xinference/model/__init__._install()` imports every model family at actor-creation
time. That chain reaches `model/image/__init__.py` → `.ocr` → `ocr/__init__.py` →
`deepseek_ocr.py`, which does a **module-level** `from torchvision import transforms`.

Net effect: torchvision is imported into **every** model subprocess — audio models,
LLMs, anything — regardless of whether the model has anything to do with OCR.

## Why this is the odd one out

Every other torchvision consumer in the tree already imports it lazily, inside the
function that needs it:

| file | import site |
|---|---|
| `model/llm/transformers/multimodal/intern_vl.py` | inside method (line ~113) |
| `model/llm/transformers/multimodal/ovis2.py` | inside method (line ~99) |
| `model/video/diffusers.py` | inside method (line ~393) |
| `model/image/ocr/deepseek_ocr.py` | **module level** ← this PR |

So this looks like a plain oversight rather than a deliberate choice.

## Change

Move the import into the only two scopes that use `transforms`:

- `normalize_transform()`
- `BasicImageTransform.__init__()`

That is the whole diff — 5 lines, no behavior change for OCR users (the import
just happens on first use instead of at module import).

## Relation to #5189

This is the "additionally" item called out at the end of #5189, filed as its own
minimal change. It does **not** claim to fix the root cause of #5189 (the
`_create_subpool` / `_prepare_virtual_env` ordering — that reorder touches the
launch flow and deserves a separate, maintainer-directed discussion).

What it does do is remove one eager **host** torchvision import from subprocesses
whose virtualenv installs its own pinned torchvision, which the issue identifies as
what aggravates the `operator torchvision::nms does not exist` failure variant for
audio models that never use torchvision at all.

### DIFF
--- a/xinference/model/image/ocr/deepseek_ocr.py
+++ b/xinference/model/image/ocr/deepseek_ocr.py
@@ -23,7 +23,6 @@ import numpy as np
 import PIL.Image
 import torch
 import torch.nn as nn
-from torchvision import transforms
 
 if TYPE_CHECKING:
     from ..core import ImageModelFamilyV2
@@ -170,6 +169,8 @@ def normalize_transform(
     std: Optional[Union[Tuple[float, float, float], List[float]]],
 ):
     """Create normalization transform."""
+    from torchvision import transforms
+
     if mean is None and std is None:
         return None
     elif mean is None and std is not None:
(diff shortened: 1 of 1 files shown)


qinxuye (approved):
LGTM

PR #5196, opened 2026-07-17
fix(worker): create model subpool after virtualenv is installed

### LINKED ISSUE #5195: vLLM launch in a virtualenv fails after venv (re)build: RuntimeError: operator torchvision::nms does not exist
### System Info / 系統信息

- OS: Ubuntu 24.04.4 LTS (kernel 6.17.0-35-generic)
- GPU: 2 x NVIDIA GeForce RTX 3090 Ti
- NVIDIA driver: 580.159.03, CUDA Version: 13.0
- Python: 3.12.13
- Base env (inference_env) at failure time:
    torch 2.13.0+cu130, torchvision 0.28.0+cu130, transformers 5.14.0
- Per-model vLLM virtualenv:
    vllm 0.23.0, torch 2.11.0+cu130, torchvision 0.26.0+cu130, transformers 4.57.6
    (vllm 0.23.0 hard-pins torch==2.11.0 / torchvision==0.26.0 / torchaudio==2.11.0)

- [ ] docker / docker
- [x] installation from source / 从源码安装

### Version info / 版本信息

2.12.0+39.g3507c59b.dirty  (source / local mode)

xinference-local --host 0.0.0.0 --port 9999

### Reproduction / 复现过程

Launch a model with the **vLLM** engine (pytorch format) inside a per-model
virtualenv, on the **first launch after the venv is (re)built**. The model
subprocess crashes at import time and the request returns 500.

Steps to reproduce the behavior / 复现步骤:
1. Start `xinference-local` from source, with a base env whose torch/torchvision
   differ from what vLLM pins (here base torch 2.13.0 vs vLLM's torch 2.11.0).
2. Remove/rebuild the model virtualenv, then launch `qwen3.5`
   (model_format=pytorch, model_engine=vLLM, model_type=LLM).
3. First launch fails; the model subprocess dies during import.

Error / stack trace:

```text
File ".../virtualenv/v4/qwen3.5/vllm/3.12.13/.../torchvision/_meta_registrations.py", line 163
    @torch.library.register_fake("torchvision::nms")
File "/.../envs/inference_env/lib/python3.12/site-packages/torch/library.py", line 493, in impl
    if torch._C._dispatch_has_kernel_for_dispatch_key(
RuntimeError: operator torchvision::nms does not exist

# subprocess dies -> parent sees:
INFO uvicorn.access ... "POST /v1/models HTTP/1.1" 500
```

Note the mixed paths in the trace: `torchvision` is loaded from the **venv**
(0.26.0), but `torch` resolves to the **base env** (2.13.0).

Root cause: the model subpool (subprocess) is created **before** the virtualenv
is populated. The subprocess boots on the venv's python, but during bootstrap
xinference imports base-env libraries (numpy, and `torch` via the model
`_install()` import chain, e.g. `xinference/model/image/ocr/deepseek_ocr.py`
-> `from torchvision import transforms`). Those get cached in `sys.modules`
before the venv packages are installed, so the venv's own torch can no longer
replace the base-env torch. Base torch 2.13.0 + venv torchvision 0.26.0 (built
for torch 2.11) -> ABI mismatch -> `operator torchvision::nms does not exist`.
(The same mechanism causes venv-numba vs host-numpy mismatches.)

The same model launched with `llama.cpp` / GGUF works, because that path does
not trigger the torch/torchvision op registration.

### Expected behavior / 期待表现

A model launched in a per-model virtualenv should load using the venv's own
packages, regardless of the base environment's torch/torchvision versions.
The first launch after a venv (re)build should succeed, not fail with a
host/venv version mismatch.

kejhz653-stack:
The model subprocess boots on the venv's python. Base-env libraries imported during its bootstrap (numpy, and torch via xinference's import chain) get cached in sys.modules from whatever is visible at spawn time; packages installed into the venv afterwards can no longer replace them. Spawning the subpool before the venv was populated therefore caused the first launch after a venv (re)build to fail with host/venv version mixes, e.g. base torch 2.13.0 paired with venv torchvision 0.26.0 -> "operator torchvision::nms does not exist", and venv numba vs host numpy.

Defer _create_subpool until after _prepare_virtual_env for single-worker launches. Sharded multi-worker launches still need the subpool address in model_kwargs before instantiation, so they keep the subprocess-first order.fix #5195

### DIFF
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -2886,21 +2886,38 @@ class WorkerActor(xo.StatelessActor):
                     subpool_envs = build_subpool_envs_for_virtual_env(
                         envs, enable_virtual_env, virtual_env_manager
                     )
-                    subpool_address, devices = await self._create_subpool(
-                        model_uid,
-                        model_type,
-                        n_gpu=n_gpu,
-                        gpu_idx=gpu_idx,
-                        start_python=subpool_python_path,
-                        env=subpool_envs,
-                    )
-                    all_subpool_addresses = [subpool_address]
+                    # The model subprocess must not be spawned before the
+                    # virtualenv is populated: the subprocess boots on the
+                    # venv's python, and base libraries imported during its
+                    # bootstrap (numpy, and torch via xinference's own import
+                    # chain) are cached in sys.modules from whatever is
+                    # visible at that moment — packages installed afterwards
+                    # cannot replace them. Spawning before install therefore
+                    # made the first launch after a venv (re)build fail with
+                    # host/venv version mixes (e.g. venv numba vs host numpy,
+                    # venv torchvision vs host torch). Single-worker launches
+                    # defer subprocess creation until after
+                    # _prepare_virtual_env; sharded multi-worker launches need
+                    # the subpool address inside model_kwargs before model
+                    # instantiation, so they keep the legacy subprocess-first
+                    # order.
+                    subpool_address: Optional[str] = None
+                    devices: List[str] = []
+                    all_subpool_addresses: List[str] = []
+                    if n_worker > 1:  # type: ignore
+                        subpool_address, devices = await self._create_subpool(
+                            model_uid,
+                            model_type,
+                            n_gpu=n_gpu,
+                            gpu_idx=gpu_idx,
+                            start_python=subpool_python_path,
+                            env=subpool_envs,
+                        )
+                        all_subpool_addresses.append(subpool_address)
                     try:
                         xavier_config: Optional[Dict] = kwargs.pop(
                             "xavier_config", None
                         )
-                        if xavier_config is not None:
-                            xavier_config["rank_address"] = subpool_address
                         model_kwargs = kwargs.copy()
                         model_kwargs["enable_virtual_env"] = enable_virtual_env
                         if n_worker > 1:  # type: ignore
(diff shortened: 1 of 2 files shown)


xinference/core/worker.py
```diff
@@ -3028,9 +3048,24 @@ def check_cancel():
                             )
                             launch_info.virtual_env_manager = virtual_env_manager
 
-                        # check before creating model actor
+                        # check before creating subpool and model actor
                         check_cancel()
 
+                        if subpool_address is None:
+                            subpool_address, devices = await self._create_subpool(
```
qinxuye: Delaying `_create_subpool` here also delays GPU allocation, because `_create_subpool` performs `allocate_devices*` before spawning the process. Previously the device was reserved before model download and virtualenv installation. With `XINFERENCE_MAX_CONCURRENT_LAUNCHES` defaulting to 5, concurrent single-worker launches can therefore be scheduled from the same stale GPU-allocation snapshot and spend the whole preparation phase targeting the same device; they only contend/reserve here at the end, which can defeat idle-first placement, fail with no slot when multi-replica-per-GPU is disabled, or overload one GPU when it is enabled. Please keep device reservation at the original point and defer only the actual `append_sub_pool`/process creation (for example, split allocation/env preparation from subpool spawning).

qinxuye (commented):
One concurrency issue remains after the cleanup fix.

qinxuye (approved):
LGTM

PR #5200, opened 2026-07-18
fix: pass torch version to offline package builds

qinxuye:
## Summary

- derive `TORCH_VERSION` from the exact torch pin in the shared runtime constraints
- pass the package-build environment to both `pip download` and `pip wheel`
- cover environment propagation through the download orchestration tests

## Root cause

The `v3.0.0` pypiserver release build resolves `gptqmodel==4.2.5`, which is only published as an sdist. Its build metadata inspects the installed torch version. The mirror downloads the torch wheel without installing it, so both amd64 and arm64 builds failed while preparing GPTQModel metadata with `No package metadata was found for torch`.

Supplying the version through GPTQModel's supported `TORCH_VERSION` build hook avoids installing the large torch wheel into the downloader while keeping the value aligned with `requirements-runtime.txt`.

Failed release run: https://github.com/xorbitsai/inference/actions/runs/29653510204

## Validation

- `pytest -q xinference/deploy/docker/pypiserver/tests/test_scripts.py` (12 passed)
- `pre-commit run --files xinference/deploy/docker/pypiserver/download_packages.py xinference/deploy/docker/pypiserver/tests/test_scripts.py`
- in an isolated Python 3.12 environment with no torch installed, successfully downloaded the `gptqmodel==4.2.5` sdist and built `gptqmodel-4.2.5-py3-none-any.whl` with `TORCH_VERSION=2.11.0`

### DIFF
--- a/xinference/deploy/docker/pypiserver/download_packages.py
+++ b/xinference/deploy/docker/pypiserver/download_packages.py
@@ -36,6 +36,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import os
 import subprocess
 import sys
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


OliverBryant (approved):
LGTM

PR #5203, opened 2026-07-18
fix(audio/indextts2): return waveform in memory to avoid TorchCodec save failure (#5201)

### LINKED ISSUE #5201: [Bug] IndexTTS2 /v1/audio/speech returns 500 at WAV save: libtorchcodec cannot load in v2.12.0 Docker
### System Info / 系统信息

- Host: Windows + Docker Desktop
- Container OS: Ubuntu 22.04.5 LTS
- GPU: NVIDIA GeForce RTX 5070 Ti, 16 GB
- NVIDIA driver: 591.86
- Image: `xprobe/xinference:v2.12.0`
- Python: 3.12
- PyTorch: `2.11.0+cu130`
- TorchAudio: `2.11.0+cu130`
- TorchCodec: `0.10.0`
- FFmpeg CLI: `6.1.2`
- SoundFile: `0.14.0`

The IndexTTS2 model reaches `READY` and inference itself completes. The failure happens only when saving the generated waveform.

### Running Xinference with Docker? / 是否使用 Docker 运行 Xinference？

- [x] docker / docker

### Version info / 版本信息

`v2.12.0`

### The command used to start Xinference / 用以启动 Xinference 的命令

```yaml
services:
  xinference:
    image: xprobe/xinference:v2.12.0
    gpus: all
    shm_size: "8gb"
    ports:
      - "127.0.0.1:9997:9997"
    environment:
      XINFERENCE_HOME: /data
      XINFERENCE_ENABLE_VIRTUAL_ENV: "1"
    command: ["xinference-local", "-H", "0.0.0.0"]
```

### Reproduction / 复现过程

1. Start the v2.12.0 Docker image.
2. Launch `IndexTTS2` and wait until the replica is `READY`.
3. Open its launch UI, provide a valid WAV speaker prompt and text such as `你好测试一下`.
4. Generate speech, or call `POST /v1/audio/speech`.
5. IndexTTS2 runs inference, reaches the final "saving audio..." stage, then the endpoint returns HTTP 500.

Relevant call path:

```text
xinference/model/audio/indextts2.py::speech
  -> thirdparty/indextts/infer_v2.py::infer
  -> torchaudio.save(output_path, wav.type(torch.int16), sampling_rate)
  -> torchaudio.save_with_torchcodec
  -> import torchcodec.encoders.AudioEncoder
  -> Could not load libtorchcodec
```

The main error is:

```text
RuntimeError: Could not load libtorchcodec

FFmpeg version 6:
OSError: libavutil.so.58: cannot open shared object file: No such file or directory

FFmpeg version 4:
OSError: libtorchcodec_core4.so: undefined symbol:
_ZN3c1013MessageLoggerC1EPKciib
```

`ldd /usr/local/lib/python3.12/dist-packages/torchcodec/libtorchcodec_core6.so` reports:

```text
libtorch.so => not found
libc10.so => not found
libavutil.so.58 => not found
libavcodec.so.60 => not found
libavformat.so.60 => not found
libtorch_cpu.so => not found
```

The container's `ffmpeg -version` command works, but TorchCodec cannot resolve the corresponding shared libraries. There also appears to be a TorchCodec/PyTorch ABI mismatch for the fallback core library.

Full request ends with:

```text
POST /v1/audio/speech HTTP/1.1" 500
```

### Expected behavior / 期待表现

IndexTTS2 should return the generated WAV/audio response successfully after inference.

### Analysis / 建议

The generated waveform already exists before the failure. This issue is specific to the output serialization path. A narrowly scoped fix would be to avoid TorchCodec for this IndexTTS2 WAV write and use `soundfile.write` (already installed in the image), or ensure the image ships a TorchCodec build and FFmpeg shared-library paths compatible with `torch==2.11.0+cu130`.

For example, replacing the final `torchaudio.save` call with an int16 SoundFile write avoids changing the inference path and avoids impacting other models that may depend on TorchCodec.

### Related issues

- #4960 reports a generic `Could not load libtorchcodec` problem for embedding/reranker models on v2.9.0, but does not cover IndexTTS2 inference succeeding and failing specifically at WAV serialization on v2.12.0.
- #4586 covers the separate IndexTTS2 `No module named 'tn.english'` dependency problem. That dependency was fixed locally before reproducing this output-save failure.

Anai-Guo:
## Summary

Fixes #5201.

`IndexTTS2` inference completes, but `/v1/audio/speech` returns **500 at the WAV save step**: the model's `infer()` writes the intermediate WAV via `torchaudio.save`, which routes through **TorchCodec**. In several images TorchCodec's FFmpeg shared libraries are missing or ABI-mismatched against `torch==2.11.0+cu130`:

```
RuntimeError: Could not load libtorchcodec
OSError: libavutil.so.58: cannot open shared object file
```

The generated waveform already exists in memory before this failure — only the on-disk serialization breaks.

## Fix

`IndexTTS2.infer()` already supports `output_path=None`, in which case it **returns `(sampling_rate, wav)`** (int16 numpy, `samples × channels`) and skips `torchaudio.save` completely. `speech()` in `xinference/model/audio/indextts2.py` then re-encodes the response with `soundfile` anyway, after reading the temp WAV back.

So this change:

- passes `output_path=None` and consumes the returned waveform directly,
- drops the temp-file creation, the `soundfile.read()` round-trip, and the two `os.unlink(output_path)` cleanups.

Net effect: the TorchCodec/FFmpeg dependency is removed from this code path; encoding is done solely by `soundfile` (already used here and already a dependency). No inference logic changes, and no other model's TorchCodec usage is touched. The int16 array returned by `infer()` is losslessly equivalent to reading back the int16 PCM WAV that `torchaudio.save` previously wrote, so the emitted audio is unchanged.

### DIFF
--- a/xinference/model/audio/indextts2.py
+++ b/xinference/model/audio/indextts2.py
@@ -138,15 +138,18 @@ class Indextts2:
                 temp_emo.write(emo_prompt_speech)
                 emo_prompt_path = temp_emo.name
 
-        # Generate complete audio first
-        with tempfile.NamedTemporaryFile(suffix=".wav", delete=False) as temp_output:
-            output_path = temp_output.name
-
         try:
-            self._model.infer(
+            # Pass output_path=None so IndexTTS2 returns the generated waveform in
+            # memory (as an int16 numpy array) instead of writing it to disk with
+            # torchaudio.save. torchaudio.save routes through TorchCodec, whose
+            # FFmpeg shared libraries are missing / ABI-mismatched in some images
+            # (see #5201), which made /v1/audio/speech fail at the save step even
+            # though inference had already succeeded. soundfile below performs the
+            # actual encoding, so the on-disk WAV round-trip is not needed.
+            sample_rate, audio = self._model.infer(
                 spk_audio_prompt=temp_prompt_path,
                 text=input,
-                output_path=output_path,
+                output_path=None,
                 emo_audio_prompt=emo_prompt_path,
                 emo_alpha=emo_alpha,
                 emo_text=emo_text,
@@ -154,9 +157,6 @@ class Indextts2:
                 emo_vector=emo_vector,
                 use_emo_text=use_emo_text,
             )
-
-            # Read generated audio
-            audio, sample_rate = soundfile.read(output_path)
 
             if stream:
                 # Streaming mode - return generator that yields chunks
@@ -167,9 +167,6 @@ class Indextts2:
                         ) as f:
                             f.write(audio)
                         complete_audio = out.getvalue()
-
-                    # Clean up temp file
-                    os.unlink(output_path)
 
                     # Yield the complete audio in chunks
                     chunk_size = 8192  # 8KB chunks
@@ -186,8 +183,6 @@ class Indextts2:
                         f.write(audio)
                     result = out.getvalue()
 
-                # Clean up temp file
-                os.unlink(output_path)
                 return result
         finally:
             # Clean up temp files


OliverBryant:
Thanks for the detailed analysis and the patch — the diagnosis of the v2.12.0 failure chain (`torchaudio.save` → TorchCodec → missing/mismatched FFmpeg shared libraries) is accurate.

That said, after verifying the conditions of the current `main` image, we are going to close this in favor of #5223:

1. **The root cause no longer exists on `main`.** The v2.12.0 failure came from the old image shipping `torchcodec==0.10.0` (built against a different torch, hence the `undefined symbol: _ZN3c1013MessageLoggerC1EPKciib` fallback error) together with an FFmpeg 6 CLI that had no shared libraries on the loader path. The current slim image pins a matched pair `torch==2.11.0` / `torchcodec==0.14.0` (`requirements-runtime.txt`) and installs apt `ffmpeg` 4.4 with its shared libraries. I reproduced the Dockerfile's deadsnakes python3.12 + ffmpeg layers in a container with exactly those pins: TorchCodec loads its FFmpeg-4 core fine and `torchaudio.save`/`load` round-trips succeed. So the WAV save step this PR bypasses works on the v3 image.

2. **The actual remaining blocker for IndexTTS2 on v3.0.0 is different.** As found in the follow-up testing on the issue (https://github.com/xorbitsai/inference/issues/5201#issuecomment-5014913961), the model cannot even launch because its virtualenv dependency declaration is incomplete. That is fixed by #5223, which completes the spec (librosa, soundfile, einops, omegaconf, sentencepiece, textstat, WeTextProcessing/wetext, `#system_torchaudio#`, …) with container verification on both amd64 and arm64.

With #5223 merged, IndexTTS2 launches and the standard `torchaudio.save` output path works end-to-end on the current image, so the in-memory bypass isn't needed anymore. Closing — and thanks again for the thorough report and analysis, it made tracking down the real state of things much easier.

OliverBryant:
Follow-up: further real-world testing on a bare-metal conda deployment (torch 2.11.0+cu130, no FFmpeg shared libraries in the environment) showed that this failure mode does persist outside the Docker image — inference completes and the save step raises `Could not load libtorchcodec` for every FFmpeg core, exactly as diagnosed here. The Docker image itself is fine, but conda/pip environments often are not.

Your commit has therefore been cherry-picked as-is (authorship preserved) into #5223, which bundles the rest of the IndexTTS2 launch fixes. Thanks again for the fix — closing this one turned out to be premature on my part with respect to non-Docker environments, and the change is landing through #5223.

PR #5215, opened 2026-07-20
fix: disambiguate replica model UIDs with -rep{n} suffix

### LINKED ISSUE #5198: parse_replica_model_uid silently corrupts bare model_uid values ending in '-<digits>' (e.g. llama-2, phi-2, gpt-2)
### Describe the bug

`parse_replica_model_uid` in `xinference/core/utils.py` (lines 178-187) is meant to split a replica uid like `"my-model-0"` back into `("my-model", 0)`, and to detect when the input has **no** replica suffix at all, returning `(uid, -1)` in that case:

```python
def parse_replica_model_uid(replica_model_uid: str) -> Tuple[str, int]:
    """
    Parse replica model uid to model uid and rep id.
    """
    parts = replica_model_uid.split("-")
    if len(parts) == 1:
        return replica_model_uid, -1
    rep_id = int(parts.pop())
    model_uid = "-".join(parts)
    return model_uid, rep_id
```

The `len(parts) == 1` branch shows the function is explicitly designed to be safe to call on a bare, non-replica-suffixed `model_uid` (several call sites do exactly that — see below). But the "no suffix" detection only works when the bare uid has **zero** hyphens. If a bare `model_uid` itself ends in `-<digits>` (no replica suffix ever appended), it's silently misparsed as if the trailing digits were a real replica index, and the last `-<digits>` segment is stripped from the uid.

This isn't a hypothetical shape: `_gen_model_uid` (`xinference/core/supervisor.py`, ~line 1853) defaults an unlaunched model's `model_uid` to the bare `model_name` on first launch, and `xinference/model/llm/llm_family.json` currently ships built-in model families whose `model_name` ends in exactly this pattern, e.g. `llama-2`, `llama-3`, `phi-2`, `gpt-2`, `glm-5`, `baichuan-2`. Launching any of these with no custom UID gives `model_uid == "llama-2"` etc. — the exact ambiguous shape.

### To Reproduce

Against a fresh clone of `main` (`xinference/core/utils.py`, unmodified):

```python
from xinference.core.utils import build_replica_model_uid, parse_replica_model_uid

# Correctly detected: no hyphen at all.
assert parse_replica_model_uid("mymodel") == ("mymodel", -1)

# Correctly round-trips: a GENUINE replica suffix was appended.
built = build_replica_model_uid("gpt-4", 0)          # "gpt-4-0"
assert parse_replica_model_uid(built) == ("gpt-4", 0)  # OK

# BUG: a bare uid (never passed through build_replica_model_uid) that
# happens to end in "-<digits>" is silently corrupted:
print(parse_replica_model_uid("llama-2"))   # -> ('llama', 2)   expected ('llama-2', -1)
print(parse_replica_model_uid("phi-2"))     # -> ('phi', 2)     expected ('phi-2', -1)
print(parse_replica_model_uid("gpt-4"))     # -> ('gpt', 4)     expected ('gpt-4', -1)
```

Real model names currently in `llm_family.json` that trigger this: `llama-2`, `llama-3`, `phi-2`, `gpt-2`, `glm-5`, `baichuan-2`, `deepseek-v3-0324`, `gemma-4`, etc.

### Expected behavior

`parse_replica_model_uid("llama-2")` should return `("llama-2", -1)`, matching the documented behavior for "uid with no replica suffix" (same as the zero-hyphen case).

### Actual behavior

The trailing `-<digits>` segment of the bare uid is misread as a fake replica index and silently stripped from the model uid, with no exception raised (so no fallback/error path is triggered downstream).

### Where this bites

`parse_replica_model_uid` has ~15 call sites across `worker.py`, `supervisor.py`, `metrics.py`, and `model/llm/core.py`. Most operate on values built by `build_replica_model_uid`/`iter_replica_model_uid`, which round-trip correctly regardless of hyphens in the base name (split-all/pop-last/rejoin is fine when a real suffix was appended — I want to be clear this part is *not* broken). The bug only fires on the subset of call sites that may receive a bare `model_uid`, e.g.:

- `WorkerActor.is_model_vllm_backend(self, model_uid: str)` (`worker.py` ~line 1445) parses its `model_uid` argument and then does `supervisor_ref.get_model(_model_uid)` with the parsed value — for a bare uid like `"llama-2"` this would look up `"llama"` instead.
- `metrics.py` (~line 543) has a defensive `except (TypeError, ValueError): base_uid = m_uid` fallback for malformed input, but that fallback never triggers here because `int("2")` succeeds — the parse "succeeds" with the wrong answer instead of raising, so the per-model GPU-memory metric silently keys/looks up under the wrong `model_uid`.

I haven't traced every one of the ~15 call sites to confirm which ones definitely receive a bare (non-suffixed) uid in production versus always a genuinely-built replica uid — flagging this as a function-level correctness bug against its own documented contract (the `len(parts) == 1` fallback) rather than claiming a specific worst-case exploit chain.

### Suggested direction

The underlying issue is that encoding the replica index as a `-<n>` string suffix is fundamentally ambiguous whenever the base `model_uid` can itself end in `-<digits>` (which is common for real model names). A robust fix likely needs either a delimiter that's reserved/disallowed in `model_uid` (e.g. validate in `is_valid_model_uid` that user-supplied uids can't end in a bare `-<digits>` segment, or use a less ambiguous suffix like `-rep<n>`), or tracking replica uids via an explicit registry rather than deriving them by parsing the string. Both are format/compatibility decisions, so I'm filing this as a bug report rather than a PR — happy to help implement whichever direction the maintainers prefer.

### Environment

- xinference: main branch, commit from 2026-07-18
- Reproduced with a pure Python unit-level test against `xinference/core/utils.py`, no GPU/model download required.

Solaris-star:
## Summary
`parse_replica_model_uid` split on `-` and treated the last segment as a replica index whenever it was numeric. Bare model UIDs that themselves end in `-<digits>` (built-in names such as `llama-2`, `phi-2`, `gpt-2`, `glm-5`, …) were therefore rewritten, e.g. `llama-2` → `("llama", 2)` instead of `("llama-2", -1)`.

That breaks any call site that may receive a bare `model_uid` (worker backend checks, metrics keys, …) even though the `len(parts) == 1` branch was clearly intended to accept non-replica UIDs.

## Change
- Encode replica UIDs as `{model_uid}-rep{n}` via `build_replica_model_uid` / `iter_replica_model_uid`
- Parse only the reserved `-rep{n}` suffix; anything else is a bare UID (`rep_id = -1`)
- Reject user-supplied UIDs that already look like a replica suffix in `is_valid_model_uid`
- Replace the one hardcoded `f"{model_uid}-{_rep_idx}"` in supervisor metrics with the helper
- Unit tests for bare digit-suffixed names + round-trip + validation

## Test plan
- [x] Pure logic assertions for `llama-2` / `phi-2` / `gpt-4` bare parse + `llama-2-rep0` round-trip
- [x] Extended `xinference/core/tests/test_utils.py`
- [ ] CI: `pytest xinference/core/tests/test_utils.py -q`

Fixes #5198

## Upgrade notes (added by maintainer)

Replica UIDs change format from `{uid}-{n}` to `{uid}-rep{n}`, which has two operational consequences:

- **Worker recovery files** written by older versions are migrated automatically: `_try_recover_models` now falls back to the legacy `{uid}-{n}` parse (only when the stripped base UID is a model the supervisor still knows) and rewrites the UID to the new format.
- **Live state is not migrated**: replicas launched by an older version and still running (supervisor restart against old workers, or mixed supervisor/worker versions) cannot be re-tracked, because the old-format UIDs are baked into running actors. Upgrade supervisor and workers together and relaunch models that were running before the upgrade.

### DIFF
--- a/xinference/core/utils.py
+++ b/xinference/core/utils.py
@@ -163,7 +163,9 @@ def log_sync(logger, level=logging.DEBUG, log_exception=True):
 # Replica suffix must not collide with bare model names that end in
 # "-<digits>" (llama-2, phi-2, gpt-2, ...). See #5198.
 _REPLICA_MODEL_UID_SEP = "-rep"
-_REPLICA_MODEL_UID_RE = re.compile(rf"^(?P<uid>.+){_REPLICA_MODEL_UID_SEP}(?P<rep>\d+)$")
+_REPLICA_MODEL_UID_RE = re.compile(
+    rf"^(?P<uid>.+){_REPLICA_MODEL_UID_SEP}(?P<rep>\d+)$"
+)
 
 
 def iter_replica_model_uid(model_uid: str, replica: int) -> Generator[str, None, None]:
@@ -197,12 +199,34 @@ def parse_replica_model_uid(replica_model_uid: str) -> Tuple[str, int]:
         return match.group("uid"), int(match.group("rep"))
     # No reserved suffix: treat as bare model uid. Do NOT fall back to the
     # legacy "-{n}" split here — that mis-parses real model names like
-    # "llama-2" as ("llama", 2). Legacy in-memory uids built before this
-    # change still round-trip via the pattern only when callers pass a
-    # value that was produced by the old builder AND the base uid itself
-    # never ended in digits; those are rare enough that we prefer correct
-    # bare-name handling.
+    # "llama-2" as ("llama", 2). UIDs built by the old builder therefore no
+    # longer parse as replicas; migration paths that may still see them
+    # (e.g. worker recovery files written by an older version) must handle
+    # the legacy format explicitly via parse_legacy_replica_model_uid.
     return replica_model_uid, -1
+
+
+_LEGACY_REPLICA_MODEL_UID_RE = re.compile(r"^(?P<uid>.+)-(?P<rep>\d+)$")
+
+
+def parse_legacy_replica_model_uid(
+    replica_model_uid: str,
+) -> Optional[Tuple[str, int]]:
+    """
+    Parse the legacy ``{model_uid}-{n}`` replica format used before the
+    reserved ``-rep{n}`` suffix was introduced.
+
+    Only intended for migration paths (e.g. worker recovery files written
+    by an older version). The legacy format is ambiguous by design — a bare
+    model name like ``llama-2`` also matches — so callers must verify that
+    the returned base uid actually refers to a known model before trusting
+    the result. Returns ``None`` when the input does not match the legacy
+    pattern.
+    """
+    match = _LEGACY_REPLICA_MODEL_UID_RE.match(replica_model_uid)
+    if match:
+        return match.group("uid"), int(match.group("rep"))
+    return None
 
 
 def is_valid_model_uid(model_uid: str) -> bool:
--- a/xinference/core/utils.py
+++ b/xinference/core/utils.py
@@ -160,37 +160,58 @@ def log_sync(logger, level=logging.DEBUG, log_exception=True):
     return decorator
 
 
+# Replica suffix must not collide with bare model names that end in
+# "-<digits>" (llama-2, phi-2, gpt-2, ...). See #5198.
+_REPLICA_MODEL_UID_SEP = "-rep"
+_REPLICA_MODEL_UID_RE = re.compile(rf"^(?P<uid>.+){_REPLICA_MODEL_UID_SEP}(?P<rep>\d+)$")
+
+
 def iter_replica_model_uid(model_uid: str, replica: int) -> Generator[str, None, None]:
     """
     Generates all the replica model uids.
     """
     replica = int(replica)
     for rep_id in range(replica):
-        yield f"{model_uid}-{rep_id}"
+        yield build_replica_model_uid(model_uid, rep_id)
 
 
 def build_replica_model_uid(model_uid: str, rep_id: int) -> str:
     """
     Build a replica model uid.
+
+    Uses a reserved ``-rep{n}`` suffix so bare model names that themselves
+    end in ``-<digits>`` (e.g. ``llama-2``) are not ambiguous with a replica
+    index.
     """
-    return f"{model_uid}-{rep_id}"
+    return f"{model_uid}{_REPLICA_MODEL_UID_SEP}{int(rep_id)}"
 
 
 def parse_replica_model_uid(replica_model_uid: str) -> Tuple[str, int]:
     """
     Parse replica model uid to model uid and rep id.
+
+    Returns ``(model_uid, -1)`` when the input has no replica suffix.
     """
-    parts = replica_model_uid.split("-")
-    if len(parts) == 1:
-        return replica_model_uid, -1
-    rep_id = int(parts.pop())
-    model_uid = "-".join(parts)
-    return model_uid, rep_id
+    match = _REPLICA_MODEL_UID_RE.match(replica_model_uid)
+    if match:
+        return match.group("uid"), int(match.group("rep"))
+    # No reserved suffix: treat as bare model uid. Do NOT fall back to the
+    # legacy "-{n}" split here — that mis-parses real model names like
+    # "llama-2" as ("llama", 2). Legacy in-memory uids built before this
+    # change still round-trip via the pattern only when callers pass a
+    # value that was produced by the old builder AND the base uid itself
+    # never ended in digits; those are rare enough that we prefer correct
+    # bare-name handling.
+    return replica_model_uid, -1
 
 
 def is_valid_model_uid(model_uid: str) -> bool:
     model_uid = model_uid.strip()
     if not model_uid or len(model_uid) > 100:
+        return False
+    # Forbid user-supplied uids that look like a replica suffix so they
+    # cannot collide with build_replica_model_uid output.
+    if _REPLICA_MODEL_UID_RE.match(model_uid):
         return False
     return True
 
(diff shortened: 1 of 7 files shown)


xinference/core/utils.py
```diff
@@ -160,38 +160,59 @@ def wrapped(*args, **kwargs):
...
+    if match:
+        return match.group("uid"), int(match.group("rep"))
+    # No reserved suffix: treat as bare model uid. Do NOT fall back to the
+    # legacy "-{n}" split here — that mis-parses real model names like
+    # "llama-2" as ("llama", 2). Legacy in-memory uids built before this
+    # change still round-trip via the pattern only when callers pass a
+    # value that was produced by the old builder AND the base uid itself
+    # never ended in digits; those are rare enough that we prefer correct
+    # bare-name handling.
+    return replica_model_uid, -1
```
OliverBryant, line 206: The old `{uid}-{n}` format can outlive a process and reach this parser, and it is now always treated as a bare UID. Concretely:

1. The worker persists launch args keyed by replica UID (`WorkerActor._persist_launch_args`), and `recover_models_on_startup` parses those keys. A recovery file written before this change holds keys like `myllm-0`; this parser now returns `("myllm-0", -1)`, `describe_model("myllm-0")` fails, and recovery is silently skipped.
2. The supervisor rebuild paths (`_get_model_uid_and_replica_index` and its callers in `_rebuild_worker_replica_state` / `_rebuild_worker_status_guard_state`) take the `replica_idx < 0` branch for pre-upgrade replicas reported by still-running workers, so those replicas drop out of tracking after a supervisor restart.
3. In a mixed-version cluster, a new supervisor sends `myllm-rep0` to an old worker, whose parser executes `int("rep0")` and raises `ValueError`.

All of these fail soft (skip + warning) or with an explicit error, but the practical consequence is: models launched before the upgrade must be relaunched, and supervisor/workers must be upgraded together. Could you either call this out in the PR description / release notes, or add a narrow legacy fallback in just the two recovery paths (accept a trailing `-{n}` only when the stripped base UID is a model the supervisor knows)?

xinference/core/utils.py
```diff
@@ -160,38 +160,59 @@ def wrapped(*args, **kwargs):
...
 
 
 def is_valid_model_uid(model_uid: str) -> bool:
     model_uid = model_uid.strip()
     if not model_uid or len(model_uid) > 100:
         return False
+    # Forbid user-supplied uids that look like a replica suffix so they
+    # cannot collide with build_replica_model_uid output.
+    if _REPLICA_MODEL_UID_RE.match(model_uid):
+        return False
```
OliverBryant, line 239: Now that `is_valid_model_uid` also rejects `-rep{n}`-shaped names, the launch-time error message in `SupervisorActor.launch_builtin_model` ("The model UID is invalid. Please specify the model UID by 0 < length <= 100.") is misleading for a user who picked e.g. `mymodel-rep1` — the UID satisfies the length rule yet gets rejected. Consider mentioning the reserved `-rep{n}` suffix in that message (and in the autostart validation) so users can tell why their UID was refused.

OliverBryant (commented):
Reviewed the change and verified it locally: all in-tree call sites already go through the three helpers (the one hardcoded f-string in `get_cluster_metrics_data` is fixed here), the Xavier `-rank0` path is unaffected, and `xinference/core/tests/test_utils.py` passes with this diff applied. The fix direction looks right to me.

Two things worth addressing before merge, left as inline comments:

1. Pre-upgrade UIDs in the old `{uid}-{n}` format can outlive a process (worker recovery file, supervisor rebuild from running workers) and are now treated as bare UIDs, so models running before the upgrade are silently dropped from recovery/tracking. This at least needs a release note.
2. The launch-time validation error message no longer matches what `is_valid_model_uid` checks.

Also +1 to the existing comment that the "legacy in-memory uids still round-trip" paragraph in `parse_replica_model_uid` is inaccurate — old-format UIDs never match the new pattern — though I would skip the suggested broad `try/except`; the new implementation no longer raises.

xinference/core/utils.py
OliverBryant: Addressed in 7396eacb2: added `parse_legacy_replica_model_uid()` and a migration fallback in the worker's `_try_recover_models()` — recovery files written by older versions with `{uid}-{n}` keys are now migrated to the reserved `-rep{n}` format when the stripped base uid is still known to the supervisor, instead of being silently skipped. The remaining operational caveats (replicas tracked by a live supervisor rebuild, mixed-version clusters) cannot be fixed in code because the old-format actor uids are baked into running xoscar actors; documented them in an "Upgrade notes" section of the PR description instead.

xinference/core/utils.py
OliverBryant: Addressed in 7396eacb2: the validation error messages in `SupervisorActor` (all three call sites) and in the autostart config validation now mention the reserved `-rep<number>` suffix in addition to the length rule.

Solaris-star:
Thanks @OliverBryant for pushing 7396eacb2 — the `parse_legacy_replica_model_uid()` + migration fallback in `_try_recover_models()` is exactly the right shape for the recovery path, and documenting the live-supervisor-rebuild / mixed-version caveats in the upgrade notes (rather than forcing a code workaround) makes sense given the old-format actor UIDs are baked into running xoscar actors. The updated validation messages that now mention the reserved `-rep<number>` suffix also cover the misleading-error concern. LGTM on my end — happy for this to go in whenever you're ready.

qinxuye (approved):
LGTM

PR #5218, opened 2026-07-21
feat(ui): add SSO/OIDC login button to login page

m199369309:
## Summary

Add a **"Sign in with Keycloak"** SSO button to the login page when OIDC is enabled on the server.

The backend already exposes `oidc_enabled` via `/v1/cluster/ui_config` and all OIDC routes are registered. The only missing piece was the frontend login button.

## Changes

- **`frontend/src/components/pages/login/index.tsx`**: Renders an "OR" divider and "Sign in with Keycloak" button below the login form, gated by `clusterUIConfig.oidc_enabled` from the existing global context (no separate API call).
- **i18n**: Added `login.or` and `login.ssoLogin` translation keys to all 4 locale files (en, zh, ja, ko).

## Background

This feature was lost during the v3.0.0 frontend rewrite from the old React MUI app (which had it) to the new Next.js app. The backend OIDC infrastructure was already complete and working.

## Testing

- Button only appears when `XINFERENCE_OIDC_ENABLED=true` is set on the server
- Clicking the button redirects to `/api/oidc/authorize` → Keycloak login → callback → JWT token → home page

### DIFF
--- a/frontend/src/components/pages/login/index.tsx
+++ b/frontend/src/components/pages/login/index.tsx
@@ -10,7 +10,7 @@ import { FormField } from '@/components/ui/form-field';
 import { Form } from '@/components/ui/form';
 import { useForm } from '@/hooks/use-form';
 import request from '@/lib/request';
-import { decodeJwtPayload } from '@/lib/utils';
+import { decodeJwtPayload, getApiUrl } from '@/lib/utils';
 import { setAccessToken, setRefreshToken } from '@/lib/auth-storage';
 import { getBrandingFromEnv } from '@/lib/branding';
 import { useI18n } from '@/contexts/i18n-context';
@@ -28,7 +28,7 @@ interface TokenResponse {
 export default function LoginPage() {
   const router = useRouter();
   const [form] = useForm();
-  const { fetchGlobalAfterAuth, clusterUIConfig } = useGlobal();
+  const { fetchGlobalAfterAuth } = useGlobal();
 
   const { t } = useI18n();
   const branding = getBrandingFromEnv();
@@ -36,6 +36,17 @@ export default function LoginPage() {
   const [loading, setLoading] = useState(false);
   const [changePasswordOpen, setChangePasswordOpen] = useState(false);
   const [changePasswordUserId, setChangePasswordUserId] = useState<number | null>(null);
+  const [oidcEnabled, setOidcEnabled] = useState(false);
+  const [authAdvanced, setAuthAdvanced] = useState(false);
+
+  useEffect(() => {
+    request.get('/v1/cluster/ui_config')
+      .then((config: any) => {
+        setOidcEnabled(config?.oidc_enabled ?? false);
+        setAuthAdvanced(config?.auth_advanced ?? false);
+      })
+      .catch(() => {});
+  }, []);
 
   useEffect(() => {
     // Setup (components/pages/setup) sets this right before navigating here,
@@ -135,7 +146,7 @@ export default function LoginPage() {
             </Button>
           </Form>
 
-          {clusterUIConfig.auth_advanced && clusterUIConfig.oidc_enabled && (
+          {authAdvanced && oidcEnabled && (
             <>
               <div className="relative my-6">
                 <div className="absolute inset-0 flex items-center">
@@ -152,7 +163,7 @@ export default function LoginPage() {
                 variant="outline"
                 type="button"
                 onClick={() => {
-                  window.location.href = '/api/oidc/authorize';
+                  window.location.href = getApiUrl() + '/api/oidc/authorize';
                 }}
               >
                 {t('login.ssoLogin')}
--- a/frontend/src/components/pages/login/index.tsx
+++ b/frontend/src/components/pages/login/index.tsx
@@ -135,7 +135,7 @@ export default function LoginPage() {
             </Button>
           </Form>
 
-          {clusterUIConfig.oidc_enabled && (
+          {clusterUIConfig.auth_advanced && clusterUIConfig.oidc_enabled && (
             <>
               <div className="relative my-6">
                 <div className="absolute inset-0 flex items-center">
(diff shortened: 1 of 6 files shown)


frontend/src/components/pages/login/index.tsx
```diff
@@ -134,6 +134,31 @@ export default function LoginPage() {
...
+                    {t('login.or')}
+                  </span>
+                </div>
+              </div>
+              <Button
+                block
+                variant="outline"
+                type="button"
+                onClick={() => {
+                  window.location.href = '/api/oidc/authorize';
```
OliverBryant: This bypasses `getApiUrl()`, so in dev (UI and API on different origins via `NEXT_PUBLIC_API_URL`) this hits the Next dev server and 404s. Prefer:

```ts
window.location.href = getApiUrl() + '/api/oidc/authorize';
```

for consistency with the rest of the request layer. (Note the OIDC callback hands tokens over via same-origin `sessionStorage`, so a cross-origin deployment can't complete the flow regardless — but this keeps the entry point consistent.)

frontend/src/i18n/locales/en.ts
```diff
@@ -617,6 +617,8 @@ const en = {
     password: 'Password',
     mustChangePasswordDesc: 'You need to change your password before continuing on first login.',
     changePasswordUserMissing: 'Unable to identify the current user. Please log in again.',
+    or: 'or',
+    ssoLogin: 'Sign in with Keycloak',
```
OliverBryant, line 621: The backend OIDC implementation is provider-agnostic (`XINFERENCE_OIDC_ISSUER` can point to Keycloak, Okta, Auth0, or any OIDC provider), but this label hardcodes Keycloak in all four locales. Consider a generic "Sign in with SSO", or making the provider display name configurable.

OliverBryant (dismissed):
Thanks for restoring this — the feature was indeed lost in the v3.0.0 frontend rewrite, and the backend flow does line up with the new UI (the callback stores `token`/`refresh_token` in `sessionStorage`, matching `frontend/src/lib/auth-storage.ts`). However, as implemented the button can never render for an unauthenticated user, so the feature doesn't work yet — see the existing thread on `login/index.tsx` for details and a suggested fix that avoids a new side effect.

One additional edge case worth closing: `/v1/cluster/ui_config` reports `oidc_enabled` purely from the `XINFERENCE_OIDC_ENABLED` env var (`xinference/api/routers/admin.py:279`), while the `/api/oidc/*` routes are only registered when advanced auth is **also** enabled (`xinference/api/restful_api.py:487-494`). With the OIDC env var set but advanced auth off, the button (once rendering is fixed) would appear and then 404 on click. Gating the button on `auth_advanced && oidc_enabled`, or making the backend report the effective state, would close that gap.

Since the current code can't show the button from an unauthenticated login page, please also re-check the manual test steps described in the PR once the fix is in.

frontend/src/components/pages/login/index.tsx
```diff
@@ -28,7 +28,7 @@ interface TokenResponse {
 export default function LoginPage() {
   const router = useRouter();
   const [form] = useForm();
-  const { fetchGlobalAfterAuth } = useGlobal();
+  const { fetchGlobalAfterAuth, clusterUIConfig } = useGlobal();
```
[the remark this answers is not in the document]
OliverBryant: Verified — this is a real blocker. In `AppInit` (`frontend/src/contexts/app-init.tsx`), `fetchGlobalAfterAuth()` is only called when auth is disabled or when a token/refresh token already exists. An unauthenticated visitor sitting on `/login` never triggers it, so `clusterUIConfig` stays `{}` and the button never renders.

However, the suggested fix above introduces a new problem: `fetchGlobalAfterAuth()` also requests `/v1/cluster/version`, which is registered with `Security(auth, scopes=["admin"])` when auth is enabled (`xinference/api/routers/admin.py:1001`). Calling it unauthenticated returns 401, and the axios interceptor emits `RequestEvents.UNAUTHORIZED`, which toasts an error, clears tokens, and redirects (`frontend/src/contexts/request-context.tsx`) — every visit to the login page would flash an error toast.

`/v1/cluster/ui_config` itself is public (registered without an auth dependency), so a cleaner fix is to fetch only that endpoint on login-page mount with a plain `fetch(getApiUrl() + '/v1/cluster/ui_config')` — the same pattern `fetchNeedsSetup()` uses in `app-init.tsx` — and feed the result into `setClusterUIConfig` (or local state).

OliverBryant (commented):
Re-reviewed after aca849ed — both functional blockers from my previous review are fixed:

1. **Unauthenticated rendering**: `AppInit` now fetches `/v1/cluster/ui_config` in the not-logged-in branch, and that endpoint is registered without auth dependencies, so anonymous access works. With the `{}` default for `clusterUIConfig`, the button simply stays hidden until the fetch resolves — safe fallback if the request fails.

2. **`auth_advanced` gating**: the button is now gated on `auth_advanced && oidc_enabled`, which exactly mirrors backend route registration (`_advanced_auth_service` + `XINFERENCE_OIDC_ENABLED`). I checked the env-var parsing on both sides (`XINFERENCE_AUTH_ADVANCED` default-true with the same falsy set; `XINFERENCE_OIDC_ENABLED` with the same truthy set) — they match, so the click-then-404 case is closed.

Also re-verified the token handoff: the OIDC callback writes `token` / `refresh_token` to sessionStorage, matching `ACCESS_TOKEN_KEY` / `REFRESH_TOKEN_KEY` in `frontend/src/lib/auth-storage.ts`.

Two earlier minor comments are still open and non-blocking — the hardcoded `window.location.href = '/api/oidc/authorize'` bypassing `getApiUrl()`, and the Keycloak-specific label in all four locales. Fine to address here or in a follow-up. Dismissing my earlier "request changes" since the functional issues are resolved.

qinxuye (approved):
LGTM

PR #5220, opened 2026-07-21
fix(api): cast replica to int to prevent TypeError on multi-replica launch

m199369309:
## Summary

Fix a `TypeError: \x27str\x27 object cannot be interpreted as an integer` when launching a model with `replica > 1` via the REST API.

## Root cause

The frontend submits `replica` as a string in the JSON body (e.g. `"replica": "2"`). `payload.get("replica", 1)` returns the raw string, and `list(range("2"))` raises `TypeError`.

Single-replica deployments are unaffected because the frontend usually omits the field entirely, so the default `1` (int) is used.

## Fix

Wrap `payload.get("replica", 1)` with `int()` in both `launch_model()` and `launch_model_by_version()`:

```python
# before
replica = payload.get("replica", 1)

# after
replica = int(payload.get("replica", 1))
```

This handles both integer (`int(1) -> 1`) and string (`int("2") -> 2`) inputs safely.

## Backward compatibility

✅ Fully compatible — `int(1)` is a no-op, `int("2")` produces the correct integer.

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -783,7 +783,7 @@ class RESTfulAPI(CancelMixin):
         model_format = payload.get("model_format")
         quantization = payload.get("quantization")
         model_type = payload.get("model_type", "LLM")
-        replica = payload.get("replica", 1)
+        replica = int(payload.get("replica", 1))
         n_gpu = payload.get("n_gpu", "auto")
         request_limits = payload.get("request_limits", None)
         peft_model_config = payload.get("peft_model_config", None)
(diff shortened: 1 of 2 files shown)


xinference/api/restful_api.py
```diff
@@ -783,7 +783,18 @@ async def launch_model(
         model_format = payload.get("model_format")
         quantization = payload.get("quantization")
         model_type = payload.get("model_type", "LLM")
-        replica = payload.get("replica", 1)
+        try:
+            replica = int(payload.get("replica", 1))
```
qinxuye: `int()` still coerces non-integer JSON values instead of validating them: `1.9` and `true` are both forwarded as replica `1`, while `1e400` parses as infinity and raises an uncaught `OverflowError`, returning 500. Please validate the raw value (allow integer strings, reject booleans and fractional/non-finite floats), catch `OverflowError`, apply the same logic to both launch endpoints, and add focused tests.

qinxuye (approved):
LGTM

PR #5222, opened 2026-07-22
feat: launch model worker multi select

### LINKED ISSUE #5206: bug: Cannot multi select worker ip address
## 2.12.0
Image

## 3.0.0
Image

maoyuehui:
1.Changed the Launch Model worker_ip field from an Input to a MultiSelect, including options fetching and submit/backfill data handling.
2.Optimized the dropdown styles for Select, MultiSelect, and AutoComplete components.

### DIFF
--- a/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
+++ b/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
@@ -517,7 +517,7 @@ export default function LaunchDialog({
       {
         name: 'n_gpu',
         type: 'select',
-        label: t('launchModel.nGPU'),
+        label: t('launchModel.nGPUDevice'),
         fieldProps: { options: nGpuOptions },
       },
       {
@@ -621,7 +621,7 @@ export default function LaunchDialog({
       {
         name: 'n_gpu',
         type: 'select',
-        label: t('launchModel.nGPU'),
+        label: t('launchModel.nGPUDevice'),
         fieldProps: { options: nGpuOptions },
       },
       {
@@ -864,7 +864,7 @@ export default function LaunchDialog({
       {
         name: 'n_gpu',
         type: 'select',
-        label: t('launchModel.nGPU'),
+        label: t('launchModel.nGPUDevice'),
         fieldProps: { options: nGpuOptions },
       },
       {
@@ -961,7 +961,7 @@ export default function LaunchDialog({
       {
         name: 'n_gpu',
         type: 'select',
-        label: t('launchModel.nGPU'),
+        label: t('launchModel.nGPUDevice'),
         fieldProps: { options: nGpuOptions },
       },
       {
@@ -1065,7 +1065,7 @@ export default function LaunchDialog({
       {
         name: 'n_gpu',
         type: 'select',
-        label: t('launchModel.nGPU'),
+        label: t('launchModel.nGPUDevice'),
         fieldProps: { options: nGpuOptions },
       },
       {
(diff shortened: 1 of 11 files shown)


frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -87,7 +94,14 @@ export default function LaunchDialog({
   const quantizationValue = toOptionValue(useWatch('quantization', form));
   const multimodalProjectorValue = toOptionValue(useWatch('multimodal_projector', form));
   const nGpuValue = useWatch('n_gpu', form);
+  const [workerOptions, setWorkerOptions] = useState<WorkerOption[]>([]);
 
+  const fetchWorkers = useCallback(async () => {
+    const data = await request.get<ClusterInfoResponse>('/v1/cluster/info', {
```
qinxuye: This endpoint is protected by the `admin` scope when authentication is enabled, while launching a model only requires `models:write`. A non-admin user who is allowed to launch models will therefore get a 403 here and an empty worker selector, and the previous free-form input is no longer available. Please expose the minimal worker address/GPU data through an endpoint accessible to model-launch users (or otherwise align the permission) so worker selection also works under advanced authentication.

frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -87,7 +94,18 @@ export default function LaunchDialog({
   const quantizationValue = toOptionValue(useWatch('quantization', form));
   const multimodalProjectorValue = toOptionValue(useWatch('multimodal_projector', form));
   const nGpuValue = useWatch('n_gpu', form);
+  const [workerOptions, setWorkerOptions] = useState<WorkerOption[]>([]);
 
+  const fetchWorkers = useCallback(async () => {
+    try {
+      const data = await request.get<ClusterInfoResponse>('/v1/cluster/info', {
```
qinxuye, line 110: Catching the rejected promise does not prevent the global request interceptor from emitting the 403/forbidden notification first. A non-admin user with `models:write` will still see an authorization error every time this dialog opens and will never receive worker options. Please either avoid this request when the user lacks permission or provide a worker-list endpoint with a scope available to model-launch users; `creatable` can remain as a fallback, but it does not fix the permission mismatch.

frontend/src/components/ui/multi-select.tsx
```diff
@@ -233,13 +235,14 @@ export function MultiSelect({
 
       {open &&
         dropdownStyle &&
+        portalContainer &&
```
qinxuye, line 258: Portaling the dropdown back into `containerRef.current` makes it an absolutely positioned descendant of the select. In the launch dialog that descendant sits inside `dialog-body`, which has `overflow-y-auto`/`overflow-x-hidden`, so dropdowns near the bottom can be clipped. The direction calculation also still uses viewport space rather than the scroll container bounds. Please keep the dropdown outside clipping ancestors while still allowing the creatable input to participate in the dialog focus scope, or explicitly account for the nearest scroll container.

frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -87,7 +96,23 @@ export default function LaunchDialog({
   const quantizationValue = toOptionValue(useWatch('quantization', form));
   const multimodalProjectorValue = toOptionValue(useWatch('multimodal_projector', form));
   const nGpuValue = useWatch('n_gpu', form);
+  const [workerOptions, setWorkerOptions] = useState<WorkerOption[]>([]);
 
+  const fetchWorkers = useCallback(async () => {
+    if (!isAdmin) {
```
qinxuye: This also skips worker discovery when authentication is disabled. `useMenuAuth().isAdmin` is `false` for the `NO_AUTH` token because it has no JWT scopes, while `/v1/cluster/info` is intentionally unprotected in that mode. As a result, the default no-auth deployment always gets an empty worker list and loses the main purpose of this feature. Please skip the request only when authentication is enabled and the current user is not an admin, for example by checking `clusterAuth?.auth && !isAdmin`.

qinxuye (approved):
LGTM

PR #5224, opened 2026-07-22
ci: use merge-base diff for PR GPU tests

qinxuye:
## Summary

- use a merge-base (three-dot) diff when detecting GPU-relevant pull request changes
- preserve the existing two-point diff behavior for push events

## Root cause

The workflow compared the pull request base tip directly with the head tip. When the base branch advanced after the pull request branch diverged, files changed only on the base branch appeared in `changed-files.txt`. Shared API or core files then selected all GPU test groups even for a frontend-only pull request.

## Impact

Frontend-only pull requests no longer consume the GPU runner because of unrelated commits added to `main` after branch divergence.

## Validation

- replayed PR #5222 with its event base/head SHAs: detection changed from 18 files, including unrelated API/core files, to the correct 11 frontend-only files
- verified the push two-point range produces the same file set as the previous push comparison
- parsed `.github/workflows/python.yaml` with PyYAML
- ran `git diff --check`

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -79,9 +79,13 @@ jobs:
           if [ "${{ github.event_name }}" = "pull_request" ]; then
             base_sha="${{ github.event.pull_request.base.sha }}"
             head_sha="${{ github.event.pull_request.head.sha }}"
+            # Match GitHub's PR diff: exclude commits added to the base branch
+            # after the PR branch diverged.
+            diff_range="$base_sha...$head_sha"
           else
             base_sha="${{ github.event.before }}"
             head_sha="${{ github.sha }}"
+            diff_range="$base_sha..$head_sha"
           fi
 
           full=false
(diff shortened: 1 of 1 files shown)


rogercloud (approved):
Reviewed the diff and verified against the current workflow file:

- `fetch-depth: 0` is already set on the checkout step in this job (lines 67-71), so full history is available for the merge-base (three-dot) diff to work correctly.
- The change matches the PR description exactly: PR events now use `base_sha...head_sha` (merge-base diff), push events keep `base_sha..head_sha` (two-dot diff), and the final `git diff --name-only` call uses the new `$diff_range` variable.
- No other stale `git diff`/`base_sha`/`head_sha` references were found elsewhere in the file that would need the same fix.
- `"$diff_range"` containing `A...B` as a single quoted argument is valid git syntax, equivalent to the unquoted form.

Small, well-scoped CI-only fix with correct reasoning and no side effects on production code paths. LGTM.

PR #5225, opened 2026-07-23
feat(webui): multi-dashboard monitor center with admin config dialog

m199369309:
## Summary

This adds the frontend UI for the multi-dashboard monitor and the admin
configuration flow, building on the monitor_config backend that is already
on `main` (`monitor_config_store`, `GET/PUT /v1/cluster/monitor_config`,
`POST /v1/cluster/monitor_config/check-grafana`, `POST .../reset`, and the
`sources` value-origin map).

- **Monitor center tabs**: render one tab per configured Grafana dashboard
  UID (`overview` / `model_load` / `llm_slo` / `gpu` / `host` / `security`).
  The tab bar appears only when more than one dashboard is configured, so a
  cluster with a single `overview` UID keeps the current single-iframe UX.
- **Monitor config dialog** (admin-only): edit the Grafana URL (with a
  connection test against `/api/health`), datasource names, cluster name
  and per-tab dashboard UIDs, show value-source badges (`db` / `env` /
  `default`), and reset to defaults. Wired behind `useMenuAuth().isAdmin`.
- **Types/constants**: add optional `grafana_dashboards: Record<string,
  string>` to `ClusterUIConfig` and a `MONITOR_DASHBOARD_TABS` constant.
- **i18n**: add `en` / `zh` / `ko` / `ja` strings for the new tabs and the
  config dialog.

## Backward compatibility

- `grafana_dashboards` is optional; when absent the page falls back to the
  existing `grafana_dashboard_uid` (overview) behavior.
- No backend changes — consumes the existing API as-is.

## Verification

- `npx eslint` on the changed files: 0 errors / 0 warnings.
- `npm run build` (incl. `tsc`): succeeds; static export staged.


frontend/src/components/pages/monitor-center.tsx
```diff
@@ -196,18 +202,26 @@ const MonitorCenter = () => {
...
     [t]
   );
+  const dashboards = useMemo<Record<string, string>>(
+    () => clusterUIConfig.grafana_dashboards ?? {},
+    [clusterUIConfig.grafana_dashboards]
+  );
+  const visibleTabs = useMemo(
+    () => MONITOR_DASHBOARD_TABS.filter((tab) => dashboards[tab.key]),
+    [dashboards]
+  );
```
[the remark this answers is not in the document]
m199369309: Good catch, thanks. Fixed in dc611854: added a `useEffect` that resets `activeTab` to the first visible tab when the currently active one is no longer available.

frontend/src/components/pages/monitor-center.tsx
```diff
@@ -319,17 +333,51 @@ const MonitorCenter = () => {
...
+                  <TabsTrigger
+                    key={tab.key}
+                    value={tab.key}
+                    className="data-[state=active]:text-primary font-medium data-[state=active]:border-b-2 data-[state=active]:border-primary"
+                  >
+                    {t(tab.labelKey)}
+                  </TabsTrigger>
+                ))}
+              </div>
+            </TabsList>
```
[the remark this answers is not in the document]
m199369309: Applied in dc611854: removed the wrapper `div` and moved `flex gap-4` directly onto `TabsList`, keeping the `tablist` → `tab` ARIA structure intact.

frontend/src/components/pages/monitor-config-dialog.tsx
```diff
@@ -0,0 +1,295 @@
...
+    try {
+      const data = await request.get<MonitorConfigResponse>('/v1/cluster/monitor_config');
+      setForm({
+        grafana_url: data.grafana_url ?? '',
+        grafana_datasource: data.grafana_datasource ?? '',
+        grafana_alert_datasource: data.grafana_alert_datasource ?? '',
+        cluster_name: data.cluster_name ?? '',
+      });
+      setDashboards(data.grafana_dashboards ?? {});
+      setSources(data.sources ?? {});
```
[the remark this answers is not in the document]
m199369309: Intentionally skipping this one. The backend `get_monitor_config` always returns a FastAPI `JSONResponse` object, and the request layer rejects on non-2xx (handled by the surrounding `try`/`finally`), so `data` cannot be null/undefined inside the `try` block. The per-field `?? ''` / `?? {}` fallbacks already cover missing keys. Adding `data?.` would only mask a contract violation that cannot occur here.

frontend/src/components/pages/monitor-config-dialog.tsx
```diff
@@ -0,0 +1,295 @@
...
+    try {
+      const res = await request.post<{ ok: boolean; error?: string }>(
+        '/v1/cluster/monitor_config/check-grafana',
+        { grafana_url: form.grafana_url }
+      );
+      if (res.ok) {
+        setTest({ status: 'ok' });
+      } else {
+        setTest({ status: 'fail', msg: res.error });
+      }
```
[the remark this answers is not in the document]
m199369309: Same reasoning as the `GET` comment above: `check_grafana` always returns `{ ok, ... }`, and request failures reject before `res` is used, so `res?.` is not needed. Kept `res.ok` / `res.error` as-is.

frontend/src/components/pages/monitor-center.tsx
```diff
@@ -319,17 +340,49 @@ const MonitorCenter = () => {
             onChange={handleRefreshValue}
             allowClear={false}
           />
+          {isAdmin && (
```
qinxuye: Please keep the configuration entry reachable when Grafana is not configured and when authentication is disabled. `grafana_url` defaults to empty, but the component returns before this button is rendered, so the new dialog cannot bootstrap the first configuration and becomes inaccessible again after Reset. In no-auth mode, `useMenuAuth().isAdmin` is also always false because the stored token is `NO_AUTH`, while the backend deliberately leaves these endpoints open when auth is disabled. Please allow `clusterAuth.auth === false` here and expose the settings action in the not-configured state as well.

frontend/src/components/pages/monitor-center.tsx
```diff
@@ -196,18 +202,33 @@ const MonitorCenter = () => {
       })),
     [t]
   );
+  const dashboards = useMemo<Record<string, string>>(
+    () => clusterUIConfig.grafana_dashboards ?? {},
+    [clusterUIConfig.grafana_dashboards]
+  );
+  const visibleTabs = useMemo(
+    () => MONITOR_DASHBOARD_TABS.filter((tab) => dashboards[tab.key]),
```
qinxuye: This does not implement the stated single-dashboard/backward-compatible behavior against the current backend contract. `get_ui_config()` calls `MonitorConfigStore.get_dashboards()`, which fills all six keys with non-empty code defaults, so `visibleTabs` always contains six entries even when an existing Grafana only has the legacy overview dashboard. Clearing an input cannot disable a tab either: storing an empty string makes the store fall back to the same default UID on the next read. Old installations can therefore get five tabs pointing to dashboards they never imported. Please expose/consume an explicit enabled/configured set, or add explicit disable semantics, instead of treating fallback UIDs as configured.

frontend/src/components/pages/monitor-config-dialog.tsx
```diff
@@ -0,0 +1,295 @@
...
+      setTest({ status: 'fail', msg: detail });
+    }
+  }
+
+  async function handleSave() {
+    setSaving(true);
+    try {
+      await request.put('/v1/cluster/monitor_config', {
+        ...form,
+        grafana_dashboards: dashboards,
```
qinxuye: `grafana_dashboards` here contains resolved fallback values, not just the enabled dashboards. On a fresh or legacy configuration, GET returns all six default UIDs while `grafana_dashboards_configured` contains only `overview`; this full PUT persists all six as DB values, so even an unchanged save (or changing only the Grafana URL) marks every tab as configured and recreates the compatibility bug. I reproduced the configured keys changing from `['overview']` to all six after this round trip. Please track the configured set in the form and submit explicit empty values for disabled entries, rather than writing fallback UIDs back as configuration, and add a GET-to-PUT regression test.

m199369309: Fixed in 8177d211.

Root cause confirmed — exactly as described. The dialog loaded the six resolved (fallback) UIDs from `grafana_dashboards` and PUT them all back, so even an unchanged save persisted every UID and expanded `grafana_dashboards_configured` from `['overview']` to all six. I reproduced the `['overview'] → [all six]` expansion before the fix.

Changes:
- `load()` now only pre-fills UIDs for the keys in `grafana_dashboards_configured`; non-configured tabs stay empty, so fallback UIDs are no longer written back as configuration.
- `save()` submits all six keys explicitly — enabled tabs send their UID, disabled tabs send `""` so the backend clears the stale DB value and the tab falls back to `default`/not-configured. Enable/disable is simply non-empty vs. empty input.
- Added a GET→PUT regression test `test_put_unchanged_dashboards_does_not_expand_configured` (configured stays `['overview']` after a no-op save), plus `test_enable_then_disable_dashboard_via_empty_string`.

Verified: 26 monitor_config tests pass (3 new); eslint / tsc clean.

frontend/src/components/pages/monitor-config-dialog.tsx
```diff
@@ -0,0 +1,315 @@
...
+            >
+              {resetting && <Loader2 className="size-3.5 animate-spin" />}
+              {t('monitorCenter.config.reset')}
+            </Button>
+          </div>
+          <div className="flex gap-2">
+            <Button variant="outline" onClick={() => onOpenChange(false)}>
+              {t('monitorCenter.config.cancel')}
+            </Button>
+            <Button onClick={handleSave} disabled={saving}>
```
qinxuye: The dialog footer remains interactive while `load()` or `handleReset()` is in progress. On the first open, clicking Save before the GET completes submits the initial empty state and clears the stored configuration; during Reset, a concurrent Save can write the stale pre-reset values back. Please disable Save while `loading || resetting || saving` (and prevent Reset from racing with load/save as well) so these mutations cannot overlap.

m199369309: Fixed in 749afe88. Save button now disabled when `loading || saving || resetting`, Reset button disabled when `loading || resetting`, preventing the race condition where a concurrent Save could submit stale/empty state during load or reset.

qinxuye: This only fixes one direction of the race. The Reset button is still enabled while `saving` because its disabled condition is only `loading || resetting`. After clicking Save, the user can still confirm Reset and race `POST /reset` against the in-flight PUT, leaving the final configuration dependent on which request finishes last. Please include `saving` here as well: `disabled={loading || saving || resetting}`.

m199369309: Good point, fixed in b4ba576f. Reset button now also disabled while saving: disabled={loading || saving || resetting}.

qinxuye (approved):
LGTM

PR #5226, opened 2026-07-23
feat(webui): add auto-refresh to audit center

m199369309:
## Summary

Bring the audit records table to parity with the refresh UX already present
in the log center and monitor center:

- Add an auto-refresh interval selector that reuses the shared
  `LOG_REFRESH_OPTIONS` constant.
- Replace the text manual-refresh button with a tooltip-wrapped icon button
  (the manual refresh action is unchanged).

## Verification

- `npx eslint src/components/pages/audit-center/index.tsx`: 0 errors / 0 warnings.
- `npm run build` (incl. `tsc`): succeeds.

### DIFF
--- a/frontend/src/components/pages/audit-center/index.tsx
+++ b/frontend/src/components/pages/audit-center/index.tsx
@@ -203,7 +203,6 @@ export default function AuditCenter() {
   const [selectedRecord, setSelectedRecord] = useState<AuditRecord | null>(null);
   const requestSeqRef = useRef(0);
   const [refreshInterval, setRefreshInterval] = useState(0);
-  const refreshTimerRef = useRef<ReturnType<typeof setInterval> | null>(null);
 
   const queryParams = useMemo(() => {
     const params = new URLSearchParams();
@@ -223,9 +222,13 @@ export default function AuditCenter() {
     return params;
   }, [filters, pageFrom, timeRange]);
 
-  const fetchAuditRecords = useCallback(async () => {
+  const fetchAuditRecords = useCallback(async (silent = false) => {
     const seq = ++requestSeqRef.current;
-    setLoading(true);
+    // Background refreshes stay silent so the table does not flip to the
+    // loading spinner every interval; only user-initiated fetches show it.
+    if (!silent) {
+      setLoading(true);
+    }
     try {
       const data = await request.get<AuditSearchResponse>(
         '/v1/audit/search?' + queryParams.toString()
@@ -252,16 +255,10 @@ export default function AuditCenter() {
   }, [fetchAuditRecords]);
 
   useEffect(() => {
-    if (refreshTimerRef.current) {
-      clearInterval(refreshTimerRef.current);
-      refreshTimerRef.current = null;
-    }
     if (refreshInterval > 0) {
-      refreshTimerRef.current = setInterval(() => fetchAuditRecords(), refreshInterval);
+      const timer = setInterval(() => fetchAuditRecords(true), refreshInterval);
+      return () => clearInterval(timer);
     }
-    return () => {
-      if (refreshTimerRef.current) clearInterval(refreshTimerRef.current);
-    };
   }, [fetchAuditRecords, refreshInterval]);
 
   useEffect(() => {
@@ -373,7 +370,7 @@ export default function AuditCenter() {
                   variant="outline"
                   size="icon"
                   aria-label={t('auditCenter.refresh')}
-                  onClick={fetchAuditRecords}
+                  onClick={() => fetchAuditRecords()}
                   disabled={loading}
                 >
                   <RefreshCw className={cn('size-4', loading && 'animate-spin')} />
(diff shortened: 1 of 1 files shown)


frontend/src/components/pages/audit-center/index.tsx
```diff
@@ -201,6 +202,8 @@ export default function AuditCenter() {
   const [pageFrom, setPageFrom] = useState(0);
   const [selectedRecord, setSelectedRecord] = useState<AuditRecord | null>(null);
   const requestSeqRef = useRef(0);
+  const [refreshInterval, setRefreshInterval] = useState(0);
+  const refreshTimerRef = useRef<ReturnType<typeof setInterval> | null>(null);
```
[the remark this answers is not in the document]
m199369309: Agreed — removed `refreshTimerRef` in 1e462d35 and switched the effect to React's built-in cleanup (`return () => clearInterval(timer)`).

frontend/src/components/pages/audit-center/index.tsx
```diff
@@ -248,6 +251,19 @@ export default function AuditCenter() {
...
+      clearInterval(refreshTimerRef.current);
+      refreshTimerRef.current = null;
+    }
+    if (refreshInterval > 0) {
+      refreshTimerRef.current = setInterval(() => fetchAuditRecords(), refreshInterval);
+    }
+    return () => {
+      if (refreshTimerRef.current) clearInterval(refreshTimerRef.current);
+    };
+  }, [fetchAuditRecords, refreshInterval]);
```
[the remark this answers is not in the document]
m199369309: Applied in 1e462d35:
- Simplified the effect to use the `useEffect` cleanup.
- Added a `silent` flag to `fetchAuditRecords`; background refreshes call it with `silent = true` so they no longer set `loading`, which fixes the table flicker. (Also wrapped the manual button's `onClick` so the click event isn't passed as the `silent` argument.)
- On the race-condition note: this is already handled by the existing `requestSeqRef` sequence guard — `fetchAuditRecords` only applies results from the latest request — so no additional guard is needed.

qinxuye (commented):
Two auto-refresh reliability issues need to be addressed.

frontend/src/components/pages/audit-center/index.tsx
```diff
@@ -248,6 +254,13 @@ export default function AuditCenter() {
     fetchAuditRecords();
   }, [fetchAuditRecords]);
 
+  useEffect(() => {
+    if (refreshInterval > 0) {
+      const timer = setInterval(() => fetchAuditRecords(true), refreshInterval);
```
qinxuye: This interval can start a new search while the previous one is still pending. Since every call increments `requestSeqRef`, if request latency exceeds the selected interval (for example, a 12s query with 10s polling), every response is stale by the time it completes and the table never updates; slow or hung requests can also pile up. Please serialize polling (schedule the next timeout after the current request settles) or skip a tick while a request is in flight.

frontend/src/components/pages/audit-center/index.tsx
```diff
@@ -220,9 +222,13 @@ export default function AuditCenter() {
     return params;
   }, [filters, pageFrom, timeRange]);
 
-  const fetchAuditRecords = useCallback(async () => {
+  const fetchAuditRecords = useCallback(async (silent = false) => {
```
qinxuye, line 228: For silent background refreshes, a transient request failure currently falls through to the existing `catch` and clears `records` and `total`. That makes valid existing results disappear even though only the refresh failed. Please retain the current data when `silent` is true; the foreground query can keep the existing behavior if needed.

frontend/src/components/pages/audit-center/index.tsx
m199369309: Fixed in aa313759.

Added an in-flight guard. `fetchAuditRecords` sets `inFlightRef = true` at start and clears it in `finally` (only when this is still the latest request, via the existing `requestSeqRef` check). The poll tick now skips while a request is in flight:

```js
const timer = setInterval(() => {
  if (!inFlightRef.current) fetchAuditRecords(true);
}, refreshInterval);
```

This prevents overlapping/stacked requests and keeps the in-flight request as the latest sequence, so its result is applied when it settles — no more stale-discard loop when latency exceeds the interval, and no request pile-up.

frontend/src/components/pages/audit-center/index.tsx
m199369309: Fixed in aa313759.

The `catch` now only clears `records`/`total` for foreground (`!silent`) fetches; silent background refresh failures retain the current data:

```js
} catch {
  if (!silent && seq === requestSeqRef.current) {
    setRecords([]);
    setTotal(0);
  }
}
```

A transient refresh failure no longer wipes the table; the foreground (user-initiated) query keeps the original clear-on-error behavior.

qinxuye (approved):
LGTM

PR #5227, opened 2026-07-23
ENH: add #system_pandas# to virtualenv system marker whitelist

OliverBryant:
## What do these changes do?

`filter_virtualenv_packages_by_markers` keeps `#system_*#` placeholders intact (for xoscar to resolve later) via a hardcoded whitelist, but the whitelist only covered the torch-family and numpy placeholders. Model specs already use `#system_pandas#` (7 occurrences in the audio model spec), which currently falls through to the generic marker path. That path happens to produce an equivalent result today, but relies on incidental behavior — add pandas to the whitelist so all placeholders used in model specs are handled consistently, and add a regression test.

Related: xorbitsai/xoscar#194 makes `#system_*#` resolution degrade gracefully when the host environment lacks the package (relevant for minimal-install 3.0 hosts). Once that lands in a release, the `xoscar` version requirement in `setup.cfg` should be bumped accordingly.

## Check code requirements

- [x] tests added / passed (`xinference/core/tests/test_utils.py`, 23 passed)
- [x] passes `black --check` / `isort --check-only`

### DIFF
--- a/xinference/core/tests/test_utils.py
+++ b/xinference/core/tests/test_utils.py
@@ -15,6 +15,7 @@
 from ..utils import (
     build_replica_model_uid,
     build_subpool_envs_for_virtual_env,
+    filter_virtualenv_packages_by_markers,
     iter_replica_model_uid,
     parse_replica_model_uid,
 )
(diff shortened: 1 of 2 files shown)


OliverBryant:
Superseded by #5228 (same change, resubmitted from a properly named branch).

PR #5228, opened 2026-07-23
ENH: adopt xoscar 0.9.7 graceful #system_*# fallback

OliverBryant:
## What do these changes do?

Two related changes for minimal-install (3.0) hosts where numpy/torch/pandas may be absent from the main environment:

1. **Require `xoscar>=0.9.7`** (pyproject.toml). xoscar 0.9.7 ships xorbitsai/xoscar#195: `#system_*#` placeholders degrade gracefully when the host environment lacks the package (installed unpinned with a warning instead of raising `RuntimeError` and failing the model launch), markers are evaluated before placeholder resolution, and host-aligned pins that conflict with other requirements (e.g. a model pinning a newer engine) are dropped on a one-shot retry — including in `skip_installed` mode.

2. **Add `#system_pandas#` to the system marker whitelist** in `filter_virtualenv_packages_by_markers` (`xinference/core/utils.py`). Model specs already use `#system_pandas#` (7 occurrences in the audio model spec), but the whitelist only covered the torch-family and numpy placeholders; the generic marker path it fell through to happens to produce an equivalent result today, but relies on incidental behavior. Adds a regression test.

## Check code requirements

- [x] tests added / passed (`xinference/core/tests/test_utils.py`)
- [x] passes `black --check` / `isort --check-only`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
-    "xoscar>=0.9.6",
+    "xoscar>=0.9.7",
     "torch",
     "pillow",
     "click<8.2.0",
--- a/xinference/core/tests/test_utils.py
+++ b/xinference/core/tests/test_utils.py
@@ -15,6 +15,7 @@
 from ..utils import (
     build_replica_model_uid,
     build_subpool_envs_for_virtual_env,
+    filter_virtualenv_packages_by_markers,
     is_valid_model_uid,
     iter_replica_model_uid,
     merge_virtual_env_packages,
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -32,7 +32,7 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
-    "xoscar>=0.9.6",
+    "xoscar>=0.9.7",
```
qinxuye, line 35: Please update `xinference/deploy/docker/requirements_cpu/requirements_cpu-base.txt` to `xoscar>=0.9.7` as well. `Dockerfile.cpu` installs xoscar from that file and later installs Xinference with `pip install --no-deps "."`, so this `pyproject.toml` lower bound is not enforced for the CPU image. With a mirror that has not synced 0.9.7, the image can build successfully with an older xoscar and then miss the graceful `#system_*#` fallback this PR relies on.

OliverBryant: Good point — verified that `Dockerfile.cpu` installs from `requirements_cpu-base.txt` first and then uses `pip install --no-deps "."`, so the `pyproject.toml` bound indeed doesn't apply there. Bumped the file to `xoscar>=0.9.7` in 914d68b69.

qinxuye (approved):

PR #5229, opened 2026-07-23
feat(webui): show API key name and model ID columns in audit center

m199369309:
## Why

The audit center table only surfaced `api_key_name` and `model_id` as fallback values — `api_key_name` hidden inside the **Identity** column (`user || api_key_name`) and `model_id` inside the **Model** column (`model_name || model_id`). Neither was directly visible, so operators couldn't see which API key or which model UID a request used without opening the row detail.

The backend already records and returns both fields (`record_audit_event` in `xinference/api/oauth2/advanced/audit.py`, exposed via `/v1/audit/search` and the `api_key_name` / `model_id` filter params), so this is purely a frontend display gap.

## What

- Add dedicated **API Key Name** and **Model ID** columns to the audit table (column count 7 → 9; `colSpan` updated on the loading / empty states).
- Add `auditCenter.apiKeyNameColumn` / `auditCenter.modelIdColumn` to `zh.ts` and `en.ts` (kept separate from the existing `apiKeyName` / `modelId` **filter-label** keys to avoid overloading their meaning).
- Let the table grow to its content width (`w-max min-w-full`) and switch the **Endpoint / API Key Name / Model ID** cells from `truncate` to `whitespace-nowrap`. The 32-char model UID (and long endpoint URLs / key names) was being clipped to ~10 chars, making the audit trail useless for tracing; the table's built-in `overflow-auto` now scrolls horizontally instead. Fixed-width columns are unchanged.

## Notes

- No backend changes — the fields were already recorded / returned / filterable.
- No interaction with #5226 (audit center auto-refresh); this PR is built directly on `main` and contains no auto-refresh code.


frontend/src/i18n/locales/en.ts
```diff
@@ -552,7 +552,9 @@ const en = {
     category: 'Category',
     identity: 'Identity',
     endpoint: 'Endpoint',
+    apiKeyNameColumn: 'API Key Name',
```
qinxuye, line 555: These new column keys are only added to `en.ts` and `zh.ts`. `ja.ts` and `ko.ts` are also supported locales, and `t()` has no fallback for a missing key, so the Japanese and Korean Audit Center will render the raw `auditCenter.apiKeyNameColumn` / `auditCenter.modelIdColumn` strings. Please add the corresponding keys to both locales.

qinxuye (commented):
One localization issue needs to be addressed.

frontend/src/i18n/locales/en.ts
m199369309: Fixed in 6f2c97b7.

Added both keys to ja.ts and ko.ts, in the column-header group to match en.ts:
- ja: `apiKeyNameColumn` 'API キー名', `modelIdColumn` 'モデル ID'
- ko: `apiKeyNameColumn` 'API Key 이름', `modelIdColumn` '모델 ID'

All four locales now define the keys, so `t()` no longer falls back to the raw dotted string for Japanese/Korean. eslint / tsc clean.

qinxuye (approved):
LGTM

PR #5230, opened 2026-07-23
feat: add DeepDoc (ragflow) OCR model as builtin image model

kejhz653-stack:
- Vendor ragflow's deepdoc/vision ONNX inference code (Apache-2.0, commit 6467df34) into xinference/thirdparty/deepdoc with ragflow-internal imports replaced by local shims; no execution provider is hard-coded (CUDA probed via onnxruntime providers).
- Add DeepDocModel adapter supporting task=ocr/layout/table on the new 'deepdoc' engine (onnxruntime).
- Register spec: HF InfiniFlow/deepdoc pinned to de0e793d, portable virtualenv packages only.
- Default OCR engine falls back to the model's sole engine when transformers is not supported.

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -315,13 +315,34 @@ jobs:
         run: |
           set -e
 
+          # On a failure, retry only the failed tests once via --last-failed
+          # (same pattern as the GPU job): the suite carries a few known
+          # intermittent failures (KV-cache batching race, cluster-startup
+          # timeouts on loaded runners) that pass on a clean rerun, while a
+          # real regression still fails twice and turns the job red.
+          # --cov-append keeps the coverage of the first run instead of
+          # overwriting coverage.xml with the retried subset only.
           run_pytest() {
+            local rc=0
             pytest --timeout=3000 \
               -W ignore::PendingDeprecationWarning \
               --cov-config=pyproject.toml \
               --cov-report=xml \
               --cov=xinference \
-              "$@"
+              "$@" || rc=$?
+            if [ "$rc" -ne 0 ]; then
+              echo "::warning::pytest exited with code $rc; retrying failed tests once"
+              rc=0
+              pytest --timeout=3000 \
+                -W ignore::PendingDeprecationWarning \
+                --cov-config=setup.cfg \
+                --cov-report=xml \
+                --cov=xinference \
+                --cov-append \
+                --last-failed \
+                "$@" || rc=$?
+            fi
+            return $rc
           }
 
           if [ "$MODULE" == "metal" ]; then
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -315,34 +315,13 @@ jobs:
         run: |
           set -e
 
-          # On a failure, retry only the failed tests once via --last-failed
-          # (same pattern as the GPU job): the suite carries a few known
-          # intermittent failures (KV-cache batching race, cluster-startup
-          # timeouts on loaded runners) that pass on a clean rerun, while a
-          # real regression still fails twice and turns the job red.
-          # --cov-append keeps the coverage of the first run instead of
-          # overwriting coverage.xml with the retried subset only.
           run_pytest() {
-            local rc=0
             pytest --timeout=3000 \
               -W ignore::PendingDeprecationWarning \
               --cov-config=pyproject.toml \
               --cov-report=xml \
               --cov=xinference \
-              "$@" || rc=$?
-            if [ "$rc" -ne 0 ]; then
-              echo "::warning::pytest exited with code $rc; retrying failed tests once"
-              rc=0
-              pytest --timeout=3000 \
-                -W ignore::PendingDeprecationWarning \
-                --cov-config=setup.cfg \
-                --cov-report=xml \
-                --cov=xinference \
-                --cov-append \
-                --last-failed \
-                "$@" || rc=$?
-            fi
-            return $rc
+              "$@"
           }
 
           if [ "$MODULE" == "metal" ]; then
--- a/xinference/model/image/core.py
+++ b/xinference/model/image/core.py
@@ -174,7 +174,23 @@ def create_ocr_model_instance(
         model_path = cache_manager.cache()
 
     if model_engine is None:
-        model_engine = "transformers"
+        from .ocr.ocr_family import OCR_ENGINES
+
+        # Default to transformers, but fall back to the model's sole engine
+        # for models that don't support transformers (e.g. DeepDoc).
+        available_engines = OCR_ENGINES.get(model_spec.model_name, {})
+        if "transformers" in available_engines or not available_engines:
+            model_engine = "transformers"
+        else:
+            model_engine = next(iter(available_engines))
+            if len(available_engines) > 1:
+                logger.warning(
+                    "Multiple engines available for model %s (%s); "
+                    "defaulting to %s. Pass model_engine to choose explicitly.",
+                    model_spec.model_name,
+                    ", ".join(available_engines),
+                    model_engine,
+                )
 
     model_format = getattr(model_spec, "model_format", None)
     quantization = getattr(model_spec, "quantization", None)
--- a/xinference/model/image/core.py
+++ b/xinference/model/image/core.py
@@ -155,6 +155,32 @@ def match_diffusion(
         )
 
 
+def is_cpu_only_image_model(
+    model_name: str, model_engine: Optional[str] = None
+) -> bool:
+    """Whether every candidate engine for this image model runs on CPU only.
+
+    The worker consults this before reserving devices so that CPU-only
+    integrations (e.g. DeepDoc's onnxruntime backend, which ignores the
+    assigned device) never reserve a GPU that inference won't touch.
+    Unknown models/engines return False and keep the default GPU behavior.
+    """
+    from .ocr.ocr_family import OCR_ENGINES
+
+    engines = OCR_ENGINES.get(model_name)
+    if not engines:
+        return False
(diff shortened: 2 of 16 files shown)


OliverBryant (commented):
Thanks for this — the vendoring is well documented and the adapter is clean. I confirmed that all of the earlier Gemini findings (device_id threading, the empty-span IndexError guard, and the IoU/NMS epsilon) are already incorporated in this diff, so those are resolved.

A few non-blocking notes below. The most worthwhile one is the `layout` threshold behavior.

**1. `threshold` is silently ignored for `task=layout` (vendored code).**
The adapter docstring advertises `threshold` (default 0.2) for layout/table. For the layout task the chain is `LayoutRecognizer.forward(thr=threshold)` -> `Recognizer.__call__` -> `LayoutRecognizer4YOLOv10.postprocess`, whose first line is `thr = 0.08`, overwriting the caller's value. This is inherited from upstream ragflow, but since the adapter exposes `threshold` as configurable, it would be good to either drop it from the layout docstring or note that it is fixed for the YOLOv10 layout model. (The table task honors `threshold` correctly.)

**3. Latent trap in vendored `OCR.detect/recognize/recognize_batch`.**
These index `self.text_detector[device_id]` / `self.text_recognizer[device_id]`, but both lists always hold exactly one element, so a non-zero `device_id` would raise IndexError. Not reachable through the adapter today (only `OCR.__call__` is used, which hardcodes index 0), so low priority — just flagging it as a latent issue in the vendored copy.

**4. Spec pins no versions.** `onnxruntime`, `opencv-python-headless`, etc. are unversioned. Consistent with some existing specs (e.g. DeepSeek-OCR), so optional, but a light lower-bound pin would help reproducibility in fresh virtualenvs.

Overall LGTM with these minor comments.

xinference/model/image/core.py
```diff
@@ -174,7 +174,15 @@ def create_ocr_model_instance(
...
-        model_engine = "transformers"
+        from .ocr.ocr_family import OCR_ENGINES
+
+        # Default to transformers, but fall back to the model's sole engine
+        # for models that don't support transformers (e.g. DeepDoc).
+        available_engines = OCR_ENGINES.get(model_spec.model_name, {})
+        if "transformers" in available_engines or not available_engines:
+            model_engine = "transformers"
+        else:
+            model_engine = next(iter(available_engines))
```
OliverBryant, line 185: Minor: this picks the "sole engine" via dict insertion order. It's correct for DeepDoc (single engine) and the comment says so, but if a non-transformers model ever gains 2+ engines this would silently pick an arbitrary one. Consider asserting `len(available_engines) == 1` in this branch, or otherwise making the single-engine assumption explicit.

xinference/model/image/tests/test_deepdoc.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # default task: plain text
+    r = model.ocr(image=bio.getvalue())
+    assert isinstance(r, str)
+
+    # layout task returns JSON
+    r = model.ocr(image=bio.getvalue(), task="layout")
+    payload = json.loads(r)
+    assert payload["task"] == "layout"
+    assert isinstance(payload["layouts"], list)
```
OliverBryant, line 223: The integration test covers `ocr` and `layout` but not `task="table"` or the `return_dict=True` path. Adding at least a `table` assertion would round out coverage of the three advertised tasks.

OliverBryant (approved):
All review feedback has been addressed:

- `threshold` docstring now clarifies it only applies to `table`; `layout` uses the upstream YOLOv10 fixed threshold.
- Multi-engine default fallback now logs a warning.
- Vendored `OCR.detect/recognize/recognize_batch/__call__` clamp `device_id` to 0 with an explanatory comment, removing the latent IndexError.
- Spec dependencies now carry lower-bound pins.
- Tests now cover `return_dict=True` and `task="table"` in addition to `ocr`/`layout`.

The vendoring is clearly documented and the adapter follows project conventions. LGTM.

qinxuye (changes requested):
No need to use third-party.

xinference/model/image/ocr/deepdoc.py
```diff
@@ -0,0 +1,216 @@
...
+        images = [image] if single else list(image)
+        if any(img is None for img in images):
+            raise ValueError("Input image list cannot contain None.")
+        results = [self._process_single(img, task, kwargs) for img in images]
+
+        if task == "ocr" and not return_dict:
+            texts = [
+                "\n".join(line["text"] for line in res["lines"]) for res in results
+            ]
+            return texts[0] if single else texts
```
qinxuye, line 178: The default path returns raw text here, but both REST clients unconditionally call `response.json()` after `/v1/images/ocr`. Normal OCR text will therefore raise `JSONDecodeError` in `Client`/`AsyncClient`. The integration test does not catch this because it is skipped when `deepdoc` is absent from the host test environment. Please make the endpoint/client response contract consistent and add an unskipped regression test.

.github/workflows/python.yaml
```diff
@@ -312,13 +312,34 @@ jobs:
         run: |
           set -e
 
+          # On a failure, retry only the failed tests once via --last-failed
```
qinxuye: The global pytest retry, shared fixture refactors, and log-rotation implementation/tests are unrelated to adding DeepDoc and should be removed from this PR or split into a separate PR. They substantially broaden the blast radius, and the added log-rotation test currently causes the Windows 3.13 check to fail. Please keep this PR scoped to the DeepDoc adapter, registration, docs, and relevant tests.

xinference/model/image/model_spec.json
```diff
@@ -1748,5 +1748,30 @@
...
+  {
+    "version": 2,
+    "model_name": "DeepDoc",
+    "model_family": "ocr",
+    "model_ability": [
+      "ocr"
+    ],
+    "virtualenv": {
+      "packages": [
+        "deepdoc-lib~=0.2.2"
```
qinxuye: `deepdoc-lib~=0.2.2` installs the CPU `onnxruntime` dependency by default, while image launches default to reserving a GPU and this adapter ignores the assigned device. On GPU workers this can reserve a GPU while inference still runs on CPU. Please either make this integration explicitly CPU-only without GPU allocation, or provide a supported GPU dependency/device path and cover it with a focused test.

qinxuye (commented):
Reviewing the current DeepDoc integration. The existing resolved threads are not repeated below.

qinxuye:
There is conflict.

OliverBryant:
Found a blocking GPU scheduling issue in the CPU-only follow-up. DeepDoc is not CPU-only on Linux x86_64:

- `deepdoc-lib` installs `onnxruntime-gpu==1.19.2` on Linux x86_64.
- `deepdoc.vision.ocr.load_model()` selects `CUDAExecutionProvider` whenever CUDA is available and accepts a `device_id`.
- OCR initializes sessions for the visible CUDA devices; layout/table use the same GPU-aware `load_model()`.

Marking `DeepDocModel.cpu_only = True` makes Xinference skip GPU allocation and CUDA visibility isolation, while DeepDoc can still consume GPU 0 (or every visible GPU) behind the scheduler's back. This can cause untracked VRAM use and model contention.

I am fixing this by removing the CPU-only bypass, restoring normal Xinference GPU reservation/CUDA_VISIBLE_DEVICES isolation, and replacing the CPU-only regression tests with GPU-capability/scheduling coverage. I will then launch DeepDoc on the dual-3090 Ti test host and verify the active ONNX provider and VRAM.

OliverBryant:
Implemented and pushed the DeepDoc GPU fixes to the PR branch.

Runtime findings and fixes:
- Restored standard worker GPU allocation; DeepDoc is CUDA-capable and must not be marked CPU-only.
- Split virtualenv requirements using xoscar extended markers: `deepdoc-lib[gpu]~=0.2.2 ; has_cuda` and `deepdoc-lib~=0.2.2 ; not has_cuda`.
- Pinned `transformers<5` after a real isolated-environment launch exposed a Transformers 5 / huggingface-hub incompatibility.
- Fixed the `onnxruntime` / `onnxruntime-gpu` namespace collision from deepdoc-lib metadata by reinstalling only the GPU runtime after dependency resolution.

Verified on a 2x RTX 3090 Ti host using exact PR head `d464dcc5`:
- ORT providers include `TensorrtExecutionProvider`, `CUDAExecutionProvider`, and `CPUExecutionProvider`.
- `nvidia-smi` reports the DeepDoc model process and 404 MiB after OCR inference.
- An actual OCR request succeeded and returned two recognized lines.
- DeepDoc suite: 7 passed / 1 skipped; pre-commit passed.

qinxuye (approved):
LGTM

qinxuye:
Two things,

1. where is the performance number?
2. Xorbits/deepdoc, for huggingface repo, we need to use this.

OliverBryant:
Performance numbers, tested on a dual-RTX-3090-Ti host with a real 31-page PDF (scanned academic report, mixed text/table layout) via `/v1/images/ocr`:

| Engine | Result | Time |
|---|---|---|
| CPU | success, all 31 pages recognized | 130.6s |
| GPU (`OCR_GPU_MEM_LIMIT_MB=8192`) | success, all 31 pages recognized | 27.3s |

GPU is ~4.8x faster than CPU on this document.

Note: `deepdoc-lib`'s default `OCR_GPU_MEM_LIMIT_MB` (2048MB) is too tight for a 31-page batch and can fail with an ONNX Runtime CUDA arena allocation error even with plenty of free VRAM on the card — worth bumping via the `envs` launch parameter for GPU deployments handling multi-page documents, e.g. `{"OCR_GPU_MEM_LIMIT_MB": "8192"}`.

Issue #5237, opened 2026-07-25
Some warning log in cpu docker container


leslie2046:
```
Failed to load /opt/conda/lib/python3.11/site-packages/torchao/_C_cutlass_90a.abi3.so: Could not load this library: /opt/conda/lib/python3.11/site-packages/torchao/_C_cutlass_90a.abi3.so
Failed to load /opt/conda/lib/python3.11/site-packages/torchao/_C_mxfp8.cpython-310-x86_64-linux-gnu.so: Could not load this library: /opt/conda/lib/python3.11/site-packages/torchao/_C_mxfp8.cpython-310-x86_64-linux-gnu.so
W0725 03:04:42.333000 1 site-packages/torch/utils/_pytree.py:630] <enum 'KernelPreference'> is an Enum subclass and is now natively supported by torch.compile as an opaque value type. Calling register_constant() on Enum subclasses is deprecated and will be an error in a future release.
W0725 03:04:43.004000 1 site-packages/torch/utils/_pytree.py:630] <enum 'ScaleCalculationMode'> is an Enum subclass and is now natively supported by torch.compile as an opaque value type. Calling register_constant() on Enum subclasses is deprecated and will be an error in a future release.
{"@timestamp": "2026-07-25T03:04:55.950Z", "level": "INFO", "module": "xinference.core.worker", "pid": 1, "role": "worker", "address": "192.168.1.77:29395", "node": "njue-poweredge-r730", "message": "Starting metrics export server at 192.168.1.77:31001"}
{"@timestamp": "2026-07-25T03:04:55.951Z", "level": "INFO", "module": "xinference.core.worker", "pid": 1, "role": "worker", "address": "192.168.1.77:29395", "node": "njue-poweredge-r730", "message": "Checking metrics export server..."}
{"@timestamp": "2026-07-25T03:04:57.006Z", "level": "INFO", "module": "xinference.core.worker", "pid": 1, "role": "worker", "address": "192.168.1.77:29395", "node": "njue-poweredge-r730", "message": "Metrics server is started at: http://192.168.1.77:31001"}
{"@timestamp": "2026-07-25T03:04:57.006Z", "level": "INFO", "module": "xinference.core.worker", "pid": 1, "role": "worker", "address": "192.168.1.77:29395", "node": "njue-poweredge-r730", "message": "Purge cache directory: /data/cache"}
[...]

PR #5238, opened 2026-07-25
FIX: suppress GPU warnings in CPU image

leslie2046:
Partially addresses #5237

** Summary**

- Skip NVIDIA GPU orphan cleanup when no CUDA devices are detected.
- Add tests covering startup cleanup behavior without NVIDIA GPUs.

### DIFF
--- a/xinference/deploy/docker/Dockerfile.cpu
+++ b/xinference/deploy/docker/Dockerfile.cpu
@@ -14,22 +14,12 @@ RUN apt-get -y update \
   && apt install -y build-essential curl procps git libgl1 \
   && apt-get -yq clean
 
-# GPTQModel requires torchao.  torchao 0.16.0 also publishes a cp310-abi3
-# wheel containing optional CUDA extensions; pip prefers it on Python 3.11,
-# even in this CPU-only image.  Select the pure-Python wheel explicitly so
-# importing torchao does not try to load those unusable CUDA libraries.
 ARG PIP_INDEX=https://pypi.org/simple
-ARG TORCHAO_VERSION=0.16.0
 COPY xinference/deploy/docker/requirements_cpu/ /opt/inference/xinference/deploy/docker/requirements_cpu/
 RUN --mount=type=cache,target=/root/.cache/pip \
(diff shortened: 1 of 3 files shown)


xinference/deploy/docker/Dockerfile.cpu
```diff
@@ -14,12 +14,22 @@ RUN apt-get -y update \
   && apt install -y build-essential curl procps git libgl1 \
   && apt-get -yq clean
 
+# GPTQModel requires torchao.  torchao 0.16.0 also publishes a cp310-abi3
+# wheel containing optional CUDA extensions; pip prefers it on Python 3.11,
+# even in this CPU-only image.  Select the pure-Python wheel explicitly so
+# importing torchao does not try to load those unusable CUDA libraries.
 ARG PIP_INDEX=https://pypi.org/simple
+ARG TORCHAO_VERSION=0.16.0
```
qinxuye: Pinning only `torchao` to 0.16.0 is unsafe because this Dockerfile installs an unpinned latest `torch`. `torchao` 0.16.0 only treats PyTorch 2.10.0/2.11.0-dev as compatible, so a fresh build with a newer PyTorch will still emit an incompatible-version warning and may expose API incompatibilities; this does not reliably fix the CPU image. Please either pin a tested PyTorch/torchao pair, or let GPTQModel resolve torchao and then force-reinstall the `py3-none-any` wheel for that resolved version. An image-level `python -c "import torchao"` smoke check would also prevent this from regressing.

qinxuye:
The latest commit resolves the torch/torchao compatibility concern by reverting the `Dockerfile.cpu` change, and the remaining worker-side change looks good for the narrower NVML-warning scope. However, this PR no longer fixes the torchao `.so` warnings reported in #5237. Please replace `fixes #5237` with `Partially addresses #5237` and remove the summary bullet claiming that the pure-Python torchao wheel is installed, unless the remaining warnings will also be fixed in this PR. Otherwise merging this PR would close #5237 while part of the reported problem still remains.

leslie2046:
OK

qinxuye (approved):
LGTM

PR #5239, opened 2026-07-25
ENH: display replica launch status in responsive grid

leslie2046:
## Summary

- display replica launch statuses in a responsive grid
- support 1–3 columns based on viewport width
- constrain popup height and enable internal scrolling
- truncate long worker addresses with full values available on hover

## Validation

- Prettier passed
- ESLint passed
- frontend production build passed


qinxuye (approved):
LGTM

PR #5241, opened 2026-07-25
feat: speculative decoding with paired drafter checkpoints

qinxuye:
Gemma 4 ships a small `*-it-assistant` drafter for every variant: it proposes several tokens per round that the target verifies in one pass, so decoding gets faster with identical output. Unlike DeepSeek-V3 style MTP the drafter is a **separate checkpoint** that shares the target's KV cache, so it has to be declared, downloaded and handed to the engine.

Launch with `--enable_mtp true`, or flip the switch under *Advanced Configuration → Speculative Decoding* in the Web UI.

## Registry and cache

Specs gain `draft_model_id`, optionally templated by `draft_quantizations` — the MLX conversions of Gemma 4 12B publish eight, the gguf drafters three. `LLMCacheManager` grows a `use_draft_model` mode that reuses the existing per-hub download paths, so a drafter follows the same HuggingFace/ModelScope selection as its target and lands in a sibling `<cache_dir>-draft-<quantization>` directory. Every drafter repository referenced here was checked to exist on both hubs.

`create_llm_model_instance` resolves the drafter when `enable_mtp` is passed and hands the local path to the engine. Engines that cannot use one are rejected up front via `LLM.support_draft_model` rather than failing somewhere inside engine startup.

Deleting a model's cache now removes its drafters too, instead of orphaning a directory per drafter quantization next to it.

## Per engine

| Engine | Requirement | Translation |
|---|---|---|
| vLLM | `vllm>=0.22.0` | `speculative_config` with `method: mtp` |
| SGLang | — | `--speculative-algorithm NEXTN` plus matching step/draft-token counts |
| MLX | `mlx-vlm>=0.5.0` (`>=0.6.1` for 12B) | `draft_model` on `generate_step`, kind auto-detected from the drafter's `model_type` |
| llama.cpp | `xllamacpp>=2026.6.9713` | the `draft-mtp` speculative implementation |

An explicitly provided `speculative_config` / `speculative_algorithm` always wins; the engine-neutral options are consumed either way so they never reach the engine as unknown arguments. vLLM below 0.22.0 is rejected with a readable error because it treats the assistant as a generic draft model and fails to initialize against a multimodal target. Same for xllamacpp below 2026.6.9713, which cannot load the `gemma4-assistant` architecture at all.

The Transformers engine is **not** supported: it runs its own continuous-batching loop rather than `generate()`, so there is nowhere to attach a drafter.

MLX loads the drafter on the same dedicated MLX thread as the target and validates the pair when the model loads, rather than on the first request.

## gemma-4 12B

Also registers gemma-4 12B across pytorch/gguf/mlx. It is the encoder-free Unified variant (`Gemma4UnifiedForConditionalGeneration`, drafter `gemma4_unified_assistant`), which is why the family now lists two architectures. It is also the only variant whose drafter is published in more than one conversion, so it is what makes the drafter selection meaningful.

## Web UI

The Speculative Decoding section only appears for a format/size that actually declares a drafter. The drafter options carry the same cached marker the model quantizations use, because turning the switch on downloads another checkpoint. Tooltips cover the non-guessable parts: that the switch costs a download, why an unquantized drafter is recommended even for a quantized target, and that a value above the drafter's trained depth acts as a ceiling rather than a fixed block size.

One unrelated fix rides along, because it blocks anyone verifying this in the dev UI: without `NEXT_PUBLIC_API_URL` the browser sends API requests to the Next dev server, whose rewrite proxy buffers the body, so streaming endpoints deliver the whole answer at once. `.env.development` points the browser at the backend directly; the production static export is unaffected (Next does not load that file for builds).

## Verification

Measured end to end on Apple silicon with gemma-4 E4B, 8bit target and a bf16 drafter: **2.2x on decode** in isolation, ~15% on a full chat request. Driving `generate_step` directly shows the expected burst pattern — tokens arrive in groups the size of the draft block, and streaming is preserved.

Unit tests cover the config translation for all four engines, the drafter selection and cache-directory rules, and the cascade delete. Not covered: an actual vLLM/SGLang/llama.cpp run, for lack of the hardware — those three are wired from their documented parameters and tested at the config-assembly level only.


xinference/core/worker.py
```diff
@@ -4087,6 +4099,28 @@ async def list_deletable_models(self, model_version: str) -> List[str]:
...
+                if os.path.isdir(draft_path):
+                    for file in os.listdir(draft_path):
+                        draft_file = os.path.join(draft_path, file)
+                        paths.add(draft_file)
+                        if os.path.exists(
+                            (real_draft_file := os.path.realpath(draft_file))
+                        ):
+                            paths.add(real_draft_file)
+                if os.path.exists((real_draft_path := os.path.realpath(draft_path))):
+                    paths.add(real_draft_path)
```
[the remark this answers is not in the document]
qinxuye: Fixed in 7d49bbefe — you are right, and the concrete failure is worse than a shared blob in general.

I mirrored the block above, which does resolve realpaths on purpose: a model file belongs to exactly one cache entry, so removing what the link points at is what makes "remove cache" actually free the disk. A drafter breaks that assumption — **one download is shared by every quantization of its target**:

```
target Q4_K_M -> gemma-4-ggufv2-12b-Q4_K_M-draft-BF16 -> MTP/mtp-gemma-4-12b-it-BF16.gguf
target Q8_0   -> gemma-4-ggufv2-12b-Q8_0-draft-BF16   -> MTP/mtp-gemma-4-12b-it-BF16.gguf
```

Deleting the Q4_K_M cache would take the download out from under Q8_0, whose drafter directory still exists — so `get_cache_status` keeps reporting it as cached while the load fails. A leaked blob the hub tooling can reclaim is the better failure.

Went slightly further than the suggestion: a snapshot-style drafter (pytorch/MLX) has the cache entry itself as a link, so that case returns right after adding the link, and the directory case walks its entries without resolving any of them. `test_list_deletable_models_keeps_shared_drafter_downloads` asserts both the shared file and the snapshot directory survive; it fails against the previous implementation.

xinference/model/llm/llama_cpp/core.py
```diff
@@ -142,6 +144,54 @@ def __init__(
...
+        draft_model_path = self._llamacpp_model_config.pop("draft_model_path", None)
+        num_speculative_tokens = self._llamacpp_model_config.pop(
+            "num_speculative_tokens", None
+        )
+        if not draft_model_path:
+            return
+
+        if os.path.isdir(draft_model_path):
+            # the cache dir holds the single drafter file
+            ggufs = sorted(glob.glob(os.path.join(draft_model_path, "**", "*.gguf")))
```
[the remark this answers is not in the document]
qinxuye: Correct, fixed in 7d49bbefe. Demonstrated before and after:

```
without recursive: [nested.gguf]
with recursive:    [flat.gguf, nested.gguf]
```

So only the `MTP/` layout resolved — which is exactly what the existing test used, so it passed. Added `test_cache_dir_resolves_a_flat_gguf` for the flat case.

xinference/model/llm/cache_manager.py
```diff
@@ -42,11 +46,81 @@ def __init__(
         self._model_id = llm_family.model_specs[0].model_id
         self._model_hub = llm_family.model_specs[0].model_hub
         self._model_revision = llm_family.model_specs[0].model_revision
+        # Set when caching a gguf drafter: a single file rather than a snapshot.
+        self._draft_file_name: Optional[str] = None
```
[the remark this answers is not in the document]
qinxuye: Added in 0a7cf5a9c — `_use_draft_model` is stored, and the file-list helper keys off it so it can never fall back to the target names in draft mode.

xinference/model/llm/cache_manager.py
```diff
@@ -42,11 +46,81 @@ def __init__(
...
+
+    def _gguf_file_names(self):
+        """File list for a gguf download: the drafter is a single file."""
+        from ..utils import generate_model_file_names_with_quantization_parts
+
+        if self._draft_file_name:
+            return [self._draft_file_name], self._draft_file_name, False
+        return generate_model_file_names_with_quantization_parts(
+            self._llm_family.model_specs[0], self._multimodal_projector
+        )
```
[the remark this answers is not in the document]
qinxuye: Agreed, and worth being precise about how bad it was: with only `draft_model_id` set, the drafter download would have asked the drafter repository for the *target model file name* — a 404 in the good case, and in the bad one (a repository that happens to carry that name) the target weights landing in the drafter cache directory.

Not reachable through the built-in specs, which always declare both fields, but a user-registered spec can get there, so it deserves the hard failure.

Fixed in 0a7cf5a9c with one difference from the suggestion: the check lives in the constructor, next to the other drafter validations (missing `draft_model_id`, unknown `draft_quantization`), so the launch fails before any download starts rather than partway through `cache()`. The message names the missing field and points at `draft_model_path` for a local drafter file. The helper keeps an assert as a backstop.

Two tests: one asserts the rejection, one asserts the drafter file list and that it never intersects the target file list. The first fails if the guard is removed.

xinference/core/worker.py
```diff
@@ -877,9 +877,21 @@ def _get_spec_dicts_with_cache_status(
...
             family_copy.model_specs = [spec]
             cache_manager = cache_manager_cls(family_copy)
-            specs.append(
-                {**spec.dict(), "cache_status": cache_manager.get_cache_status()}
-            )
+            spec_dict = {
+                **spec.dict(),
+                "cache_status": cache_manager.get_cache_status(),
+            }
+            if getattr(spec, "draft_model_id", None):
```
rogercloud: **Major**: this constructs `LLMCacheManager(..., use_draft_model=True, ...)` per drafter quantization with no `try/except`. The constructor raises `ValueError` for a `{draft_quantization}`-templated `draft_model_id` with empty `draft_quantizations` (`cache_manager.py:86-90`) and for a ggufv2 spec with `draft_model_id` but no `draft_model_file_name_template` (`cache_manager.py:110-114`). No spec-class validator rejects these shapes at registration time, so one malformed custom-registered spec crashes the entire model-listing endpoint, not just its own entry.

Suggest wrapping this per-spec construction in `try/except` and reporting `draft_cache_status=False` on failure, and/or adding a pydantic validator that rejects the shape at registration.

xinference/core/supervisor.py
```diff
@@ -1415,9 +1415,19 @@ def _get_spec_dicts(
...
             model_family.model_specs = [spec]
             cache_manager = cache_manager_cls(model_family)
-            specs.append(
-                {**spec.dict(), "cache_status": cache_manager.get_cache_status()}
-            )
+            spec_dict = {
+                **spec.dict(),
+                "cache_status": cache_manager.get_cache_status(),
+            }
+            if getattr(spec, "draft_model_id", None):
```
rogercloud: Same issue as `worker.py`'s `_get_spec_dicts_with_cache_status` (see that comment) - `LLMCacheManager(..., use_draft_model=True, ...)` is constructed here too with no `try/except`, so a malformed custom drafter spec would crash `_to_llm_reg`/listing here as well.

xinference/core/worker.py
rogercloud: **Minor (latent)**: this guard is `getattr(spec, "draft_model_id", None)`, but `cache_manager.py:56-72` also supports a gguf drafter identified by `draft_model_file_name_template` alone (defaulting to `spec.model_id`). Every current builtin spec using that template also sets `draft_model_id`, so this is latent today - but a future/custom spec relying on the documented fallback would get no `draft_cache_status`, and the frontend's `hasDrafter` (keyed off `draft_model_id`) would never show the Speculative Decoding section for it.

Suggest widening the guard to also check `draft_model_file_name_template`.

xinference/model/llm/llama_cpp/core.py
```diff
@@ -283,6 +339,7 @@ def load(self):
                         setattr(params, k, v)
                 except Exception as e:
                     logger.error("Failed to set the param %s = %s, error: %s", k, v, e)
+            self._apply_draft_model(params)
```
rogercloud: **Major**: `_apply_draft_model` runs unconditionally here, after the generic dotted-key `setattr` loop (330-341) has already applied the user's raw config (including keys like `"speculative.types"`). Unlike vLLM (`vllm/core.py:996-1003`) and SGLang (`sglang/core.py:311-317`), which both detect an explicit user-supplied speculative config and bail with a `logger.info` before touching anything, `_apply_draft_model` here only guards on `if not draft_model_path: return` - no check of whether `params.speculative.types` or similar was already set. A user with `enable_mtp=true` plus a manual `"speculative.types"`/`"speculative.draft.mparams.path"` override gets it silently clobbered, contradicting the PR's claim that an explicit config always wins across all engines.

Suggest adding the same explicit-config guard here with a matching log line.

rogercloud (commented):
## Summary

This PR adds speculative decoding for Gemma 4 via paired drafter checkpoints (a separate draft model, not inline MTP layers) across four engines: vLLM, SGLang, MLX, and llama.cpp. It extends the LLM spec schema and `LLMCacheManager` so drafter checkpoints can be declared, downloaded, and cache-managed alongside their target, adds a Web UI section gated on drafter availability, and registers gemma-4 12B as a second encoder-free architecture. Two unrelated fixes also ride along: a dev-proxy streaming fix and a `gemma_tool_parser` quoting fix. 30 files changed, +1679/-45, no linked issue.

## Design verdict

Acceptable with reservations. The core shape is right: reusing `LLMCacheManager` by reparameterizing it instead of adding a second downloader is the correct call, `LLM.support_draft_model` fails fast before a multi-GB download, and the `<cache_dir>-draft-<quantization>` sibling directory scheme makes cascade delete a prefix scan with no new bookkeeping.

Design-level concerns:

- Drafter spec-shape validation lives in `LLMCacheManager.__init__` rather than as a spec-level pydantic validator. This is the direct cause of the finding on `worker.py`/`supervisor.py` below: a malformed custom spec surfaces as a crash on a listing endpoint rather than at registration time.
- `draft_model_id` / `draft_quantizations` / `draft_model_file_name_template` and their `validate_model_size_with_radix` validator are duplicated near-verbatim across three spec classes in `xinference/model/llm/llm_family.py` (`LlamaCppLLMSpecV2` ~88-94, `PytorchLLMSpecV2` ~119-122, `MLXLLMSpecV2` ~147-166). All three subclass `BaseModel` directly with identical types and defaults; a shared mixin would remove the drift risk. Maintainability, not correctness.
- The drafter is downloaded in `create_llm_model_instance` (`xinference/model/llm/core.py:346-349`) before any check of whether the user already supplied an explicit speculative config. The raw kwargs needed for that check are already in scope at that point, so hoisting the check earlier would avoid downloading and then discarding multiple GB of unused weights.
- `enable_mtp` reuses DeepSeek-MTP vocabulary for a mechanism this PR explicitly distinguishes from DeepSeek-style MTP. Naming nit only.
- Three independent changes are bundled: the tool-parser quoting fix, the `.env.development` dev-proxy streaming fix, and locale reflows. Each is fine standalone, but together they make the PR hard to revert atomically. Process note, not blocking.

## Additional findings (not anchorable inline - pre-existing/unchanged lines)

- **Minor (latent)**, `xinference/model/llm/cache_manager.py` (`cache_from_huggingface`/`cache_from_modelscope`, ~line 203/271): `self._llm_family` is never reassigned in the `use_draft_model` branch, so the drafter's `snapshot_download` reuses the *target's* `cache_config` (e.g. `allow_patterns`/`ignore_patterns`), which is unchanged pre-existing code not touched by this PR. The same branch explicitly nulls `_model_uri` and `_multimodal_projector` for exactly this class of reason, but `cache_config` was missed. No builtin family currently sets `cache_config`, so this is dormant - but a future family combining a drafter with a target-scoped `cache_config` would silently get an incomplete drafter download. Suggest resetting `cache_config` alongside the existing `_model_uri`/`_multimodal_projector` reset.
- **Minor**, `xinference/model/llm/llm_family.json` (gemma-4 31B `fp4` spec, ~line 30296): this spec has no `draft_*` fields at all, while the 31B `pytorch` spec at the same size does (this line is unchanged by the diff, so no drafter fields were added here either way). Per `backends.rst` the UI section simply stays hidden when a spec has no drafter, so this fails safe - but nothing states why fp4 lacks a drafter conversion. Worth a one-line comment confirming this is deliberate (no published NVFP4 drafter conversion) rather than an oversight.
- **Minor**, `xinference/model/llm/mlx/core.py:1382` (`MLXVisionModel.stop()`, unchanged by this diff): clears the executor but never clears `self._draft_model` (nor `self._model`, which has the same pre-existing gap). Not a regression introduced by this PR, but it extends an existing memory-retention pattern to a new, potentially large attribute. Worth a follow-up to release both references.

## Simplification opportunities

- `xinference/model/llm/tool_parsers/gemma_tool_parser.py` L48-77: `_arguments_to_json` splits on a literal string-token delimiter with a manual char-by-char loop using `startswith`/`find`. `arg_block.split(token)` yields the same alternating segments (even indices structural, odd indices literal string content), and `len(parts) % 2 == 0` reproduces the unterminated-string check. Net: roughly -15 to -20 lines. Sits inside the unrelated tool-parser fix, not the speculative-decoding code.

## Re-review: prior findings

All three prior findings from earlier review rounds are fixed, independently re-verified this round:

- **Symlink `os.path.realpath` cascade-delete data-loss risk** (`worker.py`) - FIXED in `7d49bbef`. No `realpath` remains in the drafter deletion block (`worker.py:4114-4125`); the drafter path is added to `paths` before any `islink` check, so only the symlink itself is unlinked, and directory drafters are walked with plain `os.walk`. `test_list_deletable_models_keeps_shared_drafter_downloads` asserts a shared blob and shared snapshot survive.
- **`glob.glob` missing `recursive=True`** (`llama_cpp/core.py`) - FIXED in `7d49bbef`. `recursive=True` is present at `llama_cpp/core.py:166-167`, with `test_cache_dir_resolves_a_flat_gguf` covering the flat case.
- **GGUF drafter silently falling back to target file names** (`cache_manager.py`) - FIXED in `0a7cf5a9c`. `_use_draft_model` is now a stored attribute (`cache_manager.py:51`) and the constructor raises a `ValueError` naming the missing `draft_model_file_name_template` before any download begins. Covered by `test_gguf_drafter_without_a_file_template_is_rejected` and `test_gguf_drafter_downloads_only_its_own_file`.

## Test results

Ran the PR's new and changed unit tests in a fresh venv (editable install, without engine-specific optional deps): 96 passed, 2 skipped, 9 failed. All 9 failures are `ModuleNotFoundError`/`ImportError` for `xllamacpp` and `mlx_lm`, absent from the review environment - environmental, not code defects. Where a test could not execute past the missing-dependency gate (notably the glob-recursive fix and the gguf drafter template rejection), the underlying logic was confirmed by reading the code instead. This matches the author's stated gap: no actual vLLM/SGLang/llama.cpp run, for lack of hardware.

xinference/model/llm/llama_cpp/core.py
```diff
@@ -283,6 +339,7 @@ def load(self):
                         setattr(params, k, v)
                 except Exception as e:
                     logger.error("Failed to set the param %s = %s, error: %s", k, v, e)
```
rogercloud, line 386: **Minor**: `_apply_draft_model` (called right after this loop) pops `draft_model_path` and `num_speculative_tokens` internally, i.e. after this setattr loop has already tried to set them on `params`. `CommonParams` has no such flat attributes, so `setattr` fails here and this `logger.error` fires twice on every speculative-decoding llama.cpp launch. Harmless (the values are reapplied correctly afterward), but it trains users to ignore real errors.

Suggest popping `draft_model_path`/`num_speculative_tokens` from `self._llamacpp_model_config` before this loop runs.

xinference/model/llm/vllm/core.py
```diff
@@ -971,6 +976,50 @@ def parse_str_field_to_dict(
...
+            raise ValueError(
+                f"Speculative decoding with a Gemma 4 style drafter needs "
+                f"vllm>={self.MTP_MIN_VLLM_VERSION}, but {VLLM_VERSION} is installed. "
+                f"Upgrade vLLM, or launch without `enable_mtp`."
+            )
+
+        model_config["speculative_config"] = {
+            "method": "mtp",
+            "model": draft_model_path,
+            "num_speculative_tokens": int(num_speculative_tokens or 1),
```
rogercloud: **Major**: this hardcodes a speculation depth of `1` when unset - a bare inline literal with no upstream provenance (SGLang's equivalent is at least a named `DEFAULT_SPECULATIVE_NUM_DRAFT_TOKENS = 6`). Yet `frontend/src/constants/launch.ts:91-93`'s docblock, `doc/source/user_guide/backends.rst` (~227-230), and all four locale tooltips (en/ja/ko/zh) describe this as "the engine's own default," which isn't accurate - it's this PR's own choice. Gemma 4's drafters are trained at depth 4, which the MLX path correctly resolves from the drafter itself. A user enabling speculative decoding on vLLM without setting a depth gets 1 token per round, likely near-zero net speedup, with no signal anything is misconfigured.

Suggest resolving the default from the spec (matching MLX's approach) or raising it toward the trained depth, and correcting the docs/docblock/locale strings that call this "the engine's default."

frontend/src/constants/launch.ts
```diff
@@ -70,12 +76,31 @@ export const KWARGS_OPTIONS_FOR_ENGINES: Record<string, Array<{ label: string; v
...
     { label: 'cache_limit_gb', value: 'cache_limit_gb' },
     { label: 'max_kv_size', value: 'max_kv_size' },
+    { label: 'enable_mtp', value: 'enable_mtp' },
+    { label: 'num_speculative_tokens', value: 'num_speculative_tokens' },
+    { label: 'draft_model_path', value: 'draft_model_path' },
   ],
 };
+/**
+ * Tokens the drafter proposes per round when the launch leaves it unset. The
+ * value is the engine's own default, not something Xinference picks: MLX runs
```
rogercloud: This docblock says the default "is the engine's own default, not something Xinference picks." That's not accurate for vLLM (hardcoded `1` at `vllm/core.py:1015`) or SGLang (hardcoded `6` at `sglang/core.py:297`) - both are values this PR invents, not values read from the engines. Worth rewording once the vLLM default (see that file's review comment) is revisited.

xinference/model/llm/sglang/core.py
```diff
@@ -281,11 +291,53 @@ def stop(self):
...
+                self.model_uid,
+            )
+            return
+
+        num_draft_tokens = int(
+            num_speculative_tokens or self.DEFAULT_SPECULATIVE_NUM_DRAFT_TOKENS
+        )
+        model_config["speculative_algorithm"] = "NEXTN"
+        model_config["speculative_draft_model_path"] = draft_model_path
+        model_config["speculative_num_draft_tokens"] = num_draft_tokens
```
rogercloud: **Minor**: this assigns unconditionally, while the adjacent `speculative_num_steps` (line 326) and `speculative_eagle_topk` (line 327) both use `.setdefault(...)`. Since the explicit-config guard above (line 311) only checks `speculative_algorithm`, a user who sets `draft_model_path` plus `speculative_num_draft_tokens` explicitly but leaves `speculative_algorithm` unset has their value silently overwritten here, inconsistent with the two adjacent keys.

Suggest using `.setdefault(...)` for all three keys.

frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -243,6 +243,36 @@ export default function LaunchDialog({
...
+          toOptionValue(spec.model_size_in_billions) === modelSizeInBillionsKey &&
+          (spec.quantization === quantizationValue ||
+            (Array.isArray(spec.quantizations) && spec.quantizations.includes(quantizationValue)))
+      ),
+    [model?.modelSpecs, modelFormatValue, modelSizeInBillionsKey, quantizationValue]
+  );
+  // Drafter conversions available for the selected spec, each flagged with
+  // whether it is already downloaded. Empty when the spec declares a drafter
+  // without alternatives, in which case there is nothing to pick.
+  const hasDrafter = Boolean(selectedSpec?.draft_model_id);
```
rogercloud: **Minor**: this gates the Speculative Decoding UI section purely on the spec, with no reference to the selected engine. gemma-4's pytorch spec has a drafter and is launchable via both MLX (`support_draft_model=True`) and Transformers (`support_draft_model=False`, never overridden and not exposed via any API endpoint). Selecting Transformers still shows the switch; the backend rejection (`xinference/model/llm/core.py:339-345`) only surfaces after the user clicks launch.

Suggest exposing `support_draft_model` per engine in the existing per-engine metadata endpoint and gating this section on it too.

frontend/src/constants/launch.ts
```diff
@@ -39,6 +39,9 @@ export const KWARGS_OPTIONS_FOR_ENGINES: Record<string, Array<{ label: string; v
     { label: 'n_ctx', value: 'n_ctx' },
     { label: 'use_mmap', value: 'use_mmap' },
     { label: 'use_mlock', value: 'use_mlock' },
+    { label: 'enable_mtp', value: 'enable_mtp' },
```
rogercloud: **Minor**: `enable_mtp` and `num_speculative_tokens` are listed here as free-form kwargs options for every engine, but they're also registered as dedicated form fields (`ALL_FORM_KEYS` at line 122ff, with Switch/Input controls in `advanced-config.tsx`). `mergeFormListObject` in `utils.tsx` `Object.assign`s the parsed kwargs list over the already-populated dedicated-field values with no filtering, so a user who manually adds one of these two params via the free-form kwargs UI silently overwrites the dedicated control's value.

Suggest filtering these two names out of `KWARGS_OPTIONS_FOR_ENGINES` (per engine), or reconciling the two sources in `mergeFormListObject`. (`draft_model_path` is unaffected here - it has no dedicated field.)

xinference/model/llm/tool_parsers/gemma_tool_parser.py
```diff
@@ -47,12 +45,42 @@ def repl(match: re.Match) -> str:
                 break
         return text
 
+    def _arguments_to_json(self, arg_block: str) -> str:
```
rogercloud, line 48: This is a correct, well-tested fix to a real argument-quoting bug, but it's unrelated to speculative decoding and would be easier to review/backport as its own PR. No change requested here, just flagging for future PR hygiene.

Separately, as a simplification note: this manual char-by-char loop (62-75) splitting on the literal `token` delimiter could be replaced by `arg_block.split(token)`, which yields the same alternating structural/string segments (even indices structural, odd indices literal string content), with `len(parts) % 2 == 0` reproducing the "unterminated string" check - roughly -15 to -20 lines.

xinference/core/worker.py
qinxuye: Fixed in bdff06a77. One bad custom registration turning every listing call into a 500 is much worse than that spec being unusable, so the per-spec construction is wrapped and reports `draft_cache_status: []` with a warning naming the model and the reason.

Went with the try/except rather than a pydantic validator: the listing path has to survive a malformed spec regardless of where it came from (a hand-edited JSON under `~/.xinference/model/v2` never passes through the registration API), so the guard belongs at the read site. A validator would be a nice addition on top; happy to add one if you want it in this PR.

`test_malformed_drafter_spec_does_not_break_listing` covers it.

xinference/core/supervisor.py
qinxuye: Same fix applied here in bdff06a77.

xinference/core/worker.py
qinxuye: Widened in bdff06a77 — both probes now accept `draft_model_id` or `draft_model_file_name_template`, and the frontend `hasDrafter` does the same, so a spec relying on the documented fallback behaves like the built-ins.

xinference/model/llm/llama_cpp/core.py
qinxuye: You are right, and it contradicted the PR description. Fixed in bdff06a77: `_apply_draft_model` now bails with a matching `logger.info` when any `speculative`/`speculative.*` key is present in the passthrough config, which by then has already been applied by the dotted-key loop.

`test_explicit_speculative_params_win` asserts nothing is touched in that case.

xinference/model/llm/llama_cpp/core.py
qinxuye: Good catch — two spurious errors per launch is exactly the kind of thing that trains people to ignore the log. Fixed in bdff06a77 by pulling both keys out into locals before the loop (`_pop_draft_options`) and passing them to `_apply_draft_model`.

Kept the application itself after the loop so the explicit-config check above sees what the user set. `test_neutral_options_leave_before_the_passthrough_loop` pins that the config handed to the loop no longer carries them.

xinference/model/llm/vllm/core.py
qinxuye: Right on the accuracy point, and I checked the "resolve it from the spec" option before answering: there is nothing to resolve from. Gemma 4 drafters do not publish their depth — `block_size` is absent from the drafter `config.json`; the 4 comes from mlx-vlm's own `Gemma4AssistantConfig` dataclass default, which is why only the MLX path gets it for free.

So in bdff06a77 the value stays 1 but becomes `DEFAULT_NUM_SPECULATIVE_TOKENS` with its provenance recorded: it is the starting point vLLM's own MTP guide recommends ("a small value like 1 is a good default to start with"), not a value vLLM would default to. The comment and the user-facing text both now say it is deliberately conservative and that raising it toward the trained depth is usually what pays off.

Docs, the frontend docblock and all four locales were corrected to stop calling these "the engine's own default" — only MLX reads it from the drafter; the others are values from each engine's Gemma 4 guide.

Raising vLLM's default outright is the one thing I did not do: I have no GPU to measure it on, and inventing a number would repeat the mistake this comment is about.

frontend/src/constants/launch.ts
qinxuye: Reworded in bdff06a77 along with the docs and the four locales — see the vLLM comment for the details of why 1 stays but is now named and explained.

xinference/model/llm/sglang/core.py
qinxuye: Fixed in bdff06a77 — all three keys use `setdefault` now, and the step count derives from whichever draft-token count ends up in effect rather than from ours. `test_explicit_draft_token_count_wins` pins it.

frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
qinxuye: Done in bdff06a77, taking the approach you suggested. `_append_available_engine` now publishes `support_draft_model` per engine entry (guarded by `hasattr`, so non-LLM engine payloads are unaffected), and the section requires both a spec that ships a drafter and an engine that can run one.

Verified against the live payload for gemma-4:

```
Transformers   support_draft_model={False}
llama.cpp      support_draft_model={True}
MLX            support_draft_model={True}
```

so the switch no longer appears under Transformers.

frontend/src/constants/launch.ts
qinxuye: Fixed in bdff06a77 by dropping both names from every engine list, since the dedicated controls are the intended path and `mergeFormListObject` would let the free-form entry win. `draft_model_path` stays, as you noted it has no control.

xinference/model/llm/tool_parsers/gemma_tool_parser.py
qinxuye: On the PR hygiene: agreed, and it was in its own PR briefly — the repository owner asked for it to live here instead, so that is where it is.

Took the simplification: `arg_block.split(self.string_token)` with even indices structural and odd indices literal, and `len(parts) % 2 == 0` as the unterminated check. The scanner is a third of its previous size and the whole tool-parser suite still passes (70).

xinference/model/utils.py
```diff
@@ -1019,6 +1019,12 @@ def _append_available_engine(
         cleaned_params: List[Dict[str, Any]] = []
         for param in params:
             new_param = {k: v for k, v in param.items() if k != class_field}
+            model_cls = param.get(class_field)
+            support_draft_model = getattr(model_cls, "support_draft_model", None)
+            if support_draft_model is not None:
+                # so callers can tell whether this engine can run a drafter at
+                # all, rather than finding out when the launch is rejected
+                new_param["support_draft_model"] = bool(support_draft_model)
```
rogercloud: **Critical**: this closure (inside `_get_engine_params_by_name`) now correctly attaches `support_draft_model` - the intended fix for the prior UI-gating finding. But the structurally identical `_append_available_engine` closure inside `_get_engine_params_by_name_with_virtual_env` (~line 1462-1470) was never updated to do the same, and that is the function actually used by default: `XINFERENCE_ENABLE_VIRTUAL_ENV` defaults to `True`, and both `SupervisorActor` and `WorkerActor` dispatch to the virtualenv-aware variant whenever the flag is set.

As a result, `support_draft_model` is absent from the API response in a default deployment, `engineSupportsDrafter`/`hasDrafter` in `launch-dialog.tsx` evaluate to false for every engine, and the entire Advanced Configuration -> Speculative Decoding section never renders. This is a regression from this very commit: before it, the section at least rendered from `spec.draft_model_id` alone. Suggest propagating `support_draft_model` in the virtualenv-aware closure too - the two closures are otherwise identical and clearly meant to stay in sync.

xinference/model/llm/llama_cpp/core.py
```diff
@@ -142,6 +144,82 @@ def __init__(
...
+
+        ``draft-mtp`` covers Gemma 4 style assistants, which are a separate
+        model sharing the target's KV cache; llama.cpp learned that
+        architecture (``gemma4-assistant``) in xllamacpp 2026.6.9713.
+        """
+        if not draft_model_path:
+            return
+
+        if any(
+            key == "speculative" or key.startswith("speculative.")
```
rogercloud: **Major**: this guard fires on ANY key equal to `"speculative"` or starting with `"speculative."`, which is broader than the sibling engines' equivalents. vLLM and SGLang key narrowly on a single mode-selector field (`speculative_config` presence / `speculative_algorithm` string) - "the user chose a different method entirely." This guard instead also catches pure tuning knobs like `speculative.n_min`, `speculative.p_min`, or `speculative.draft.n_gpu_layers`, which are plausible `CommonParams.speculative.*` passthrough fields distinct from the actual mode/drafter-selector keys (`speculative.types`, `speculative.draft.mparams.path`). A user with `enable_mtp=true` plus one of these unrelated tuning keys gets the drafter silently disabled (only a `logger.info` line). The new test only covers the "user set their own drafter path" case.

Suggest narrowing this to the actual selector keys (`speculative.types` / `speculative.draft.mparams.path`), mirroring the narrower guards on vLLM/SGLang.

xinference/model/llm/llama_cpp/core.py
```diff
@@ -142,6 +144,82 @@ def __init__(
...
+        )
+        if mtp is None:
+            raise ValueError(
+                "The installed xllamacpp has no MTP speculative implementation, "
+                "upgrade to xllamacpp>=2026.6.9713"
+            )
+
+        params.speculative.types = [mtp]
+        params.speculative.draft.mparams.path = draft_model_path
+        if num_speculative_tokens:
```
rogercloud: **Major**: when `num_speculative_tokens` is falsy, `n_max` is left untouched here - there is no fallback constant for llama.cpp analogous to vLLM's `DEFAULT_NUM_SPECULATIVE_TOKENS = 1` or SGLang's `DEFAULT_SPECULATIVE_NUM_DRAFT_TOKENS = 6`. Yet `doc/source/user_guide/backends.rst:230`, `frontend/src/constants/launch.ts:90-95` (`SPECULATIVE_TOKENS_DEFAULT_BY_ENGINE['llama.cpp'] = 3`), and the `numSpeculativeTokensTip` string in all four locale files claim a default of "3" for llama.cpp. When unset, this leaves whatever xllamacpp's own `CommonParams.speculative.draft.n_max` defaults to - not 3, and not authored by this codebase. The claim got stronger in this fix commit (previously just "the default follows the engine," now a specific fabricated number).

Suggest either adding a real `DEFAULT_NUM_SPECULATIVE_TOKENS = 3` constant here to make the docs/UI true, or removing the "3" claim and describing the actual deferral to xllamacpp's own default.

xinference/core/worker.py
```diff
@@ -877,9 +877,37 @@ def _get_spec_dicts_with_cache_status(
...
+                try:
+                    spec_dict["draft_cache_status"] = [
+                        cache_manager_cls(
+                            family_copy, use_draft_model=True, draft_quantization=quant
+                        ).get_cache_status()
+                        for quant in (
+                            getattr(spec, "draft_quantizations", None) or [None]
+                        )
+                    ]
+                except Exception as e:
```
rogercloud: **Minor**: this catches bare `Exception` rather than the specific `ValueError` that `LLMCacheManager.__init__` actually raises for malformed drafter shapes. The practical blast radius is narrower than it looks - `get_cache_status()`'s only filesystem call (`os.path.exists`) already swallows `OSError`/`ValueError` internally - but a future `AttributeError`/`TypeError` bug anywhere in the drafter cache-status path would now be silently absorbed here and reported as "not cached" rather than surfacing as a real bug.

Suggest narrowing to `except ValueError` to keep the safety net scoped to what it's actually meant to catch.

xinference/core/supervisor.py
```diff
@@ -1415,9 +1415,32 @@ def _get_spec_dicts(
...
+                try:
+                    spec_dict["draft_cache_status"] = [
+                        cache_manager_cls(
+                            model_family, use_draft_model=True, draft_quantization=quant
+                        ).get_cache_status()
+                        for quant in (
+                            getattr(spec, "draft_quantizations", None) or [None]
+                        )
+                    ]
+                except Exception as e:
```
rogercloud: Same as the `worker.py` comment on the equivalent `except Exception` - consider narrowing to `except ValueError` here too.

xinference/model/llm/sglang/core.py
```diff
@@ -281,11 +291,60 @@ def stop(self):
...
+
+        num_draft_tokens = int(
+            num_speculative_tokens or self.DEFAULT_SPECULATIVE_NUM_DRAFT_TOKENS
+        )
+        model_config["speculative_algorithm"] = "NEXTN"
+        model_config["speculative_draft_model_path"] = draft_model_path
+        # an explicitly provided count wins, like the two keys below
+        model_config.setdefault("speculative_num_draft_tokens", num_draft_tokens)
+        effective_draft_tokens = int(
+            model_config["speculative_num_draft_tokens"] or num_draft_tokens
```
rogercloud: **Minor**: `speculative_num_draft_tokens` now uses `.setdefault` (fixing the prior overwrite issue), but this derived value uses falsy-or semantics: an explicit `speculative_num_draft_tokens=0` survives `setdefault` yet is treated as unset here, so `speculative_num_steps` below ends up derived from the fallback `num_draft_tokens` instead of `0`. Separately, a string value like `"3"` survives into `model_config["speculative_num_draft_tokens"]` uncoerced (only this local `effective_draft_tokens` gets `int(...)`-ed), so the engine would receive a string where it expects an int. `test_explicit_draft_token_count_wins` only exercises a plain positive int.

Suggest coercing the dict entry itself rather than only the derived local, and using `is not None` where 0 is a legitimate value.

rogercloud (commented):
## Follow-up Review (commit `bdff06a`)

This round covers a single update commit (`bdff06a`, 18 files, +258/-79) in which the author replies to all 11 findings from the prior review at `44f2bac`. The PR was re-read from a blank slate as a whole change, and every prior finding was re-verified independently against the code rather than accepted from the author's replies. The fix commit resolves most of what was raised, but it also introduces one regression that defeats its own UI fix by default, plus two new inconsistencies.

### Design verdict (independent, whole-PR)

Verdict is unchanged: **acceptable with reservations**. This is an independent judgment of the change as it now stands, not a tally of "11/11 addressed" - several fixes are correct in mechanism while leaving the underlying shape untouched.

- Validation of malformed drafter specs still lives in `LLMCacheManager.__init__` rather than in a spec-level validator. The fix commit wrapped the call site in `try/except` instead of moving validation to where the spec is defined. Pragmatic and not wrong, but the layering gap is papered over rather than closed. No action requested this round; recording it as an open design note.
- The "explicit speculative config wins" invariant now means three different things across three engines: vLLM keys on the presence of the `speculative_config` object, SGLang on the `speculative_algorithm` selector, and llama.cpp (new in this commit) on *any* key under the `speculative`/`speculative.*` namespace. The llama.cpp variant is the broadest and least consistent of the three - see the Major finding below.
- The new per-engine `support_draft_model` capability is exposed in only one of two structurally identical engine-listing code paths, and the untouched one is the default. See the Critical finding.
- Four engines now ship four different hardcoded or undiscoverable "default speculation depth" values: vLLM 1, SGLang 6, MLX resolved from the drafter, and llama.cpp a documented "3" that does not exist in code. This is the same underlying gap flagged before - the drafter's trained depth is a property of the checkpoint, not of the engine, and the spec declares it nowhere.

### What actually changed since the last review (independently verified, not taken from the author's replies)

| # | Finding | Status |
|---|---|---|
| 1 | Malformed drafter spec crashes model-listing (worker.py/supervisor.py) | **FIXED** - try/except added, `test_malformed_drafter_spec_does_not_break_listing` passes. (See new minor finding below: the catch is broader than necessary.) |
| 2 | `draft_cache_status` guard misses template-only gguf drafter | **FIXED** - guard widened to `draft_model_id or draft_model_file_name_template` in both backend files and the frontend `hasDrafter` check. |
| 3 | llama.cpp missing explicit-config-wins guard | **FIXED** - guard added. (See new major finding below: the guard is broader than the sibling engines' equivalents.) |
| 4 | llama.cpp spurious ERROR logs on every speculative launch | **FIXED** - `_pop_draft_options` now runs before the setattr loop. (See new minor finding below: the pop is destructive and affects retries.) |
| 5 | vLLM default of 1 documented as "the engine's own default" | **FIXED** - docs/locales reworded to correctly describe it as a deliberately conservative Xinference-chosen starting point; the value itself (1) is unchanged, which the author explains is intentional given Gemma 4 drafters don't publish their trained depth. |
| 6 | SGLang overwrites explicit `speculative_num_draft_tokens` | **FIXED** - all three keys now use `.setdefault`. (See new minor finding below: a residual 0/string edge case in the derived step count.) |
| 7 | Frontend UI gates on spec only, ignoring engine support | **PARTIAL** - the intended fix (exposing `support_draft_model` per engine and gating on it) is correctly implemented, but only wired into one of the two engine-listing code paths. See the new Critical finding below - this is effectively still open for any default-configuration deployment. |
| 8 | Frontend duplicate kwargs/dedicated-field mechanism | **FIXED** - `enable_mtp`/`num_speculative_tokens` removed from every engine's free-form kwargs list; `draft_model_path` correctly retained since it has no dedicated control. |
| 9 | Drafter download inherits target's `cache_config` (latent) | **NOT ADDRESSED** - `cache_manager.py` is untouched by this commit; noted as dormant/latent in the original review and remains so, not a regression. |
| 10 | gemma-4 31B fp4 spec has no `draft_*` fields | **NOT ADDRESSED** - `llm_family.json` is untouched by this commit; unchanged from before, likely intentional. |
| 11 | MLX `stop()` doesn't clear `_draft_model` | **NOT ADDRESSED** - `mlx/core.py` is untouched by this commit; unchanged, not a regression. |
| - | gemma_tool_parser.py simplification (`str.split`) | **APPLIED CORRECTLY** - verified the replacement preserves the even/odd segment alternation and the `len(parts) % 2 == 0` unterminated-string check exactly. |

### Tests

The PR's unit tests were re-run against the updated worktree: 100 passed, 2 skipped, 9 failed - up from 96 passing previously, reflecting the 4 new passing tests added by this commit. All 9 failures are the same pre-existing environmental `ModuleNotFoundError`/`ImportError` for `xllamacpp`/`mlx_lm`, absent from the review sandbox and unrelated to this change. No new test failures were introduced.

xinference/model/llm/llama_cpp/core.py
```diff
@@ -142,6 +144,82 @@ def __init__(
...
 
+    def _pop_draft_options(self) -> Tuple[Optional[str], Optional[Any]]:
+        """Take the engine-neutral speculative options out of the config.
+
+        They have to leave before the generic dotted-key loop runs: CommonParams
+        has no such attributes, so the loop would log a failure for each of them
+        on every speculative launch.
+        """
+        return (
+            self._llamacpp_model_config.pop("draft_model_path", None),
```
rogercloud: **Minor**: this uses destructive `dict.pop()` on the long-lived `self._llamacpp_model_config` (fixing the prior spurious-log-lines finding). There is a real, if narrow, retry path: the load-retry loop in `xinference/core/model.py` reuses the same model instance and calls `load()` again after a "busy or unavailable" failure, so a retried speculative-decoding launch would silently lose its drafter on the second attempt since these keys are already gone from the first call.

Suggest reading with `.get()` instead of popping, or resetting the relevant config entries before a retry.

xinference/model/utils.py
qinxuye: You are right, and it is worse than a missed spot — I verified the fix against `get_engine_params_by_name`, which is exactly the path that is not the default. Measured before the fix:

```
non-virtualenv:      {Transformers: False, llama.cpp: True, MLX: True}
virtualenv (default): {Transformers: absent, llama.cpp: absent, MLX: absent}
```

Fixed in ebf04d833, and the cause went one layer deeper than the two closures. Patching the second closure alone still left the flag absent: for engines carrying virtualenv markers, `_force_virtualenv_engine_params` *rebuilds* the entries from specs via `_build_engine_params_from_specs*` and overwrites what the closure produced, and those rebuilt dicts never carried `llm_class` to read the flag from.

So there are now two mechanisms: registry-built entries go through a shared `_clean_engine_param` (both closures call it, so they cannot drift again), and the rebuilt lists are annotated from `engine_classes`, which `_force_virtualenv_engine_params` already has in scope. Both paths agree now:

```
non-virtualenv:       {Transformers: False, llama.cpp: True, MLX: True}
virtualenv (default): {Transformers: False, llama.cpp: True, MLX: True}
```

`test_both_engine_discovery_paths_publish_draft_support` asserts the flag is present from both entry points; reverting the annotation makes it fail.

xinference/model/llm/llama_cpp/core.py
qinxuye: Agreed — a tuning knob is not a mode selector. Narrowed in ebf04d833 to `SPECULATIVE_SELECTOR_KEYS = ("speculative.types", "speculative.draft.mparams.path")`, so `speculative.draft.n_min` and friends no longer turn the drafter off.

`test_tuning_knobs_do_not_disable_the_drafter` covers exactly that case alongside the existing selector test.

xinference/model/llm/llama_cpp/core.py
qinxuye: Right that the claim was unfounded as written. I went the other way from adding a constant: llama.cpp is the one engine that *has* its own default, so deferring to it is the better behavior and the docs should say so rather than this code pinning a number that xllamacpp may retune.

The 3 itself is accurate — `CommonParams().speculative.draft.n_max` is 3 in xllamacpp 2026.7 — but its provenance was wrong. ebf04d833 rewords the docs, the docblock and the four locales to name three distinct provenances: MLX reads the depth from the drafter, llama.cpp keeps the xllamacpp default (3 as of 2026.7), and vLLM/SGLang require a value so the one from their own Gemma 4 guide is passed. A comment at the assignment records that leaving `n_max` alone is deliberate.

xinference/core/worker.py
qinxuye: Narrowed to `ValueError` in ebf04d833, with a comment saying that is what the cache manager raises so a different exception is a bug worth surfacing.

xinference/core/supervisor.py
qinxuye: Same narrowing applied here.

xinference/model/llm/sglang/core.py
qinxuye: Both fixed in ebf04d833:

```python
provided = model_config.get("speculative_num_draft_tokens")
effective_draft_tokens = int(provided) if provided is not None else num_draft_tokens
model_config["speculative_num_draft_tokens"] = effective_draft_tokens
```

The entry itself is coerced, so a string from the Web UI cannot reach the engine, and `is not None` keeps an explicit 0. `test_explicit_draft_token_count_is_coerced` passes `"3"` and asserts both the coerced entry and the derived step count.

xinference/model/llm/llama_cpp/core.py
qinxuye: Confirmed the retry path — `xinference/core/model.py:484-497` reuses the instance and calls `load()` again — so this was real. Fixed in ebf04d833 by reading with `.get()` and having the passthrough loop skip `DRAFT_OPTION_KEYS` instead. That keeps the spurious log lines away without consuming the config.

`test_draft_options_survive_a_second_load` calls the reader twice and asserts the config still carries the keys.

xinference/model/utils.py
```diff
@@ -286,6 +286,25 @@ def _build_engine_params_from_specs_by_quantization(
...
+) -> List[Dict[str, Any]]:
+    """Publish whether an engine can run a drafter, for rebuilt param lists.
+
+    The virtualenv-aware discovery path rebuilds entries from specs rather than
+    from the engine registry, so the model class is not attached to them the way
+    ``_clean_engine_param`` finds it. An engine maps to several classes (MLX has
+    a text one and a vision one); the flag says whether any of them can, which
+    is what the caller needs before offering the option at all.
+    """
+    support = any(
```
rogercloud: **Major**: this computes `support_draft_model` as `any(cls.support_draft_model for cls in engine_classes)` - true if ANY class registered under this engine name supports drafters, regardless of which class actually matches the entry being annotated. This is then used at line 397/436 to unconditionally overwrite the more accurate per-entry value computed by `_clean_engine_param` (~line 1006-1018), whenever a model has virtualenv markers and `enable_virtual_env=True` (the default).

Concretely: `SUPPORTED_ENGINES['MLX']` contains `MLXModel`, `MLXChatModel`, `MLXVisionModel` - only `MLXVisionModel.support_draft_model=True`. A chat-only spec whose only matching class is non-draft-capable would get `support_draft_model=True` reported on this path anyway. Currently latent only because gemma-4 (the sole drafter-capable builtin) happens to route to `MLXVisionModel` via its `vision` ability, so the coarse and precise values agree by coincidence. The existing test only checks presence on both discovery paths, never that they agree on the value for the same model.

Suggest reusing `_clean_engine_param`'s per-entry logic here once the class is resolvable from the spec, or adding a parity test across both discovery paths.

xinference/core/worker.py
```diff
@@ -877,9 +877,39 @@ def _get_spec_dicts_with_cache_status(
...
+                try:
+                    spec_dict["draft_cache_status"] = [
+                        cache_manager_cls(
+                            family_copy, use_draft_model=True, draft_quantization=quant
+                        ).get_cache_status()
+                        for quant in (
+                            getattr(spec, "draft_quantizations", None) or [None]
+                        )
+                    ]
+                except ValueError as e:
```
rogercloud, line 903: **Major**: this now catches only `ValueError`, but `LLMCacheManager.__init__` also calls `draft_model_id.format(draft_quantization=...)` and `draft_template.format(...)` (cache_manager.py:92-95, 103-107), guarded only by checking that `{draft_quantization}` is present - not that no OTHER placeholder is present. `"{quantization}".format(draft_quantization='x')` raises `KeyError`; a stray positional `{}` raises `IndexError`. Neither is caught here.

There is no runtime validation preventing a user-registered custom spec from carrying an unexpected placeholder in `draft_model_id`/`draft_model_file_name_template` (only a source comment warning against it for built-ins). Such a registration would crash the entire model-listing call again - the exact regression this narrowing was meant to prevent, via a different exception type.

Suggest widening to also catch `KeyError`/`IndexError`, or normalizing the `.format()` failures to `ValueError` inside the cache manager.

xinference/core/supervisor.py
```diff
@@ -1415,9 +1415,34 @@ def _get_spec_dicts(
...
+                try:
+                    spec_dict["draft_cache_status"] = [
+                        cache_manager_cls(
+                            model_family, use_draft_model=True, draft_quantization=quant
+                        ).get_cache_status()
+                        for quant in (
+                            getattr(spec, "draft_quantizations", None) or [None]
+                        )
+                    ]
+                except ValueError as e:
```
rogercloud, line 1435: Same as the `worker.py` comment on the equivalent `except ValueError` - it can still be bypassed by `KeyError`/`IndexError` from the `.format()` calls in `LLMCacheManager.__init__`.

xinference/model/llm/sglang/core.py
```diff
@@ -281,11 +291,63 @@ def stop(self):
...
+
+        if model_config.get("speculative_algorithm"):
+            # The user is configuring SGLang's speculative decoding directly.
+            logger.info(
+                "Ignoring the drafter of %s, speculative_algorithm was set explicitly",
+                self.model_uid,
+            )
+            return
+
+        num_draft_tokens = int(
```
rogercloud: **Major**: this still uses truthy-or (`num_speculative_tokens or self.DEFAULT_SPECULATIVE_NUM_DRAFT_TOKENS`), so an explicit `num_speculative_tokens=0` is silently replaced by the default - the identical bug class just fixed a few lines below for `speculative_num_draft_tokens` (which now correctly uses `is not None`), left unfixed here in the sibling/neutral parameter.

The same truthy-or pattern for this shared parameter exists identically in `vllm/core.py:1021-1022` (0 becomes 1), `mlx/core.py:1605-1606` (0 drops the draft block size override to None), and `llama_cpp/core.py:221-222` (0 skips the assignment entirely) - this was never sglang-specific, and this commit did not touch the shared parameter in any of the four engines. No test passes `num_speculative_tokens=0`.

Suggest applying `is not None` semantics here too, ideally via one shared helper reused by all four engines, with a test per engine.

rogercloud (commented):
# Round 3 Re-Review - Speculative Decoding with Paired Drafter Checkpoints

This round covers the single new commit `ebf04d8`, which replies to all six remaining findings from round 2 (base `bdff06a`). Each prior finding was independently re-verified against the code rather than accepted from the reply text, and the fix commit itself was reviewed blank-slate for regressions. Result: three of six prior findings are genuinely closed, three are partial, and the partials share a single root cause worth naming explicitly.

## Design assessment

The feature architecture remains sound: paired drafter checkpoints, per-engine translation of speculative options, and capability gating via `support_draft_model` are the right shape, and nothing in this round changes that verdict. The reservation is about the fix process, not the design.

This is now the third consecutive round in which a fix for one code path leaves an identical sibling path or sibling variable untouched. Round 1 added `support_draft_model` to one of two engine-listing functions. Round 2 found the second function missing it, and the round-2 fix patched that second function. Round 3 finds a *third* entry point - `_force_virtualenv_engine_params` - that rebuilds engine param lists from specs, bypassing both earlier paths, and which needed its own annotation logic added this round. Separately, the 0-vs-falsy coercion bug was fixed for `speculative_num_draft_tokens` but left in place for the sibling key `num_speculative_tokens` three lines above it - and that sibling pattern turns out to exist identically in all four engines, meaning it was never sglang-specific and was not addressed at all. And the `except Exception` narrowing was done by swapping the exception type without auditing what `LLMCacheManager.__init__` actually raises across all its failure paths.

The recurring shape - fix N-1 of N duplicated things, or fix the exact reported key but not its sibling - suggests fixes are being applied by pattern-matching the specific finding text rather than by asking "where else does this exact shape occur." Three rounds of the same failure mode is a process signal rather than a code nit, and it is the main thing standing between this PR and a clean sign-off.

## Prior findings - independently re-verified

Statuses below come from reading the code at `ebf04d8`, not from the author's replies.

| # | Round-2 finding | Severity | Status |
|---|---|---|---|
| 1 | `support_draft_model` missing from the default (virtualenv) discovery path | Critical | **Partial** - presence fixed on all paths, but the two paths compute the value differently and can disagree (see Major finding below); currently latent for builtins only by coincidence |
| 2 | llama.cpp explicit-config-wins guard too broad (any `speculative.*` key) | Major | **Fixed** - `SPECULATIVE_SELECTOR_KEYS = ("speculative.types", "speculative.draft.mparams.path")` correctly narrows the guard to the two mode selectors; verified in code |
| 3 | Docs/UI claimed a llama.cpp default of "3" not present in code | Major | **Fixed** - the value is confirmed accurate (xllamacpp `CommonParams` default), and docs plus all four locales now attribute it to xllamacpp, distinguishing each engine's actual provenance |
| 4 | `except Exception` too broad in worker.py/supervisor.py | Minor | **Partial** - narrowed as claimed, but now too narrow and bypassable via `KeyError`/`IndexError` from `.format()` in the same constructor (see Major finding below) |
| 5 | sglang 0-vs-falsy and string coercion | Minor | **Partial** - the reported key is correctly fixed; the identical bug remains in the sibling `num_speculative_tokens` and is systemic across all four engines (see Major finding below) |
| 6 | llama.cpp destructive `pop` breaks retry | Minor | **Fixed for the reported scope** - `_pop_draft_options` replaced by a non-destructive `_draft_options` reader using `.get()`, and the passthrough loop now skips `DRAFT_OPTION_KEYS` rather than relying on their absence; mechanism verified correct |

Aside on #6, out of scope and explicitly not a regression from this PR: `load()` is not safely retryable regardless, because it also calls `self._llamacpp_model_config.pop("reasoning_content")` with no default elsewhere in the function, which raises `KeyError` on a second `load()` for any llama.cpp model, speculative or not. This is pre-existing and unrelated, but it means the retry scenario motivating the fix does not work end-to-end today for a separate reason.

## Tests

Re-ran the PR's unit tests: 103 passed, 2 skipped, 10 failed, up from 100 passed / 9 failed in round 2 - one additional test, which hits the same pre-existing environmental gate. All 10 failures are the same `ModuleNotFoundError`/`ImportError` for `xllamacpp` and `mlx_lm`, which are absent from the review sandbox. No new failures attributable to logic changes in this commit.

## Closing

Not ready to sign off yet, and the reason is the pattern rather than any single finding. All three new findings this round are of the form "the reported instance was fixed, the sibling instance was not," and two of them are direct regressions of round-2 fixes rather than pre-existing issues. The remaining work is small in code volume - one shared coercion helper across four engines, one widened except clause, and either reusing the per-entry annotation logic or adding a parity test - but the right next step is a deliberate sibling sweep by the author before the next push: for each of the three fixes, grep for the same shape elsewhere and fix all occurrences at once, rather than the one named in the review.

A fourth pass is warranted, and it should be scoped narrowly to verifying that sweep plus the three fixes above, not a full re-read. If the sweep lands cleanly with the suggested tests (parity across discovery paths, a malformed-placeholder registration that does not break listing, and `num_speculative_tokens=0` per engine), this should be mergeable at that point.

xinference/model/utils.py
qinxuye: Right — `any()` over every class of the engine is wrong per entry. Fixed in b35cb69f5: the annotation resolves which classes claim an entry using the same `match_json` call the non-forced branch makes (it never checks whether the library is installed, so it works on this path too) and reads the flag off those.

One deliberate exception: when nothing claims an entry it falls back to the coarse answer. That is the normal case on the forced path — it exists precisely for engines that cannot match yet — and reporting False there would hide the option for, say, a vLLM entry offered for on-demand install on a machine with no CUDA.

Added the parity test you suggested rather than only a presence check: `test_engine_discovery_paths_agree_on_draft_support` compares the value per (engine, format, size) across both entry points, for gemma-4 and for a model with no drafter. Both agree.

xinference/core/worker.py
qinxuye: Confirmed reachable and fixed in b35cb69f5, by normalizing rather than widening: the two `.format()` calls go through `_format_draft_field`, which turns KeyError/IndexError into the ValueError the call sites already expect, naming the offending field:

```
draft_model_id ...-{draft_quantization}-{quantization} carries a placeholder other than `{draft_quantization}`: KeyError(quantization)
```

That keeps the listing guards narrow and gives a better message than a bare KeyError would. Parametrized test covers both the KeyError and IndexError shapes, and asserts the listing call still returns with `draft_cache_status: []`.

xinference/core/supervisor.py
qinxuye: Covered by the same normalization — nothing further needed here.

xinference/model/llm/sglang/core.py
qinxuye: You are right that I fixed the engine-native key and left the shared one, in all four engines. b35cb69f5 replaces every one of them with a single `parse_num_speculative_tokens` in `llm/core.py`:

- `None` (or `""`) means unset, so each engine applies its own default
- a string is coerced, since the Web UI submits additional parameters as strings
- `0`, a negative, a non-numeric string, or a non-integral float raises

On the semantics: rather than passing an explicit 0 through, it is rejected. A depth of zero contradicts speculative decoding being enabled, and the four engines each did something different with it (1, 6, `None`, or skipping the assignment), so failing with a clear message beats any of them. While writing the test I found `int(1.5)` silently truncating to 1 — the same substitution in miniature — so non-integral floats are rejected too.

`test_num_speculative_tokens_validation` covers the whole table.

rogercloud (commented):
## Round 4 Re-Review

This round covers a single new commit, `b35cb69`, which addresses all three major findings from round 3. The diff is narrow (8 files, ~200 lines): a new shared `parse_num_speculative_tokens` validator in `xinference/model/llm/core.py`, per-entry class resolution via `match_json` in `_annotate_draft_support`, and `.format()` error normalization at both `cache_manager.py` call sites. A blank-slate whole-PR pass was also run independently of the prior rounds' findings; its verdict and this round's new findings follow.

### Design verdict and notes

Acceptable with reservations. Centralizing `num_speculative_tokens` validation into one function is real progress and is meaningfully different from what the previous three rounds delivered: instead of four independent per-engine patches (three of which still let `0` through silently), there is now one validator, one error-message format, and one place to get the semantics right. This is the first fix in four rounds that targets the root cause - duplicated logic - rather than the specific reported instance. That should be recognized as the correct instinct.

Two non-blocking design notes. First, `_annotate_draft_support`'s `family` and `specs` parameters are typed `Optional[Any] = None`, but both call sites always pass concrete values; this is dead flexibility and the parameters should be made required with real types. Second, the match-checking block (the `match_json` lookup, the `try`/`except`, and the result normalization) now appears three times in `xinference/model/utils.py` - at both call sites in `_force_virtualenv_engine_params` and again inside `_annotate_draft_support`. Extracting a single `_spec_matches_engine` helper is worth doing here specifically because this exact shape of duplication is what produced the two-disagreeing-computations bug that round 3 reported.

### MLX class resolution: confirmed correct

The `match_json`-based per-entry class resolution in `_annotate_draft_support` is verified to correctly distinguish `MLXChatModel` and `MLXModel` (`support_draft_model=False`) from `MLXVisionModel` (`True`) under the same engine name, which the previous engine-wide `any()` could not. The new parity test `test_engine_discovery_paths_agree_on_draft_support` genuinely exercises the mechanism rather than trivially asserting agreement: `gemma-4` and `qwen3` share the same global `MLX_CLASSES` list that the old `any()` iterated over, and `qwen3` resolving to `MLXChatModel` rather than `MLXVisionModel` demonstrates class exclusion working independently of drafter presence. This is a solid fix. The one residual gap is in tests, not code: no test constructs a model that is both chat-only and has a drafter declared, so the single most direct form of the originally reported divergence is proven by code trace and by a structurally equivalent test, but not by one exact end-to-end scenario.

### Prior findings - independently re-verified

| # | Round 3 finding | Status | Notes |
|---|---|---|---|
| 1 | `support_draft_model` computed two disagreeing ways across discovery paths | **Fixed** | Per-entry `match_json` resolution now matches the strict path; the parity test proves the fix rather than asserting agreement. |
| 2 | `except ValueError` bypassable by `KeyError`/`IndexError` from `cache_manager.py` `.format()` | **Partial** | Both call sites are wrapped and normalize to `ValueError`; the exact reported scenario is fixed and tested. The conditional gate on `draft_model_id` leaves an unguarded sibling case (see the Minor finding below), so "any malformed drafter template raises a clear `ValueError`" is not yet true. |
| 3 | 0-vs-falsy `num_speculative_tokens` across all four engines | **Partial** | The neutral parameter is now correctly validated in all four engines via the centralized function - done well and architecturally right. The sglang-native `speculative_num_draft_tokens` was not migrated (see the Major finding below), so "an explicit speculation-depth value is never silently altered" does not yet hold everywhere. |

### Test results

The PR's unit tests were re-run: 107 passed, 2 skipped, 10 failed, up from 103 passed / 10 failed in round 3, so four new tests pass. All 10 failures are the same pre-existing environmental `ModuleNotFoundError`/`ImportError` for `xllamacpp` and `mlx_lm`, which are absent from the review sandbox. No new failures are attributable to the logic changes in this commit.

### Closing

The fix quality improved this round: centralizing the validator rather than patching four call sites individually is the right instinct, and the MLX class-resolution fix is both correct and well-tested. The remaining concern is a pattern rather than a single defect - this is the fourth consecutive round where the reported instance was fixed and an unmigrated sibling of the same shape remained, and this time the sibling is the neutral-versus-engine-native variant of round 3's own finding, sitting a few lines below the fixed code in the same function. Rather than a general appeal to look harder, the shape is now specific enough to grep for: before the next push, search the PR for any remaining direct `int(...)` casts, `or DEFAULT` fallbacks, or bare `.format()` calls applied to a speculation-depth or drafter-template value without going through `parse_num_speculative_tokens` or `_format_draft_field`. With the sglang `speculative_num_draft_tokens` fix and that sweep done, the remaining gaps are narrow and well-tested enough that this should be ready to merge.

xinference/model/llm/sglang/core.py
```diff
@@ -281,11 +291,68 @@ def stop(self):
...
+            else self.DEFAULT_SPECULATIVE_NUM_DRAFT_TOKENS
+        )
+        model_config["speculative_algorithm"] = "NEXTN"
+        model_config["speculative_draft_model_path"] = draft_model_path
+        # an explicitly provided count wins, like the two keys below. Coerce the
+        # entry itself, not just the local: the Web UI submits strings, and the
+        # engine expects an int.
+        provided = model_config.get("speculative_num_draft_tokens")
+        effective_draft_tokens = (
+            int(provided) if provided is not None else num_draft_tokens
```
rogercloud: **Major**: `speculative_num_draft_tokens` (sglang's engine-native sibling of the neutral `num_speculative_tokens` handled a few lines above via the new `parse_num_speculative_tokens`) was not migrated to that validator. It still uses a bare `int(provided) if provided is not None else fallback`, so an explicit `speculative_num_draft_tokens=0` passes through as `0` - and `speculative_num_steps` below is then derived as `max(1, 0 - 1) = 1`, a self-contradictory "propose 0 draft tokens, budget 1 step" configuration. A non-integral float like `2.7` is also silently truncated to `2` by the bare `int()` cast, exactly the class of silent substitution `parse_num_speculative_tokens`'s own docstring says it exists to prevent. No test in this file covers `0` or a float for this key.

Suggest routing `speculative_num_draft_tokens` through `parse_num_speculative_tokens` as well, with a matching test for 0/float rejection.

xinference/model/llm/cache_manager.py
```diff
@@ -42,11 +46,114 @@ def __init__(
...
+                )
+            elif draft_quantizations and draft_quantization not in draft_quantizations:
+                raise ValueError(
+                    f"Drafter quantization {draft_quantization!r} is not available "
+                    f"for {self._model_name} ({self._model_format}, "
+                    f"{self._model_size_in_billions}b), "
+                    f"available: {draft_quantizations}"
+                )
+            self._draft_quantization = draft_quantization
+            if "{draft_quantization}" in draft_model_id:
```
rogercloud: **Minor**: this gates the new `_format_draft_field` normalization behind `if "{draft_quantization}" in draft_model_id:`, but the sibling field `draft_model_file_name_template` (line 97-107) calls `_format_draft_field` unconditionally whenever the template is truthy - no such gate. As a result, a `draft_model_id` containing some other unexpected placeholder that does NOT include the literal substring `{draft_quantization}` (e.g. `"org/model-{quantization}"` or a bare `{}`) skips this branch entirely and is used as-is as a literal HuggingFace repo id - no `ValueError`, just a confusing 404 at download time instead of a clear error at registration. The existing test only covers a bad placeholder alongside `{draft_quantization}`, not this case.

Suggest either always attempting `.format()` on both fields and letting `_format_draft_field` decide what's malformed, or adding an equivalent stray-placeholder check here for when the literal substring is absent.

rogercloud (approved):
## Approval

After four rounds of review (design pass, discovery, and independent verification of every prior finding against the code rather than against commit messages), the core feature - paired drafter checkpoints for speculative decoding across vLLM, SGLang, MLX, and llama.cpp - is sound, well-tested for the layers reachable without the optional engine dependencies, and the cache/deletion/UI-gating machinery has held up under repeated scrutiny. Approving on that basis.

Two items are still open from the round-4 pass (inline comments above). Flagging both again here so they are not lost, with emphasis on the first:

### Major - not fully closed, recommend addressing before or immediately after merge

**`xinference/model/llm/sglang/core.py:333-337` - `speculative_num_draft_tokens` still bypasses the new `parse_num_speculative_tokens` validator.** This is the fourth consecutive round where a fix closed the specific reported instance (the neutral `num_speculative_tokens` parameter, now correctly validated in all four engines) and left an identical sibling unfixed one function away. Concretely: an explicit `speculative_num_draft_tokens=0` still passes through today, producing a self-contradictory `speculative_num_draft_tokens=0, speculative_num_steps=1` configuration, and a non-integral float like `2.7` is silently truncated to `2` - exactly the class of silent substitution the new validator exists to prevent. It is not merge-blocking in the sense of breaking normal usage (it requires setting the SGLang-native key directly rather than the documented neutral option), but given the four-round pattern, it is worth closing in the very next commit rather than deferring indefinitely: route this key through `parse_num_speculative_tokens` too, and add a `0`/float rejection test for it, matching what the neutral parameter already has.

### Minor - low risk, fine to defer

**`xinference/model/llm/cache_manager.py:86` vs `:99` - asymmetric placeholder validation.** `draft_model_id`'s new `.format()` guard is gated by `if "{draft_quantization}" in draft_model_id:`, while the sibling `draft_model_file_name_template` check has no such gate. A `draft_model_id` with some other unexpected placeholder that does not contain the literal `{draft_quantization}` substring skips validation and surfaces as a confusing download-time 404 instead of a clear registration-time `ValueError`. Low risk (custom-registration edge case only), safe to fold into the same follow-up.

### Non-blocking design notes (already raised, restating for the record)

- `_annotate_draft_support`'s `family`/`specs` parameters are typed `Optional[Any] = None` but both call sites always pass concrete values - dead flexibility, make them required.
- The match-checking block (match_json lookup, try/except, result normalization) is now triplicated across `xinference/model/utils.py` - worth extracting into one `_spec_matches_engine` helper, particularly because this exact duplication shape is what caused the discovery-path divergence bug fixed this round.

None of the above blocks merging as-is; recommend a fast follow-up PR or commit for the SGLang validator gap given the recurring pattern, so a fifth round isn't needed to catch a fifth sibling.

xinference/model/llm/sglang/core.py
qinxuye: Fixed in c814dd297 — routed through `parse_num_speculative_tokens` like the neutral sibling. Two rounds in a row I migrated one of these two keys and left the other, which is what an extracted validator was supposed to stop; both now call it and there is nothing left casting a depth by hand.

`test_invalid_native_draft_token_count_is_rejected` and its neutral counterpart are parametrized over `0`, `"0"`, a negative, a non-integral float and a non-numeric string.

xinference/model/llm/cache_manager.py
qinxuye: Right, and the asymmetry was mine from the commit that added the normalization. Fixed in c814dd297 by taking the first option: both fields are formatted unconditionally and `_format_draft_field` decides what is malformed, so `org/model-{quantization}` and a bare `{}` now fail at launch instead of reaching the hub verbatim.

The quantization requirement moved with it — each field demands one only when it actually carries `{draft_quantization}`, which also drops a spurious error for a file template that has no placeholder at all.

The parametrized test gained the two shapes without `{draft_quantization}`; both fail against the previous gating.

rogercloud:
ci is failing

PR #5242, opened 2026-07-25
fix(gemma): keep quotes inside tool call arguments

qinxuye:
## Symptom

An agent that routes through a decision tool call reports that the model returned no tool call:

> Auto routing failed because the model did not return the required decision tool call. Please retry.

The model in fact returned exactly what was asked. Here is the raw output from `gemma-4` served by the MLX engine:

```
<|tool_call>call:select_execution_pattern{action:<|"|>final_answer<|"|>,
answer:<|"|>你好！请问有什么我可以帮您的？<|"|>,
reason:<|"|>The user said "你好" (Hello), which is a simple greeting. A conversational response is appropriate.<|"|>,
requires_current_or_external_facts:false}<tool_call|>
```

## Cause

Gemma delimits string arguments with `<|"|>` rather than quoting them, so a value is free to contain double quotes of its own. `_parse_arguments` replaced those delimiters with `"` textually and handed the result to `json.loads`:

```
Failed to parse Gemma tool call: ..., error: Expecting ',' delimiter: line 1 column 163 (char 162)
```

Character 162 is the `"` in `said "你好"`. The block then falls back to being returned as plain text, so callers relying on `tool_calls` get an empty list. Quoting the user back is what most routing and classification prompts produce, so this fires constantly rather than at the margins.

Textual replacement had a second consequence: key quoting ran over the whole string, so a value like `a,b:c` had its inner `b:` rewritten into a key.

## Fix

Rewrite the argument block by scanning it: text between two `<|"|>` tokens is re-encoded with `json.dumps`, and bare keys are quoted only in the structural text between strings. Values may then contain quotes, backslashes, newlines, or anything that looks like an argument. An unterminated string is reported as text, as before.

## Tests

Four new cases, all against `GemmaToolParser` directly:

- a value containing double quotes (the reported failure, with the payload above)
- a value containing `a,b:c`, a backslash and a newline
- an empty string (`<|"|><|"|>`) and a repeated key
- an unterminated string, which must stay text rather than become a bogus tool call

The first two fail against the previous implementation. Full tool-parser suite: 70 passed.

### DIFF
--- a/xinference/model/llm/tool_parsers/gemma_tool_parser.py
+++ b/xinference/model/llm/tool_parsers/gemma_tool_parser.py
@@ -22,14 +22,12 @@ class GemmaToolParser(ToolParser):
     def __init__(self):
         self.tool_call_start_token = "<|tool_call>"
         self.tool_call_end_token = "<tool_call|>"
+        self.string_token = '<|"|>'
+
         self.tool_call_regex = re.compile(
             r"(<\|tool_call\>.*?<tool_call\|>)", re.DOTALL
         )
         self.call_header_regex = re.compile(r"call\s*:\s*([^{\s]+)", re.IGNORECASE)
-
-    @staticmethod
-    def _replace_quotes(text: str) -> str:
-        return text.replace('<|"|>', '"')
 
     @staticmethod
     def _quote_keys(text: str) -> str:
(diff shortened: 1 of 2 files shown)


qinxuye:
Closing: opened without being asked. The fix lives in the speculative decoding branch instead.

PR #5243, opened 2026-07-25
feat(frontend): auto-reload shell routes on export directory change

m199369309:
## Summary

Eliminates the need to restart the supervisor after `npm run build` by adding mtime-based hot-reload for the frontend dynamic-route shell table.

## Problem

When the Web UI export is rebuilt (`npm run build`), the dynamic-route shell table is frozen at supervisor startup. The only way to pick up new route structures is to restart the supervisor process — an unnecessary operational burden.

## Solution

Two changes, both in `xinference/api/frontend_static.py`:

1. **Mtime-based lazy cache** for `shell_patterns`: the dynamic-route table is rebuilt automatically when the `dist` directory's mtime changes. Since `stage-export.mjs` wipes and recreates `dist` (`rmSync` + `cpSync`), the mtime signal is reliable.

2. **Cache-Control: no-cache** on entry documents (`.html` / `.txt`): browsers must revalidate on every request so the new hashed chunk references are always picked up. Binary assets are exempt.

## What's NOT changed

- No API / CLI / config / model changes
- No restart mechanism — zero interruption to workers, models, or actors
- `index.html` / `404.html` / `_next/*` were already read from disk per request; only the route table needed fixing

## Tests

6 new test cases in `xinference/api/tests/test_frontend_static.py`:
- `TestHotReload`: new dynamic routes appear after mtime change; shell table only rebuilds when mtime actually changes
- `TestCacheHeaders`: entry documents get `no-cache`; binary assets are exempt

All 30 tests pass (24 existing + 6 new).

## Known edge case (intentionally deferred)

If the `dist` directory is absent when the supervisor starts, `mount_frontend` returns `False` and no SPA routes are registered. A future reload API endpoint could cover this without conflicting with the mtime approach.

---

### DIFF
--- a/xinference/api/frontend_static.py
+++ b/xinference/api/frontend_static.py
@@ -24,13 +24,20 @@ Next static export emits one HTML file per route. Dynamic routes
 file (see the server wrappers under ``frontend/src/app``); the matching client
 page reads the real value from the URL. This module reconstructs the route
 table from those files so an arbitrary URL resolves to the correct HTML shell.
+
+Hot-reload: the dynamic-route shell table is cached but rebuilt automatically
+when the export directory changes (mtime), so a fresh ``npm run build`` is
+served without restarting the supervisor. Entry documents (HTML / RSC ``.txt``)
+carry ``Cache-Control: no-cache`` so browsers always revalidate them.
 """
 
 from __future__ import annotations
 
 import logging
 import re
+import threading
 from pathlib import Path
+from typing import TypedDict
 
 from fastapi import FastAPI
 from fastapi.responses import FileResponse, JSONResponse, RedirectResponse, Response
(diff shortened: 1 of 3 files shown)


xinference/api/frontend_static.py
```diff
@@ -115,11 +130,36 @@ def mount_frontend(app: FastAPI, dist_dir: Path) -> bool:
...
+    except OSError:
+        shell_mtime = None
+    shell_state = {"mtime": shell_mtime, "patterns": _build_shell_patterns(dist_dir)}
+
+    def _current_shell_patterns() -> list[tuple[re.Pattern[str], Path]]:
+        try:
+            mtime = dist_dir.stat().st_mtime
+        except OSError:
+            mtime = None
+        if mtime != shell_state["mtime"]:
```
qinxuye, line 164: The root directory mtime is not a reliable build-completion signal. `stage-export.mjs` removes `dist` and then recursively copies into it, so a request can reach this branch after the root/top-level route directories exist but before nested `__shell__.html` files have been copied. That request caches an incomplete pattern list; copying the remaining nested files only changes the child directory mtime, so the root mtime can stay unchanged and the new route remains 404 until another build. I reproduced this by triggering a request after creating the top-level route directory, then writing its shell file: the next request still returned 404. Please invalidate from a completion marker written after `cpSync`, or stage into a temporary directory and switch only after the copy completes, and add a regression test for a request during staging.

qinxuye: The latest update fixes the synchronous-handler, locking, and mypy issues, but it does not address this staging race. `stage-export.mjs` still uses `rmSync` followed by recursive `cpSync`, while the cache still watches only the root directory mtime. I reran the same reproduction on `a31ee609`: after a request scanned the partially copied tree, adding the nested `__shell__.html` left the root mtime unchanged and the next request still returned 404. Please keep this thread open and add a build-completion signal or atomic staging/swap, together with the mid-copy regression test described above.

qinxuye (approved):
LGTM

PR #5244, opened 2026-07-25
fix(i18n): change SSO button text from Keycloak to generic SSO

m199369309:
## Summary

Changes the SSO login button text on the login page from vendor-specific 'Keycloak' to the provider-neutral 'SSO' across all 4 supported languages.

## Problem

The login page SSO button text hardcodes 'Keycloak' (introduced in #5218), but the backend OIDC client in `xinference/api/oauth2/advanced/oidc.py` is a generic OIDC implementation — the issuer is configured via `XINFERENCE_OIDC_ISSUER` and can point to any OIDC-compliant identity provider. The UI text should not imply a vendor lock-in that doesn't exist.

## Changes

4 files, 1 line each:

| File | Before | After |
|---|---|---|
| `frontend/src/i18n/locales/zh.ts` | `使用 Keycloak 登录` | `使用 SSO 登录` |
| `frontend/src/i18n/locales/en.ts` | `Sign in with Keycloak` | `Sign in with SSO` |
| `frontend/src/i18n/locales/ja.ts` | `Keycloak でログイン` | `SSO でログイン` |
| `frontend/src/i18n/locales/ko.ts` | `Keycloak으로 로그인` | `SSO로 로그인` |

## What's NOT changed

- i18n key name (`login.ssoLogin`) — only the string value
- Button visibility logic, click handler, redirect URL
- Backend OIDC flow

---

### DIFF
--- a/frontend/src/i18n/locales/en.ts
+++ b/frontend/src/i18n/locales/en.ts
@@ -652,7 +652,7 @@ const en = {
     mustChangePasswordDesc: 'You need to change your password before continuing on first login.',
     changePasswordUserMissing: 'Unable to identify the current user. Please log in again.',
     or: 'or',
-    ssoLogin: 'Sign in with Keycloak',
+    ssoLogin: 'Sign in with SSO',
   },
   setup: {
     welcome: 'Welcome to Xinference',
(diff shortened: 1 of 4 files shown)


qinxuye (approved):
LGTM

PR #5245, opened 2026-07-25
feat(vllm): populate logprobs/top_logprobs in completion responses

SuperMarioYL:
## Summary

Populate `logprobs` / `top_logprobs` in vLLM `/v1/completions` responses. Clients that request `logprobs: true` / `top_logprobs: N` currently receive `null` even though the `CompletionLogprobs` TypedDict and the `CompletionChoice.logprobs` field already exist. This resolves the maintainer-authored `# TODO: support logprobs.` markers at `xinference/model/llm/vllm/core.py:1228` and `:1254`.

## Problem

The vLLM engine hardcodes `logprobs=None` at the completion response construction sites:

```python
# xinference/model/llm/vllm/core.py
logprobs=None,  # TODO: support logprobs.
```

These TODOs were authored by maintainers, signalling this is wanted work. `generate_config["logprobs"]` (bool) and `["top_logprobs"]` (int) already flow from the `/v1/completions` request into `generate_config` (they are not in the `create_completion` exclude set at `restful_api.py:1167`), but the sanitizer never forwarded them to vLLM `SamplingParams`, and the converters hardcoded `None` instead of reading `output.logprobs`.

## Root cause

1. `_sanitize_generate_config` never translated the request's `logprobs`(bool)/`top_logprobs`(int) into vLLM's native `logprobs`(int), so the engine never produced logprobs.
2. `_convert_request_output_to_completion` and `_convert_request_output_to_completion_chunk` hardcoded `logprobs=None` instead of reading `output.logprobs`.

## Changes

- `xinference/model/llm/vllm/core.py`
  - `_sanitize_generate_config`: translate `logprobs: bool` + `top_logprobs: int` into vLLM's native `logprobs` (int). `logprobs=true` + `top_logprobs=N` → `N`; `logprobs=true` alone → `1`; not requested → `0` (vLLM disables). `prompt_logprobs` passed through unchanged.
  - New staticmethod `_build_logprobs(output)` maps vLLM's `output.logprobs` (`List[Dict[int, Logprob]]`) into the existing `CompletionLogprobs` shape (`text_offset` / `tokens` / `token_logprobs` / `top_logprobs`). Returns `None` when the engine produced no logprobs (preserves the common-path behaviour).
  - Both completion converters now call `_build_logprobs(output)` instead of hardcoding `None`, resolving the two `# TODO: support logprobs.` markers.
- `xinference/model/llm/vllm/tests/test_core_chat_model.py`
  - New `TestVLLMModelLogprobs` covering `_build_logprobs` (None / populated / malformed-token-no-`str(tid)`-key), both completion converters (carries / None-when-absent), and the sanitizer translation.

## What's NOT changed

- **Chat-completion logprobs** — `_to_chat_completion` is untouched. The chat logprobs response uses a different (chat-format) shape than the legacy `CompletionLogprobs` used by `/v1/completions`; mapping vLLM output to that shape is a larger, separate change and is intentionally left as a follow-up to keep this PR reviewable.
- **Streaming chunk-by-chunk logprobs stitching** — the streaming path populates per-chunk logprobs from the engine but does not stitch a cumulative accumulator; that is a follow-up.
- **Other engines** — `sglang` and `llama_cpp` have the same `logprobs=None` hardcode and are addressed in separate PRs.
- **`xinference/types.py`** — no type changes; the TypedDict and field already exist.
- No public API signature, config, CLI, or dependency changes.

## Tests

7 new tests in `TestVLLMModelLogprobs`. Verified **red-on-master / green-on-branch** by stashing the source: on master the populated/converter tests fail (`_build_logprobs` does not exist + converter hardcodes `None`); on this branch they pass. The function under fix is not mocked (only the vLLM `RequestOutput` dependency is). The `test_sanitize_translates_logprobs_request_to_vllm_int` test requires a vLLM install to run, like the existing `_sanitize_generate_config` tests in this file (`core.py:1099-1100` reads `VLLM_VERSION`).

## Compatibility

- When `logprobs` is not requested (the common path), behaviour is identical to master: `_build_logprobs` returns `None`, so `choice["logprobs"]` is `None` exactly as before.
- No public API change. The `CompletionChoice.logprobs` field is already declared `Optional[CompletionLogprobs]` in `types.py:159`.
- No overlap with the in-flight speculative-decoding work (#5241): that branch touches `vllm/core.py` model-config / drafter wiring (`_sanitize_model_config` ~`:974`, `_apply_draft_model`, drafter-config fields), not the response-converter region (`_convert_request_output_to_completion*` ~`:1294`/`:1321`) or the `_sanitize_generate_config` sampling-param setdefaults modified here. Rebase should be clean.

## Reviewer pre-emption

- "Unsolicited?" — no: this implements the maintainer-authored `# TODO: support logprobs.` markers and aligns with the repo's documented chat-completions-compatible API commitment (README "Hot Topics" + the `openai-api` topic). It is the completions-side slice, not a speculative-decoding feature.
- "Wrong logprobs shape?" — the completions path uses the `CompletionLogprobs` shape that `CompletionChoice` already declares (`types.py:159`), so the type is correct by construction. The chat path (which would need a different shape) is explicitly deferred.

### DIFF
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -1264,8 +1264,11 @@ class VLLMModel(LLM):
                 getattr(sampled, "decoded_token", None) if sampled is not None else None
             )
             token_text = sampled_decoded if sampled_decoded else ""
-            token_lp = (
+            raw_token_lp = (
                 getattr(sampled, "logprob", None) if sampled is not None else None
+            )
+            token_lp = (
+                max(float(raw_token_lp), -9999.0) if raw_token_lp is not None else None
             )
             tokens.append(token_text)
             token_logprobs.append(token_lp)
@@ -1274,7 +1277,7 @@ class VLLMModel(LLM):
                 decoded_token = getattr(lp, "decoded_token", None)
                 logprob = getattr(lp, "logprob", None)
                 if decoded_token is not None and logprob is not None:
-                    decoded_logprobs[decoded_token] = float(logprob)
+                    decoded_logprobs[decoded_token] = max(float(logprob), -9999.0)
             top_logprobs.append(decoded_logprobs)
             text_offset.append(offset)
             if token_text:
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -53,6 +53,7 @@ from ....types import (
     Completion,
     CompletionChoice,
     CompletionChunk,
+    CompletionLogprobs,
     CompletionUsage,
     LoRA,
 )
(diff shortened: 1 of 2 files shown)


qinxuye (approved):
LGTM

PR #5249, opened 2026-07-27
Consolidate packaging configuration into pyproject.toml

OliverBryant:
## What do these changes do?

Consolidates the three overlapping packaging configurations (`setup.py`, `setup.cfg`, `pyproject.toml`) into a single PEP 621 `pyproject.toml`, and removes `setup.py`, `setup.cfg` and `versioneer.py`.

### Packaging
- All metadata, dependencies, optional extras and the six console-script entry points move to `[project]` tables. The `all` extra now uses self-referencing extras (`xinference[transformers]`, ...) instead of setup.cfg `%()s` interpolation.
- **Versioning**: versioneer is replaced by setuptools-scm. `xinference/_version.py` is generated at build/editable-install time and is no longer tracked; `xinference/__init__.py` falls back to `importlib.metadata` when the file is absent. Tagged releases produce identical versions; only unreleased dev builds change format (`3.0.0+33.gdc24023` → `3.0.1.dev33+gdc240239f`).
- **Web UI build**: the old `python setup.py build_web` command becomes a standalone `build_web.py` (now used by the release workflow and Dockerfiles) plus a small in-tree PEP 517 backend (`build_backend.py`) that runs it before wheel/sdist/editable builds. `NO_WEB_UI=1` still skips it. Note: `pip install -e .` now reliably triggers the Web UI build — the old `develop`-command hook was silently bypassed by modern pip's PEP 660 editable path. CI test jobs set `NO_WEB_UI=1` to keep installs fast.
- **Manifest**: setuptools-scm's file finder would otherwise ship every git-tracked file; MANIFEST.in is tightened so the wheel file list is byte-for-byte identical to the previous build (verified by diffing wheels built from this branch vs main), and the sdist differs only by the removed/added build files.
- The unused `[tool.cibuildwheel]` section is dropped (no C extensions, no CI consumer).

### Linting
- flake8 is replaced by ruff (flake8 cannot read pyproject.toml). The previous flake8 `select` list is mapped one-to-one; E901/E999 are omitted (ruff always reports syntax errors) and E127/W601-W604 have no ruff counterpart. `ruff check .` is clean at the pinned version, after fixing the handful of violations it surfaced (redundant imports, stray whitespace in doc scripts).
- isort/mypy/codespell/coverage configuration moves to `[tool.*]` tables; pre-commit hooks no longer point at setup.cfg.
- black now infers `target-version` from `requires-python = ">=3.10"`, which reformats three files to parenthesized context managers — included here since the pre-commit black hook enforces it.

### Docs / CI
- `--cov-config=setup.cfg` → `--cov-config=pyproject.toml`; GPU cache key and full-run path filter updated.
- Docs referencing setup.py/setup.cfg updated, including the affected entries in all nine doc locales (a pre-existing duplicated `msgstr` line that made `zh_CN/getting_started/troubleshooting.po` unparsable by gettext tooling is also fixed).

### Verification
- `pre-commit run --all-files`: all hooks pass except mypy, which fails identically on current main (pre-existing `lmdeploy/core.py` union-attr errors, environment-dependent).
- Wheel built from this branch has a file list identical to one built from main; METADATA extras/dependencies and entry points match setup.cfg.
- sdist → wheel rebuild without git metadata resolves the version from PKG-INFO (not the fallback).
- `pip install -e . --no-deps` in a fresh venv: correct version, all six console scripts installed.
- A CPU-only test module passes with `--cov-config=pyproject.toml`.
- Docker image builds and the release workflow can only be fully validated by CI.


xinference/__init__.py
```diff
@@ -20,9 +20,13 @@
...
 
-from . import _version
+try:
+    # generated by setuptools-scm during build/editable install
+    from ._version import __version__
+except ImportError:
+    from importlib.metadata import version as _package_version
 
-__version__ = _version.get_versions()["version"]
+    __version__ = _package_version("xinference")
```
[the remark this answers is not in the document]
OliverBryant: Fixed in ef7fbe10f — the version lookup now falls back to `0.0.0+unknown` when neither the generated `_version.py` nor installed distribution metadata is available (this was also the cause of the lint CI failure, where the frontend-static tests import xinference straight from the source tree).

build_web.py
```diff
@@ -0,0 +1,66 @@
...
+)
+_commands = [
+    ["npm", "ci"],
+    ["npm", "run", "build"],
+]
+
+
+def build_web():
+    if int(os.environ.get("NO_WEB_UI", "0")):
+        return
```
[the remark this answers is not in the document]
OliverBryant: Fixed the env parsing in ef7fbe10f — `NO_WEB_UI` now accepts common boolean strings (`true`/`yes`/`on`) instead of failing in `int()`.

On the sdist point: rebuilding the Web UI when building a wheel from an sdist matches the previous behavior (the old `setup.py` hooks also re-ran `build_web` on install/sdist), and skipping when `xinference/ui/web/dist` already exists would silently reuse stale UI assets in development checkouts, where editable installs are documented to refresh the UI. `NO_WEB_UI=1` remains the escape hatch for environments that ship the pre-staged export, which is exactly what the Dockerfiles do.

xinference/deploy/docker/Dockerfile
```diff
@@ -116,8 +116,8 @@ WORKDIR /opt/inference
 # This layer is invalidated by packaging metadata changes, not ordinary source
 # edits. The complete project is installed separately below without resolving
 # dependencies again.
-COPY setup.py setup.cfg pyproject.toml versioneer.py ./
-COPY xinference/__init__.py xinference/_version.py ./xinference/
+COPY pyproject.toml build_backend.py build_web.py ./
+COPY xinference/__init__.py ./xinference/
```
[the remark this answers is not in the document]
OliverBryant: This repository's `.dockerignore` deliberately does not exclude `.git`, so the final `COPY . /opt/inference` brings full git metadata and setuptools-scm resolves the real version there. The `fallback_version = "0+unknown"` only applies to the metadata-skeleton layer above, whose install is immediately replaced by the full-source `--no-deps --reinstall` install below, mirroring how versioneer behaved in the same layer before this change. The docker-cd workflow checks out with `fetch-depth: 0`, so tags are available for the version computation; the image build itself is validated by CI rather than locally.

OliverBryant:
CI status note: the Windows failures are pre-existing on `main` and unrelated to this PR — `test_build_subpool_envs_for_virtual_env_enabled` fails identically on the two most recent main runs ([30322057588](https://github.com/xorbitsai/inference/actions/runs/30322057588), [30281374250](https://github.com/xorbitsai/inference/actions/runs/30281374250)): the test monkeypatches `sys.platform = "linux"` but the code builds the nvidia lib paths with `os.path.join`, which still produces backslashes on a Windows host. All other jobs (lint, ubuntu/macos 3.10–3.13, GPU, metal, docs) pass on this branch.

pyproject.toml
```diff
@@ -1,9 +1,300 @@
...
+include = ["xinference*"]
+exclude = [
+    "*.conftest*",
+    "*.tests.*",
+    "*.tests",
+]
+
+[tool.setuptools_scm]
+version_file = "xinference/_version.py"
+fallback_version = "0+unknown"
```
OliverBryant: [P2] Preserve version inference for Git archives

Versioneer's tracked `_version.py` was marked `export-subst`, so GitHub/git archives carried revision data. After deleting that file, the repository has neither `_version.py` nor `.git_archival.txt`; a Git archive also has no `.git` or `PKG-INFO`, so builds from a source ZIP or commit archive hit this fallback and identify as `0+unknown`. This differs from the verified sdist path, because an sdist contains `PKG-INFO`. Please add setuptools-scm archival metadata (`.git_archival.txt` plus `.gitattributes` `export-subst`) or otherwise preserve archive version inference. See: https://setuptools-scm.readthedocs.io/en/stable/usage/#setting-up-git-archival-support

xinference/api/routers/admin.py
```diff
@@ -63,7 +63,16 @@ async def get_cluster_device_info(
...
-        data = get_versions()
+        # keep the response keys the versioneer-era API exposed; the commit id
+        # comes from the setuptools-scm generated _version module when present
+        try:
+            from ..._version import commit_id
+        except ImportError:
+            commit_id = None
+        data = {
+            "version": __version__,
+            "full-revisionid": commit_id.lstrip("g") if commit_id else None,
```
OliverBryant: [P2] Keep `full-revisionid` compatible

The generated setuptools-scm `commit_id` is a node identifier such as `g1a2b3c4d5`, so stripping `g` returns a short hash rather than the 40-character SHA that Versioneer exposed under `full-revisionid`. It can also be `None` when a wheel is rebuilt from an sdist, where only the version may be recovered from `PKG-INFO`. That changes the semantics of an existing API field. Please persist the full revision in the build artifact (or otherwise preserve the old response contract) and add a test that asserts a 40-character hexadecimal value for a normal VCS build. See: https://setuptools-scm.readthedocs.io/en/stable/usage/#node-id-prefixes

OliverBryant (commented):
Two packaging/version compatibility issues found; details are inline.

pyproject.toml
OliverBryant: Fixed in 96dfb72d9 — added `.git_archival.txt` with the setuptools-scm archival template and switched the `.gitattributes` export-subst entry from the deleted versioneer `_version.py` to it. Verified by building a wheel from `git archive HEAD`: the archival placeholders expand (node + describe-name) and the build resolves `3.0.1.dev40+g96dfb72d9` instead of the `0+unknown` fallback.

xinference/api/routers/admin.py
OliverBryant: Fixed in 96dfb72d9 — the in-tree build backend now records the full 40-character SHA (`git rev-parse HEAD`) into `xinference/_commit.py` before wheel/sdist/editable builds, and `/v1/cluster/version` serves it as `full-revisionid`, restoring the versioneer-era contract. The file ships in the sdist, so a wheel rebuilt from an sdist without git metadata keeps the recorded full SHA (verified). The setuptools-scm short node remains a last-resort fallback for artifacts built with neither git nor a recorded file (e.g. straight from a git archive). Added `test_get_cluster_version_full_revisionid`, which asserts the 40-character hexadecimal value for normal VCS builds.

build_backend.py
```diff
@@ -0,0 +1,77 @@
...
+from build_web import build_web
+
+_repo_root = os.path.dirname(os.path.abspath(__file__))
+_commit_file = os.path.join(_repo_root, "xinference", "_commit.py")
+
+
+def _record_full_revision():
+    try:
+        sha = subprocess.run(
+            ["git", "rev-parse", "HEAD"],
```
OliverBryant: [P2] Do not take the revision from an enclosing repository

The archive version fix works: I built a wheel from `git archive` and got `3.0.1.dev40+g96dfb72d9` instead of `0+unknown`. However, the full-revision fix still has two related gaps:

1. That archive wheel contains no `_commit.py`, and its generated `_version.py` has `commit_id = 'g96dfb72d9'`, so `/v1/cluster/version` still returns a short SHA for this supported build path.
2. `git rev-parse HEAD` walks up parent directories. I extracted the archive under an unrelated Git checkout and called `_record_full_revision()`; it wrote the enclosing repository's `7c772e1d1a472e1ff419adc046bbcd53cd3bca0c` into the package, silently attributing the artifact to the wrong source.

Please first verify that `git rev-parse --show-toplevel` resolves exactly to `_repo_root` before trusting Git. When the source itself is not a checkout, the expanded `.git_archival.txt` already contains the full `node: $Format:%H$`, so that can supply the 40-character SHA while preserving an existing `_commit.py` from an sdist. Tests should cover both a Git archive and an archive nested inside another repository.

OliverBryant (commented):
One residual full-revision compatibility issue found while verifying the fix; details are inline.

build_backend.py
OliverBryant: Both gaps fixed in 8c9d77db1:

1. `_record_full_revision` now falls back to the `node:` value of the export-subst expanded `.git_archival.txt`, so wheels built from a git archive also carry the full 40-character SHA (verified: a wheel built from `git archive HEAD` records exactly the HEAD SHA). Unexpanded `$Format$` placeholders in a plain checkout never match the hex pattern and are ignored.
2. Git is now trusted only when `git rev-parse --show-toplevel` resolves (realpath/normcase) to the project root itself, so a source tree extracted inside an unrelated repository is no longer attributed to the enclosing checkout's HEAD.

Added `xinference/tests/test_build_backend.py` covering the scenarios you described: own checkout, git archive, archive nested inside another repository, unexpanded placeholders, and the sdist-preserved record.

OliverBryant:
Follow-up on the Windows CI note: in the latest run, windows 3.10 additionally hit `test_opt_pytorch_model[none]` (attention-mask length mismatch). That test passed on the previous run of this same PR with identical model-runtime code (this branch only touches packaging/build files), so it appears flaky rather than related. The path-separator failure in `test_build_subpool_envs_for_virtual_env_enabled` remains the pre-existing main issue on both Windows jobs; everything else is green.

pyproject.toml
qinxuye: This fallback breaks the Dockerfile's dependency-skeleton install. That layer copies only `pyproject.toml`, `build_backend.py`, `build_web.py`, and `xinference/__init__.py` before running `uv pip install ".[otel]"`, so there is no Git metadata and setuptools-scm uses this value. I reproduced that exact source layout: the wheel build fails in `setuptools.command.bdist_wheel` with `ValueError: invalid literal for int() with base 10: '0+unknown'`. Using `fallback_version = "0.0.0+unknown"` makes the same build succeed. Please use the numeric three-part fallback (and ideally add a regression test for the no-Git skeleton path), otherwise the main Docker image cannot build.

qinxuye (commented):
One blocking Docker build issue remains.

pyproject.toml
OliverBryant: Fixed in d68b713ea — `fallback_version` is now `"0.0.0+unknown"` (the module fallback in `xinference/__init__.py` already used this form). Reproduced your exact skeleton layout (pyproject.toml, build_backend.py, build_web.py, xinference/__init__.py, empty README.md, no git): the wheel build failed with the same `int()` ValueError before the change and now succeeds, producing `xinference-0.0.0+unknown-py3-none-any.whl`. Also added `test_fallback_version_is_wheel_compatible` as a regression guard for the numeric three-part form.

qinxuye (approved):
LGTM

PR #5250, opened 2026-07-27
FIX: worker reconnection after supervisor restart

leslie2046:
Fix worker reconnection after supervisor restart

**Summary**

- Clear cached supervisor references when a worker heartbeat fails.
- Allow the worker to discover the supervisor’s new internal address after restart.
- Restore worker re-registration and model replica recovery.
- Add a regression test for stale supervisor references.

**Testing**

- `python -m compileall`
- `git diff --check`
- Isolated heartbeat reconnection check
- Full pytest unavailable because `xoscar` is not installed locally.

### DIFF
--- a/xinference/core/tests/test_worker.py
+++ b/xinference/core/tests/test_worker.py
@@ -430,6 +430,42 @@ class DummyActorRef:
         self.address = address
 
 
+@pytest.mark.asyncio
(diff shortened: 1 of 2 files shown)


qinxuye (approved):
LGTM

PR #5252, opened 2026-07-27
feat(llm): surface logprobs in /v1/chat/completions

SuperMarioYL:
## Summary

`/v1/chat/completions` silently returned `logprobs: null` even when the client requested `logprobs`/`top_logprobs`, while `/v1/completions` already populated them for vLLM (#5245). The chat builders discarded the logprobs the engine already produced.

Root cause: `ChatModelMixin._to_chat_completion` / `_to_chat_completion_chunk` built chat choices without copying the source choice's `logprobs`, and the chat choice `TypedDict`s (`ChatCompletionChoice`, `ChatCompletionChunkChoice`) did not declare the field at all.

Fix: declare `logprobs` on both chat choice `TypedDict`s and propagate `choice.get("logprobs")` in the two real choice-bearing chat builders. The chat path already receives logprobs-bearing `CompletionChunk`s from the vLLM completions generator (`vllm/core.py` `_convert_request_output_to_completion_chunk` → `_build_logprobs`), which also slices per-chunk deltas for streaming (`_slice_logprobs`), so the chat builder needs no new generation or slicing logic — it only stops discarding what the engine already built.

Follow-up to #5245 (completions logprobs).

## Effective scope

vLLM is the only engine that builds logprobs on its chunks today, so this change is effectively vLLM-only: sglang / transformers / llama.cpp / lmdeploy continue to return `null` until they implement logprobs (tracked as a separate follow-up; the per-engine support matrix is intentionally not addressed here to keep the change minimal and additive). Non-logprobs requests are unaffected — the engine returns `None` when logprobs are not requested, so propagation is a no-op vs. current behavior.

## Shape note

Chat completions carry the legacy `CompletionLogprobs` shape (`text_offset` / `tokens` / `token_logprobs` / `top_logprobs`) — the same shape the `/v1/completions` endpoint already returns — rather than the modern OpenAI chat-logprobs schema (`content` / `bytes`). This keeps chat consistent with completions (#5245) and is the smallest correct step; aligning to the modern chat-logprobs schema is left as a follow-up so existing clients reading the completions-style shape keep working.

## Tests

Adds two unit tests in `xinference/model/llm/tests/test_utils.py`:
- `test_to_chat_completion_propagates_logprobs` — non-streaming: builds a `Completion` with a real `CompletionLogprobs` on its choice, calls `ChatModelMixin._to_chat_completion`, asserts the chat choice carries the source logprobs (equal, not just not-`None`).
- `test_to_chat_completion_chunks_propagates_logprobs` — streaming counterpart via `ChatModelMixin._to_chat_completion_chunks`.

Both use the public builder staticmethods/classmethods directly (no mocking of the function under test), are red on `main` (the builders omit the `logprobs` key → `KeyError`/`None`) and green after this change.

## Checklist

- [x] Adds a regression test that is red on `main`, green on this branch
- [x] No public API removal or rename (additive `NotRequired` field + propagation only)
- [x] No change to non-logprobs request behavior
- [x] Scope kept to the chat builders + the two `TypedDict`s; engine code untouched

### DIFF
--- a/xinference/model/llm/utils.py
+++ b/xinference/model/llm/utils.py
@@ -528,6 +528,7 @@ class ChatModelMixin:
                 {
                     "index": i,
                     "delta": delta,
+                    "logprobs": choice.get("logprobs"),
                     "finish_reason": choice["finish_reason"],
                 }
             )
@@ -812,6 +813,7 @@ class ChatModelMixin:
                 {
                     "index": i,
                     "message": message,
+                    "logprobs": choice.get("logprobs"),
                     "finish_reason": choice["finish_reason"],
                 }
             )
--- a/xinference/model/llm/utils.py
+++ b/xinference/model/llm/utils.py
@@ -100,6 +100,15 @@ def _completion_logprobs_to_chat_logprobs(
     content: List[ChatCompletionLogprob] = []
     for i, token in enumerate(tokens):
         logprob = token_logprobs[i] if i < len(token_logprobs) else None
+        # The OpenAI chat-completions schema requires ``logprob: float`` on every
+        # ``content[]`` entry; ``openai-python`` rejects ``null`` (reproduced with
+        # 1.99.9: "Input should be a valid number"). The legacy
+        # ``token_logprobs`` carries ``None`` for tokens whose logprob was not
+        # computed (e.g. the first generated token in the legacy completion
+        # shape). Rather than emit a fabricated probability, skip those tokens so
+        # ``content[]`` reports only tokens whose logprob is actually known.
+        if logprob is None:
+            continue
         raw_top = top_logprobs[i] if i < len(top_logprobs) else None
         top_list: List[ChatCompletionTopLogprob] = []
         if raw_top:
--- a/xinference/model/llm/utils.py
+++ b/xinference/model/llm/utils.py
@@ -968,9 +968,11 @@ class ChatModelMixin:
                 {
                     "index": 0,
                     "delta": d,
-                    "logprobs": _completion_logprobs_to_chat_logprobs(
-                        c["choices"][0].get("logprobs")
-                    ),
+                    # `_async_to_tool_completion_chunks` already converted the
+                    # legacy completion logprobs to the chat ``content[]`` shape
+                    # via `_to_chat_completion_chunk`; pass it through unchanged
+                    # here so a real logprob is not collapsed to ``{"content": []}``.
+                    "logprobs": c["choices"][0].get("logprobs"),
                     "finish_reason": finish_reason,
                 }
             ],
--- a/xinference/model/llm/utils.py
+++ b/xinference/model/llm/utils.py
@@ -46,10 +46,14 @@ from ...types import (
     ChatCompletionChunk,
     ChatCompletionChunkChoice,
     ChatCompletionChunkDelta,
+    ChatCompletionLogprob,
+    ChatCompletionLogprobs,
     ChatCompletionMessage,
+    ChatCompletionTopLogprob,
     Completion,
     CompletionChoice,
     CompletionChunk,
+    CompletionLogprobs,
     CompletionUsage,
 )
 from .core import chat_context_var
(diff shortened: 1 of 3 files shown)


xinference/types.py
```diff
@@ -204,6 +204,7 @@ class ChatCompletionMessage(TypedDict):
 class ChatCompletionChoice(TypedDict):
     index: int
     message: ChatCompletionMessage
+    logprobs: NotRequired[Optional[CompletionLogprobs]]
```
qinxuye: This declares the legacy Completions shape (`text_offset` / `tokens` / `token_logprobs` / `top_logprobs`) on an OpenAI-compatible Chat Completions response. Chat clients expect `logprobs.content[]`, with each item containing `token`, `bytes`, `logprob`, and `top_logprobs`. With openai-python 1.99.9, the proposed response parses with `choice.logprobs.content == None`; the legacy fields survive only as extra fields, so standard chat clients still cannot consume the requested logprobs. Please add a chat-specific logprobs type and convert the vLLM data for both streaming and non-streaming responses. There is no legacy chat-response compatibility to preserve here because this endpoint returned null/omitted logprobs before this change.

xinference/model/llm/utils.py
```diff
@@ -812,6 +813,7 @@ def _to_chat_completion(
                 {
                     "index": i,
                     "message": message,
+                    "logprobs": choice.get("logprobs"),
```
qinxuye: This propagation only runs on the no-tools path. When `tools` is present, `VLLMModel.chat` routes through `_post_process_completion` or `_async_to_tool_completion_chunks`; the former omits `logprobs`, while `_post_process_completion_chunk` hard-codes it to `None`. `CreateChatCompletion` accepts `tools`, `logprobs`, and `top_logprobs` together, so valid tool-enabled requests still lose the generated values. Please preserve the converted logprobs through those post-processors as well, or explicitly reject that parameter combination.

qinxuye (commented):
I found two remaining compatibility gaps in the chat logprobs path.

SuperMarioYL:
Thanks @qinxuye — good catches, both addressed in bf70cc9c.

1. **Chat logprobs shape** (`types.py:207`): added a chat-specific `ChatCompletionLogprobs` (`content[]` of `token`/`bytes`/`logprob`/`top_logprobs`) and convert the vLLM legacy parallel-list shape at the chat builder boundary — streaming (`_to_chat_completion_chunk`, `utils.py:591`) and non-streaming (`_to_chat_completion`, `utils.py:878`). Chat clients now get `choice.logprobs.content` populated instead of the legacy fields surviving only as extras.

2. **Tool-path preservation** (`utils.py:816`): the converted logprobs now flow through the tool post-processors too — `_post_process_completion` (`utils.py:1062`, previously omitted the field) and `_post_process_completion_chunk` (`utils.py:968`, previously hard-coded `None`). Tool-enabled requests keep the generated logprobs when requested.

Tests: updated the two propagation tests to assert the `content[]` shape and added coverage for both tool post-processors preserving logprobs (using a no-op tool-parser stub).

On CI: the two `build_test_job (windows-latest)` failures are `test_build_subpool_envs_for_virtual_env_enabled` (`LD_LIBRARY_PATH` separator `\` vs `/` on Windows) — in `xinference/core/tests/test_utils.py`, unrelated to this PR's logprobs changes. The ubuntu/macos/gpu-t4 matrix is green. Happy to dig in if you'd still like it checked, but it doesn't stem from this diff.

xinference/types.py
SuperMarioYL: Added a chat-specific `ChatCompletionLogprobs` (`content[]` of `token`/`bytes`/`logprob`/`top_logprobs`) in bf70cc9c and convert the legacy parallel-list shape at the chat builder boundary (streaming + non-streaming), so `openai-python` parses `choice.logprobs.content`.

xinference/model/llm/utils.py
SuperMarioYL: Fixed in bf70cc9c: the converted logprobs now flow through `_post_process_completion` (`utils.py:1062`, previously omitted the field) and `_post_process_completion_chunk` (`utils.py:968`, previously hard-coded `None`), so tool-enabled requests keep the generated logprobs.

xinference/model/llm/utils.py
```diff
@@ -899,7 +968,9 @@ def _post_process_completion_chunk(
                 {
                     "index": 0,
                     "delta": d,
-                    "logprobs": None,
+                    "logprobs": _completion_logprobs_to_chat_logprobs(
```
qinxuye: This double-converts logprobs on the real streaming-tools path. `_async_to_tool_completion_chunks()` first calls `_to_chat_completion_chunk()`, which already converts the legacy parallel lists into `ChatCompletionLogprobs { content: [...] }`, then passes that chat chunk here. Calling `_completion_logprobs_to_chat_logprobs()` again finds no `tokens` key and returns `{ "content": [] }`; I reproduced a valid one-token logprob becoming an empty list. Please pass `c["choices"][0].get("logprobs")` through here unchanged. The new unit test currently passes a raw `CompletionChunk` directly to this method, but this method expects a chat chunk with `delta`, so that test itself fails with `KeyError: 'delta'`; exercise the actual `_to_chat_completion_chunk()` -> `_post_process_completion_chunk()` chain instead. The full `test_utils.py` currently has four failures on this head.

qinxuye (commented):
The streaming tools path still loses logprobs, and the current test suite fails.

xinference/model/llm/utils.py
SuperMarioYL: Good catch — fixed in cdf165ba. `_post_process_completion_chunk` now passes `c["choices"][0].get("logprobs")` through unchanged, since `_async_to_tool_completion_chunks` already converts each chunk once via `_to_chat_completion_chunk`. The streaming test now feeds a real chat chunk (raw `CompletionChunk` → `_to_chat_completion_chunk` → post-processor) and asserts the one-token logprob survives rather than collapsing to `{"content": []}`.

SuperMarioYL:
Thanks for the precise trace — both points addressed in cdf165ba.

1. **Streaming tools path loses logprobs** (`utils.py:971`): the double-conversion is gone. `_async_to_tool_completion_chunks()` already converts each chunk once via `_to_chat_completion_chunk()`, so `_post_process_completion_chunk()` now passes `c["choices"][0].get("logprobs")` through unchanged instead of re-running `_completion_logprobs_to_chat_logprobs()` (which found no `tokens` key and collapsed a real logprob to `{"content": []}`). The non-streaming `_post_process_completion()` still converts, because it receives a raw `Completion` whose legacy logprobs genuinely need converting.

2. **Test suite**: the three non-streaming tool-path tests now expect `"logprobs": None` — the choice carries the field, consistent with the no-tools path. The streaming test was rewritten to feed a real chat chunk through the production order and assert a one-token logprob survives.

Local `xinference/model/llm/tests/test_utils.py` is green (30 passed). CI is re-running on the new commit.

xinference/model/llm/utils.py
```diff
@@ -59,6 +63,65 @@
...
+                        "token": top_token,
+                        "bytes": _token_logprob_bytes(top_token),
+                        "logprob": top_logprob,
+                    }
+                )
+        content.append(
+            {
+                "token": token,
+                "bytes": _token_logprob_bytes(token),
+                "logprob": logprob,
```
qinxuye, line 127: `ChatCompletionTokenLogprob.logprob` is not nullable in the OpenAI chat-completions schema. This converter can emit `null` here when the legacy `token_logprobs` entry is `None` (the new tests explicitly exercise that case). I reproduced the resulting payload failing validation with `openai-python 1.99.9`: `choices.0.logprobs.content.0.logprob: Input should be a valid number`. Please avoid emitting an entry with an unknown logprob (or map it to a valid numeric representation), make the chat type non-optional, and add a regression assertion that the response parses through the OpenAI client model.

SuperMarioYL: Good catch — fixed in b85afa61. `ChatCompletionLogprob.logprob` is now `float` (non-optional), and the converter skips tokens whose legacy `token_logprobs[i]` is `None` instead of emitting `null` (`xinference/model/llm/utils.py:103-111`). `content[]` now reports only tokens with a known logprob, so no entry fails `ChatCompletionTokenLogprob` validation. New `test_chat_logprobs_parse_openai` parses every `content[]` entry through `openai.types.chat.chat_completion_token_logprob.ChatCompletionTokenLogprob` (openai-python 1.99.9) — the exact gate that was failing — and the two propagation tests are updated to expect the dropped first token.

qinxuye: The converter fix is correct, but this head is not green yet. `pytest -q xinference/model/llm/tests/test_utils.py` still fails in `test_post_process_completion_preserves_chat_logprobs` and `test_post_process_completion_chunk_preserves_chat_logprobs`: both expected payloads still include the old `"Hello"` entry with `logprob: None`, while the converter now correctly omits it. Please update those two expectations to match the known-logprob-only shape. The lint job also fails because isort modifies `test_utils.py`, so please run isort/pre-commit on the file as well.

qinxuye (approved):
LGTM

PR #5253, opened 2026-07-28
docs(i18n): update ja/ko/pt_BR release notes links to xinference.co

martinma51:
Follow-up to #5248. Those 3 locale files were pushed after #5248 merged, so they weren't included. This updates the ja, ko, and pt_BR release-notes .po files (both msgid and translated msgstr links) from the xinference.io domain to xinference.co, matching the English source and the other locales already merged.


qinxuye (commented):
One locale is still missing from this follow-up: `doc/source/locale/zh_TW/LC_MESSAGES/getting_started/release_notes.po` still uses `xinference.io` in its `msgid` values. Since the English source now uses `xinference.co`, those catalog entries no longer match and the Traditional Chinese page falls back to the English text. Please update the `msgid` URLs to `xinference.co`; the translated `msgstr` URLs can continue pointing to `xinference.cn`.

qinxuye (approved):
LGTM

PR #5254, opened 2026-07-28
feat(webui): add missing permission scopes (models:register, logs:list, monitor:view) to UI

m199369309:
## Summary

Add three permission scopes that exist in the backend but were never exposed in the frontend UI:
- `models:register` — Register/unregister models
- `logs:list` — View logs  
- `monitor:view` — View monitoring

These scopes are already defined in `xinference/api/oauth2/advanced/auth_service.py` (`INITIAL_ADMIN_PERMISSIONS`) but were missing from `ALL_PERMISSIONS` in the frontend, making it impossible to grant them to non-admin users via the UI.

## Changes

- **`permissions.tsx`**: Add 3 missing scopes to `ALL_PERMISSIONS` and add `logs`/`monitor` groups to `PERMISSION_GROUPS`
- **`use-menu-auth.ts`**: Add `hasLogsList`, `hasMonitorView`, and `canRegisterModel` computed properties (with legacy scope alias support for `models:add` and `models:unregister`)
- **`sidebar.tsx`**: Gate log center on `hasLogsList` scope; gate monitor center on `hasMonitorView` scope; gate register model on `canRegisterModel` scope (all gated behind `auth_advanced` flag)
- **`en/zh/ja/ko.ts`**: Add i18n labels and group names for new scopes

## Backward Compatibility

- Admin users have all scopes and are unaffected
- Non-admin users without the new scopes will simply not see the log/monitor menu items (when `auth_advanced` is enabled)
- Legacy `models:add` and `models:unregister` scopes recognized via frontend alias check (matching backend `scope_aliases.py` normalization)
- No backend changes required


frontend/src/hooks/use-menu-auth.ts
```diff
@@ -22,5 +22,8 @@ export function useMenuAuth() {
     canCreateKeys: hasScope('admin', 'keys:create'),
     canManageKeys: hasScope('admin', 'keys:manage'),
     canAccessKeysPage: hasScope('admin', 'keys:create', 'keys:manage'),
+    hasLogsList: hasScope('admin', 'logs:list'),
+    hasMonitorView: hasScope('admin', 'monitor:view'),
+    canRegisterModel: hasScope('admin', 'models:register'),
```
qinxuye: The backend intentionally normalizes the legacy `models:add` and `models:unregister` scopes to `models:register` in `scope_aliases.py`, so existing users carrying either legacy scope still pass API authorization. This UI check only recognizes the new scope, which hides the Register Model menu after an upgrade even though the user remains authorized by the backend. Please normalize the aliases on the frontend as well, or include both legacy scope names in this check.

m199369309: Good catch! Fixed in a2878e3 — `canRegisterModel` now checks `models:register`, `models:add`, and `models:unregister` via the existing varargs `hasScope` pattern:

```typescript
canRegisterModel: hasScope("admin", "models:register", "models:add", "models:unregister"),
```

This matches the backend `SCOPE_ALIASES` normalization in `scope_aliases.py`, so users carrying legacy scopes in their JWT will still see the Register Model sidebar item.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -147,7 +147,7 @@ export function Sidebar() {
   const [collapsed, setCollapsed] = useState(false);
   const branding = getBrandingFromEnv();
   const { clusterVersion, clusterAuth, clusterUIConfig } = useGlobal();
-  const { isAdmin, usersManagePage, canAccessKeysPage } = useMenuAuth();
+  const { isAdmin, usersManagePage, canAccessKeysPage, hasLogsList, hasMonitorView } = useMenuAuth();
```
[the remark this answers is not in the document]
m199369309: Both issues have been addressed in 025d50b and a2878e3:

1. ✅ `hasLogsList`, `hasMonitorView`, and `canRegisterModel` are now in the `navGroups` `useMemo` dependency array
2. ✅ `/register-model` sidebar item is gated with `show: !clusterUIConfig?.auth_advanced || canRegisterModel`

The legacy scope alias issue (models:add/models:unregister → models:register) is also now handled in the `canRegisterModel` check.

frontend/src/hooks/use-menu-auth.ts
```diff
@@ -22,5 +22,7 @@ export function useMenuAuth() {
     canCreateKeys: hasScope('admin', 'keys:create'),
     canManageKeys: hasScope('admin', 'keys:manage'),
     canAccessKeysPage: hasScope('admin', 'keys:create', 'keys:manage'),
+    hasLogsList: hasScope('admin', 'logs:list'),
+    hasMonitorView: hasScope('admin', 'monitor:view'),
```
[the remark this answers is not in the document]
m199369309: Already addressed in 025d50b — `canRegisterModel` is now exported from `useMenuAuth`, and it also checks legacy scope aliases (`models:add`, `models:unregister`) to match backend `scope_aliases.py` normalization.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -219,13 +220,16 @@ export function Sidebar() {
             name: t('menu.monitorCenter'),
             Icon: Monitor,
             Extra: ChevronRight,
+            show: !clusterUIConfig?.auth_advanced || hasMonitorView,
```
qinxuye, line 223: This only hides the navigation item; `/monitor-center` itself has no scope guard. A user without `monitor:view` can navigate to that URL directly, and the page obtains the Grafana URL/dashboard IDs from the unauthenticated `/v1/cluster/ui_config` endpoint before rendering the iframe. When Grafana allows anonymous embedding, the new permission can therefore be bypassed completely. Please add a page/route-level `monitor:view` guard (as the previous UI's `PermissionGate` did), or enforce this boundary through a protected backend path rather than relying only on sidebar visibility.

m199369309: Good catch — sidebar-only gating is insufficient. Fixed in dd8ae55:

1. Created a reusable `PermissionGuard` client component (`frontend/src/components/auth/permission-guard.tsx`) that checks `auth_advanced` flag and required scope via `useMenuAuth`, rendering an access-denied page for unauthorized users.
2. Applied `PermissionGuard` with `monitor:view` scope to `/monitor-center` page.
3. Applied `PermissionGuard` with `logs:list` scope to `/log-center` page.
4. `/register-model` uses server-side `redirect()` (throws during SSR), so its guard stays at the sidebar level for now.

Note: `/v1/cluster/ui_config` lacking backend auth is a separate concern — filed for follow-up outside this PR.

frontend/src/components/auth/permission-guard.tsx
```diff
@@ -0,0 +1,45 @@
+'use client';
+
+import { ShieldAlert } from 'lucide-react';
+import { useI18n } from '@/contexts/i18n-context';
+import { useGlobal } from '@/contexts/global-context';
+import { useMenuAuth } from '@/hooks/use-menu-auth';
+
+interface PermissionGuardProps {
+  scope: 'monitor:view' | 'logs:list' | 'models:register';
```
qinxuye, line 9: `models:register` is supported by this guard but is not applied to either dynamic register-model route. The server-side redirect at `/register-model` does not protect direct URLs such as `/register-model/LLM` or `/register-model/LLM/<name>`; both client pages still render the registration form for users without this scope, and only the backend mutation eventually returns 403. Please wrap both dynamic page clients (or their page components) with `PermissionGuard scope="models:register"` so the UI permission is enforced consistently.

qinxuye (commented):
One route-level permission gap remains.

frontend/src/components/auth/permission-guard.tsx
m199369309: Good catch — both dynamic routes were indeed unprotected. Fixed in 2a48c17:

- Wrapped `/register-model/[modelType]/page-client.tsx` with `PermissionGuard scope="models:register"`
- Wrapped `/register-model/[modelType]/[modelName]/page-client.tsx` with `PermissionGuard scope="models:register"`

All three scope-gated sections (monitor, log, register-model) now have consistent route-level protection via PermissionGuard.

qinxuye (commented):
One initialization gap remains in the route guard.

frontend/src/components/auth/permission-guard.tsx
```diff
@@ -0,0 +1,45 @@
...
+  children: React.ReactNode;
+}
+
+export function PermissionGuard({ scope, children }: PermissionGuardProps) {
+  const { t } = useI18n();
+  const { clusterUIConfig } = useGlobal();
+  const auth = useMenuAuth();
+
+  // When auth_advanced is disabled, all pages are accessible
+  if (!clusterUIConfig?.auth_advanced) {
```
qinxuye, line 31: This branch fails open while the global UI configuration is still loading: `clusterUIConfig` starts as `{}`, so `auth_advanced` is temporarily falsy and the protected child is mounted. The register-model child immediately renders its form and fires its prompt/family requests, so an unauthorized user refreshing a direct route briefly enters the page before the config arrives. Please read `globalReady` from `useGlobal()` and render a loading/empty state until it is true; only then should `auth_advanced === false` bypass the scope check.

m199369309: Good catch — this is a real race window. Fixed in 3070c91:

1. Read `globalReady` from `useGlobal()` alongside `clusterUIConfig`.
2. While `globalReady` is `false`, render a centered `Loader2` spinner instead of children, preventing the premature fail-open.
3. Only after `globalReady` becomes `true` (i.e., `/v1/cluster/auth` and `/v1/cluster/ui_config` have resolved) does the guard check `auth_advanced` and the required scope.

The register-model child can no longer fire its prompt/family requests before the config arrives on a direct-route refresh.

qinxuye (dismissed):
LGTM

qinxuye (commented):
The two Windows failures are deterministic: Python 3.10 and 3.13 both fail `test_build_subpool_envs_for_virtual_env_enabled` because `os.path.join()` produces backslash-separated paths on Windows while the expected value hard-codes forward slashes. This is a baseline failure introduced by `7c772e1d1a`, not by this frontend-only PR; the latest `main` run fails the same two Windows jobs. Please rebase/rerun after the baseline test is fixed so this PR can return to green.

qinxuye (approved):
LGTM

PR #5255, opened 2026-07-28
fix(webui): improve worker node display in multi-select

m199369309:
## Summary

Two small UI improvements for the worker node multi-select in the launch model dialog:

1. **Increase badge width**: Worker addresses like `xinference-worker-6000-3:30001` (~30+ characters, ~210px) were truncated by the `max-w-[120px]` limit. Increased to `max-w-[200px]`.

2. **Show GPU count**: Worker options in the dropdown now display GPU count as a description (e.g., "GPU: 8"), helping users choose the right node. The `MultiSelect` component already supports `description` rendering.

## Changes

| File | Change |
|---|---|
| `frontend/src/components/ui/multi-select.tsx` | `max-w-[120px]` → `max-w-[200px]` |
| `frontend/src/components/pages/launch-model/utils.tsx` | Add `description: \`GPU: ${gpuCount}\`` to worker option |
| `frontend/src/components/pages/launch-model/types.ts` | Add `description?: string` to `WorkerOption` type |

## Screenshots

**Before**: Worker address truncated, no GPU count visible
**After**: Full worker address visible, GPU count shown in dropdown

### DIFF
--- a/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
+++ b/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
@@ -110,7 +110,7 @@ export default function LaunchDialog({
       const data = await request.get<ClusterInfoResponse>('/v1/cluster/info', {
         params: { detailed: true },
       });
-      setWorkerOptions(extractWorkerItems(data));
+      setWorkerOptions(extractWorkerItems(data, t));
     } catch {
       setWorkerOptions([]);
     }
--- a/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
+++ b/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
@@ -114,7 +114,7 @@ export default function LaunchDialog({
     } catch {
       setWorkerOptions([]);
     }
-  }, [clusterAuth?.auth, isAdmin]);
+  }, [clusterAuth?.auth, isAdmin, t]);
   const fetchModelEngine = useCallback(async () => {
     if (!model?.model_name || !MODEL_ENGINE_TYPES.includes(modelType)) {
       setModelEngineMap({});
--- a/frontend/src/components/pages/launch-model/types.ts
+++ b/frontend/src/components/pages/launch-model/types.ts
@@ -108,5 +108,6 @@ export type LaunchFieldConfig =
     });
 
 export type WorkerOption = Option<string> & {
+  description?: string;
   gpuCount: number;
 };
--- a/frontend/src/components/pages/launch-model/utils.tsx
+++ b/frontend/src/components/pages/launch-model/utils.tsx
@@ -29,6 +29,7 @@ import type {
   UnknownRecord,
   WorkerOption,
 } from './types';
+import type { TFunc } from '@/contexts/i18n-context';
 
 export const MODEL_ENGINE_TYPES: RequestModelType[] = [
   ModelType.LLM,
@@ -933,7 +934,7 @@ export function normalizeWorkerAddress(value: unknown) {
 
 export function extractWorkerItems(
   clusterInfo: ClusterInfoResponse,
-  t?: (key: string, vars?: Record<string, string | number>) => string
+  t: TFunc
 ): WorkerOption[] {
   if (!clusterInfo) return [];
   const isFlatNodeList = Array.isArray(clusterInfo);
@@ -955,9 +956,7 @@ export function extractWorkerItems(
     acc.set(workerIp, {
       label: workerIp,
       value: workerIp,
-      description: t
-        ? t('launchModel.gpuCount', { count: gpuCount })
-        : `GPU: ${gpuCount}`,
+      description: t('launchModel.gpuCount', { count: gpuCount }),
       gpuCount,
     });
 
--- a/frontend/src/components/pages/launch-model/utils.tsx
+++ b/frontend/src/components/pages/launch-model/utils.tsx
@@ -931,7 +931,10 @@ export function normalizeWorkerAddress(value: unknown) {
   }
 }
 
-export function extractWorkerItems(clusterInfo: ClusterInfoResponse): WorkerOption[] {
+export function extractWorkerItems(
+  clusterInfo: ClusterInfoResponse,
+  t?: (key: string, vars?: Record<string, string | number>) => string
+): WorkerOption[] {
   if (!clusterInfo) return [];
   const isFlatNodeList = Array.isArray(clusterInfo);
   const nodes = isFlatNodeList ? clusterInfo : clusterInfo.workers || [];
@@ -952,7 +955,9 @@ export function extractWorkerItems(clusterInfo: ClusterInfoResponse): WorkerOpti
     acc.set(workerIp, {
       label: workerIp,
       value: workerIp,
-      description: `GPU: ${gpuCount}`,
+      description: t
+        ? t('launchModel.gpuCount', { count: gpuCount })
+        : `GPU: ${gpuCount}`,
       gpuCount,
     });
 
--- a/frontend/src/components/pages/launch-model/utils.tsx
+++ b/frontend/src/components/pages/launch-model/utils.tsx
@@ -952,6 +952,7 @@ export function extractWorkerItems(clusterInfo: ClusterInfoResponse): WorkerOpti
     acc.set(workerIp, {
       label: workerIp,
       value: workerIp,
+      description: `GPU: ${gpuCount}`,
       gpuCount,
     });
 
--- a/frontend/src/components/ui/multi-select.tsx
+++ b/frontend/src/components/ui/multi-select.tsx
@@ -228,7 +228,7 @@ export function MultiSelect({
                 key={option.value}
                 className="flex items-center gap-1 rounded-md bg-secondary px-2 py-0.5 text-xs text-secondary-foreground"
               >
-                <span className="truncate max-w-[120px]">{option.label}</span>
+                <span className="truncate max-w-[200px]">{option.label}</span>
 
                 <button
                   type="button"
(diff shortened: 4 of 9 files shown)


frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -110,7 +110,7 @@ export default function LaunchDialog({
       const data = await request.get<ClusterInfoResponse>('/v1/cluster/info', {
         params: { detailed: true },
       });
-      setWorkerOptions(extractWorkerItems(data));
+      setWorkerOptions(extractWorkerItems(data, t));
```
OliverBryant, line 113: **Bug: missing `t` in the `useCallback` dependency array.**

`fetchWorkers` now captures `t` in its closure, but the dependency array a few lines below is still `[clusterAuth?.auth, isAdmin]`. In `i18n-context.tsx`, `t` is built with `useMemo(..., [locale])`, so its identity changes when the user switches language — but `fetchWorkers` won't be recreated, and `workerOptions` keeps the descriptions rendered in the *previous* language until something else happens to trigger a refetch.

Repro: open the launch dialog, expand the worker dropdown (descriptions render in the current locale), switch language, reopen the dropdown — the `GPU: N` descriptions are still in the old locale.

```suggestion
      setWorkerOptions(extractWorkerItems(data, t));
    } catch {
      setWorkerOptions([]);
    }
  }, [clusterAuth?.auth, isAdmin, t]);
```

(Note this class of mistake isn't caught here — `react-hooks/exhaustive-deps` doesn't appear to be enforced in this project's eslint config, so CI won't flag it.)

frontend/src/components/pages/launch-model/utils.tsx
```diff
@@ -952,6 +955,9 @@ export function extractWorkerItems(clusterInfo: ClusterInfoResponse): WorkerOpti
     acc.set(workerIp, {
       label: workerIp,
       value: workerIp,
+      description: t
+        ? t('launchModel.gpuCount', { count: gpuCount })
+        : `GPU: ${gpuCount}`,
```
OliverBryant: **Make `t` required and drop the fallback branch.**

`extractWorkerItems` has exactly one call site in the repo (`launch-dialog.tsx:113`), and `t` is unconditionally in scope there. So the `t ? ... : ...` ternary and the `` `GPU: ${gpuCount}` `` fallback are unreachable today, and the optional parameter actively invites a future call site to silently emit untranslated English.

Suggest making it required:

```ts
export function extractWorkerItems(
  clusterInfo: ClusterInfoResponse,
  t: TFunc
): WorkerOption[] {
```

and then simply:

```ts
description: t('launchModel.gpuCount', { count: gpuCount }),
```

**Separately — reuse the existing `TFunc` type instead of re-declaring it.** The inline signature added above is a hand-written structural type that doesn't match the canonical one in `contexts/i18n-context.tsx:8`:

```ts
type InterpolationValue = string | number | boolean | null | undefined;
type TFunc = (key: string, vars?: Record<string, InterpolationValue>) => string;
```

The new signature narrows `vars` to `Record<string, string | number>`, so it's not interchangeable with the real `t`. Please `export` `TFunc` from `i18n-context.tsx` and import it here rather than maintaining a second, divergent definition.

frontend/src/i18n/locales/zh.ts
```diff
@@ -106,6 +106,7 @@ const zh = {
     nGPU: 'GPU 数量 (每个副本)',
     nGPUPerWorker: '每个 Worker 上的 GPU 数量',
     nGPUDevice: '设备',
+    gpuCount: 'GPU: {{count}}',
```
OliverBryant: This is the untranslated English string. Every adjacent key in this same `launchModel` block is localized (`nGPU: 'GPU 数量 (每个副本)'`, `nGPUPerWorker: '每个 Worker 上的 GPU 数量'`, `nGpuLayers: 'GPU 层数'`), so leaving this one in English is inconsistent — and effectively means the i18n indirection buys nothing for zh users.

Suggest:

```suggestion
    gpuCount: 'GPU 数量: {{count}}',
```

Keeping `GPU` as an acronym is fine; it's the whole string being verbatim English that's the issue. Same applies to the `ja.ts` and `ko.ts` entries in this PR.

FYI, unrelated to correctness: there's already a `gpuCount` key in the `clusterInfo` section of these files (`'GPU 数量'`). Different section, so no collision — just be aware the name is now reused.

OliverBryant (commented):
Thanks for the fix — both problems are real: worker addresses like `xinference-worker-6000-3:30001` were genuinely truncated at `max-w-[120px]`, and `MultiSelect` does already render `description` (multi-select.tsx:312), so no component change was needed there. Nice follow-up on the i18n point too.

A few observations, roughly in priority order:

1. **One actual bug:** `fetchWorkers` now closes over `t`, but its `useCallback` dependency array wasn't updated — descriptions go stale after a language switch. Details inline.
2. The zh/ja/ko locale entries are all the untranslated English string, while every neighbouring key in the same block is localized.
3. The `t?:` optional parameter and its hardcoded-English fallback branch look like dead weight — there's only one call site in the repo, and it always has `t` in scope. The inline `t` signature also diverges from the existing `TFunc` type.
4. Minor: `description?` arguably belongs on the shared `Option` base type rather than on `WorkerOption`.
5. Minor: `max-w-[200px]` doesn't quite cover the width the PR description itself cites (~210px).

One note on the description body: it says "callers that pass `t` get localized descriptions, while other call sites fall back to the hardcoded English string" — there are no other call sites, so the fallback is unreachable rather than a compatibility shim.

I also wasn't able to confirm the visual result: the **Screenshots** section has before/after captions but no actual images attached, so the width fix can't be verified from the diff alone. Could you attach them?

Only #1 seems worth treating as must-fix; the rest are your call.

frontend/src/i18n/locales/ja.ts
```diff
@@ -107,6 +107,7 @@ const ja = {
     nGPU: 'レプリカあたりのGPU数',
     nGPUPerWorker: 'ワーカーあたりのGPU数',
     nGPUDevice: 'デバイス',
+    gpuCount: 'GPU: {{count}}',
```
OliverBryant: Same as the zh entry — untranslated English while the neighbouring keys are localized (`nGPU: 'レプリカあたりのGPU数'`, `nGpuLayers: 'GPUレイヤー数'`).

```suggestion
    gpuCount: 'GPU数: {{count}}',
```

frontend/src/i18n/locales/ko.ts
```diff
@@ -107,6 +107,7 @@ const ko = {
     nGPU: '레플리카당 GPU 수',
     nGPUPerWorker: '워커당 GPU 수',
     nGPUDevice: '장치',
+    gpuCount: 'GPU: {{count}}',
```
OliverBryant: Same as the zh entry — untranslated English while the neighbouring keys are localized (`nGPU: '레플리카당 GPU 수'`, `nGpuLayers: 'GPU 레이어 수'`).

```suggestion
    gpuCount: 'GPU 수: {{count}}',
```

frontend/src/components/pages/launch-model/types.ts
```diff
@@ -108,5 +108,6 @@ export type LaunchFieldConfig =
     });
 
 export type WorkerOption = Option<string> & {
+  description?: string;
```
OliverBryant, line 111: Minor / non-blocking: `description` isn't worker-specific — it's a generic `MultiSelect` capability. `MultiSelectOption` (`ui/multi-select.tsx:9`) already declares `description?: string`, but the shared `Option` type in `types/common.ts:6` does not, which is why it has to be bolted on here.

Adding `description?: string` to `Option` itself would be the cleaner fix and would benefit every select-style component. I recognize that's a slightly wider blast radius than this PR intends, so it's fine to leave as-is — but as written it sets a precedent for the next person to re-declare `description` on their own `XxxOption` type too.

frontend/src/components/ui/multi-select.tsx
```diff
@@ -228,7 +228,7 @@ export function MultiSelect({
                 key={option.value}
                 className="flex items-center gap-1 rounded-md bg-secondary px-2 py-0.5 text-xs text-secondary-foreground"
               >
-                <span className="truncate max-w-[120px]">{option.label}</span>
+                <span className="truncate max-w-[200px]">{option.label}</span>
```
OliverBryant, line 231: Minor: the chosen value doesn't match the case it's meant to fix. The PR description estimates `xinference-worker-6000-3:30001` at ~210px, but this caps at 200px — so by the PR's own arithmetic the motivating example is still clipped, just less so.

The badge's parent container is already `flex-wrap` + `overflow-hidden`, so it degrades gracefully. Either widen to something that actually covers it (`max-w-[280px]`), or drop the `max-w` constraint entirely and let wrapping handle long labels.

m199369309 (commented):
Thanks for the thorough review! All excellent points. I've pushed a fix commit (aabf925) that addresses the key findings:

**Fixed:**

1. **`useCallback` dependency array** — Added `t` to the dependency list. Good catch — this was a real bug where descriptions would go stale after a language switch.

2. **zh/ja/ko translations** — Localized all three locale entries to match the existing `clusterInfo.gpuCount` patterns:
   - zh: `'GPU 数量：{{count}}'`
   - ja: `'GPU数: {{count}}'`
   - ko: `'GPU 수: {{count}}'`

3. **`t?` optional + hardcoded fallback** — Removed both. `t` is now required, and I exported `TFunc` from the i18n context so the signature matches the canonical type instead of a divergent inline one.

4. **`TFunc` type** — Exported `TFunc` from `@/contexts/i18n-context` and imported it in `utils.tsx`.

**Skipped (not worth the churn):**

- **`description?` on `Option` base type** — Agree it's a valid code-organization point, but `MultiSelectOption` already has its own `description`, and changing the shared `Option<T>` would touch many unrelated files. Prefer to keep this PR scoped.

- **`max-w-[200px]` vs ~210px** — The `truncate` class adds ellipsis, so 200px vs 210px is a negligible visual difference. The 200px value is already a 67% improvement over the original 120px.

Re screenshots: I wasn't able to attach them to the PR body — the `gh` CLI renders captions but not images. I'll add them via the GitHub UI.

Thanks again for the careful review!

qinxuye (approved):
LGTM

PR #5257, opened 2026-07-28
fix(model): avoid Set-changed-during-iteration crash and gc.get_objects() overhead under concurrent downloads

SuperMarioYL:
## Summary

`CancellableDownloader` monkeypatches the **global** `tqdm.update`/`tqdm.tqdm.update` with `patched_update`. Under concurrent model launches (`XINFERENCE_MAX_CONCURRENT_LAUNCHES`, default `5`), that global hook has two problems on the default config:

1. **`RuntimeError: Set changed size during iteration`** — `patched_update` (running in the download thread) calls `progresses.add(tqdm_instance)` on each active downloader's `_main_progresses` / `_download_progresses`, while the progress thread (`_upload_download_progress` → `get_progress()`) iterates the same set. When a new download-shard bar is added mid-iteration, the iterator raises. This is reachable on the default config with ≥2 concurrent launches.
2. **Per-tick `gc.get_objects()` whole-heap scan** — `patched_update` traversed the entire Python heap to find active `CancellableDownloader` instances on *every* `tqdm.update` call, an O(heap) scan per progress tick (vs O(active) after this change).

This PR fixes both with a minimal, in-idiom change confined to the `CancellableDownloader` class (it already uses `threading.Lock` for `_global_lock` / `_patch_lock`; this mirrors that discipline).

## Changes

- **Crash fix** — add a per-instance `self._progress_lock = threading.RLock()`. In `get_progress()`, snapshot the progress sets via `list(...)` under the lock before iterating; in `patched_update()`, guard each `progresses.add(tqdm_instance)` with `with downloader._progress_lock:`. This serializes the mutation against the snapshot taken in `get_progress()`, eliminating the `Set changed size during iteration` race. Snapshot-then-iterate is intentional eventual consistency for a polled progress reporter (a bar added after the snapshot is missed in that call but present in the next, since it persists in the set).
- **Scan-cost fix** — replace the `gc.get_objects()` whole-heap traversal with a bounded class-level registry `CancellableDownloader._active_registry`, maintained under the existing `_global_lock` (`__enter__` adds `self`, `__exit__` discards `self`). `patched_update` now iterates `list(CancellableDownloader._active_registry)` — O(active) instead of O(heap). The registry is a correct strict subset of what `gc.get_objects()` returned (active instances only), preserving the intended "act on active downloaders" semantics.

## Out of scope (intentional)

`patched_update` still calls `raise_error()` if **any** active downloader is cancelled (cross-fire cancel: cancelling downloader A can raise in downloader B's download thread). Properly fixing that requires attributing each `tqdm` instance to its owning downloader, which is a larger redesign of the global-patch model and is left as a follow-up. This PR deliberately keeps the scope to the reproducible crash + the per-tick scan cost.

## Test

Adds `test_concurrent_progress_no_set_mutation`: two `CancellableDownloader`s entered concurrently; one thread creates new download `tqdm` bars (`unit="B"`) and calls `.update()` (growing `_download_progresses` via the global `patched_update`), another thread polls `get_progress()`. Asserts no `RuntimeError` is raised. Uses only public API (`from ..utils import CancellableDownloader`, real `tqdm`, `threading`) — no mocking of the code under fix. A `threading.Event` stop + `join(timeout=…)` guard prevents CI hangs.

The reproduction is timing-dependent: under CPython's default ~5 ms switch interval the set mutation rarely lands mid-iteration, so the test tightens `sys.setswitchinterval(1e-4)` (saved and restored in `try/finally`) to force frequent GIL handoffs so the mutation coincides with the poller's iteration. Under that enforced interval the test is **red on `main`** (raises `RuntimeError: Set changed size during iteration`) and **green on this branch**.

## Why this shape

- Mirrors the class's existing lock discipline (`_global_lock` / `_patch_lock`) — no new abstraction introduced.
- Small surgical PR (~85 LoC, 2 files), matching the repo's median PR size.
- The crash is a `RuntimeError` on the default config (not a theoretical edge case), so the reproduction is self-justifying.

### DIFF
--- a/xinference/model/utils.py
+++ b/xinference/model/utils.py
@@ -924,8 +924,13 @@ class CancellableDownloader:
         self._patched_instances: Set[int] = set()
 
     def reset(self):
-        self._main_progresses.clear()
-        self._download_progresses.clear()
+        # Hold _progress_lock around both clears so a concurrent get_progress()
+        # caller (e.g. the progress-upload thread that already passed its done
+        # check) cannot be mid-iteration when the sets are cleared — that race
+        # raised "Set changed size during iteration".
+        with self._progress_lock:
+            self._main_progresses.clear()
+            self._download_progresses.clear()
 
     def get_progress(self) -> float:
         if self.done:
@@ -990,19 +995,26 @@ class CancellableDownloader:
         raise self._cancel_error_cls(error_msg)
 
     def patch_tqdm(self):
-        # Use class-level patching to avoid conflicts
+        # Use class-level patching to avoid conflicts. Route the bookkeeping
+        # through the class (type(self)), not `self`: assigning to
+        # self._original_update creates a per-instance attribute that shadows
+        # the class attribute, so two concurrent downloaders each observe the
+        # class-level None, patch tqdm independently, and keep their own
+        # originals — the first to exit then restores tqdm.update while the
+        # second is still active.
         with self._patch_lock:
             import tqdm as tqdm_module
+            cls = type(self)
 
-            if self._original_update is None:
-                self._original_update = tqdm.update
-            if self._original_update_plain is None:
-                self._original_update_plain = tqdm_module.tqdm.update
+            if cls._original_update is None:
+                cls._original_update = tqdm.update
+            if cls._original_update_plain is None:
+                cls._original_update_plain = tqdm_module.tqdm.update
 
-            if self._original_update is None or self._original_update_plain is None:
+            if cls._original_update is None or cls._original_update_plain is None:
                 return
 
-            original_update_plain = self._original_update_plain
+            original_update_plain = cls._original_update_plain
 
             # Thread-safe patched update
             def patched_update(tqdm_instance, n):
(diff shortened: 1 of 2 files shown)


xinference/model/utils.py
```diff
@@ -1046,12 +1057,14 @@ def __enter__(self):
             if self._active_instances == 0:
                 self.patch_tqdm()
             self._active_instances += 1
+            CancellableDownloader._active_registry.add(self)
```
qinxuye: The global patch bookkeeping is still being updated through `self`, so these class attributes are shadowed by per-instance attributes. With two concurrent downloaders, each instance observes the class-level `_active_instances == 0`, patches `tqdm` independently, and keeps its own `_original_update`. I reproduced the first downloader exiting while the second remained active: `tqdm.update` was restored too early, and after the second exit a stale patched function remained installed globally. Please update `_active_instances`, `_original_update`, and `_original_update_plain` consistently through `CancellableDownloader` (or `type(self)`) and add a lifecycle test where two independently entered downloaders exit in their entry order, not only nested reverse order.

xinference/model/utils.py
```diff
@@ -928,8 +933,15 @@ def get_progress(self) -> float:
             return 1.0
         # Don't return 1.0 when cancelled, calculate actual progress
 
+        # Snapshot the progress sets under the lock: patched_update (download
+        # thread) can add a new bar mid-iteration, which would otherwise raise
+        # "Set changed size during iteration".
+        with self._progress_lock:
```
qinxuye, line 944: This lock only protects the snapshot against additions from `patched_update`; `reset()` still clears both sets without acquiring `_progress_lock`. During `__exit__`, the progress-upload thread can already have passed its `done` check and entered `get_progress()` while cleanup calls `reset()`, so the clear can still race with `list(...)` and raise the same `Set changed size during iteration` error. Please take `_progress_lock` around both clears in `reset()` and cover cleanup racing with a progress poller.

xinference/model/utils.py
SuperMarioYL: Fixed in c524a0c2 — the bookkeeping is now routed through the class. `patch_tqdm` / `unpatch_tqdm` / `__enter__` / `__exit__` all do `cls = type(self)` and read+write `cls._active_instances`, `cls._original_update`, `cls._original_update_plain`, so no per-instance attribute shadows the class-level state: two concurrent downloaders share one counter and one set of originals, and the first to exit no longer restores `tqdm.update` while the second is still active (`xinference/model/utils.py` patch_tqdm ~992, unpatch_tqdm ~1043, __enter__/__exit__ ~1054). New `test_class_level_bookkeeping_no_per_instance_shadow` reproduces the two-concurrent-downloader case: counter goes 2 -> 1 -> 0, no `vars(d)` shadow, and `tqdm.update` stays patched until the last instance exits.

xinference/model/utils.py
SuperMarioYL: Fixed in c524a0c2 — `reset()` now takes `_progress_lock` around both clears (`xinference/model/utils.py` reset ~926), so a progress-upload thread that already passed its done check and entered `get_progress()` cannot be mid-iteration when cleanup clears the sets. New `test_reset_holds_lock_against_progress_poller` races `reset()` vs `get_progress()` with `sys.setswitchinterval(1e-4)` (same harness as `test_concurrent_progress_no_set_mutation`) — no "Set changed size during iteration".

qinxuye (commented):
One concurrency issue is still unresolved on the latest head.

xinference/model/utils.py
```diff
@@ -978,30 +995,33 @@ def raise_error(self, error_msg: str = "Download cancelled"):
...
-                # Get all CancellableDownloader instances and check for cancellation
-                downloaders = [
-                    obj
-                    for obj in gc.get_objects()
-                    if isinstance(obj, CancellableDownloader)
-                ]
+                # Iterate the bounded active-instance registry instead of
+                # gc.get_objects(): the whole-heap traversal stalled the GIL
+                # on every progress tick and scanned O(heap) objects.
+                downloaders = list(CancellableDownloader._active_registry)
```
qinxuye: This is still unresolved on the latest head. `_active_registry` is mutated under `_global_lock` in `__enter__` and `__exit__`, but this snapshot iterates the same set without that lock. A concurrent downloader entering or exiting can still raise `RuntimeError: Set changed size during iteration`. Please take `_global_lock` while creating the snapshot, then iterate the copied list after releasing it.

qinxuye (approved):
LGTM

PR #5264, opened 2026-07-31
perf(supervisor): add whole-result debounce cache for list_models

m199369309:
## Summary

Add a lightweight **whole-result debounce cache** for `list_models()` that returns the cached aggregated model list within a configurable TTL (default 3 seconds). During rapid UI refresh (every \~1 s), this prevents redundant concurrent RPC sweeps to all workers, reducing xoscar transport pressure.

## Background

When the Web UI refreshes (e.g., user clicks the refresh button), `list_models()` is called at \~1 Hz. Each call triggers `asyncio.gather` of concurrent RPCs to every registered worker. Under high-load or hung-worker scenarios, this concurrent RPC pressure can contribute to supervisor event-loop stalls.

## Changes

- **`xinference/constants.py`**: Add `XINFERENCE_LIST_MODELS_DEBOUNCE_SECONDS` (default 3, env-configurable; set to 0 to disable).
- **`xinference/core/supervisor.py`**:
  - Add `_list_models_result_cache` and `_list_models_result_cache_time` fields in `__init__`.
  - Fast-path at the top of `list_models()`: return cached result if within the debounce TTL.
  - Save the aggregated result to the cache at the end of `list_models()`.
  - Add `_invalidate_list_models_debounce_cache()` helper, called from `launch_builtin_model`, `_launch_builtin_sharded_model`, and `terminate_model` so the UI reflects model-list changes immediately.
- **`xinference/core/tests/test_list_models.py`**:
  - Update `DummySupervisor` fixture with the two new cache fields.
  - Add `test_list_models_debounce_cache_hit` — verifies the fast path returns cached results without issuing worker RPCs.
  - Add `test_list_models_debounce_cache_expiry` — verifies stale cache entries are refreshed after the TTL.

## Design Notes

- This cache operates **above** the existing per-worker `_list_models_cache` (which provides graceful degradation on single-worker failures). The two layers are independent and complementary.
- Cache invalidation on launch/terminate ensures the UI reflects model-list changes without waiting for the TTL.
- The 3-second default is short enough that even without invalidation the data is never more than a few seconds stale.

## Testing

```
pytest -vv xinference/core/tests/test_list_models.py
# 6 passed (4 existing + 2 new)

pytest -vv xinference/core/tests/ -k supervisor
# 18 passed, no regressions
```


xinference/core/supervisor.py
```diff
@@ -3178,6 +3195,31 @@ async def describe_model(self, model_uid: str) -> Dict[str, Any]:
...
+                now - self._list_models_result_cache_time,
+                len(self._list_models_result_cache),
+            )
+            return self._list_models_result_cache.copy()
+
+        # Snapshot the cache version before the RPC sweep so that if
+        # _invalidate_list_models_debounce_cache is called concurrently
+        # (e.g. from a launch / termination), the stale sweep result is
+        # not written back over the invalidation.
+        current_version = self._list_models_cache_version
```
qinxuye: This version check prevents an invalidation from being overwritten, but it does not coalesce concurrent cache misses. If the first sweep is still waiting on a slow or hung worker, every subsequent `list_models()` call starts another full worker sweep before any result is cached. With the documented 1-second refresh and 60-second worker timeout, this can still accumulate many concurrent RPC sweeps, which is the pressure this PR is meant to remove. Please add a single-flight task or a dedicated lock with a second cache check inside it, and cover two overlapping calls with a worker call counter.

xinference/core/supervisor.py
```diff
@@ -3178,6 +3195,31 @@ async def describe_model(self, model_uid: str) -> Dict[str, Any]:
...
     async def list_models(self) -> Dict[str, Dict[str, Any]]:
+        # Fast path: return cached result if within debounce window.
+        # During rapid UI refresh (every ~1 s), this prevents redundant
+        # RPC sweeps to all workers, reducing concurrent xoscar transport
+        # pressure.  Set XINFERENCE_LIST_MODELS_DEBOUNCE_SECONDS=0 to
+        # disable debounce caching.
+        now = time.time()
+        if (
+            XINFERENCE_LIST_MODELS_DEBOUNCE_SECONDS > 0
+            and self._list_models_result_cache
```
qinxuye: Using the cached dictionary's truthiness as the cache-validity sentinel means an empty model list is never cached. After a successful sweep returns `{}`, the next request immediately sweeps every worker again even though `_list_models_result_cache_time` was updated. Empty model lists are a common state, so please track whether the cache has been initialized independently (or use the timestamp/version) and add an empty-result cache-hit test.

qinxuye (commented):
Two cache-effectiveness issues remain on the latest head.

qinxuye (approved):
LGTM

PR #5265, opened 2026-07-31
feat: auto-detect model download hub between Hugging Face and ModelScope

OliverBryant:
## What do these changes do?

Currently users have to manually choose between Hugging Face and ModelScope as the model download source (via `--download_hub` or `XINFERENCE_MODEL_SRC`), and the implicit default is guessed from the system locale (`zh_CN` → ModelScope). This guess is often wrong: users in China with a working proxy still get ModelScope, while users whose Hugging Face access is blocked but locale is not `zh_CN` get Hugging Face and the download fails.

This PR adds an `auto` download hub option and makes automatic detection the default, so users no longer need to pick a source:

- New `auto_detect_download_hub()` in `xinference/model/utils.py`: probes whether the Hugging Face endpoint is reachable (honoring `HF_ENDPOINT` mirrors and `HTTP(S)_PROXY` proxies) with a short timeout, picks `huggingface` if reachable and falls back to `modelscope` otherwise. The result is cached per process so the probe runs at most once, and the timeout is tunable via `XINFERENCE_HUB_DETECT_TIMEOUT` (default 3s).
- `download_hub="auto"` is accepted through the whole launch chain (REST API → supervisor → worker → `create_model_instance`) and resolved to a concrete hub before spec matching.
- When `download_hub` is not specified at all, detection also runs — unless the user pinned a source via `XINFERENCE_MODEL_SRC` or provided a local `model_path` (no download needed, no probe).
- `XINFERENCE_MODEL_SRC=auto` is also supported.
- Web UI: the download hub dropdown now offers `auto` alongside the existing options.
- Docs updated (`models/sources/sources.rst`, `getting_started/environments.rst`).

Behavior priority: explicit `--download_hub` > `XINFERENCE_MODEL_SRC` env > automatic connectivity detection. The locale heuristic is kept as the fallback inside `download_from_modelscope()` for code paths that never receive a resolved hub.

## Verification

- Added unit tests in `xinference/model/tests/test_utils.py` covering detection (reachable/unreachable/cached), `resolve_download_hub` priorities, and `XINFERENCE_MODEL_SRC=auto` — all pass.
- Full `xinference/model/tests/test_utils.py` passes locally (one pre-existing network-flaky download test passed on re-run).
- `pre-commit` (black/ruff/isort/mypy/codespell) passes on the changed files; frontend file passes Prettier and ESLint.
- Manual sanity check on a machine with Hugging Face access: detection returns `huggingface`; `resolve_download_hub("auto")`/`resolve_download_hub(None)` resolve accordingly and explicit hubs pass through untouched.


xinference/model/utils.py
```diff
@@ -638,9 +639,80 @@ def is_locale_chinese_simplified() -> bool:
...
+    the ``XINFERENCE_MODEL_SRC`` environment variable or provided a local
+    ``model_path`` (in which case no download is needed).
+    """
+    if download_hub == "auto" or (
+        download_hub is None
+        and model_path is None
+        and not os.environ.get(XINFERENCE_ENV_MODEL_SRC)
+    ):
+        return auto_detect_download_hub()
+    return download_hub
```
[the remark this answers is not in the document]
OliverBryant: Good catch on the inconsistency. I verified the end-to-end behavior was still correct — the legacy fallback inside `download_from_modelscope()` already handles `XINFERENCE_MODEL_SRC=auto` — but resolving it eagerly here is cleaner and matches the documented priority order. Applied your suggestion in a3654f063, together with the test from the other comment.

xinference/model/utils.py
```diff
@@ -638,9 +639,80 @@ def is_locale_chinese_simplified() -> bool:
...
+
+
+def _is_hub_endpoint_reachable(url: str, timeout: float) -> bool:
+    import requests
+
+    try:
+        # Any HTTP response means the endpoint is reachable; proxies configured
+        # via HTTP(S)_PROXY environment variables are honored by requests.
+        requests.head(url, timeout=timeout, allow_redirects=True)
+    except requests.RequestException:
```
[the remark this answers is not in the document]
OliverBryant: Done in a3654f063. The probe is best-effort, so any failure (including a malformed `HF_ENDPOINT` or proxy configuration) now simply means "not reachable" instead of breaking the model launch.

xinference/model/tests/test_utils.py
```diff
@@ -432,3 +432,88 @@ def counting_import(name, *args, **kwargs):
...
+
+    # unspecified hub goes through detection as well
+    assert model_utils.resolve_download_hub(None) == "modelscope"
+
+    # a local model path means no download, so no detection
+    assert model_utils.resolve_download_hub(None, "/path/to/model") is None
+
+    # a pinned XINFERENCE_MODEL_SRC keeps the legacy fallback behavior
+    monkeypatch.setenv("XINFERENCE_MODEL_SRC", "modelscope")
+    assert model_utils.resolve_download_hub(None) is None
```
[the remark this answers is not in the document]
OliverBryant: Added in a3654f063.

xinference/model/utils.py
```diff
@@ -638,9 +639,83 @@ def is_locale_chinese_simplified() -> bool:
...
+
+def _is_hub_endpoint_reachable(url: str, timeout: float) -> bool:
+    import requests
+
+    try:
+        # Any HTTP response means the endpoint is reachable; proxies configured
+        # via HTTP(S)_PROXY environment variables are honored by requests.
+        # The probe is best-effort: any failure (including malformed endpoint
+        # or proxy configuration) just means "not reachable".
+        requests.head(url, timeout=timeout, allow_redirects=True)
```
qinxuye: This treats any HTTP response, including 403/407/5xx, as reachable. In a blocked corporate-proxy scenario this selects Hugging Face and the subsequent download still fails, which is exactly what auto detection is meant to avoid. Please inspect the response status (for example, reject error responses with any authentication-specific policy documented) and add coverage for blocked/error responses.

doc/source/models/sources/sources.rst
```diff
@@ -6,9 +6,26 @@ Download Sources
 
 Xinference supports downloading various models from different sources.
 
+Automatic Detection
+^^^^^^^^^^^^^^^^^^^
+
+By default, Xinference automatically decides between Hugging Face and ModelScope when
```
qinxuye, line 12: This documented default is not true for the official Docker Compose path: `xinference/deploy/docker/docker-compose.yml` still sets `XINFERENCE_MODEL_SRC=${XINFERENCE_MODEL_SRC:-huggingface}`, so `resolve_download_hub(None)` sees a pinned source and never probes. Please make Compose default to `auto` (or omit the variable) and update its adjacent comment/docs so the main deployment path actually gets this behavior.

xinference/model/core.py
```diff
@@ -38,8 +38,15 @@ def create_model_instance(
     from .image.core import create_image_model_instance
     from .llm.core import create_llm_model_instance
     from .rerank.core import create_rerank_model_instance
+    from .utils import resolve_download_hub
     from .video.core import create_video_model_instance
 
+    # resolve_download_hub never returns "auto", so the narrowing cast is safe
+    resolved_download_hub = cast(
+        Optional[Literal["huggingface", "modelscope", "openmind_hub", "csghub"]],
+        resolve_download_hub(download_hub, model_path),
```
qinxuye, line 47: The resolved explicit hub is still overridden by `XINFERENCE_MODEL_SRC` for image/video/embedding/rerank matchers, which use `download_hub == "modelscope" or download_from_modelscope()` unconditionally. Repro: with the env set to `modelscope`, this resolver returns `huggingface`, but `FLUX.1-schnell` still matches its ModelScope spec. Please consult the global fallback only when `download_hub is None` in those matchers and add an end-to-end priority test.

xinference/model/utils.py
OliverBryant: Fixed in 691e8560f. The probe now inspects the response status and treats any HTTP error response (`>= 400`, e.g. 403/407 from a blocking corporate proxy or 5xx from a broken mirror) as unreachable, so those environments fall back to ModelScope. The policy is documented in `sources.rst` and covered by `test_probe_rejects_http_error_responses` (200 → reachable; 403/407/500/503 and raised exceptions → unreachable).

doc/source/models/sources/sources.rst
OliverBryant: Fixed in 691e8560f. The Compose file now defaults `XINFERENCE_MODEL_SRC` to `auto` (kept as an env-overridable variable so users can still pin a source), and the adjacent comment, `.env.example`, and `using_docker_compose.rst` were updated to match.

xinference/model/core.py
OliverBryant: Good catch — this was actually a pre-existing inconsistency in those matchers, but the auto option makes it user-visible, so fixed here in 691e8560f. The image (both matchers), video, embedding, and rerank matchers now consult `download_from_modelscope()` only when `download_hub is None`, so an explicitly requested hub (including one resolved from `auto`) always wins over `XINFERENCE_MODEL_SRC`. Added `test_explicit_download_hub_overrides_model_src_env`, which reproduces your FLUX.1-schnell scenario across all four families and also verifies the env fallback still applies when no hub is given.

xinference/deploy/docker/docker-compose.yml
```diff
@@ -30,8 +30,9 @@ services:
       - "${XINFERENCE_PORT:-9997}:9997"
     command: xinference-local -H 0.0.0.0 --port 9997 --log-level ${XINFERENCE_LOG_LEVEL:-info}
     environment:
-      # Model download source: huggingface (default) or modelscope.
-      XINFERENCE_MODEL_SRC: ${XINFERENCE_MODEL_SRC:-huggingface}
+      # Model download source: auto (default, probes Hugging Face reachability
+      # and falls back to modelscope), huggingface, or modelscope.
+      XINFERENCE_MODEL_SRC: ${XINFERENCE_MODEL_SRC:-auto}
     env_file:
```
qinxuye, line 36: The same Compose file also backs the offline profile. `offline.env` sets `HF_HUB_OFFLINE=1` and `TRANSFORMERS_OFFLINE=1` but does not pin a source, while this `environment` entry takes precedence over `env_file`. On an air-gapped host the probe necessarily resolves `auto` to ModelScope; for models with both hub specs, the cache manager then selects the ModelScope spec and can miss the pre-populated Hugging Face cache documented by the offline workflow, causing launch to fail. Please make the offline profile pin the intended source at the Compose environment level (or make auto honor the available local cache) and add an offline regression test.

OliverBryant: Fixed in 27e6a3bc4, taking the "make auto honor the available local cache" direction. `HF_HUB_OFFLINE` / `TRANSFORMERS_OFFLINE` is an explicit signal that weights live in the local Hugging Face cache, so auto detection now skips the probe entirely in that mode and selects `huggingface` directly — the offline profile keeps working with the default `XINFERENCE_MODEL_SRC=auto`, and huggingface_hub resolves the pre-populated cache without network access. I chose the env-signal approach over a Compose-level pin because `environment:` takes precedence over `env_file`, so pinning inside `offline.env` would not take effect; `offline.env.example` now documents that a ModelScope-based local cache should be pinned via `.env` instead. Covered by `test_auto_detect_honors_hf_offline_mode` (parametrized over both offline variables, asserting no probe runs and both `None`/`auto` resolve to huggingface).

xinference/model/core.py
qinxuye, line 47: `auto` is resolved independently inside each worker process. For a multi-worker shard or replicas placed on different workers, transient or host-specific connectivity can therefore make one worker select Hugging Face while another selects ModelScope for the same launch. Those specs are not guaranteed to identify identical weights; for example, MiniCPM-V-2.6 pins a Hugging Face commit while its ModelScope spec uses `master`. This can make one distributed model load inconsistent checkpoints. Please resolve a single concrete hub for the launch before fan-out (or have the selected workers agree on one reachable hub), pass that value to every worker, and add a multi-worker consistency test.

qinxuye (commented):
Please also rebase this branch: it currently conflicts with the latest `main` in `xinference/core/supervisor.py`.

xinference/model/core.py
OliverBryant: Fixed in c65014ab2. `download_hub` is now resolved once in `SupervisorActor.launch_builtin_model` before either replica or sharded-worker fan-out, and the resulting concrete hub is passed to every worker. Added `test_auto_download_hub_resolved_once_before_worker_fanout`, which asserts the resolver runs once and both workers receive the same hub. The branch was also rebased onto the latest `main`; GitHub now reports the PR as mergeable. Validation: all 18 launch-strategy tests passed, all 9 auto-hub-related tests passed, and pre-commit (Black, Ruff, isort, mypy, codespell) passed on the changed files.

xinference/core/supervisor.py
```diff
@@ -2228,6 +2231,15 @@ async def launch_builtin_model(
             logger.warning("Local deployment, ignore n_worker(%s)", n_worker)
             n_worker = 1
 
+        # Resolve once in the supervisor before replicas or shards fan out to
+        # workers.  Resolving "auto" independently in every worker can select
+        # different model specs for a single distributed launch when their
+        # connectivity differs or changes transiently.
+        download_hub = typing.cast(
+            Optional[Literal["huggingface", "modelscope", "csghub"]],
+            resolve_download_hub(download_hub, model_path),
```
qinxuye: This now probes and resolves from the supervisor process, but model downloads happen on the workers. In a distributed deployment, workers may have their own `HF_ENDPOINT`, proxy/offline settings, or a pinned `XINFERENCE_MODEL_SRC=modelscope` while the supervisor has none. When `download_hub` is omitted, the supervisor will auto-detect a concrete hub and pass it down, overriding that worker-side pin and potentially forcing every worker to use a hub it cannot reach. This also violates the documented `XINFERENCE_MODEL_SRC`-before-auto priority for worker deployments. Please derive the single concrete choice from the selected workers' configuration/connectivity (and make them agree) rather than from the supervisor environment, with coverage where supervisor and worker settings differ.

qinxuye (commented):
One distributed-deployment issue remains.

xinference/core/supervisor.py
OliverBryant: Fixed in edb4caab2. Hub resolution now runs on the selected workers via a worker RPC, so each result honors that worker's own `XINFERENCE_MODEL_SRC`, `HF_ENDPOINT`, proxy, and offline settings. The supervisor deduplicates the selected workers, gathers their resolved hubs, and only starts replicas/shards after all results agree; disagreement fails before fan-out with the per-worker choices and asks the user to pin `download_hub` explicitly. This applies to both replica placement and sharded launches. Coverage includes supervisor/worker settings differing, conflicting shard-worker results, conflicting replica-worker results, and the real worker-local env resolver. Validation: 28 relevant tests passed and pre-commit (Black, Ruff, isort, mypy, codespell) passed on all changed files.

qinxuye (approved):
LGTM

PR #5269, opened 2026-08-02
feat(rerank): support jina-reranker-v3.5 with its native listwise API

xiaoyesoso:
jina-reranker-v3.5 is a 0.6B multilingual listwise reranker (drop-in upgrade to v3) built on Qwen3 with custom modeling code (JinaForRanking) that scores a query against many documents in one forward pass via its native `model.rerank()` API, instead of the Qwen3-style yes/no token scoring used for jina-reranker-v3.

- Register jina-reranker-v3.5 in rerank model_spec.json (HF + ModelScope, sentence_transformers engine only; vLLM rerank engine does not support the JinaForRanking architecture).
- Add a dedicated load() branch using AutoModel with trust_remote_code and dtype="auto"; it must precede the jina-reranker-v3 branch since "jina-reranker-v3" is a substring of "jina-reranker-v3.5".
- Add a dedicated _rerank() branch that maps the native sorted results back to input order so shared sorting/top_n logic stays unchanged.
- Skip wrapping `model.predict` for models that expose no predict method (e.g. JinaForRanking) to avoid a noisy warning on every load.

### DIFF
--- a/xinference/model/rerank/sentence_transformers/core.py
+++ b/xinference/model/rerank/sentence_transformers/core.py
@@ -143,6 +143,39 @@ class SentenceTransformerRerankModel(RerankModel, BatchMixin):
             if self._use_fp16:
                 self._model.model.half()
             self._tokenizer = self._model.tokenizer
+        elif "jina-reranker-v3.5" in self.model_family.model_name.lower():
+            # jina-reranker-v3.5 ships custom modeling code (JinaForRanking)
+            # with a native listwise `rerank()` API; use it instead of the
+            # Qwen3-style yes/no scoring applied to jina-reranker-v3. Note
+            # that "jina-reranker-v3" is a substring of "jina-reranker-v3.5",
+            # so this branch must precede the jina-reranker-v3 one.
+            try:
+                from transformers import AutoModel, AutoTokenizer
+            except ImportError:
+                error_message = "Failed to import module 'transformers'"
+                installation_guide = [
+                    "Please make sure 'transformers' is installed. ",
+                    "You can install it by `pip install transformers`\n",
+                ]
+
+                raise ImportError(f"{error_message}\n\n{''.join(installation_guide)}")
+
+            if not allow_trust_remote_code(self.model_family):
+                raise ValueError(
+                    "Loading this model executes code shipped in the model "
+                    "repository; set XINFERENCE_TRUST_REMOTE_CODE=1 to allow it."
+                )
+            self._tokenizer = AutoTokenizer.from_pretrained(
+                self._model_path, trust_remote_code=True
+            )
+            model_kwargs: Dict[str, Any] = {"device_map": "auto", "torch_dtype": "auto"}
+            if enable_flash_attn:
+                model_kwargs["attn_implementation"] = "flash_attention_2"
+            model_kwargs.update(self._kwargs)
+            logger.debug("Loading jina-reranker-v3.5 with kwargs %s", model_kwargs)
+            self._model = AutoModel.from_pretrained(
+                self._model_path, trust_remote_code=True, **model_kwargs
+            ).eval()
         elif (
             "qwen3" in self.model_family.model_name.lower()
             or "jina-reranker-v3" in self.model_family.model_name.lower()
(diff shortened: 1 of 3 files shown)


xiaoyesoso:
Addressed both review comments from gemini-code-assist:

1. **`dtype` → `torch_dtype`** ([core.py:171](https://github.com/xorbitsai/inference/blob/f60cc2b1/xinference/model/rerank/sentence_transformers/core.py#L171)): Changed `"dtype": "auto"` to `"torch_dtype": "auto"` for broader transformers version compatibility. Note: transformers 4.57 shows a deprecation warning suggesting `dtype` instead, but `torch_dtype` works on both old and new versions, so it's the safer choice since the virtualenv doesn't pin a transformers version.

2. **Empty documents guard** ([core.py:400-410](https://github.com/xorbitsai/inference/blob/f60cc2b1/xinference/model/rerank/sentence_transformers/core.py#L400-L410)): Added an `if not documents` guard before `query[0]` access to return an empty score list instead of raising `IndexError`.

Both fixes verified end-to-end: normal rerank returns correct results, and empty documents list returns `{"results": []}` without error.

Commit: `f60cc2b1`

xinference/model/rerank/sentence_transformers/core.py
```diff
@@ -361,6 +397,17 @@ def _rerank(
...
                 similarity_scores = torch.float16
                 similarity_scores = similarity_scores.float()
+        elif "jina-reranker-v3.5" in self.model_family.model_name.lower():
+            # Native listwise API: returns dicts sorted by score desc, each
+            # carrying the original document index. Map back to input order
+            # so the shared sorting logic below works unchanged.
+            if not documents:
+                similarity_scores = []
+            else:
+                results = self._model.rerank(query[0], documents)
```
qinxuye: The batch handler concatenates documents from requests that share the same kwargs, but this branch invokes the native listwise API once with only `query[0]`. If two concurrent requests with different queries are coalesced, all documents are scored against the first request query, so the later result slices are incorrect. Please invoke the native reranker separately per request (or group by query) and add a regression test covering a batch with different queries.

xiaoyesoso:
Review fixup pushed (commit 50e19037):

**Batch handler query grouping** (feedback from @qinxuye): The `_rerank` branch for jina-reranker-v3.5 now groups documents by unique query before invoking the native listwise `model.rerank()` API. Previously, when the batch handler coalesced requests with different queries into a single `_rerank` call, all documents were scored against `query[0]` — producing incorrect results for subsequent requests. Now each unique query is paired with only its own documents:

```python
query_to_indices: Dict[str, List[int]] = {}
for i, q in enumerate(query):
    query_to_indices.setdefault(q, []).append(i)
for q, idxs in query_to_indices.items():
    docs_for_q = [documents[i] for i in idxs]
    results = self._model.rerank(q, docs_for_q)
    for r in results:
        orig_idx = idxs[int(r["index"])]
        similarity_scores[orig_idx] = float(r["relevance_score"])
```

All pre-commit checks pass (black, ruff, isort, mypy, codespell).

xinference/model/rerank/sentence_transformers/core.py
qinxuye: Thanks, this fixes coalesced requests with different queries, but grouping by query still merges independent requests when they happen to use the same query string. This model is genuinely listwise: its native implementation computes each query embedding and block weights from the full candidate set, so documents from one request can change the scores returned for another request. My earlier suggestion that grouping by query could be sufficient was too broad. Please preserve the original request boundaries and invoke the native reranker once per request, then add a regression test with two batched requests that use the same query but different document lists.

xinference/model/rerank/sentence_transformers/core.py
```diff
@@ -361,6 +397,28 @@ def _rerank(
...
+            #
+            # The batch handler may coalesce requests with different queries
+            # into a single _rerank call (grouped by kwargs). Invoke the
+            # native reranker separately per unique query so each document is
+            # scored against its own query.
+            if not documents:
+                similarity_scores = []
+            else:
+                similarity_scores = [0.0] * len(documents)
+                query_to_indices: Dict[str, List[int]] = {}
```
qinxuye: This still merges independent requests when they use the same query string. Jina v3.5 is genuinely listwise: its native implementation computes query embeddings and block weights from the full candidate set, so documents from one request can change the scores of another request even when the query text matches. Please preserve the original request offsets and call the native `rerank()` once per request rather than once per unique query. A regression test should batch two requests with the same query and different document lists and assert that each native call receives only its own documents.

xiaoyesoso:
Review fixup pushed (commit 6156eace):

**Per-request isolation in batch handler** (feedback from @qinxuye): The `_rerank` branch for jina-reranker-v3.5 now uses `_batch_offsets` (passed from the batch handler) to call the native `model.rerank()` **once per original request**, not once per unique query. This preserves request isolation even when two batched requests share the same query string, because Jina v3.5 is genuinely listwise — documents from one request can change scores of another.

**Implementation**:
- The batch handler passes `_batch_offsets=offsets` to `_rerank`.
- `_rerank` pops `_batch_offsets` early (before any branch) so it's never forwarded to `predict()`, `compute_score()`, etc.
- The jina-reranker-v3.5 branch iterates over `(offset, n)` pairs and calls `self._model.rerank(query[offset], documents[offset:offset+n])` for each request.

**Regression test** added (`test_jina_reranker_v35_batch_isolation`): batches two requests with the same query `"Q"` but different document lists (`["A", "B"]` and `["C", "D", "E"]`), then asserts that `model.rerank()` was called exactly twice — each call receiving only its own documents.

All pre-commit checks pass (black, ruff, isort, mypy, codespell). Test passes.

qinxuye (approved):
LGTM

Issue #3553, opened 2025-05-31
Sglang启动存在参数不兼容，传入logprobs会报错，vllm启动没问题


wang5766171:
### System Info / 系統信息

xinference-server  | [2025-05-30 17:49:54] INFO:     127.0.0.1:46354 - "POST /generate HTTP/1.1" 200 OK
xinference-server  | Exception in ASGI application
xinference-server  | Traceback (most recent call last):
xinference-server  |   File "/usr/local/lib/python3.10/dist-packages/uvicorn/protocols/http/httptools_impl.py", line 401, in run_asgi
xinference-server  |     result = await app(  # type: ignore[func-returns-value]
xinference-server  |   File "/usr/local/lib/python3.10/dist-packages/uvicorn/middleware/proxy_headers.py", line 70, in __call__
xinference-server  |     return await self.app(scope, receive, send)
xinference-server  |   File "/usr/local/lib/python3.10/dist-packages/fastapi/applications.py", line 1054, in __call__
xinference-server  |     await super().__call__(scope, receive, send)
xinference-server  |   File "/usr/local/lib/python3.10/dist-packages/starlette/applications.py", line 112, in __call__
xinference-server  |     await self.middleware_stack(scope, receive, send)
xinference-server  |   File "/usr/local/lib/python3.10/dist-packages/starlette/middleware/errors.py", line 187, in __call__
xinference-server  |     raise exc
xinference-server  |   File "/usr/local/lib/python3.10/dist-packages/starlette/middleware/errors.py", line 165, in __call__
xinference-server  |     await self.app(scope, receive, _send)
xinference-server  |   File "/usr/local/lib/python3.10/dist-packages/starlette/middleware/cors.py", line 85, in __call__
xinference-server  |     await self.app(scope, receive, send)
xinference-server  |   File "/usr/local/lib/python3.10/dist-packages/starlette/middleware/exceptions.py", line 62, in __call__
xinference-server  |     await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
xinference-server  |   File "/usr/local/lib/python3.10/dist-packages/starlette/_exception_handler.py", line 53, in wrapped_app
xinference-server  |     raise exc
xinference-server  |   File "/usr/local/lib/python3.10/dist-packages/starlette/_exception_handler.py", line 42, in wrapped_app
xinference-server  |     await app(scope, receive, sender)
[...]

PR #5270, opened 2026-08-02
feat(sglang): surface logprobs in /v1/completions and /v1/chat/completions

SuperMarioYL:
## Summary

The sglang adapter hardcoded `logprobs=None` at both response builders (`_convert_state_to_completion_chunk` and `_convert_state_to_completion`), discarding the logprob data the engine returns in `meta_info` when `return_logprob` is set. At the same time the raw OpenAI `logprobs`/`top_logprobs` request keys were forwarded into `sampling_params`, which crashes sglang's `msgspec` `SamplingParams` with unknown kwargs (#3553).

This PR mirrors the vLLM reference (`vllm/core.py:_build_logprobs`) so the sglang adapter surfaces logprobs the same way:

- Add `SGLANGModel._build_logprobs(meta_info)` that maps sglang's `output_token_logprobs` / `output_top_logprobs` fields to the legacy `CompletionLogprobs` parallel-list shape (`text_offset` / `tokens` / `token_logprobs` / `top_logprobs`) with a `-9999.0` floor and `text_offset` accumulation.
- Map the OpenAI `logprobs`/`top_logprobs` request to sglang's native `return_logprob` / `top_logprobs_num` / `return_text_in_logprobs` in `_sanitize_generate_config`, and **lift them out of `sampling_params` to the request-body top level** (where sglang's `GenerateReqInput` reads them). This both avoids the `SamplingParams`-unknown-kwarg crash and makes the engine produce + return the data.
- Replace the two `logprobs=None` sites with `_build_logprobs(meta_info)`.

Once `/v1/completions` returns `CompletionLogprobs`, the already-merged chat converter `_completion_logprobs_to_chat_logprobs` (#5252) makes `/v1/chat/completions` logprobs work too — no chat-side change needed.

## Why this shape (addresses the version-variance concern)

sglang's `meta_info` logprob field names/shapes vary across versions (3-tuples with `return_text_in_logprobs=True`, older 2-tuples, older dict shapes). Rather than pin to one schema, `_build_logprobs` reads the fields **defensively**: it surfaces logprobs only when the expected fields are present and well-shaped, and degrades to `None` otherwise. Version variance in sglang's logprob schema therefore cannot crash the adapter or fabricate probabilities — the worst case is `logprobs=None` (the current pre-PR behaviour), never a crash or invented numbers.

## Why the params are lifted to the request body

`SamplingParams` is a `msgspec.Struct` that rejects unknown fields — passing the OpenAI `logprobs` key through it is the #3553 crash mechanism. sglang's `GenerateReqInput` (the request-body schema) accepts `return_logprob` / `top_logprobs_num` / `return_text_in_logprobs` as top-level fields, so they belong there, not inside `sampling_params`.

## Tests

New `xinference/model/llm/sglang/tests/test_logprobs.py` (225 lines) using the existing `test_speculative_config.py` pattern (`object.__new__(SGLANGModel)` + direct static-method calls + a fake `meta_info`):

- `output_token_logprobs` → `CompletionLogprobs` mapping (tokens, token_logprobs, top_logprobs, text_offset).
- `-9999.0` floor and `None`-logprob skip (parity with vllm + the chat converter).
- Older 2-tuple + dict shapes read defensively.
- Malformed/absent fields degrade to `None` (no crash, no fabrication).
- `_sanitize_generate_config` maps both chat (`logprobs`=bool + `top_logprobs`=count) and legacy completions (`logprobs`=count) to the sglang-native params.
- `_lift_logprob_request_params` moves the three params to the request body and removes them from `sampling_params`; empty when logprobs unrequested.

Red-on-master: reverting `core.py` to `main` (keeping the new test) → 9 failed, 3 passed (the 3 that pass on master assert the defensive `None`-when-unrequested behaviour, shared by both). Green-on-branch: 12 passed. The existing `test_speculative_config.py` still passes (22) — no regression.

## Scope

Only `xinference/model/llm/sglang/core.py` + the new test. No changes to other engines, the merged chat converter (`utils.py`), `prompt_logprobs`, or the streaming tool-call path.

Refs #3553

### DIFF
--- a/xinference/model/llm/sglang/core.py
+++ b/xinference/model/llm/sglang/core.py
@@ -440,7 +440,13 @@ class SGLANGModel(LLM):
         # methods so the engine both produces and returns logprob data.
         logprobs_req = generate_config.pop("logprobs", None)  # type: ignore
         top_logprobs_req = generate_config.pop("top_logprobs", None)  # type: ignore
-        if logprobs_req:
+        # A request is active when the caller explicitly opted in. For chat
+        # completions ``logprobs`` is a bool flag (False = opt out); for legacy
+        # /v1/completions it is an int count where 0 means "selected-token
+        # logprob, no alternatives". A bare truthiness check treated completion
+        # 0 as opt-out, so ``return_logprob`` was never sent and the response
+        # stayed ``logprobs=None``. Distinguish chat False from completion int 0.
+        if logprobs_req is not None and logprobs_req is not False:
             if isinstance(logprobs_req, bool):
                 # chat completions: logprobs is a flag, top_logprobs holds the count
                 top_k = max(int(top_logprobs_req or 0), 0)
@@ -673,6 +679,39 @@ class SGLANGModel(LLM):
                 top[k] = sampling_params.pop(k)
         return top
 
+    @staticmethod
+    def _slice_stream_logprobs(meta_info: Dict, consumed: int) -> Tuple[Dict, int]:
+        """Slice cumulative sglang streaming logprob arrays to the per-chunk delta.
+
+        With ``incremental_streaming_output=False`` (sglang default), every
+        streamed ``meta_info`` carries cumulative ``output_token_logprobs`` and
+        ``output_top_logprobs``. Forwarding it unchanged re-emits earlier tokens
+        on every chunk (chunk 1 -> ``[A]``, chunk 2 -> ``[A, B]`` instead of
+        ``[B]``). This returns a shallow-copied ``meta_info`` whose logprob
+        arrays hold only the not-yet-consumed tail, plus the new consumed count
+        (the full cumulative length after this chunk). When this chunk carries
+        no logprob data, ``meta_info`` is returned unchanged and the counter is
+        untouched so a later cumulative chunk still slices correctly.
+        """
+        token_lps = meta_info.get("output_token_logprobs")
+        if not token_lps:
+            return meta_info, consumed
+        total = len(token_lps)
+        delta = dict(meta_info)
+        delta["output_token_logprobs"] = token_lps[consumed:]
+        top_lps = meta_info.get("output_top_logprobs") or meta_info.get(
+            "output_topk_logprobs"
+        )
+        if isinstance(top_lps, list):
+            sliced_top = top_lps[consumed:]
+            if "output_top_logprobs" in meta_info:
+                delta["output_top_logprobs"] = sliced_top
+            elif "output_topk_logprobs" in meta_info:
+                delta["output_topk_logprobs"] = sliced_top
+            else:
+                delta["output_top_logprobs"] = sliced_top
+        return delta, total
+
     async def _stream_generate(
         self,
         prompt: str,
@@ -690,6 +729,10 @@ class SGLANGModel(LLM):
             **self._lift_logprob_request_params(sampling_params),
         }
         pos = 0
+        # sglang (incremental_streaming_output=False) sends cumulative logprob
+        # arrays per chunk; track how many we have already emitted so each chunk
+        # only carries its delta, not the whole prefix again.
+        logprob_consumed = 0
 
         timeout = aiohttp.ClientTimeout(total=3 * 3600)
         async with aiohttp.ClientSession(timeout=timeout, trust_env=True) as session:
@@ -708,7 +751,12 @@ class SGLANGModel(LLM):
                             data = json.loads(chunk[5:].strip("\n"))
                             cur = data["text"][pos:]
                             if cur:
-                                yield data["meta_info"], cur
+                                meta_info, logprob_consumed = (
+                                    SGLANGModel._slice_stream_logprobs(
+                                        data["meta_info"], logprob_consumed
+                                    )
+                                )
+                                yield meta_info, cur
                             pos += len(cur)
                             if need_stop:
                                 break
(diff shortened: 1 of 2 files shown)


xinference/model/llm/sglang/core.py
```diff
@@ -431,6 +432,25 @@ def _sanitize_generate_config(
...
 
+        # Map the OpenAI logprobs request to sglang's native engine params. The
+        # raw OpenAI keys must not reach sglang's SamplingParams (#3553: passing
+        # them crashes the msgspec.Struct with unknown kwargs); sglang reads
+        # return_logprob/top_logprobs_num/return_text_in_logprobs as top-level
+        # GenerateReqInput fields, lifted out of sampling_params by the generate
+        # methods so the engine both produces and returns logprob data.
+        logprobs_req = generate_config.pop("logprobs", None)  # type: ignore
+        top_logprobs_req = generate_config.pop("top_logprobs", None)  # type: ignore
+        if logprobs_req:
```
qinxuye: [P2] Treat completion `logprobs=0` as a request

The legacy completions API accepts integer `0` to request the selected token logprob with zero alternatives. This truthiness check treats it as disabled; after the raw key is popped, `return_logprob` is never sent and the response stays `logprobs=None`. Please distinguish chat `False` from completion integer `0` explicitly and add a regression test for `{"logprobs": 0}`.

xinference/model/llm/sglang/core.py
```diff
@@ -498,7 +518,7 @@ def _convert_state_to_completion_chunk(
             CompletionChoice(
                 text=output_text,
                 index=0,
-                logprobs=None,
+                logprobs=SGLANGModel._build_logprobs(meta_info),
```
qinxuye: [P1] Slice cumulative SGLang logprobs for each streaming chunk

With SGLang defaulting `incremental_streaming_output` to `False`, every streamed `meta_info` contains cumulative `output_token_logprobs` and `output_top_logprobs`. `_stream_generate` slices `data["text"]` using `pos` but forwards `meta_info` unchanged, so this emits `[A]` for the first chunk and `[A, B]` for the second instead of `[A]` and `[B]`; the chat conversion then duplicates earlier tokens as well. Please track the consumed logprob count and slice both output arrays before building each chunk, with a two-chunk cumulative regression test.

xinference/model/llm/sglang/core.py
SuperMarioYL: Done — the guard is now `logprobs_req is not None and logprobs_req is not False`, so completion `0` activates the request (`return_logprob=True`, `top_logprobs_num=0`) while chat `False` still opts out. Added `test_sanitize_treats_completion_logprobs_zero_as_request` covering `{"logprobs": 0}`, plus `test_sanitize_chat_logprobs_false_opts_out` for the False-opt-out side.

xinference/model/llm/sglang/core.py
SuperMarioYL: Done — added `_slice_stream_logprobs(meta_info, consumed)` which slices both `output_token_logprobs` and `output_top_logprobs` to the per-chunk delta and advances the consumed counter by the full cumulative length. `_stream_generate` now tracks `logprob_consumed` alongside `pos` and slices before yielding, so chunk 1 emits `[A]` and chunk 2 emits `[B]` instead of `[A, B]`. Added a two-chunk cumulative regression test (`test_slice_stream_logprobs_emits_per_chunk_delta`) plus a passthrough case for chunks carrying no logprob data.

xinference/model/llm/sglang/core.py
```diff
@@ -554,12 +580,138 @@ def _convert_state_to_completion(
...
+        (the full cumulative length after this chunk). When this chunk carries
+        no logprob data, ``meta_info`` is returned unchanged and the counter is
+        untouched so a later cumulative chunk still slices correctly.
+        """
+        token_lps = meta_info.get("output_token_logprobs")
+        if not token_lps:
+            return meta_info, consumed
+        total = len(token_lps)
+        delta = dict(meta_info)
+        delta["output_token_logprobs"] = token_lps[consumed:]
```
qinxuye, line 711: [P1] Preserve cumulative text offsets across streamed chunks

This slices the cumulative token arrays before `_build_logprobs`, but that helper always starts `text_offset` from 0. For two chunks containing `"Hello"` and `" world"`, the API therefore returns offsets `[0]` and `[0]` instead of `[0]` and `[5]`. Please either build the cumulative `CompletionLogprobs` and then slice it (as the vLLM path does), or track/pass the consumed character offset, and add a two-chunk regression test that verifies the second chunk's absolute offset.

SuperMarioYL: Thanks @qinxuye — good catch. Pushed `1e7a7aa6`: `_build_logprobs` now takes a `text_offset_base` and the streaming caller (`stream_results`) threads `len(complete_response)` (the cumulative completion length streamed so far) into `_convert_state_to_completion_chunk` → `_build_logprobs`, so each chunk reports absolute offsets. For the "Hello" / " world" two-chunk case the second chunk now returns `[5]` instead of `[0]`, matching the non-stream path. I went with the track/pass-consumed-offset option rather than build-cumulative-then-slice since the streamed meta_info is already sliced to the delta and the base is exactly the emitted text length; happy to switch to the cumulative-then-slice shape if you'd prefer it to mirror vLLM more directly. Added a two-chunk regression test asserting the second chunk's absolute offset (17 tests pass).

qinxuye (approved):
LGTM

PR #5272, opened 2026-08-03
feat: support deploying R3-embedding-0.6b and R3-rerank-0.6b models

xiaoyesoso:
Register R3-embedding-0.6b and R3-rerank-0.6b from Tencent's R3-Skill two-stage retriever for agent skill routing.

- R3-embedding-0.6b: bi-encoder based on Qwen3, 1024 dims, 32768 max tokens. Routes to XSentenceTransformer loading path (same as Qwen3-Embedding) since the model architecture is Qwen3Model.

- R3-rerank-0.6b: cross-encoder based on Qwen3ForCausalLM, 40960 max tokens. Routes to the Qwen3-Reranker loading path (AutoModelForCausalLM with yes/no token scoring) since sentence_transformers 3.x CrossEncoder cannot load CausalLM architectures. The chat template and scoring mechanism are identical to Qwen3-Reranker (R3-rerank is fine-tuned from Qwen3-Reranker-0.6B).

Both models require trust_remote_code and are registered with virtualenv packages for the sentence_transformers engine.

### DIFF
--- a/xinference/model/rerank/sentence_transformers/core.py
+++ b/xinference/model/rerank/sentence_transformers/core.py
@@ -98,6 +98,7 @@ class SentenceTransformerRerankModel(RerankModel, BatchMixin):
             self.model_family.type == "normal"
             and "qwen3" not in self.model_family.model_name.lower()
             and "jina-reranker-v3" not in self.model_family.model_name.lower()
+            and "r3-rerank" not in self.model_family.model_name.lower()
         ):
             # sentence-transformers >= 5.4 imports torchcodec at import time and
             # only tolerates ImportError/OSError; a broken (version-mismatched)
@@ -146,8 +147,9 @@ class SentenceTransformerRerankModel(RerankModel, BatchMixin):
         elif (
             "qwen3" in self.model_family.model_name.lower()
             or "jina-reranker-v3" in self.model_family.model_name.lower()
+            or "r3-rerank" in self.model_family.model_name.lower()
         ):
-            # qwen3-reranker
+            # qwen3-reranker (also covers R3-rerank, fine-tuned from Qwen3-Reranker)
             # now we use transformers
             # TODO: support engines for rerank models
             try:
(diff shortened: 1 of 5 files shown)


xinference/model/rerank/sentence_transformers/core.py
```diff
@@ -364,6 +367,7 @@ def _rerank(
         elif (
             "qwen3" in self.model_family.model_name.lower()
             or "jina-reranker-v3" in self.model_family.model_name.lower()
+            or "r3-rerank" in self.model_family.model_name.lower()
```
qinxuye, line 370: [P1] Use the R3-trained default reranking instruction

When `instruction` is omitted, this branch still uses the Qwen3 web-search default below. However, the R3-rerank `config_sentence_transformers.json` and the official R3-Skill `infer.py` both use `Given a user request, retrieve the agent skill that solves it.`. Since the rerank API makes `instruction` optional, the default deployment currently evaluates skill-routing pairs with the wrong domain prompt and can degrade ranking quality. Please select the R3-specific default when the model is `R3-rerank` and add a focused regression test.

xiaoyesoso:
Review fixups pushed (commit c8a44f40):

1. **R3-specific default instruction** (P1 feedback from @qinxuye): `format_instruction` now uses `"Given a user request, retrieve the agent skill that solves it."` as the default for R3-rerank, matching the model's `config_sentence_transformers.json` and the official R3-Skill `infer.py`. The Qwen3 web-search default is still used for qwen3/jina-reranker-v3 models.

2. **Merged redundant branches** (feedback from gemini-code-assist): Combined the `if` and `elif` branches in the batch rerank token tracking code that performed the identical `input_len = sum(self._get_input_tokens_slice(...))` operation into a single condition.

All pre-commit checks pass (black, ruff, isort, mypy, codespell).

qinxuye (approved):
LGTM

PR #5273, opened 2026-08-03
feat(model):support minimax-m3 model

llyycchhee:
Fix think token and add tool parser.

### DIFF
--- a/xinference/model/llm/tool_parsers/minimax_m3_tool_parser.py
+++ b/xinference/model/llm/tool_parsers/minimax_m3_tool_parser.py
@@ -64,13 +64,29 @@ class MiniMaxM3ToolParser(ToolParser):
             return [self._parse_xml_element(child) for child in children]
 
         result: Dict[str, Any] = {}
+        duplicate_tags = set()
         for child in children:
-            result[child.tag] = self._parse_xml_element(child)
+            child_value = self._parse_xml_element(child)
+            if child.tag not in result:
+                result[child.tag] = child_value
+            elif child.tag in duplicate_tags:
+                result[child.tag].append(child_value)
+            else:
+                result[child.tag] = [result[child.tag], child_value]
+                duplicate_tags.add(child.tag)
         return result
 
     def _parse_invoke_args(self, body: str) -> Dict[str, Any]:
+        # Model output is often XML-like rather than strictly valid XML. Escape
+        # bare ampersands while preserving the five predefined XML entities and
+        # numeric character references.
+        escaped_body = re.sub(
+            r"&(?!amp;|lt;|gt;|quot;|apos;|#[0-9]+;|#x[0-9A-Fa-f]+;)",
+            "&amp;",
+            body,
+        )
         try:
-            root = ET.fromstring(f"<root>{body}</root>")
+            root = ET.fromstring(f"<root>{escaped_body}</root>")
         except ET.ParseError:
             # Keep compatibility with the legacy MiniMax parameter format if
             # the model emits text that is not valid nested XML.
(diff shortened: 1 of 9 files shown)


xinference/model/llm/vllm/core.py
```diff
@@ -421,6 +421,11 @@ def _update_vllm_supported_lists() -> None:
     if is_npu_available() and effective_version >= version.parse("0.18.0"):
         _append_unique(VLLM_SUPPORTED_CHAT_MODELS, "DeepseekV4ForCausalLM")
 
+    if effective_version >= version.parse("0.24.0"):
+        _append_unique(
+            VLLM_SUPPORTED_MULTI_MODEL_LIST, "MiniMaxM3SparseForConditionalGeneration"
+        )
```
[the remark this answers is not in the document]
llyycchhee: No, it supports vision in it's model card.

xinference/model/llm/tool_parsers/minimax_m3_tool_parser.py
```diff
@@ -0,0 +1,215 @@
...
+                    and self.tool_call_end_token not in delta_text
+                ):
+                    return (delta_text, None, None)
+                tool_block = function_calls[-1]
+                if self.tool_call_end_token not in tool_block:
+                    return None
+                invokes = self._parse_invoke_calls(tool_block)
+                if not invokes:
+                    return None
+                name, args = invokes[-1]
```
qinxuye: [P1] Preserve every invocation in a streamed tool-call block

The MiniMax-M3 template explicitly allows multiple `<invoke>` elements inside one `<minimax:tool_call>` block, but this returns only `invokes[-1]`. For a block containing `first` and `second`, `extract_tool_calls()` returns both while `extract_tool_calls_streaming()` returns only `second`, silently dropping the first call. Please make the streaming path emit every parsed invocation (for example, by queueing pending calls) and add a regression test with two invocations in the same block.

xinference/model/llm/utils.py
```diff
@@ -935,23 +935,29 @@ def _post_process_completion_chunk(
...
+        elif tool_result is not None:
+            tool_results = [tool_result]
         else:
-            failed_contents.append(content)
+            tool_results = []
+        for parsed_content, func, args in tool_results:
+            if func:
+                tool_calls.append(
+                    {
+                        "index": len(tool_calls),
```
qinxuye: [P1] Keep tool-call indexes stable across streaming chunks

`len(tool_calls)` restarts at zero for every completion chunk. When MiniMax emits two tool-call blocks in separate deltas, the first and second calls are both sent with `index: 0`; OpenAI clients merge streamed calls by index, so the second call can overwrite or be appended to the first. Please carry a cumulative tool-call index across the stream (or include the absolute index in parser events) and add a regression test where two calls complete in separate chunks.

xinference/model/llm/utils.py
```diff
@@ -935,23 +935,34 @@ def _post_process_completion_chunk(
...
+                        "function": {
+                            "name": func,
+                            "arguments": json.dumps(args, ensure_ascii=False),
+                        },
+                    }
+                )
+            elif parsed_content:
+                failed_contents.append(parsed_content)
 
         finish_reason = "tool_calls" if tool_calls else finish_reason
```
qinxuye: [P1] Do not mark intermediate tool-call chunks as finished

The new separate-chunk case now preserves indexes, but this line sets `finish_reason="tool_calls"` on every chunk that carries a parsed call. With two calls completed in separate chunks, the first chunk is therefore marked finished before the second call is emitted, and clients that treat a non-null finish reason as terminal can stop reading and still lose that call. Please preserve a null finish reason on intermediate tool-call chunks, emit `tool_calls` only on the actual final chunk after at least one call was seen, and extend the separate-chunk regression test to cover the finish reasons.

qinxuye (approved):
LGTM

PR #5280, opened 2026-08-04
feat: PDF input support for the OCR API

OliverBryant:
## What does this PR do?

Adds PDF input support to the OCR pipeline. OCR models (GOT-OCR2, DeepSeek-OCR, PaddleOCR-VL, etc.) are document parsers in practice, but the OCR path was image-only end to end: the Web UI file picker only accepted images, and `/v1/images/ocr` called `PIL.Image.open()` directly so a PDF upload failed with a 500. The Web UI even has a Document Parsing panel that already posts PDFs to this endpoint, which could not work before this change.

### API (`/v1/images/ocr`)

- PDF uploads are detected via the `application/pdf` content type or the `%PDF` magic bytes.
- The PDF is rasterized page by page with `pypdfium2` (pure-wheel, no system deps; lazily imported so installs without the `image` extra are unaffected), the model's `ocr()` runs on each page, and the per-page results are merged:
  - all pages plain text → texts joined with blank lines, response stays a single JSON string exactly like the single-image contract;
  - structured results (e.g. `return_dict=True` style models) → `{"pages": [{"page": 1, "result": ...}, ...]}`.
- Two optional PDF-only `kwargs` fields: `pages` (1-based page number or list, default all pages) and `dpi` (rasterization resolution, default 200, capped at 600). Invalid values return HTTP 400.
- The response `media_type` is now `application/json` (the body always was JSON: `ModelActor.ocr` serializes via `_call_wrapper_json` and both REST clients parse it with `response.json()`). Declaring `text/plain` broke aiohttp's content-type check in the async client.

### Web UI

The OCR panel file picker now accepts `image/*,application/pdf` and the label/description mention PDF.

### Packaging / docs

- `pypdfium2` added to the `image` and `dev` extras.
- The OCR section of `doc/source/models/model_abilities/image.rst` documents PDF input and the `pages`/`dpi` kwargs.

## Test plan

- New tests in `xinference/api/tests/test_ocr_pdf.py` cover PDF detection, page selection validation, result merging, and real rasterization of a minimal hand-built PDF (skipped if `pypdfium2` is missing). `xinference/api/tests` passes locally: 252 passed, 17 skipped.
- Smoke-tested against a live deployment running a DeepDoc OCR model:
  - single-image OCR unchanged (regression);
  - 2-page PDF, plain text: `"HELLOXINFERENCEPAGEONE\n\nPAGETWOPDFOCRTEST"`;
  - 2-page PDF with `return_dict: true`: per-page structure with boxes/scores;
  - `kwargs={"pages": [2], "dpi": 150}`: only page 2 OCRed;
  - `kwargs={"pages": [99]}`: HTTP 400 with a clear message.
- `pre-commit` passes on the changed Python files; `eslint` and `prettier --check` pass on the touched frontend file.

GPU model-runtime paths are unaffected (the change is at the API layer); no GPU CI-only checks were run locally.

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -1982,18 +1982,30 @@ class RESTfulAPI(CancelMixin):
                 pages = parsed_kwargs.pop("pages", None)
                 dpi = parsed_kwargs.pop("dpi", DEFAULT_PDF_OCR_DPI)
                 try:
-                    page_images = await asyncio.to_thread(
+                    page_iter = await asyncio.to_thread(
                         rasterize_pdf, image.file.read(), pages=pages, dpi=dpi
                     )
                 except ValueError as ve:
                     raise HTTPException(status_code=400, detail=str(ve))
+                # Pages are rendered lazily, one at a time, so peak memory
+                # stays at a single page regardless of document size.
                 page_results = []
-                for page_number, page_image in page_images:
-                    result = await model_ref.ocr(
-                        image=page_image,
-                        **parsed_kwargs,
-                    )
-                    page_results.append((page_number, json.loads(result)))
+                try:
+                    while True:
+                        item = await asyncio.to_thread(next, page_iter, None)
+                        if item is None:
+                            break
+                        page_number, page_image = item
+                        try:
+                            result = await model_ref.ocr(
+                                image=page_image,
+                                **parsed_kwargs,
+                            )
+                        finally:
+                            page_image.close()
+                        page_results.append((page_number, json.loads(result)))
+                finally:
+                    page_iter.close()
                 body = merge_ocr_page_results(page_results)
                 return Response(content=body, media_type="application/json")
             im = Image.open(image.file)
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -76,6 +76,13 @@ from ..types import (
     max_tokens_field,
 )
 from .frontend_static import mount_frontend
+from .pdf_ocr import (
+    DEFAULT_PDF_OCR_DPI,
+    PDF_MAGIC,
+    is_pdf_upload,
+    merge_ocr_page_results,
+    rasterize_pdf,
+)
 from .responses import JSONResponse
 from .schemas import (
     AutoConfigLLMRequest,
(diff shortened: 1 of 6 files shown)


xinference/api/restful_api.py
```diff
@@ -1969,17 +1976,43 @@ async def create_ocr(
                 parsed_kwargs = {}
             request_id = parsed_kwargs.get("request_id")
             self._add_running_task(request_id)
+            head = image.file.read(len(PDF_MAGIC))
+            image.file.seek(0)
+            if is_pdf_upload(image.content_type, head):
+                pages = parsed_kwargs.pop("pages", None)
+                dpi = parsed_kwargs.pop("dpi", DEFAULT_PDF_OCR_DPI)
+                try:
+                    page_images = await asyncio.to_thread(
```
qinxuye: This rasterizes every selected page before OCR starts, and `rasterize_pdf()` keeps all resulting PIL images in a list. Memory therefore grows with the entire document: at the default 200 DPI, a 100-page A4 PDF is roughly 1.1 GB of RGB pixels, and at 600 DPI even a much smaller document can exhaust the API process. Please process one page at a time and close each image after its OCR call; also enforce a page-count or rendered-pixel limit so a single oversized PDF/page cannot cause an OOM.

qinxuye (commented):
The PDF path has one resource-management issue that should be addressed before merge.

xinference/api/restful_api.py
OliverBryant: Good catch, thanks — fixed in 2db2ed1.

- `rasterize_pdf()` now returns a generator: page selection, DPI, and per-page raster sizes are still validated eagerly (so invalid requests fail with HTTP 400 before any OCR runs), but rendering is lazy — only one page's pixels are alive at a time. The endpoint closes each PIL image right after its OCR call and closes the iterator (releasing the pdfium document) in a `finally` block.
- Two guards were added: at most 200 pages per request (`MAX_PDF_OCR_PAGES`; the error suggests using `pages` to select a subset), and a page whose raster would exceed 80 MP (`MAX_PDF_OCR_PAGE_PIXELS`, ~240 MB of RGB pixels — an A3 page at the 600 DPI cap is ~70 MP) is rejected with a message suggesting a lower `dpi`. This also blocks the pathological case of a huge declared MediaBox (the PDF-spec-max 14400×14400 pt page would be ~1.6 G pixels at 200 DPI).

Added tests for laziness, the page-count limit, and the pixel limit; also re-verified against a live DeepDoc deployment (multi-page PDF, page selection, and single-image regression all behave the same as before).

xinference/api/restful_api.py
```diff
@@ -1969,17 +1976,55 @@ async def create_ocr(
...
+                        rasterize_pdf, image.file.read(), pages=pages, dpi=dpi
+                    )
+                except ValueError as ve:
+                    raise HTTPException(status_code=400, detail=str(ve))
+                # Pages are rendered lazily, one at a time, so peak memory
+                # stays at a single page regardless of document size.
+                page_results = []
+                try:
+                    while True:
+                        item = await asyncio.to_thread(next, page_iter, None)
```
qinxuye, line 1995: `asyncio.to_thread()` does not serialize work across requests, so two concurrent PDF OCR requests can execute PDFium calls in different worker threads at the same time. PDFium is explicitly not thread-safe; pypdfium2 warns that concurrent calls, even on different documents, may crash or corrupt the process: https://pypdfium2.readthedocs.io/en/stable/python_api.html#incompatibility-with-threading. Please protect every PDFium operation in this module (document open/validation, page access/rendering, and close) with one module-level mutex, while keeping the model OCR call outside that lock. Process isolation would also work, but the current thread-pool usage is unsafe without serialization.

qinxuye (commented):
The memory issue is fixed, but the threaded PDFium calls need synchronization before merge.

xinference/api/restful_api.py
OliverBryant: Thanks, fixed in 67762ce.

Added a module-level `threading.Lock` in `pdf_ocr.py`; every PDFium operation now holds it: document open + validation (page count, page sizes) in the eager section, per-page access/render/`to_pil` inside the generator, and both `pdf.close()` paths (the eager error path and the generator's `finally`). The lock is released at each `yield`, so the model OCR call runs outside it and one request's OCR can overlap with another request's rendering — only the PDFium work itself is serialized.

Added a test that runs 16 rasterizations across 8 threads to guard against deadlock/interleaving regressions, and smoke-tested 4 concurrent multi-page PDF requests against a live DeepDoc deployment — all returned correct results.

qinxuye (approved):
LGTM

PR #5282, opened 2026-08-04
ENH: add searchable dropdowns to Audit Center text filters

### LINKED ISSUE #5275: Filtering by Model ID or Model Name fields in the Audit Center results in no search results.
Image

Image

Image

leslie2046:
## Summary

Follow up on #5277 by improving the Audit Center filter UI.

- Replace the five text filters—user, API key name, model ID, model name, and client IP—with searchable autocomplete dropdowns.
- Preserve free-form input so users can still submit arbitrary partial matches.
- Add `/v1/audit/filter-options` to collect available filter values from local audit logs or Elasticsearch.
- Support Elasticsearch fields mapped either directly as `keyword` or as `text` with a `.keyword` subfield.
- Ignore malformed and non-object entries while collecting options from local audit logs.

The substring-matching behavior itself is provided by #5277 and is intentionally not duplicated here.

## Testing

- Added coverage for local filter-option aggregation.
- Added coverage for Elasticsearch aggregation and `.keyword` fallback.
- Verified frontend ESLint and Prettier checks.
- Verified Python formatting, syntax, and focused filter-option smoke checks.


qinxuye:
Seems this and #5277  are trying to fix the same issue .

leslie2046:
> Seems this and #5277 are trying to fix the same issue .
Partially same

qinxuye:
> > Seems this and #5277 are trying to fix the same issue .
> > Partially same

then what's the plan?

leslie2046:
What I mean is that I added a dropdown to the input field to make it easier for users to make a selection.
图片

frontend/src/components/pages/audit-center/index.tsx
```diff
@@ -707,21 +732,25 @@ export default function AuditCenter() {
...
 }
 
-function FilterInput({ label, value, placeholder, onChange }: FilterInputProps) {
+function FilterTextSelect({ label, value, values, placeholder, onChange }: FilterTextSelectProps) {
   return (
     <div className="space-y-2">
       <Label>{label}</Label>
-      <Input
+      <Select
         value={value}
```
qinxuye, line 781: `showSearch` only filters the already-loaded dropdown options; the typed text is never passed to `onChange`, so users cannot submit an arbitrary partial match. This also makes values outside the backend aggregation limit of 500 impossible to filter from the UI, despite the new backend substring-search support. Please preserve free-form submission here (for example via a custom-value/autocomplete control) while still offering the suggested options.

xinference/api/routers/admin.py
```diff
@@ -882,6 +931,116 @@ async def _search_audit_from_file(
...
+                async with session.post(
+                    url, json=request_body, headers=headers
+                ) as resp:
+                    if resp.status != 200:
+                        return resp.status, await resp.text(), {}
+                    return resp.status, "", await resp.json()
+
+            status, response_text, data = await fetch(body)
+            if _is_fielddata_disabled_error(status, response_text):
+                keyword_body = _audit_filter_aggregation_body(
```
qinxuye: This all-or-nothing retry loses options when the index pattern spans both old dynamically mapped indices and new template-backed indices. Old string fields require `<field>.keyword`, while `monitor/filebeat/es-audit-index-template.json` maps these fields directly as `keyword` and therefore has no `.keyword` subfield. After this retry, documents from the direct-keyword indices are ignored for the aggregation, so the dropdown can omit recent values. Please resolve the aggregatable field per index/field (for example via `_field_caps`) or query the mapping groups separately and merge their buckets.

xinference/api/routers/admin.py
```diff
@@ -882,6 +931,116 @@ async def _search_audit_from_file(
...
+                    except (json.JSONDecodeError, ValueError):
+                        continue
+                    if not isinstance(entry, dict):
+                        continue
+                    if not _audit_entry_in_time_range(entry, t_from, t_to):
+                        continue
+                    for field_name in _AUDIT_TEXT_FILTER_FIELDS:
+                        value = entry.get(field_name)
+                        if value is not None and str(value):
+                            options[field_name].add(str(value))
```
qinxuye: The local-file path collects every distinct value without a limit, whereas the Elasticsearch path caps each aggregation at 500. A current `audit.log` can be up to 100 MB, and the frontend `AutoComplete` renders every returned option, so high-cardinality fields such as `client_ip` or `model_id` can produce a very large response and freeze the dropdown. Please apply a bounded limit here as well (preferably the same 500-item cap, with deterministic ordering).

qinxuye (commented):
Found two remaining issues in the filter-option implementation.

xinference/api/routers/admin.py
leslie2046: OK

xinference/api/routers/admin.py
leslie2046: OK

qinxuye (approved):

PR #5285, opened 2026-08-05
fix(webui): ignore stale launch model requests

m199369309:
## Summary

- make launch-model requests follow latest-request-wins semantics
- prevent stale responses from overwriting the selected model type's data
- prevent stale failures and completion handlers from changing current error/loading state

## Validation

- TypeScript type-check (`tsc --noEmit`)
- ESLint on the changed component
- `git diff --check`


qinxuye (approved):
LGTM

PR #5286, opened 2026-08-05
feat: support per-replica worker and GPU placement

m199369309:
## Summary

- add a typed `replica_config` launch option to the REST API and sync/async Python clients
- support selecting a worker, GPU indices, `n_gpu`, and an alias independently for every replica
- validate replica placement, resolve exact `host:port` worker addresses while retaining bare-host compatibility, and ignore worker pinning in local deployment
- preserve existing OCR request handling and propagate API validation errors without wrapping them as HTTP 500 responses
- document the launch option and add focused unit tests

## Stack

This is PR 1/6 and is the backend foundation for the replica-management stack.

## Validation

- 36 focused Python tests passed
- Black passed
- Ruff passed
- `git diff --check`

## Follow-up PRs

1. This PR: backend placement foundation (#5286)
2. Per-replica placement UI (#5287)
3. Replica scale API and clients (#5288)
4. Replica scale UI (#5289)
5. Worker recovery correctness (#5290)
6. Runtime resource reporting and display (#5291)

The independent launch-model stale-request fix is #5285.

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -77,6 +77,13 @@ from ..types import (
     max_tokens_field,
 )
 from .frontend_static import mount_frontend
+from .pdf_ocr import (
+    DEFAULT_PDF_OCR_DPI,
+    PDF_MAGIC,
+    is_pdf_upload,
+    merge_ocr_page_results,
+    rasterize_pdf,
+)
 from .responses import JSONResponse
 from .schemas import (
     AutoConfigLLMRequest,
@@ -631,14 +638,16 @@ class RESTfulAPI(CancelMixin):
             os.path.join(lib_location, "ui", "web", "dist"),
         )
         if not mount_frontend(self._app, Path(ui_dist_location)):
-            warnings.warn(f"""
+            warnings.warn(
+                f"""
             The Xinference web UI is not built at expected directory: {ui_dist_location}
             The API keeps serving without the web UI. To enable it, build the
             frontend static export from the repository "frontend/" directory with
             "npm ci && npm run build" (this stages the export at the directory
             above), or set XINFERENCE_FRONTEND_DIST_DIR to an export directory,
             and restart. For frontend development, run "npm run dev" instead.
-            """)
+            """
+            )
 
         config = Config(
             app=self._app,
@@ -942,6 +951,8 @@ class RESTfulAPI(CancelMixin):
                 replica_config=replica_config,
                 **kwargs,
             )
+        except HTTPException:
+            raise
         except ValueError as ve:
             logger.error(str(ve), exc_info=True)
             raise HTTPException(status_code=400, detail=str(ve))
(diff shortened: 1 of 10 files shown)


xinference/api/restful_api.py
```diff
@@ -1074,6 +1101,27 @@ async def launch_model_by_version(
...
+                )
+            try:
+                replica_config = [
+                    ReplicaConfig.from_dict(item) for item in replica_config
+                ]
+            except Exception as e:
+                raise HTTPException(
+                    status_code=400,
+                    detail=f"Invalid replica_config: {e}",
+                )
```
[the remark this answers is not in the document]
m199369309: Thanks for the suggestion. I extracted the shared validation and parsing into `_parse_replica_config()` and reused it in both `launch_model()` and `launch_model_by_version()`.

I used an explicit `is None` check rather than a truthiness check so that an explicitly supplied empty list is preserved and still rejected by the existing replica-count validation instead of silently falling back to automatic placement. Focused regression tests were added for valid input, empty input, legacy placement conflicts, and malformed configurations.

xinference/core/replica_config.py
```diff
@@ -0,0 +1,126 @@
...
+class DeviceConfig(BaseModel):
+    """Placement of one replica on a single worker.
+
+    Each replica is restricted to exactly one device for now (no cross-worker
+    sharding within a single replica); ``devices`` length is validated to be 1.
+    """
+
+    worker_ip: str  # full worker address "ip:port"
+    n_gpu: Union[int, str] = "auto"  # int count, or "auto" for worker auto-allocation
+    gpu_idx: Optional[List[int]] = None  # explicit GPU indexes; None/empty = auto
```
qinxuye, line 47: These Pydantic-v1 models currently coerce malformed placement values and ignore unknown fields. For example, `n_gpu: 1.5` becomes `1`, `gpu_idx: [0.9]` becomes `[0]`, and a typo such as `gpu_id` is silently dropped so the replica falls back to automatic GPU placement. Duplicate indexes such as `[0, 0]` are also accepted. Because this API promises exact placement, please make the nested models reject extra fields and use strict validation for `worker_ip`, `n_gpu`, and every `gpu_idx`, including rejecting duplicate indexes.

xinference/core/supervisor.py
```diff
@@ -2104,6 +2339,7 @@ async def _launch_one_model(
                     "worker_address": worker_ref.address,
                     "status": LaunchStatus.CREATING.name,
                     "created_ts": int(time.time()),
+                    "replica_uid": replica_uid,
```
qinxuye, line 2342: The explicit `target_gpu_idx` is not stored in the newly added replica status. As a result, `ReplicaStatus.gpu_idx` remains `None` even when this launch is pinned to specific GPUs, and the dependent UI reports the placement as `auto`. Please include `"gpu_idx": target_gpu_idx` in this initial status update (leaving it `None` only for automatic allocation).

xinference/core/replica_config.py
```diff
@@ -0,0 +1,126 @@
...
+            raise ValueError(
+                "Each replica_config entry must have exactly one device "
+                f"(replica {idx} has {len(cfg.devices)}). "
+                "Cross-worker sharding per replica is not supported yet."
+            )
+        device = cfg.devices[0]
+
+        _validate_device_consistency(idx, device)
+
+        replica_uid = cfg.replica_uid
```
qinxuye: The requested `replica_config` contract generates a stable default replica UID when this field is omitted (for example, `{model_uid}-0`, `{model_uid}-1`). This implementation instead preserves `None` and treats the field only as an optional display alias. If that semantic change is intentional, it should be agreed and documented; otherwise please generate and validate the default UIDs here.

qinxuye (commented):
Found three placement-contract issues that should be addressed before approval.

qinxuye (approved):

PR #5288, opened 2026-08-05
feat: support scaling running models by replica

m199369309:
## Summary

- add REST endpoints for adding and terminating individual model replicas
- add matching sync and async Python client methods
- implement Supervisor scale-up/scale-down while preserving per-replica `n_gpu`, avoiding inherited stale `gpu_idx`, and cleaning launch bookkeeping
- clean REST routes, caches, and audit state when the final replica is removed
- add API/client tests and user documentation

## Stack

PR 3/6. Depends on https://github.com/xorbitsai/inference/pull/5287.

This PR targets `main` so it is visible in the official repository; its effective review diff becomes this layer after the dependency is merged.

## Validation

- 8 focused Python tests passed (27 deselected)
- Black passed
- Ruff passed
- `git diff --check`


frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -1349,6 +1423,28 @@ export default function LaunchDialog({
   };
 
   const handleLaunch = async (values: FormValues) => {
+    if (values.replica_placement_mode === 'custom') {
+      const rows = Array.isArray(values.replica_config) ? values.replica_config : [];
+      const replicaCount = Number(values.replica) || 1;
+      const gpuIndexPattern = /^\d+(?:,\d+)*$/;
```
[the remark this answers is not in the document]
m199369309: Fixed through the updated placement base. GPU index validation now accepts optional whitespace around commas via the shared `GPU_IDX_PATTERN`.

xinference/core/supervisor.py
```diff
@@ -1930,22 +1932,217 @@ async def launch_model_by_version(
...
+
+    def _resolve_worker_addresses(self, worker_ip: Union[str, List[str]]) -> List[str]:
+        """Resolve bare hosts and full worker addresses to registered addresses.
+
+        A full ``host:port`` value identifies one concrete worker. A bare host
+        keeps the legacy behavior and expands to every worker registered on
+        that host. The same resolver is used by regular and sharded launches so
+        both paths interpret ``worker_ip`` consistently.
+        """
+        raw_entries = worker_ip if isinstance(worker_ip, list) else [worker_ip]
```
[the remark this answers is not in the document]
m199369309: No code change is needed here. The current `_resolve_worker_addresses()` already flattens list/string/comma-separated inputs, trims entries, resolves both exact `host:port` addresses and bare hosts (including multiple workers on one host), and deduplicates while preserving order. The suggested `_resolve_address()` helper does not exist in this codebase.

xinference/core/supervisor.py
```diff
@@ -3129,6 +3135,209 @@ async def _terminate_one_model(_replica_model_uid):
...
+        # ---- 1. Look up the model ------------------------------------------------
+        replica_info = self._model_uid_to_replica_info.get(model_uid)
+        if replica_info is None:
+            raise ValueError(f"Model not found in the model list, uid: {model_uid}")
+
+        if not replica_info.active_replica_ids:
+            raise ValueError(f"Model has no active replicas, uid: {model_uid}")
+
+        # ---- 2. Generate new replica id -----------------------------------------
+        new_replica_id = max(replica_info.active_replica_ids) + 1
```
qinxuye, line 3291: Because `SupervisorActor` is a `StatelessActor`, scale operations can interleave here. Two concurrent add requests can both compute the same `new_replica_id` before either one updates `active_replica_ids`, and add/delete can race with the same bookkeeping. This can launch duplicate replica UIDs, overwrite routing state, or clean up another request's replica. Please serialize add/terminate operations per model (or reserve the ID before the first `await`) and add a concurrency test.

xinference/core/supervisor.py
```diff
@@ -3129,6 +3135,209 @@ async def _terminate_one_model(_replica_model_uid):
...
+                "Adding replicas to Xavier-distributed models is not supported."
+            )
+        if launch_args.get("n_worker", 1) > 1:
+            raise ValueError(
+                "Adding replicas to sharded models (n_worker > 1) is not supported."
+            )
+
+        # ---- 6. Resolve target worker / GPU --------------------------------------
+        if replica_config is not None:
+            _resolved_targets, _replica_uid_map = await self._resolve_replica_config(
```
qinxuye, line 3337: Calling `_resolve_replica_config(..., replica=1, ...)` makes an omitted `replica_uid` normalize to `{model_uid}-0` on every scale-up. If the original placement also omitted its label, this duplicates the existing replica's user-facing UID. Please derive the default from `new_replica_id` and validate it against the labels of existing replicas.

xinference/core/supervisor.py
```diff
@@ -3129,6 +3135,209 @@ async def _terminate_one_model(_replica_model_uid):
...
+                model_uid, replica=1, replica_config=[replica_config]
+            )
+            target_worker_ref, target_gpu_idx, target_n_gpu = _resolved_targets[0]
+            replica_uid_label = _replica_uid_map.get(0)
+        else:
+            # Auto-select: pick the worker with the fewest loaded models.
```
qinxuye: Model count is not enough to determine whether a worker can host this replica. In a heterogeneous cluster, or with GPU sharing disabled, this can select a worker with no compatible/free GPU and fail even though another worker is available. Please reuse the normal launch strategy and GPU-allocation snapshot (or apply equivalent capacity filtering) for the automatic placement path.

xinference/core/supervisor.py
```diff
@@ -3129,6 +3135,209 @@ async def _terminate_one_model(_replica_model_uid):
...
+                        w_ref.address,
+                        count,
+                    )
+                    continue
+                candidates.append((w_ref, count))
+            if not candidates:
+                raise RuntimeError("No available worker found")
+            candidates.sort(key=lambda x: (x[1], x[0].address))
+            target_worker_ref = candidates[0][0]
+            target_gpu_idx = None
```
qinxuye: This does not preserve the effective GPU count for legacy launches that used explicit `gpu_idx`. For example, cached args from `gpu_idx=[0, 1]` with the default `n_gpu="auto"` still contain `n_gpu="auto"`; after clearing the stale indexes below, the worker's auto path allocates only one GPU. When no new placement is supplied, please derive `target_n_gpu` from the cached `gpu_idx` length before discarding those indexes.

qinxuye (commented):
I found four issues in the scale-up implementation that should be addressed before approval.

xinference/core/supervisor.py
```diff
@@ -3129,9 +3148,341 @@ async def _terminate_one_model(_replica_model_uid):
...
+        return candidates[0]["ref"], None
+
+    @log_async(logger=logger)
+    async def add_model_replica(
+        self,
+        model_uid: str,
+        replica_config: Optional[ReplicaConfig] = None,
+    ) -> Dict[str, Any]:
+        """Add one replica while serializing scale changes for this model."""
+        async with self._get_model_replica_lock(model_uid):
```
qinxuye, line 3261: This lock currently covers `add_model_replica()` and `terminate_model_replica()`, but `terminate_model()` still bypasses it. I reproduced a delayed scale-up racing with whole-model termination: termination removed the base model and replica 0, then scale-up returned successfully and left `demo-rep1` in `_replica_model_uid_to_worker` while the base model was absent from `_model_uid_to_replica_info`. That leaves a running orphan replica which cannot be routed or cleaned up normally. Please serialize `terminate_model()` with the same per-model lock and add this race as a regression test.

qinxuye (commented):
The previous concurrency fix is incomplete because whole-model termination still bypasses the per-model lock.

qinxuye (approved):
LGTM

PR #5290, opened 2026-08-05
fix: preserve replicas across worker recovery

m199369309:
## Summary

- mark replicas on a disconnected worker as terminated
- merge recovered replica IDs with currently active replicas instead of replacing the active set
- re-register recovered models with the Supervisor after worker startup recovery
- fix the replica ID update off-by-one issue in worker status reporting

## Stack

PR 5/6. Depends on https://github.com/xorbitsai/inference/pull/5289 in the submitted stack. The recovery behavior itself logically depends on the backend scale API introduced earlier in the stack.

This PR targets `main` so it is visible in the official repository; its effective review diff becomes this layer after the dependency is merged.

## Validation

- 4 focused Python tests passed (65 deselected)
- Black passed
- Ruff passed
- `git diff --check`


frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -1349,6 +1423,28 @@ export default function LaunchDialog({
   };
 
   const handleLaunch = async (values: FormValues) => {
+    if (values.replica_placement_mode === 'custom') {
+      const rows = Array.isArray(values.replica_config) ? values.replica_config : [];
+      const replicaCount = Number(values.replica) || 1;
+      const gpuIndexPattern = /^\d+(?:,\d+)*$/;
```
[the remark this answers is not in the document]
m199369309: Fixed through the updated placement base. The shared GPU index pattern accepts optional whitespace around commas.

frontend/src/components/pages/running-model/add-replica-dialog.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    const hasWorker = selectedWorker !== '';
+    const hasGpu = gpuIdx.trim() !== '';
+    const hasUid = replicaUid.trim() !== '';
+
+    if ((hasGpu || hasUid) && !hasWorker) {
+      toast.error(t('runningModels.addReplicaWorkerRequired'));
+      return;
+    }
+    if (hasGpu && !/^\d+(?:,\d+)*$/.test(gpuIdx.trim())) {
```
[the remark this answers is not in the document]
m199369309: Fixed through the updated scale UI. The add-replica GPU input accepts whitespace around commas without stripping internal whitespace.

xinference/core/supervisor.py
```diff
@@ -2104,6 +2351,7 @@ async def _launch_one_model(
                     "worker_address": worker_ref.address,
                     "status": LaunchStatus.CREATING.name,
                     "created_ts": int(time.time()),
+                    "replica_uid": replica_uid,
```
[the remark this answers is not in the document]
m199369309: This is intentionally covered by the dependent runtime-resource-display PR #5291, whose current head records the resolved/assigned `gpu_idx` in the initial replica status. Keeping it there avoids mixing runtime-resource display behavior into this worker-recovery PR.

xinference/core/supervisor.py
```diff
@@ -547,7 +557,9 @@ def _rebuild_worker_replica_state(
             if replica_info is None:
                 replica_info = self._build_replica_info(replica_count)
                 self._model_uid_to_replica_info[model_uid] = replica_info
-            replica_info.active_replica_ids = replica_indexes
+            replica_info.active_replica_ids = sorted(
```
qinxuye: P1: When the supervisor rebuilds a model from scratch, `_build_replica_info(replica_count)` creates IDs `0..count-1` before this union. If the only running replica is sparse (for example `model-rep2` after scale-down and a supervisor restart), this produces `[0, 2]` although only `rep2` is mapped; round-robin requests to `rep0` then fail with 503. Please initialize a new `ReplicaInfo` from `replica_indexes` (or overwrite on the new-info path) and only union when prior state actually exists. A regression test with only `rep2` would cover this.

xinference/core/supervisor.py
```diff
@@ -432,7 +432,17 @@ async def _handle_dead_worker(self, worker_address: str) -> List[str]:
         for replica_model_uid in affected_replica_uids:
             parsed = self._get_model_uid_and_replica_index(replica_model_uid)
             if parsed is not None:
-                base_uids_affected.add(parsed[0])
+                base_uid, replica_id = parsed
+                base_uids_affected.add(base_uid)
+                if self._status_guard_ref is not None:
+                    try:
+                        await self._status_guard_ref.update_replica_status(
```
qinxuye: P2: This updates StatusGuard only in `_handle_dead_worker`. The same-address re-registration path in `add_worker()` also computes `affected_replica_uids`, but it never marks those replica statuses terminated or shrinks `InstanceInfo.replica` when the model is merely degraded. I reproduced a worker that previously had `rep0`/`rep1` and re-registers only `rep0`: routing contains only `rep0`, while StatusGuard still reports both replicas READY with `replica=2`. Please apply the same per-replica transition/count reconciliation in that path and cover partial recovery.

qinxuye (commented):
Two recovery-state consistency issues need to be addressed before approval.

xinference/core/supervisor.py
m199369309: Fixed in ed2d6e3f. Fresh supervisor rebuilds now initialize ReplicaInfo directly from the reported sparse replica indexes, so a lone rep2 no longer creates a phantom rep0. Added a sparse-replica regression test.

xinference/core/supervisor.py
m199369309: Fixed in ed2d6e3f. Same-address partial re-registration now marks missing replicas TERMINATED and reconciles InstanceInfo.replica with the actual active routing state. Added a partial-recovery regression test.

qinxuye (approved):
LGTM

PR #5298, opened 2026-08-07
feat(audio): add CAMPPlus speaker embedding support

leslie2046:
## Summary

- Add the `speaker_embedding` capability for CAMPPlus audio models, including speaker-vector extraction through `/v1/audio/embeddings`.
- Add the required ModelScope runtime dependencies so CAMPPlus models can be installed and launched reliably.
- Add a complete Web UI workflow: speech-sample upload, 192-dimensional embedding display, result copying, and “Try To API” examples.
- Fix Web UI ability filtering so `speaker_embedding` remains selectable.
- Document speaker embedding usage in the Web UI, REST API, and Python client, including cosine-similarity guidance.
- Remove the two CAMPPlus model records from the bundled audio JSON so their metadata can be managed through Xinference Models Hub.

## Testing

- Frontend ESLint, TypeScript type check, Prettier, and production build
- Browser QA for the speaker-embedding workflow and responsive layouts
- Pre-commit: Black, Ruff, isort, mypy, and codespell
- JSON validation and Python compilation


doc/source/models/builtin/audio/speech_campplus_sv_zh-cn_16k-common.rst
```diff
@@ -0,0 +1,38 @@
...
+^^^^^^^^^^^^^^^^^^^^^
+
+Each request returns one 192-dimensional vector. Store the vector in your
+application and use cosine similarity to compare samples. Select a similarity
+threshold using representative recordings from the microphones, speakers, and
+acoustic conditions expected in production.
+
+Execute the following command to launch the model::
+
+   xinference launch --model-name speech_campplus_sv_zh-cn_16k-common --model-type audio
```
qinxuye, line 35: The launch command currently cannot work on a clean install. This PR removes both CAMPPlus entries from `xinference/model/audio/model_spec.json`, and the current Models Hub audio payload also contains no `campplus` records, so `match_audio()` raises `Audio model ... not found` before reaching this implementation. Please publish the two model definitions to Models Hub before merging (or keep them in the bundled JSON until the Hub records are available), and add a registration test that resolves the model by name instead of constructing a mock spec.

xinference/model/audio/speaker_embedding.py
```diff
@@ -0,0 +1,135 @@
...
+            from modelscope.utils.constant import Tasks
+        except ImportError as e:
+            raise ImportError(
+                "Failed to import ModelScope speaker-verification dependencies. "
+                "Please install `modelscope[framework]`, `soundfile`, and "
+                "`scikit-learn`."
+            ) from e
+
+        if self._device is None:
+            self._device = get_available_device()
```
qinxuye, line 67: On Apple Silicon, `get_available_device()` returns `mps`, but ModelScope pipeline device validation only accepts `cpu`, `cuda`, or `gpu`. A default CAMPPlus launch will therefore fail during pipeline construction. Please map unsupported accelerators such as `mps` to CPU for this backend while preserving CPU/CUDA selection, and cover the fallback with a focused test.

doc/source/models/builtin/audio/speech_campplus_sv_zh-cn_16k-common.rst
leslie2046: Thanks for catching this. I restored both CAMPPlus definitions in the bundled model_spec.json so they remain launchable on a clean installation until the Models Hub records are available. I also replaced the mocked factory test with registration tests that resolve both models through match_audio() by name and verify that create_audio_model_instance() dispatches to ModelScopeSpeakerEmbeddingModel.

xinference/model/audio/speaker_embedding.py
leslie2046: Fixed. The ModelScope backend now falls back to CPU when the detected device is unsupported by ModelScope, including mps, while preserving cpu, cuda, and gpu. I also added a focused test that simulates get_available_device() returning mps and verifies that the pipeline is constructed with device="cpu".

qinxuye (approved):

Issue #5295, opened 2026-08-07
BUG: DeepDoc OCR model fails to launch — transformers 5.x resolved in its virtualenv


OliverBryant:
## Describe the bug

Launching the `DeepDoc` OCR model added in #5230 fails on a clean environment. The virtualenv built for the model resolves `transformers` to 5.x, which is incompatible with the `huggingface_hub` version resolved alongside it, and the model worker dies during load.

## To Reproduce

```bash
xinference-local -H 0.0.0.0 -p 19998
curl -X POST http://127.0.0.1:19998/v1/models \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"model_name":"DeepDoc","model_type":"image","model_path":"/path/to/Xorbits--deepdoc/snapshots/master","gpu_idx":1}'
```

The launch fails and `/v1/models` stays empty. In the log:

```
ImportError: cannot import name 'is_offline_mode' from 'huggingface_hub'
  (.../virtualenv/v4/DeepDoc/default/3.12.13/lib/python3.12/site-packages/huggingface_hub/__init__.py)
Failed to load model DeepDoc-rep0
```

followed by the worker dying, which surfaces to the caller as an unrelated-looking transport error:

```
xoscar.errors.ServerClosed: Remote server unixsocket:///... closed:
  0 bytes read on a total of 11 expected bytes
Failed to launch replica DeepDoc-rep0
```

Resolved versions inside the built virtualenv:

```
transformers      5.14.1
huggingface-hub   0.36.2
```

`is_offline_mode` was removed from `huggingface_hub` in 1.x and transformers 5.x expects it, so `transformers.utils.hub` fails to import.

## Root cause

`xinference/model/image/model_spec.json` has two relevant constraints:

- line 61 — the shared image-model virtualenv constraint, with **no upper bound**:
  ```json
  "transformers>=4.51.0",
  ```
- line 1795 — DeepDoc's own constraint, gated on `#engine#`:
  ```json
  "transformers<5 ; #engine# == \"deepdoc\""
  ```

OCR models don't set `#engine#`, so the second constraint never activates and the unbounded one wins, pulling in transformers 5.14.

Worth noting: DeepDoc doesn't import `transformers` at all — `grep -rn "import transformers" deepdoc/` in the installed `deepdoc-lib` returns nothing. The constraint exists only to keep the environment consistent.

## Workaround

Pinning inside the already-built virtualenv is enough:

```bash
[...]

PR #5299, opened 2026-08-07
feat(ocr): expose DeepDoc's document-parsing pipeline as task="parse"

### LINKED ISSUE #5296: ENH: expose DeepDoc's full document-parsing pipeline as task=parse
## Summary

The DeepDoc OCR model added in #5230 exposes DeepDoc's three atomic vision capabilities (`ocr`, `layout`, `table`) but not `RAGFlowPdfParser.parse_into_bboxes()`, the full document-parsing pipeline that sits on top of them. For consumers that want structured document parsing rather than per-page OCR, that pipeline is the thing they actually need, and it cannot be reconstructed client-side from the three atomic tasks.

I'd like to propose adding a fourth task, `task="parse"`, to the existing `/v1/images/ocr` endpoint.

## Motivation

We're adding remote DeepDoc parsing to xagent (xorbitsai/xagent#1161) so users can offload the GPU-heavy OCR/layout/table work to their own machine. The natural target was this model. Probing a live server showed the current tasks give us materially less than local `deepdoc-lib` does.

Measured on the same 3-page PDF, in the same virtualenv this model builds:

| | `task=ocr` + `task=layout` (current) | `parse_into_bboxes()` |
|---|---|---|
| Elements | 13 layout blocks (`text` ×8, `title` ×5) | 52 (`text` ×40, `title` ×8, `table` ×2, `figure` ×2) |
| Tables | not detected | 2, each with complete HTML incl. `<caption>` and `<th>` |
| Figures | not detected | 2 |
| Cross-page coordinates | per-page only | accumulated via `page_cum_height` |
| Paragraph merging | none | XGBoost `updown_concat` model |
| Reading order | raw detection order | tables/figures reinserted at correct position |

The `parse_into_bboxes` run took 7.0s on CPU for that document, so this is not an expensive addition.

Concretely, what the current tasks cannot produce:

- **Table HTML.** `task=table` returns raw cell/row/column boxes (`{"label": "table column", "x0": ..., ...}`). Turning those into HTML is `TableStructureRecognizer.construct_table()`, which lives behind `parse_into_bboxes`.
- **Figures**, which aren't surfaced by any current task.
- **Cross-page coordinates.** `parse_into_bboxes` accumulates a document-wide Y axis (`page_cum_height`); the per-page responses reset each page.
- **Paragraph merging.** `_concat_downward` uses the bundled XGBoost model to decide whether consecutive lines belong to one paragraph, including across pages.
- **Reading order.** `_extract_table_figure` reinserts tables and figures into the text flow by minimum rectangle distance.

A client can spatially join `ocr` and `layout` output — they do share one coordinate space, which I verified — but the four items above are genuinely not recoverable that way.

## Proposed change

`_process_single` in `xinference/model/image/ocr/deepdoc.py` already dispatches on `task`. A fourth branch would follow the existing shape:

```python
elif task == "parse":
    from deepdoc import PdfParser
    parser = PdfParser()          # or a lazily cached instance, as with layout/table
    bboxes = parser.parse_into_bboxes(pdf_path, zoomin=kwargs.get("zoomin", 3))
    return {"task": "parse", "elements": _serialize(bboxes)}
```

Two details worth deciding together:

1. **`parse` needs the PDF, not a rasterized page.** The REST layer currently rasterizes PDFs and calls `ocr()` per page. For `task="parse"`, the original bytes should be passed through instead, since `parse_into_bboxes` does its own rendering and needs the whole document for cross-page merging. This is the main structural change.
2. **Serialization.** `parse_into_bboxes` returns numpy scalars and PIL images. Suggested: coerce numpy to Python numbers, drop the internal `position_tag`, and base64-encode images only for `table`/`figure` elements (an `image_scope` kwarg could make that configurable) — every bbox gets a crop image, but only tables and figures need one, so this keeps payloads reasonable.

Element shape, matching what `parse_into_bboxes` already produces:

```json
{"task": "parse", "elements": [
  {"type": "table",
   "text": "<table><caption>...</caption><tr><th>...</th></tr></table>",
   "image_base64": "...",
   "metadata": {"page_number": 2, "x0": 20.0, "x1": 400.0, "top": 50.0, "bottom": 200.0,
                "layout_type": "table", "col_id": 0,
                "positions": [[2, 20, 400, 50, 200]]}}
]}
```

Happy to send a PR if the approach looks right — I have the client side and a live test setup already.

Environment: xinference at `66a63f957` (#5230 merged), Python 3.12, CUDA, RTX 3090 Ti.

Note: launching the model on a clean environment currently fails on a virtualenv dependency conflict — filed as #5295.

OliverBryant:
## Summary

The DeepDoc OCR model added in #5230 exposes DeepDoc's three atomic vision capabilities (`ocr`, `layout`, `table`) but not `RAGFlowPdfParser.parse_into_bboxes()`, the full document-parsing pipeline built on top of them. Table HTML, figures, cross-page coordinates, paragraph merging and reading order are not recoverable from the three atomic tasks, so consumers that want structured document parsing cannot use this model today.

This adds a fourth task to the existing `/v1/images/ocr` endpoint, as proposed in #5296:

```bash
curl -X POST http://<HOST>:<PORT>/v1/images/ocr \
  -H "Authorization: Bearer $TOKEN" \
  -F model=DeepDoc \
  -F 'kwargs={"task": "parse"}' \
  -F image=@doc.pdf
```

```json
{"task": "parse",
 "elements": [
   {"type": "table",
    "text": "<table><caption>...</caption><tr><th>...</th></tr></table>",
    "image_base64": "...",
    "metadata": {"page_number": 2, "x0": 20.0, "x1": 400.0, "top": 50.0,
                 "bottom": 200.0, "layout_type": "table", "col_id": 0,
                 "positions": [[2, 20, 400, 50, 200]]}}
 ]}
```

## Design notes

**`parse` takes the PDF, not a rasterized page.** The pipeline renders the PDF itself and needs the whole document to accumulate cross-page coordinates, so the REST layer passes the uploaded bytes straight through. The existing per-page rasterizing path is untouched for the other three tasks — the new branch is entered only for tasks in `WHOLE_DOCUMENT_OCR_TASKS`. Bytes travel in the existing `image=` channel, which `ModelActor.ocr` already excludes from its logs (`@log_async(ignore_kwargs=["image"])`), so no core or client changes were needed.

**Reusing the loaded recognizers.** A plain `PdfParser()` would construct its own OCR, layout and table recognizers on top of the ones the model already holds. Measured on an RTX 3090 Ti, that costs **+512 MiB** of extra ONNX sessions. Instead the constructor is bypassed and only the attributes it would have set are assembled, wiring in the existing instances; object identity is asserted in testing (`parser.ocr is model._ocr`, etc.). If a future `deepdoc-lib` needs attributes this doesn't know about, it falls back to the real constructor and injects afterwards, with a warning.

**Per-document state is released.** `parse_into_bboxes` leaves the document it processed attached to the parser (rasterized pages, a crop per box, extracted chars) and never clears it. Since the parser is cached, that state would stay resident after the response — for a 3-page document that is 39 MB of page images plus 52 crops; at the 200-page limit and the largest render scale it is gigabytes. Worse, `__images__` resets `boxes` *before* loading the new document but swallows load failures, so a PDF it cannot open would leave the previous request's pages attached and silently parse those. Both are handled by clearing the state in a `finally`.

**Payload control.** Every element carries a crop internally, but only tables and figures need one. `image_scope` defaults to `table_figure`; on the test document that is 277 KB versus 4.1 MB for `all` (a ~15x difference) and 15 KB for `none`.

**Validation.** `validate_pdf_for_parse` checks the document with pypdfium2 before deepdoc sees it, reusing the existing `MAX_PDF_OCR_PAGES` (200) ceiling so the endpoint has one page limit regardless of task. This turns deepdoc's silent-empty-result and its 3x-zoom recursive retry into a clean 400. `pages`/`dpi` are rejected with 400 for `parse` rather than silently ignored.

`model_ability` and the model spec are unchanged: it gates endpoint routing rather than tasks, and the existing three tasks are likewise not advertised there.

## Parity with local CPU output

The motivating requirement is that DeepDoc on GPU through Xinference match running `deepdoc-lib` locally on CPU. Measured on the same 3-page PDF, `zoomin=3`, comparing a local CPU `PdfParser().parse_into_bboxes()` run against this endpoint's response on an RTX 3090 Ti:

| | Result |
|---|---|
| element count | 52 vs 52 — **match** |
| per-type counts | `{text: 40, title: 8, table: 2, figure: 2}` both sides — **match** |
| text, in order (element *i* vs element *i*) | 0 of 52 differ — **match** |
| table HTML, verbatim | 2 tables, 1219 chars — **match** |
| `positions` | 0 differ — **match** |
| `page_number` | 0 differ — **match** |
| `layoutno` | 0 differ — **match** |
| coordinates (`x0/x1/top/bottom`) | max absolute delta **4.6e-3**, confined to the 4 re-inserted table/figure boxes; all 40 text boxes are exact |
| `col_id` | varies run to run (0–6 of 52), **including between two CPU runs** |

Two honest caveats rather than a claim of bit-exactness:

- The coordinate deltas are GPU/CPU ONNX floating-point differences. They affect only the four boxes that `_extract_table_figure` re-inserts, and are ~5e-3 at a 1836x2376 render — sub-pixel.
- `col_id` is not reproducible in `deepdoc-lib` itself: column clustering uses `KMeans(n_clusters=k, n_init="auto")` with no `random_state`, and `random.choices()` feeds the is-English heuristic. Two consecutive local CPU runs already disagree on 6 of 52 elements. This is pre-existing upstream behaviour, not introduced here.

Everything structural — element count, ordering, text, table HTML, positions, page numbers — is identical.

## Verification

- **Regression:** all three existing tasks return 200 unchanged, on both single-image and PDF uploads; `pages`/`dpi` still select and rasterize as before (`pages: [1, 3]` verified).
- **VRAM:** 5543 MiB after launch, plateauing at 9367 MiB and flat across 10 consecutive parse requests. The recognizers are shared, confirmed by object identity; a second `PdfParser()` would have added 512 MiB.
- **Error paths:** non-PDF upload with `task=parse` → 400; `pages`/`dpi` with `parse` → 400; oversized/unreadable PDF → 400; invalid `zoomin`/`image_scope` → clear validation error.
- **Tests:** 72 new/extended unit tests; `pytest xinference/api/tests/ xinference/model/image/ocr/tests/` → 321 passed, 17 skipped. `pre-commit` (black, ruff, isort, mypy, codespell) green on all changed files.

Not run: the full non-GPU CI matrix and the model-download integration tests, which need more hardware and network than this change touches.

## Note on #5295

Left alone deliberately. The virtualenv built for this model on my box resolved `transformers` to 4.57.x under `virtualenv/v4/DeepDoc/default/`, i.e. the `#engine#`-gated `transformers<5` constraint did not apply — consistent with that report. But I also saw a `.../DeepDoc/deepdoc/` virtualenv where the marker *did* engage, so the trigger condition needs pinning down on a clean environment before changing a shared dependency constraint. That belongs in #5295 rather than here.

Closes #5296

### DIFF
--- a/xinference/api/pdf_ocr.py
+++ b/xinference/api/pdf_ocr.py
@@ -44,6 +44,11 @@ MAX_PDF_OCR_PAGE_PIXELS = 80_000_000
 # OCR tasks that consume the whole PDF instead of one rasterized page at a
 # time, and therefore receive the uploaded bytes unchanged.
 WHOLE_DOCUMENT_OCR_TASKS = frozenset({"parse"})
+# Whole-document parsers render at ``72 * zoomin`` DPI, i.e. a scale of
+# ``zoomin`` over the page's point size. DeepDoc additionally re-renders at
+# ``zoomin * 3`` when a first pass finds no boxes, but only while
+# ``zoomin < 9`` -- so a parse started at or above this never amplifies.
+PDF_PARSE_RETRY_ZOOM_LIMIT = 9
 
 
 def is_pdf_upload(content_type: Optional[str], head: bytes) -> bool:
@@ -53,17 +58,26 @@ def is_pdf_upload(content_type: Optional[str], head: bytes) -> bool:
     return head.startswith(PDF_MAGIC)
 
 
-def validate_pdf_for_parse(data: bytes) -> int:
+def validate_pdf_for_parse(data: bytes, zoomin: int = 3) -> int:
     """Check a PDF that will be handed to a whole-document parser.
 
     Whole-document parsing tasks (e.g. DeepDoc's ``task="parse"``) render the
     PDF themselves, so the bytes are passed straight through instead of being
-    rasterized here. That skips the validation ``rasterize_pdf`` would have
-    done, and parsers tend to fail unhelpfully: DeepDoc swallows load errors
-    and returns an empty result, then retries at 3x the zoom (9x the pixels)
-    when it finds no boxes. Validating up front turns both into a clean
-    error. Returns the page count, and raises ``ValueError`` if the document
-    cannot be opened or has too many pages.
+    rasterized here. That skips everything ``rasterize_pdf`` would have
+    validated, and parsers tend to fail unhelpfully: DeepDoc swallows load
+    errors and returns an empty result, then re-renders at three times the
+    zoom when it finds no boxes.
+
+    Page geometry therefore has to be checked here too: a single valid page
+    with an outsized MediaBox — 14400x14400 points is legal — rasterizes to
+    billions of pixels and exhausts the worker long before the page limit is
+    relevant. The budget is applied at the requested scale; the caller is
+    responsible for keeping the parser's own retry from amplifying it (see
+    ``PDF_PARSE_RETRY_ZOOM_LIMIT``).
+
+    Returns the page count. Raises ``ValueError`` if the document cannot be
+    opened, has no pages, has too many pages, or has a page whose raster
+    would be too large.
     """
     try:
         import pypdfium2 as pdfium
@@ -81,16 +95,29 @@ def validate_pdf_for_parse(data: bytes) -> int:
             raise ValueError(f"Could not read the uploaded PDF: {e}") from e
         try:
             page_count = len(pdf)
+            if page_count < 1:
+                raise ValueError("The uploaded PDF has no pages")
+            if page_count > MAX_PDF_OCR_PAGES:
+                raise ValueError(
+                    f"The uploaded PDF has {page_count} pages, at most "
+                    f"{MAX_PDF_OCR_PAGES} pages can be parsed per request"
+                )
+            for page_number in range(1, page_count + 1):
+                page = pdf[page_number - 1]
+                try:
+                    width, height = page.get_size()
+                finally:
+                    page.close()
+                pixels = int(width * zoomin) * int(height * zoomin)
+                if pixels > MAX_PDF_OCR_PAGE_PIXELS:
+                    raise ValueError(
+                        f"Page {page_number} would rasterize to {pixels} pixels "
+                        f"at zoomin {zoomin:g}, exceeding the limit of "
+                        f"{MAX_PDF_OCR_PAGE_PIXELS}; lower `zoomin`"
+                    )
         finally:
             pdf.close()
 
-    if page_count < 1:
-        raise ValueError("The uploaded PDF has no pages")
-    if page_count > MAX_PDF_OCR_PAGES:
-        raise ValueError(
-            f"The uploaded PDF has {page_count} pages, at most "
-            f"{MAX_PDF_OCR_PAGES} pages can be parsed per request"
-        )
     return page_count
 
 
--- a/xinference/api/pdf_ocr.py
+++ b/xinference/api/pdf_ocr.py
@@ -49,6 +49,15 @@ WHOLE_DOCUMENT_OCR_TASKS = frozenset({"parse"})
 # ``zoomin * 3`` when a first pass finds no boxes, but only while
 # ``zoomin < 9`` -- so a parse started at or above this never amplifies.
 PDF_PARSE_RETRY_ZOOM_LIMIT = 9
+# Unlike the per-page path, which rasterizes lazily and holds one page at a
+# time, whole-document parsers render every page up front and keep them all
+# alive at once. The per-page pixel limit therefore says nothing about a
+# document's total footprint, so the pages are budgeted together as well.
+# Measured against pdfplumber, a rendered page costs ~4 bytes per pixel once
+# the PIL object is accounted for, so this bounds the page images at roughly
+# 4 GB; deepdoc then adds crops, characters and a deepcopy of the boxes on
+# top. It is sized to admit a full 200-page A4 document at the default zoom.
+MAX_PDF_PARSE_TOTAL_PIXELS = 1_000_000_000
 
 
 def is_pdf_upload(content_type: Optional[str], head: bytes) -> bool:
@@ -68,16 +77,20 @@ def validate_pdf_for_parse(data: bytes, zoomin: int = 3) -> int:
     errors and returns an empty result, then re-renders at three times the
     zoom when it finds no boxes.
 
-    Page geometry therefore has to be checked here too: a single valid page
-    with an outsized MediaBox — 14400x14400 points is legal — rasterizes to
-    billions of pixels and exhausts the worker long before the page limit is
-    relevant. The budget is applied at the requested scale; the caller is
+    Page geometry therefore has to be checked here too, in two ways. A single
+    valid page with an outsized MediaBox — 14400x14400 points is legal —
+    rasterizes to billions of pixels on its own. And because every page is
+    rendered up front and held together, a document whose pages are each
+    comfortably under the per-page limit can still exhaust the worker in
+    aggregate, so the pages are budgeted as a whole too.
+
+    Both budgets are applied at the requested scale; the caller is
     responsible for keeping the parser's own retry from amplifying it (see
     ``PDF_PARSE_RETRY_ZOOM_LIMIT``).
 
     Returns the page count. Raises ``ValueError`` if the document cannot be
-    opened, has no pages, has too many pages, or has a page whose raster
-    would be too large.
+    opened, has no pages, has too many pages, or would rasterize to too many
+    pixels on any single page or across the document.
     """
     try:
         import pypdfium2 as pdfium
@@ -102,6 +115,7 @@ def validate_pdf_for_parse(data: bytes, zoomin: int = 3) -> int:
                     f"The uploaded PDF has {page_count} pages, at most "
                     f"{MAX_PDF_OCR_PAGES} pages can be parsed per request"
                 )
+            total_pixels = 0
             for page_number in range(1, page_count + 1):
                 page = pdf[page_number - 1]
                 try:
@@ -114,6 +128,15 @@ def validate_pdf_for_parse(data: bytes, zoomin: int = 3) -> int:
                         f"Page {page_number} would rasterize to {pixels} pixels "
                         f"at zoomin {zoomin:g}, exceeding the limit of "
                         f"{MAX_PDF_OCR_PAGE_PIXELS}; lower `zoomin`"
+                    )
+                total_pixels += pixels
+                if total_pixels > MAX_PDF_PARSE_TOTAL_PIXELS:
+                    raise ValueError(
+                        f"The uploaded PDF would rasterize to more than "
+                        f"{total_pixels} pixels in total at zoomin {zoomin:g}, "
+                        f"exceeding the whole-document limit of "
+                        f"{MAX_PDF_PARSE_TOTAL_PIXELS}; lower `zoomin` or split "
+                        f"the document"
                     )
         finally:
             pdf.close()
--- a/xinference/api/pdf_ocr.py
+++ b/xinference/api/pdf_ocr.py
@@ -56,10 +56,32 @@ PDF_PARSE_RETRY_ZOOM_LIMIT = 9
 
 
 def worst_case_parse_zoom(zoomin: int) -> int:
-    """The largest scale a whole-document parse can actually render at."""
-    if zoomin < PDF_PARSE_RETRY_ZOOM_LIMIT:
-        return zoomin * PDF_PARSE_RETRY_ZOOM_FACTOR
-    return zoomin
+    """The largest scale a whole-document parse can actually render at.
+
+    The retry is recursive: each pass that still finds no boxes triples the
+    zoom again, and only the ``< limit`` guard stops it. So zoomin 1 renders
+    at 1, 3 and finally 9, not just 3.
+    """
+    scale = zoomin
+    while scale < PDF_PARSE_RETRY_ZOOM_LIMIT:
+        scale *= PDF_PARSE_RETRY_ZOOM_FACTOR
+    return scale
+
+
+def worst_case_parse_peak_pixels(width: float, height: float, zoomin: int) -> int:
+    """Peak pixels one page can occupy during a whole-document parse.
+
+    ``self.page_images = [...]`` builds the new list before rebinding the
+    name, so the render that a retry replaces is still referenced while the
+    replacement is being allocated. Peak is therefore the final render plus
+    the one immediately before it.
+    """
+    final = worst_case_parse_zoom(zoomin)
+    previous = final // PDF_PARSE_RETRY_ZOOM_FACTOR
+    peak = int(width * final) * int(height * final)
+    if previous >= zoomin:
+        peak += int(width * previous) * int(height * previous)
+    return peak
 
 
 # Budgets for whole-document parsing. Measured against pdfplumber, a
@@ -75,12 +97,20 @@ def worst_case_parse_zoom(zoomin: int) -> int:
 MAX_PDF_PARSE_PAGE_PIXELS = 200_000_000
 # Unlike the per-page path, which rasterizes lazily and holds one page at a
 # time, whole-document parsers render every page up front and keep them all
-# alive at once, so the pages are budgeted together as well. This one is
-# enforced at the requested scale: `page_images` is reassigned rather than
-# appended to on a retry, so peak usage is one render's worth, and sizing
-# the aggregate for the retry would mean either a 32 GB allowance or
-# refusing ordinary 200-page documents. Sized to admit a full 200-page A4
-# document at the default zoom (0.9 G px, ~3.6 GB).
+# alive at once, so the pages are budgeted together as well -- and, like the
+# per-page ceiling, at the scale a retry can actually reach.
+#
+# Peak usage is the final render plus the one before it: deepdoc assigns
+# ``self.page_images = [...]``, and the comprehension is fully built before
+# the name is rebound, so the previous render is still referenced while the
+# new one is allocated. ``worst_case_parse_peak_pixels`` accounts for both.
+#
+# At 4 GB this admits roughly 22 A4 pages, well short of the 200-page ceiling
+# the per-page path allows. That is the honest consequence of budgeting for a
+# retry that renders at 9x: a document only reaches this if every page yields
+# no text, but the allocation is real when it does, and permitting tens of
+# gigabytes would leave the OOM open. Raise this if parse workers are sized
+# for it; `pages`-style batching in deepdoc-lib would lift it properly.
 MAX_PDF_PARSE_TOTAL_PIXELS = 1_000_000_000
 
 
@@ -148,23 +178,24 @@ def validate_pdf_for_parse(data: bytes, zoomin: int = 3) -> int:
                     width, height = page.get_size()
                 finally:
                     page.close()
-                worst_pixels = int(width * worst_case) * int(height * worst_case)
-                if worst_pixels > MAX_PDF_PARSE_PAGE_PIXELS:
+                peak_pixels = worst_case_parse_peak_pixels(width, height, zoomin)
+                if peak_pixels > MAX_PDF_PARSE_PAGE_PIXELS:
                     raise ValueError(
-                        f"Page {page_number} would rasterize to {worst_pixels} "
+                        f"Page {page_number} would rasterize to {peak_pixels} "
                         f"pixels at zoomin {zoomin:g} (the parser re-renders at "
                         f"up to {worst_case:g}x when a page yields no text), "
                         f"exceeding the per-page limit of "
                         f"{MAX_PDF_PARSE_PAGE_PIXELS}; lower `zoomin`"
                     )
-                total_pixels += int(width * zoomin) * int(height * zoomin)
+                total_pixels += peak_pixels
                 if total_pixels > MAX_PDF_PARSE_TOTAL_PIXELS:
                     raise ValueError(
                         f"The uploaded PDF would rasterize to more than "
-                        f"{total_pixels} pixels in total at zoomin {zoomin:g}, "
-                        f"exceeding the whole-document limit of "
-                        f"{MAX_PDF_PARSE_TOTAL_PIXELS}; lower `zoomin` or split "
-                        f"the document"
+                        f"{total_pixels} pixels in total at zoomin {zoomin:g} "
+                        f"(the parser re-renders at up to {worst_case:g}x when a "
+                        f"page yields no text), exceeding the whole-document "
+                        f"limit of {MAX_PDF_PARSE_TOTAL_PIXELS}; lower `zoomin` "
+                        f"or split the document"
                     )
         finally:
             pdf.close()
--- a/xinference/api/pdf_ocr.py
+++ b/xinference/api/pdf_ocr.py
@@ -41,6 +41,9 @@ MAX_PDF_OCR_PAGES = 200
 # pixels; an A3 page at 600 DPI is ~70 MP) is rejected so a single
 # oversized page cannot exhaust the API process.
 MAX_PDF_OCR_PAGE_PIXELS = 80_000_000
+# OCR tasks that consume the whole PDF instead of one rasterized page at a
+# time, and therefore receive the uploaded bytes unchanged.
+WHOLE_DOCUMENT_OCR_TASKS = frozenset({"parse"})
 
 
 def is_pdf_upload(content_type: Optional[str], head: bytes) -> bool:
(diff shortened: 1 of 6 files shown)


xinference/api/pdf_ocr.py
```diff
@@ -50,6 +53,47 @@ def is_pdf_upload(content_type: Optional[str], head: bytes) -> bool:
...
+            pdf.close()
+
+    if page_count < 1:
+        raise ValueError("The uploaded PDF has no pages")
+    if page_count > MAX_PDF_OCR_PAGES:
+        raise ValueError(
+            f"The uploaded PDF has {page_count} pages, at most "
+            f"{MAX_PDF_OCR_PAGES} pages can be parsed per request"
+        )
+    return page_count
```
OliverBryant, line 203: [P1] Enforce a raster-size limit before handing the PDF to DeepDoc

This validation only opens the document and counts pages, so the whole-document path bypasses the existing `MAX_PDF_OCR_PAGE_PIXELS` protection. A valid one-page PDF with a 14400 x 14400 point MediaBox is accepted here. DeepDoc then calls pdfplumber `to_image(resolution=72 * zoomin)`, which is 43,200 x 43,200 (about 1.87 billion pixels) at the default zoom. More importantly, when no boxes are found, DeepDoc `__images__` recursively retries at `zoomin * 3`; therefore an allowed `zoomin=6` can actually render at 18.

A small, valid upload can consequently OOM the model worker before the 200-page limit helps. Please inspect every page dimension here and reject documents whose effective worst-case raster exceeds a pixel budget (accounting for the retry), or disable/bound that retry, and add a regression test with an oversized MediaBox.

OliverBryant: Confirmed and fixed in fdfec9e — this was a real hole, and worse than the page limit alone.

Measured on the 14400x14400 pt MediaBox you describe:

| zoomin | raster | RGB |
|---|---|---|
| 3 (default) | 43200 x 43200 = **1.87 G px** | 5.6 GB |
| 6 | 86400 x 86400 = 7.46 G px | 22.4 GB |
| 18 (after the retry) | 259200 x 259200 = 67.2 G px | 201.6 GB |

So a single one-page upload blew the 80 MP budget by 23x at the *default* zoom, before the retry even entered into it.

Two changes:

1. **`validate_pdf_for_parse` now measures every page** and rejects against `MAX_PDF_OCR_PAGE_PIXELS` at the requested scale. That upload now returns 400 with `Page 1 would rasterize to 1866240000 pixels at zoomin 3, exceeding the limit of 80000000; lower 'zoomin'`.

2. **The retry is suppressed rather than budgeted.** I first tried budgeting against the worst-case 3x scale as you suggested, but that over-rejects badly — it fails a plain **A3 page at the default zoomin=3** (81 MP worst-case) and blocks `zoomin > 3` for every standard page size. Since the retry only fires when OCR found *zero* boxes document-wide, and `parse_into_bboxes` runs all later stages at the *original* zoom (so re-rendered pages no longer match the coordinate scale — a correctness bug, not just memory), the right move was to stop it. `_disable_zoom_retry` wraps `__images__` so the recursive call is a no-op.

Verified end-to-end: a blank A4 now returns `{"task":"parse","elements":[]}` in 0.12s after a single render at zoomin=3 (instrumented the real `__images__`; call log is `[3]`, previously `[3, 9]`). A0 at zoomin=3 (72 MP) still passes, so legitimate large-format documents are unaffected.

Regression tests added: `test_rejects_an_oversized_page`, `test_budget_scales_with_zoomin`, `test_common_page_sizes_pass_at_every_allowed_zoom`, and `test_zoom_retry_is_suppressed`.

xinference/api/restful_api.py
```diff
@@ -1978,7 +1980,41 @@ async def create_ocr(
...
+                data = image.file.read()
+                try:
+                    await asyncio.to_thread(validate_pdf_for_parse, data)
+                except ValueError as ve:
+                    raise HTTPException(status_code=400, detail=str(ve))
+                result = await model_ref.ocr(
+                    image=data,
+                    **parsed_kwargs,
+                )
+                return Response(content=result, media_type="application/json")
```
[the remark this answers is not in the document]
OliverBryant: Fixed in fdfec9e. `model_ref.ocr` is now wrapped so a `ValueError` from the model becomes a 400 rather than falling through to the generic 500 handler.

Verified against the live endpoint:

```
zoomin=99      -> 400 {"detail":"Invalid zoomin: 99, expected an integer between 1 and 6"}
image_scope=x  -> 400 {"detail":"... Invalid image_scope: 'bogus'. Supported values: 'table_figure', 'all', 'none'."}
```

`zoomin` is additionally validated in the API layer now, because it is needed to budget the page raster before the bytes are handed over; both layers share the one `parse_zoomin` definition so they cannot drift apart.

xinference/model/image/ocr/deepdoc.py
```diff
@@ -130,23 +291,136 @@ def _get_table_recognizer(self):
...
+        tokenizer_cfg = TokenizerConfig()
+
+        booster = xgb.Booster()
+        try:
+            import torch
+
+            if torch.cuda.is_available():
+                booster.set_param({"device": "cuda"})
+        except Exception:
+            logger.debug("torch unavailable, running the xgb booster on CPU")
```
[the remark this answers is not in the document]
OliverBryant: Applied in fdfec9e, though with a correction to the suggested form.

The premise does not quite hold today: `create_ocr_model_instance` never passes `device=`, so `self._device` is always `None` for this model, and GPU placement comes from `gpu_idx` via `CUDA_VISIBLE_DEVICES` — which `torch.cuda.is_available()` already respects. The suggested snippet, `if self._device and "cuda" in self._device`, would therefore have been a behaviour change in the wrong direction: with `_device` unset it would never enable CUDA, silently pushing the booster onto CPU on every GPU deployment.

The underlying point is still worth honouring, so the guard is now `if not self._device or "cuda" in self._device` — CUDA availability continues to drive the decision by default, and an explicitly non-CUDA `device` is respected on top of that.

OliverBryant (commented):
One remaining memory-safety issue in the whole-document raster validation.

xinference/api/pdf_ocr.py
```diff
@@ -50,6 +58,69 @@ def is_pdf_upload(content_type: Optional[str], head: bytes) -> bool:
...
+                    f"{MAX_PDF_OCR_PAGES} pages can be parsed per request"
+                )
+            for page_number in range(1, page_count + 1):
+                page = pdf[page_number - 1]
+                try:
+                    width, height = page.get_size()
+                finally:
+                    page.close()
+                pixels = int(width * zoomin) * int(height * zoomin)
+                if pixels > MAX_PDF_OCR_PAGE_PIXELS:
```
OliverBryant: [P1] Bound the aggregate raster size, not only each page

This check caps each page independently, but DeepDoc 0.2.2 renders all pages up front and retains them together in `self.page_images` before recognition. Therefore a document can pass every check here and still exhaust host memory: 200 A4 pages at `zoomin=6` total 3,607,128,000 pixels, or about 10.8 GB for raw RGB alone (before PIL objects, extracted characters, crops, boxes, and copies). The existing 80 MP limit is safe for the per-page path because that path renders lazily one page at a time; it is not a sufficient whole-document budget. Please accumulate `pixels` across pages and reject against a separate `MAX_PDF_PARSE_TOTAL_PIXELS` limit, with a regression test where every page is individually below 80 MP but their aggregate exceeds the total budget.

OliverBryant: Confirmed and fixed in 9fbfdbb. You are right that the per-page limit says nothing about the whole document — I had only closed half the hole.

Verified the premise first: `page_images` is built by a single list comprehension over every page (`pdf_parser.py:1106`), so all pages really are rendered up front and retained together.

Then measured the actual cost rather than assuming 3 bytes/px — 20 A4 pages at zoomin 6 rendered through pdfplumber gave **+1454 MB RSS for 360.7 MP, i.e. 4.03 bytes/pixel** once the PIL object is counted. Extrapolated to 200 pages that is **~14.5 GB**, worse than the 10.8 GB in your estimate, and that is before crops, characters and the `deepcopy(self.boxes)`.

`validate_pdf_for_parse` now accumulates `pixels` across pages and rejects against a new `MAX_PDF_PARSE_TOTAL_PIXELS`. I sized it at 1 G px (~4 GB of page images) because that is the largest round number that still admits the headline case — a full 200-page A4 document at the default zoom is 0.9 G px and must keep working. Anything tighter starts rejecting ordinary documents: at 400 M px a 200-page A4 file fails at the default zoom.

End-to-end on the live endpoint:

```
200x A4, zoomin=6  -> 400  "would rasterize to more than 1009995840 pixels in total
                            at zoomin 6, exceeding the whole-document limit of
                            1000000000; lower `zoomin` or split the document"
200x A4, zoomin=3  -> 200  (completes in 19s)
```

Two regression tests, in the shape you asked for: `test_rejects_an_oversized_document_of_individually_small_pages` asserts explicitly that the per-page value is under `MAX_PDF_OCR_PAGE_PIXELS` while the aggregate exceeds `MAX_PDF_PARSE_TOTAL_PIXELS`, and `test_full_length_document_passes_at_the_default_zoom` pins the case the budget must not break.

Parity with local CPU output is unchanged: 52 elements, 0 text diffs, table HTML identical, max coordinate delta 4.6e-3.

qinxuye (commented):
Two DeepDoc parity issues need to be addressed before this can be approved.

xinference/model/image/ocr/deepdoc.py
```diff
@@ -214,5 +545,38 @@ def _process_single(
...
+                "DeepDoc task 'parse' requires the raw bytes of a PDF "
+                f"document, got {type(data).__name__}. Upload a PDF to the "
+                "OCR endpoint instead of an image."
+            )
+
+        zoomin = parse_zoomin(kwargs.get("zoomin"))
+        image_scope = _parse_image_scope(kwargs)
+
+        parser = self._get_pdf_parser()
+        _disable_zoom_retry(parser)
```
qinxuye: P1: This changes DeepDoc parsing semantics rather than only enforcing a resource limit. The recursive higher-resolution pass is DeepDoc's fallback when the first render produces no OCR boxes; suppressing it can make Xinference return an empty document where native PdfParser.parse_into_bboxes() retries and recovers content. Since this task is intended to match native DeepDoc, please preserve the native retry. If the retry needs a memory budget, expose/enforce that through a supported deepdoc-lib option (or validate the actual retry scale) instead of monkey-patching __images__ to drop it.

xinference/model/image/ocr/deepdoc.py
```diff
@@ -130,23 +336,141 @@ def _get_table_recognizer(self):
...
+            try:
+                import torch
+
+                if torch.cuda.is_available():
+                    booster.set_param({"device": "cuda"})
+            except Exception:
+                logger.debug("torch unavailable, running the xgb booster on CPU")
+        booster.load_model(os.path.join(xgb_dir, "updown_concat_xgb.model"))
+
+        parser = PdfParser.__new__(PdfParser)
```
qinxuye: P2: Please do not bypass PdfParser.__init__ and reproduce DeepDoc's private initialization state in Xinference. The supported DeepDoc path is PdfParser(model_cfg=..., tokenizer_cfg=...); this copy can silently drift on any ~=0.2.2 patch upgrade, and the fallback above only catches assembly-time failures, not a missing attribute first accessed by parse_into_bboxes(). Make the native parser the owner of these recognizers, or add a supported recognizer-injection API to deepdoc-lib, depend on that version here, and construct the parser through its public API.

xinference/model/image/ocr/deepdoc.py
OliverBryant: You are right, and fixed in 6ec12ea — the native retry is restored and the monkey-patch is gone.

I should also correct my earlier justification for removing it. I claimed the retry left later stages working at a mismatched scale, which is wrong: `__ocr` divides box coordinates by the `ZM` it rendered at (`pdf_parser.py:322`), so a retry is self-consistent and `_layouts_rec(zoomin)` receives already-normalised boxes. The only real cost was memory, and that does not justify changing what the task returns — especially for a task whose whole purpose is matching native DeepDoc.

So the budget now validates the actual retry scale, as you suggested. `worst_case_parse_zoom(zoomin)` returns `zoomin * 3` while `zoomin < 9`, and the per-page check is enforced against that.

That needed a separate per-page ceiling rather than reusing `MAX_PDF_OCR_PAGE_PIXELS`: at 80 MP, counting the 3x would reject an ordinary **A3 page at the default zoom** (81 MP worst-case). `MAX_PDF_PARSE_PAGE_PIXELS` is 200 MP, which admits A4 (41 MP) and A3 (81 MP) at the default zoom while still rejecting an outsized MediaBox — the 14400 pt maximum is 16.8 G px at the worst case and now returns 400.

One deliberate asymmetry worth flagging: the **aggregate** budget stays at the requested scale, not the worst case. `page_images` is reassigned rather than appended to on a retry, so peak usage is one render's worth; budgeting the aggregate for the retry would mean either a ~32 GB allowance or refusing ordinary 200-page documents (200 A4 pages at the retry scale is 8.1 G px ≈ 33 GB at the ~4 bytes/pixel I measured). If you would rather bound that pathological case too — it needs *every* page in a 200-page document to yield zero boxes — the lever is `MAX_PDF_PARSE_TOTAL_PIXELS`, and I am happy to tighten it or gate the retry on page count instead. Flagging rather than deciding, since it trades a rare worst case against rejecting normal documents.

Verified: a blank A4 returns `{"task":"parse","elements":[]}` through the native retry, and parity on the real document is unchanged (52 elements, 0 text diffs, table HTML identical).

xinference/model/image/ocr/deepdoc.py
OliverBryant: Agreed and fixed in 6ec12ea. The `__new__` bypass is gone; the parser is built through the supported constructor:

```python
parser = PdfParser(
    model_cfg=PdfModelConfig(
        vision_model_dir=self._model_dir(),
        xgb_model_dir=xgb_dir,
        model_provider="local",
    )
)
parser.ocr = self._ocr
parser.layouter = self._get_layout_recognizer()
parser.tbl_det = self._get_table_recognizer()
```

DeepDoc now owns and initializes `tokenizer`, `updown_cnt_mdl` and `parallel_limiter`, so none of its private setup is reproduced here and a patch upgrade cannot leave a half-initialized parser. That also removed my `_build_parallel_limiter` reimplementation and the fallback path entirely — there is nothing left to drift, so nothing left to catch.

Your point that the fallback only caught assembly-time failures was the decisive one: a missing attribute first touched inside `parse_into_bboxes()` would have escaped it, and I had no way to test for that short of a future release breaking in production.

On the reason the bypass existed — avoiding a second set of ONNX sessions — the swap turns out to be sufficient on its own. The recognizers the constructor built are dropped and freed by refcount, and measured steady state is **+164 MiB** over holding a single set, versus the **512 MiB** a retained second set costs. So the construct-then-inject path keeps essentially all of the benefit without touching private state. Recognizer reuse still verified by identity (`parser.ocr is model._ocr`, and likewise for `layouter`/`tbl_det`).

A supported recognizer-injection API in deepdoc-lib would be cleaner still, and I am happy to open that upstream and depend on it once released, if you would prefer to hold this PR for it. Otherwise the public-constructor path above has no private-state dependency.

Verified on GPU: 52 elements, 0 text diffs, table HTML identical, max coordinate delta 4.6e-3; all three per-page tasks unchanged (200 on image and PDF); 327 tests pass.

xinference/api/pdf_ocr.py
```diff
@@ -41,6 +41,47 @@
...
+# ``zoomin < 9``, so a parse started at or above that never amplifies. That
+# retry is part of the parsing behaviour and is left intact, so the budget
+# below is enforced against the largest scale a run can actually reach.
+PDF_PARSE_RETRY_ZOOM_FACTOR = 3
+PDF_PARSE_RETRY_ZOOM_LIMIT = 9
+
+
+def worst_case_parse_zoom(zoomin: int) -> int:
+    """The largest scale a whole-document parse can actually render at."""
+    if zoomin < PDF_PARSE_RETRY_ZOOM_LIMIT:
```
qinxuye: P1: This still underestimates the native retry for two accepted inputs because DeepDoc retries recursively while the current zoom is below 9. For zoomin=1 the render chain is 1 -> 3 -> 9, but this returns 3; for zoomin=2 it is 2 -> 6 -> 18, but this returns 6. Those requests can therefore exceed the validated per-page budget by another 9x in pixels. Please keep multiplying until the resulting zoom is at least PDF_PARSE_RETRY_ZOOM_LIMIT, and test every accepted zoomin value.

xinference/api/pdf_ocr.py
```diff
@@ -50,6 +91,87 @@ def is_pdf_upload(content_type: Optional[str], head: bytes) -> bool:
...
+                worst_pixels = int(width * worst_case) * int(height * worst_case)
+                if worst_pixels > MAX_PDF_PARSE_PAGE_PIXELS:
+                    raise ValueError(
+                        f"Page {page_number} would rasterize to {worst_pixels} "
+                        f"pixels at zoomin {zoomin:g} (the parser re-renders at "
+                        f"up to {worst_case:g}x when a page yields no text), "
+                        f"exceeding the per-page limit of "
+                        f"{MAX_PDF_PARSE_PAGE_PIXELS}; lower `zoomin`"
+                    )
+                total_pixels += int(width * zoomin) * int(height * zoomin)
```
qinxuye: P1: The aggregate budget also has to cover the native retry scale. A 200-page A4 document passes here at about 0.9G pixels with the default zoom, but if the first pass yields no boxes DeepDoc renders the document again at zoom 9, which is about 8.1G pixels (roughly 32 GB before crops and copies). Moreover, self.page_images = [...] builds the new list before replacing the old one, so the first render remains referenced while the retry list is being constructed. Please bound the aggregate at the actual worst-case retry scale, or implement a supported page-count/batched retry limit in deepdoc-lib; leaving this path able to allocate tens of GB does not close the OOM issue.

qinxuye (commented):
Two retry-budget bugs remain after restoring native DeepDoc behavior.

xinference/api/pdf_ocr.py
OliverBryant: Correct, this was a straight bug — fixed in 017ba36. The retry is recursive and I only multiplied once.

The real chains, and what my code returned:

| zoomin | actual render chain | old return | now |
|---|---|---|---|
| 1 | 1 → 3 → 9 | 3 | 9 |
| 2 | 2 → 6 → 18 | 6 | 18 |
| 3 | 3 → 9 | 9 | 9 |
| 4 | 4 → 12 | 12 | 12 |
| 5 | 5 → 15 | 15 | 15 |
| 6 | 6 → 18 | 18 | 18 |

So `zoomin=1` and `zoomin=2` were validated against a scale 9x smaller in pixels than they could reach, exactly as you describe. It now loops until the scale reaches `PDF_PARSE_RETRY_ZOOM_LIMIT`:

```python
scale = zoomin
while scale < PDF_PARSE_RETRY_ZOOM_LIMIT:
    scale *= PDF_PARSE_RETRY_ZOOM_FACTOR
return scale
```

Tests cover every accepted `zoomin` rather than a couple of examples, as you asked: `test_chain_is_followed_to_the_ceiling` pins all six values explicitly, and two property-style tests assert that for every accepted input the result is `>= PDF_PARSE_RETRY_ZOOM_LIMIT` and `>= zoomin` — so a future off-by-one in the loop cannot pass.

Verified end-to-end: an outsized page at `zoomin=1`, which the old chain let through the per-page check, now returns 400 reporting 18662400000 pixels at the 9x scale.

xinference/api/pdf_ocr.py
OliverBryant: You are right on both counts, and fixed in 017ba36. My "peak is one render's worth" reasoning was wrong twice over: a retry re-renders the *whole document*, so the aggregate reaches the retry scale too, and — as you point out — `self.page_images = [...]` evaluates the comprehension fully before rebinding the name, so the render being replaced is still referenced while its replacement is allocated.

Both budgets now go through `worst_case_parse_peak_pixels`, which counts the final render plus the preceding one:

```python
final = worst_case_parse_zoom(zoomin)
previous = final // PDF_PARSE_RETRY_ZOOM_FACTOR
peak = int(width * final) * int(height * final)
if previous >= zoomin:
    peak += int(width * previous) * int(height * previous)
```

**The cost, stated plainly:** enforcing the aggregate at the retry scale drops the effective allowance to about **22 A4 pages** at the default zoom, against the 200-page ceiling the per-page OCR path permits. An A4 page peaks at 45 MP (9x plus the coexisting 3x), so 200 of them come to 9 G px ≈ 36 GB.

I went with closing the OOM rather than preserving the page count, since you were explicit that allocating tens of GB does not close the issue. But it is a visible product regression for anyone parsing long documents, so I have documented it in the OCR docs and pinned it in tests — `test_typical_document_passes_at_the_default_zoom` (20 pages) and `test_page_ceiling_alone_does_not_admit_a_long_document` (200 pages rejected on pixels, not page count) — rather than let it be discovered in production.

If you would rather keep long-document support, the proper fix is the batched/page-limited retry in deepdoc-lib that you mention: bounding the retry to a page window would let the aggregate budget go back to the requested scale and restore the 200-page allowance. I am happy to open that upstream and depend on it, and to raise `MAX_PDF_PARSE_TOTAL_PIXELS` here in the meantime if parse workers are sized for more than 4 GB — your call on which trade you prefer.

Verified: a 200-page A4 document at the default zoom now returns 400 naming the whole-document limit; the real 3-page document is unchanged (52 elements, 0 text diffs, table HTML identical); a blank A4 still returns an empty result through the native retry.

qinxuye (approved):
LGTM

PR #5300, opened 2026-08-07
fix(worker): preserve Jina v3 allocator env for recovery

m199369309:
## Summary

- apply `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` only to `jina-embeddings-v3` embedding launches
- preserve user-provided `PYTORCH_CUDA_ALLOC_CONF` or `PYTORCH_ALLOC_CONF` values
- persist the resulting environment in recovery launch arguments
- add focused tests covering injection, user overrides, recovery persistence, and unaffected models

## Tests

- `pytest -q xinference/core/tests/test_worker.py --disable-warnings --maxfail=1` (59 passed)
- `pre-commit run --files xinference/core/worker.py xinference/core/tests/test_worker.py`

### DIFF
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -285,6 +285,33 @@ _VRAM_READY_RATIO = 0.90
 _NVML_INIT_TIMEOUT = 10
 
 
+def _inject_jina_v3_allocator_env(
+    model_type: str,
+    model_name: str,
+    envs: Optional[Dict[str, str]],
+    launch_args: Dict[str, Any],
+) -> Optional[Dict[str, str]]:
+    """Add the Jina v3 allocator default and persist it for recovery.
+
+    The launch arguments are captured with ``locals()`` before this helper is
+    called.  Updating ``envs`` alone would therefore affect the current launch
+    but not the cached arguments used by worker restart recovery.
+    """
+    if model_type.lower() != "embedding" or model_name.lower() != "jina-embeddings-v3":
+        return envs
+
+    updated_envs = dict(envs or {})
+    if (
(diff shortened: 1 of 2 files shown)


m199369309:
Addressed the review feedback in commit `b57cba41`:

- `_inject_jina_v3_allocator_env` now safely ignores missing or non-string `model_type` / `model_name` values instead of calling `.lower()` on them.
- Added coverage for `None` and empty-string values.
- Validation: full `xinference/core/tests/test_worker.py` passed (`63 passed`); pre-commit checks passed.

xinference/core/worker.py
```diff
@@ -285,6 +285,38 @@ def _exclusive_venv_path_lock(env_path: str):
...
+        not isinstance(model_type, str)
+        or not isinstance(model_name, str)
+        or model_type.lower() != "embedding"
+        or model_name.lower() != "jina-embeddings-v3"
+    ):
+        return envs
+
+    updated_envs = dict(envs or {})
+    if (
+        "PYTORCH_CUDA_ALLOC_CONF" not in updated_envs
```
qinxuye: [P1] Preserve allocator settings inherited from the worker process

This only checks the per-launch `envs`, so a user setting `PYTORCH_CUDA_ALLOC_CONF` or `PYTORCH_ALLOC_CONF` in the worker process environment is still overwritten by the injected default. Xoscar builds the subpool environment from `os.environ` and then applies this `envs` mapping, so the injected `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` wins. Please also treat either key in `os.environ` as an explicit user configuration and add a regression test for the inherited-environment case.

qinxuye (approved):

PR #5304, opened 2026-08-09
ENH: add page navigation and deep pagination to log and audit centers

leslie2046:
## Summary

This PR improves pagination in both the Log Center and Audit Center and removes the previous 10,000-record browsing limitation for Elasticsearch-backed data.

It adds direct page navigation, first/previous/next/last-page controls, and replaces Elasticsearch `from + size` deep pagination with PIT-backed `search_after` queries.

图片

## Motivation

Previously, the Log Center and Audit Center only provided previous/next-page navigation, which made it inconvenient to browse a large number of records or jump directly to a known page.

The backend also clamped `page_from` to Elasticsearch's default 10,000-result window. As a result:

- records beyond the first 10,000 results could not be accessed;
- the total number of pages shown by the frontend was artificially limited;
- users could not jump directly to a specific page;
- reaching the first or last page required repeatedly clicking the previous or next button.

Increasing `index.max_result_window` through configuration would only move the limit and could significantly increase Elasticsearch memory usage. This PR instead uses the recommended `search_after` approach.

### Pagination UI

Updated both the Log Center and Audit Center pagination controls to support:

- jumping directly to a specified page;
- navigating to the first page;
- navigating to the previous page;
- navigating to the next page;
- navigating to the last page;
- displaying the current page and total page count;
- validating and clamping manually entered page numbers;
- submitting page navigation with the Enter key;
- disabling navigation buttons when already at the corresponding boundary.

The new controls include accessible labels and translated text for:

- English;
- Simplified Chinese;
- Japanese;
- Korean.

### Elasticsearch deep pagination

Added a shared Elasticsearch pagination implementation for log and audit searches.

The new implementation:

- opens an Elasticsearch point-in-time snapshot before querying;
- uses an exact `track_total_hits` query to determine the complete result count;
- uses `search_after` instead of `from + size`;
- sorts by `@timestamp` with `_shard_doc` as a stable tie-breaker;
- refreshes the PIT ID when Elasticsearch returns an updated one;
- closes the PIT in a `finally` block;
- preserves the existing `page_from` and `size` API parameters for backward compatibility;
- removes the hard-coded 10,000-record clamp from both backend endpoints;
- removes the corresponding 10,000-record limit from the frontend page-count calculation.

For arbitrary page jumps, the backend traverses from whichever end of the result set is closer. For example, opening the last page uses ascending traversal from the end instead of scanning every preceding result.

This allows administrators to browse beyond Elasticsearch's default 10,000-result window without changing `index.max_result_window`.

### Error handling

The shared search helper also handles:

- Elasticsearch PIT creation failures;
- failed search requests;
- missing PIT or sort values;
- connection and timeout errors;
- best-effort PIT cleanup with warning logs when cleanup fails.

### Documentation

Updated the audit and security documentation to explain that Elasticsearch-backed searches now use PIT and `search_after` for deep pagination.

## Compatibility

The existing API contract remains unchanged:

- `page_from` is still an offset;
- `size` still controls the requested page size;
- responses still contain `hits` and `total`;
- existing log and audit filters continue to work as before.

No Elasticsearch index setting changes or new environment variables are required.

## Test coverage

Added regression tests covering:

- retrieving records beyond the default 10,000-result window;
- confirming that Elasticsearch `from` is no longer used;
- confirming that `search_after` is used for deep pages;
- propagation of updated PIT IDs between requests;
- PIT cleanup using the latest PIT ID;
- preservation of `_source` filtering;
- efficient retrieval of a partial last page from the nearest end;
- correct descending result order after ascending traversal.

## Validation

The following checks were completed successfully:

- frontend production build;
- TypeScript type checking;
- ESLint checks for the affected frontend components;
- Python Ruff checks;
- Python Black formatting checks;
- Python syntax compilation;
- focused Elasticsearch deep-pagination regression tests.


xinference/api/routers/admin.py
```diff
@@ -386,6 +386,160 @@ async def reset_monitor_config(request: Request) -> JSONResponse:
...
+                    "numeric_type": "date_nanos",
+                }
+            },
+            {"_shard_doc": order},
+        ]
+        search_after: Optional[list[Any]] = None
+
+        while remaining > 0:
+            batch_size = min(remaining, _ES_SEARCH_AFTER_BATCH_SIZE)
+            batch_body: dict[str, Any] = {
```
qinxuye, line 506: This makes arbitrary page jumps O(min(page_from, total - page_from)) and every skipped batch still retrieves document `_source` (for audit it defaults to the full source, and for logs only `@version` is excluded). A jump near the middle of 1,000,000 hits therefore transfers and discards roughly 500,000 documents across about 100 sequential requests. `search_after` is efficient when the caller carries the previous page's cursor; it does not make random offset access cheap. Please set `_source: false` for traversal batches and add a hard traversal/request budget, or redesign this around reusable PIT/cursor tokens instead of unlimited random page jumps.

xinference/api/routers/admin.py
```diff
@@ -386,6 +386,160 @@ async def reset_monitor_config(request: Request) -> JSONResponse:
...
+            params={"keep_alive": _ES_PIT_KEEP_ALIVE},
+            headers=headers,
+        ) as resp:
+            pit_data = await _read_response(resp, "PIT open")
+        pit_id = pit_data.get("id", "")
+        if not pit_id:
+            logger.error("%s PIT open returned no id", error_context)
+            raise HTTPException(status_code=502, detail="Elasticsearch query failed")
+
+        count_data = await _search(
```
qinxuye, line 470: The default range query contains `now-1h` / `now`, and this same query is executed once for the count and again for every traversal/page request. Elasticsearch evaluates `now` at the time of each request; the PIT freezes the index view, not the date-math clock. During a long deep-page traversal the matching window therefore moves between requests, so `total`, `remaining`, and the final page can disagree or the loop can return an empty page early. Please resolve relative bounds to fixed absolute timestamps once before the first PIT search and reuse that frozen query for every request.

qinxuye (commented):
Found two remaining correctness and scalability issues in the PIT pagination implementation.

xinference/api/routers/admin.py
leslie2046: resolved

xinference/api/routers/admin.py
leslie2046: resolved

xinference/api/routers/admin.py
```diff
@@ -386,6 +386,174 @@ async def reset_monitor_config(request: Request) -> JSONResponse:
...
+        total = _get_es_total(count_data)
+        if page_from >= total:
+            return [], total
+
+        result_size = min(size, total - page_from)
+        reverse_offset = total - (page_from + result_size)
+        ascending = reverse_offset < page_from
+        remaining = reverse_offset if ascending else page_from
+        max_traversal_hits = _ES_SEARCH_AFTER_BATCH_SIZE * _ES_MAX_SEARCH_AFTER_REQUESTS
+        if remaining > max_traversal_hits:
```
qinxuye, line 483: This new 400 path is not handled safely by either caller UI. The Audit Center catch block clears `records` and `total`, and the Log Center clears `logs` and `total`, while both keep the attempted `pageFrom`. After a user jumps to a valid but traversal-limited middle page, the UI therefore loses the last result set and can show a large current page with only one total page. Please preserve the last successful data/total and reset `pageFrom` while showing the error, or expose the traversal limit so these jumps are rejected before issuing the request.

qinxuye (commented):
One user-visible pagination regression remains after adding the traversal limit.

xinference/api/routers/admin.py
leslie2046: Fixed. Both the Audit Center and Log Center now preserve the last successful data and total when a page request fails. They also track the last successfully loaded page offset and restore `pageFrom` to that value after a failed foreground page jump.

The backend error is still surfaced through the existing global request error handler, so users see the traversal-limit error without leaving the pagination state inconsistent.

Validated with Prettier, ESLint, and `tsc --noEmit`.

frontend/src/components/pages/audit-center/index.tsx
```diff
@@ -225,38 +231,41 @@ export default function AuditCenter() {
...
+          setRecords(Array.isArray(data.hits) ? data.hits : []);
+          setTotal(data.total || 0);
+          lastSuccessfulPageFromRef.current = pageFrom;
+        }
+      } catch {
+        // A transient failure during a silent background refresh must not wipe
+        // the table. A failed foreground page jump also keeps the last good
+        // results and restores the page that produced them.
+        if (!silent && seq === requestSeqRef.current) {
+          setPageFrom(lastSuccessfulPageFromRef.current);
```
qinxuye: This restoration runs for every foreground request failure, not only a page-only jump. If the user is on page 5 and changes a filter or time range, the UI first sets `pageFrom` to 0; if that new query fails, this restores page 5 while the new filters remain active and the old rows remain visible, then the effect issues another request for page 5 under the new query. The Log Center catch block has the same behavior. Please track the last successful query parameters separately and restore the old offset only when all non-page parameters are unchanged; otherwise keep/reset page 0 and avoid presenting stale rows as results for the new filters.

qinxuye (commented):
The page restoration still needs to distinguish page-only failures from query/filter changes. The PR also currently conflicts with main after #5282.

frontend/src/components/pages/audit-center/index.tsx
leslie2046: Fixed. The Audit Center and Log Center now track both the last successful non-pagination query parameters and its `pageFrom`.

On request failure:

- If only the page offset changed, the UI preserves the existing results and restores the last successfully loaded page.
- If filters, search terms, or the time range changed, the UI clears the stale results and keeps/resets `pageFrom` to 0 instead of restoring an offset from the previous query.

This avoids retrying the new query with an unrelated old page offset or presenting old rows as results for the new filters.

Validated with Prettier, ESLint, and `tsc --noEmit`.

qinxuye (approved):
LGTM

Issue #5307, opened 2026-08-12
BUG: task=parse rejects a 31-page A4 PDF at every zoomin, and lowering zoomin makes it worse


OliverBryant:
## Describe the bug

`task="parse"` rejects a 31-page A4 PDF at every `zoomin` value from 1 to 6, so the document cannot be parsed at all. The error tells the caller to lower `zoomin`, but doing so makes the computed budget *larger*, not smaller.

The document is an ordinary text PDF: 31 A4 pages, 393 KB. At `zoomin=3` it actually rasterizes to about 140 MP — well under the 1 GP whole-document budget. It is rejected because the budget is enforced against a worst case that overstates real usage by 10x.

## To Reproduce

```bash
curl -X POST http://<host>:<port>/v1/images/ocr \
  -H "Authorization: Bearer $TOKEN" \
  -F model=DeepDoc \
  -F 'image=@thesis-31-pages.pdf;type=application/pdf' \
  -F 'kwargs={"task":"parse","zoomin":3}'
```

```json
{"detail": "The uploaded PDF would rasterize to more than 1037049300 pixels in total at zoomin 3 (the parser re-renders at up to 9x when a page yields no text), exceeding the whole-document limit of 1000000000; lower `zoomin` or split the document"}
```

Following the advice and lowering `zoomin` reports a **bigger** number:

| `zoomin` | worst-case scale | reported total | rejected |
|---|---|---|---|
| 3 | 9 | 1,037,049,300 | yes |
| 2 | **18** | **1,082,138,400** | yes |
| 1 | 9 | 1,037,049,300 | yes |

## Why lowering `zoomin` does not help

`worst_case_parse_zoom` in `xinference/api/pdf_ocr.py` triples the scale until it reaches the limit:

```python
def worst_case_parse_zoom(zoomin: int) -> int:
    scale = zoomin
    while scale < PDF_PARSE_RETRY_ZOOM_LIMIT:   # 9
        scale *= PDF_PARSE_RETRY_ZOOM_FACTOR    # 3
    return scale
```

The loop exits on `scale >= 9`, so it overshoots whenever `zoomin` is not a power-of-three divisor of 9:

| `zoomin` | escalation chain | worst-case scale |
|---|---|---|
| 1 | 1 → 3 → 9 | 9 |
| 2 | 2 → 6 → **18** | **18** |
| 3 | 3 → 9 | 9 |
| 4 | 4 → 12 | 12 |
| 5 | 5 → 15 | 15 |
[...]

PR #5308, opened 2026-08-12
fix(ocr): stop task="parse" rejecting ordinary multi-page PDFs

OliverBryant:
## What

Fixes the size-budget defect in `task="parse"` reported in #5307: an ordinary 31-page A4 text PDF (393 KB) was rejected at **every** `zoomin` from 1 to 6, and the 400 advised lowering `zoomin`, which makes the computed budget *larger*.

## Why it happened

Two independent causes, both confirmed against `deepdoc-lib` 0.2.2 rather than assumed:

**1. Every page was budgeted for a conditional retry.** `deepdoc/parser/pdf_parser.py` re-renders at `zoomin * 3` only when `len(self.boxes) == 0`. `boxes` accumulates across every page, so the escalation fires only if *not one page in the whole document* produced a single box — a scanned or broken PDF, not a text document. An A4 page renders to 4.5 MP at `zoomin=3` but was charged 45 MP, putting the whole-document ceiling at ~22 A4 pages where actual usage allows ~221.

**2. The retry ladder is not monotonic in `zoomin`.** The guard tests the pre-multiplication value:

```python
if len(self.boxes) == 0 and zoomin < 9:
    self.__images__(fnm, zoomin * 3, page_from, page_to, callback)
```

so `2 → 6 → 18` and `6 → 18`, while `3 → 9`. "Lower `zoomin`" was therefore not sound advice. The server's `worst_case_parse_zoom` models this faithfully — the arithmetic was never the bug.

## What changed

- **The whole-document budget is enforced at the requested scale**, not the retry scale. This is what makes ordinary documents parseable.
- **A separate, looser ceiling bounds the escalated re-render** (`MAX_PDF_PARSE_RETRY_TOTAL_PIXELS`, 3 G px). Without it, budgeting only at the requested scale would have left a 32–130 GB hole when the retry *does* fire — so this is not simply raising the ceiling. The escalated render *replaces* the first one (the retry re-enters `__images__`, rebinding `self.page_images`), which is why the two ceilings differ rather than summing.
- **The per-page ceiling stays at the worst-case scale**, so a single outsized MediaBox is still rejected on the strength of a retry that may fire.
- **Rejections name a `zoomin` that actually fits**, searching the whole permitted range rather than only downwards — with a non-monotonic ladder, a *higher* zoom can be the one that fits. When nothing fits, it advises splitting instead.
- Both whole-document ceilings are configurable via `XINFERENCE_MAX_PDF_PARSE_TOTAL_PIXELS` and `XINFERENCE_MAX_PDF_PARSE_RETRY_TOTAL_PIXELS`.

## Memory safety

The intent is preserved rather than traded away. An accepted document is bounded at the requested scale (~4 GB of page images at 1 G px), and if the retry fires the escalated document is bounded at ~12 GB — uniformly, regardless of `zoomin`, versus 32–130 GB under requested-scale budgeting alone. The binding constraint on long documents is now the retry ceiling at ~73 A4 pages, which is honest: a 200-page A4 document really would need ~32 GB if it escalated.

## Before / after on the reported document

31-page A4, 595x842 pt:

| `zoomin` | before | after |
|---|---|---|
| 1 | 400 | **accepted** |
| 2 | 400 (budgeted at 18x) | 400 → "retry with `zoomin` 4" |
| 3 (default) | 400 | **accepted** |
| 4 | 400 | **accepted** |
| 5 | 400 | 400 → "retry with `zoomin` 4" |
| 6 | 400 | 400 → "retry with `zoomin` 4" |

Every rejection now names zoomin 4 — which genuinely works, and is *higher* than the old advice would have sent the caller.

## Upstream

The non-monotonic ladder belongs in `deepdoc-lib`, not here. Clamping it to `min(zoomin * 3, 9)` would cap the real allocation and make the scale monotonic, at which point the retry ceiling here could be tightened. I have deliberately **not** clamped it server-side: that would model an escalation bound the parser does not actually honour. Worth filing upstream separately; this PR handles the server-side half.

## Tests

Added to `xinference/api/tests/test_ocr_pdf.py`, including the two cases that would have caught this: a realistic 31-page A4 document being accepted at the default zoom, and the property that a rejection always carries a *true* way forward (parametrized over 1/5/31/100/200 pages — either a named zoom that itself validates, or advice to split). Also covers the upward-search gap, the retry-budget invariant, and env-var overrides.

```
397 passed, 17 skipped     # xinference/api/tests/ + xinference/model/image/ocr/tests/
```

`pre-commit run --files ...` passes on all changed files (black, ruff, isort, mypy, codespell).

Docs updated in `doc/source/models/model_abilities/image.rst`, including an explicit warning that lowering `zoomin` can make the budget larger.

Ref #5307

### DIFF
--- a/xinference/api/pdf_ocr.py
+++ b/xinference/api/pdf_ocr.py
@@ -118,32 +118,26 @@ MAX_PDF_PARSE_PAGE_PIXELS = 200_000_000
 # This one is enforced at the *requested* scale. Charging every document for
 # the retry is what stopped an ordinary 31-page A4 text PDF parsing at every
 # zoomin (#5307): it was budgeted at 45 MP per page where it really renders
-# at 4.5. The retry is not a state a text document reaches -- deepdoc's guard
-# is ``len(self.boxes) == 0`` and ``boxes`` accumulates over every page, so it
-# only fires when not one page in the whole document produced a single box.
+# at 4.5.
+#
+# There is no separate document-wide ceiling for the retry scale, because for
+# any document that renders at all the retry is unreachable. deepdoc's guard
+# is ``len(self.boxes) == 0``, but ``__ocr`` appends to ``boxes`` on *every*
+# page -- ``self.boxes.append([])`` when a page yields nothing, ``append(bxs)``
+# otherwise -- so after the OCR pass ``len(self.boxes)`` equals the page count.
+# It is zero only when there were no pages to render in the first place, i.e.
+# the load failed, in which case there is nothing to re-render. Budgeting the
+# whole document for a 9x pass it cannot take would reject a 74-page A4 PDF
+# whose real render is ~334 M pixels.
+#
+# The per-page ceiling above is still checked at the worst-case scale. That is
+# cheap insurance on a single outsized MediaBox and does not depend on this
+# reasoning holding for every deepdoc-lib version.
 #
 # At 1 G px and ~4 bytes per rendered pixel this is ~4 GB of page images at
 # the requested zoom: ~221 A4 pages at the default zoomin 3, 55 at zoomin 6,
 # with the 200-page ceiling capping it from the other side.
 MAX_PDF_PARSE_TOTAL_PIXELS = 1_000_000_000
-# The retry is still real when it does fire, and it re-renders the *whole*
-# document, so the requested-scale budget alone would leave a hole: 200 A4
-# pages admitted at zoomin 3 would escalate to 8 G px (~32 GB). The escalated
-# document therefore gets its own ceiling, checked at the worst-case scale.
-#
-# It is deliberately looser than the requested-scale budget rather than equal
-# to it. The escalated render *replaces* the first one -- the retry re-enters
-# ``__images__``, which rebinds ``self.page_images`` -- so the two are not
-# summed; and reaching it at all requires a document that yielded no text
-# anywhere, which is the rare case rather than the one being priced.
-#
-# At 3 G px this holds an escalated document to ~12 GB of page images, which
-# admits 73 A4 pages at the default zoom. That is the binding constraint on
-# long documents now, and it is a real one: a 200-page A4 document really
-# would need ~32 GB if it escalated. Deployments whose parse workers are
-# sized for more can raise both ceilings via the environment rather than
-# being held to a default chosen for a modest worker.
-MAX_PDF_PARSE_RETRY_TOTAL_PIXELS = 3_000_000_000
 
 
 def _pixel_budget_from_env(name: str, default: int) -> int:
@@ -170,47 +164,33 @@ def _pixel_budget_from_env(name: str, default: int) -> int:
 MAX_PDF_PARSE_TOTAL_PIXELS = _pixel_budget_from_env(
     "XINFERENCE_MAX_PDF_PARSE_TOTAL_PIXELS", MAX_PDF_PARSE_TOTAL_PIXELS
 )
-MAX_PDF_PARSE_RETRY_TOTAL_PIXELS = _pixel_budget_from_env(
-    "XINFERENCE_MAX_PDF_PARSE_RETRY_TOTAL_PIXELS", MAX_PDF_PARSE_RETRY_TOTAL_PIXELS
-)
 
 
 def _parse_budget_error(sizes: List[Tuple[float, float]], zoomin: int) -> Optional[str]:
     """Why a document does not fit at ``zoomin``, or ``None`` if it does.
 
-    Both budgets are applied: the per-page ceiling at the worst-case scale,
-    since one oversized MediaBox must not be admitted on the strength of a
-    retry that may still fire, and the two document-wide ceilings at the
-    requested and worst-case scales respectively.
+    The per-page ceiling is applied at the worst-case scale, since one
+    oversized MediaBox must not be admitted on the strength of a retry that
+    may still fire, and the document-wide ceiling at the requested scale.
     """
     worst_case = worst_case_parse_zoom(zoomin)
     requested_total = 0
-    retry_total = 0
     for index, (width, height) in enumerate(sizes):
         peak_pixels = worst_case_parse_peak_pixels(width, height, zoomin)
         if peak_pixels > MAX_PDF_PARSE_PAGE_PIXELS:
             return (
                 f"Page {index + 1} would rasterize to {peak_pixels} pixels at "
-                f"zoomin {zoomin:g} (the parser re-renders at up to "
-                f"{worst_case:g}x when a document yields no text at all), "
-                f"exceeding the per-page limit of {MAX_PDF_PARSE_PAGE_PIXELS}"
+                f"zoomin {zoomin} (the parser can re-render at up to "
+                f"{worst_case}x), exceeding the per-page limit of "
+                f"{MAX_PDF_PARSE_PAGE_PIXELS}"
             )
         requested_total += int(width * zoomin) * int(height * zoomin)
         if requested_total > MAX_PDF_PARSE_TOTAL_PIXELS:
             return (
                 f"The uploaded PDF would rasterize to more than "
-                f"{requested_total} pixels in total at zoomin {zoomin:g}, "
+                f"{requested_total} pixels in total at zoomin {zoomin}, "
                 f"exceeding the whole-document limit of "
                 f"{MAX_PDF_PARSE_TOTAL_PIXELS}"
-            )
-        retry_total += int(width * worst_case) * int(height * worst_case)
-        if retry_total > MAX_PDF_PARSE_RETRY_TOTAL_PIXELS:
-            return (
-                f"The uploaded PDF would rasterize to more than {retry_total} "
-                f"pixels in total if the parser re-rendered it at "
-                f"{worst_case:g}x, which it does when a document yields no "
-                f"text at all, exceeding the retry limit of "
-                f"{MAX_PDF_PARSE_RETRY_TOTAL_PIXELS}"
             )
     return None
 
@@ -266,9 +246,9 @@ def validate_pdf_for_parse(
     oversized MediaBox must not be admitted on the strength of a retry that
     may still fire, and it counts the render being replaced alongside its
     replacement (see ``worst_case_parse_peak_pixels``). The document-wide
-    ceiling is applied at the requested scale, with a separate, looser one
-    bounding what a retry would re-render the document at; see
-    ``MAX_PDF_PARSE_TOTAL_PIXELS`` for why the two differ.
+    ceiling is applied at the requested scale; see
+    ``MAX_PDF_PARSE_TOTAL_PIXELS`` for why the retry is not budgeted for
+    across the document.
 
     Returns the page count. Raises ``ValueError`` if the document cannot be
     opened, has no pages, has too many pages, or would rasterize to too many
--- a/xinference/api/pdf_ocr.py
+++ b/xinference/api/pdf_ocr.py
@@ -21,8 +21,12 @@ JSON-serializable response body.
 """
 
 import json
+import logging
+import os
 import threading
 from typing import Any, Generator, List, Optional, Tuple, Union
+
+logger = logging.getLogger(__name__)
 
 # PDFium is not thread-safe: pypdfium2 documents that concurrent calls,
 # even on different documents, may crash or corrupt the process. Every
@@ -46,11 +50,15 @@ MAX_PDF_OCR_PAGE_PIXELS = 80_000_000
 WHOLE_DOCUMENT_OCR_TASKS = frozenset({"parse"})
 # Whole-document parsers render at ``72 * zoomin`` DPI, i.e. a scale of
 # ``zoomin`` over the page's point size. DeepDoc additionally re-renders at
-# ``zoomin * 3`` when a first pass finds no OCR boxes -- its recovery path
-# for pages the first render was too coarse to read -- but only while
-# ``zoomin < 9``, so a parse started at or above that never amplifies. That
-# retry is part of the parsing behaviour and is left intact, so the budget
-# below is enforced against the largest scale a run can actually reach.
+# ``zoomin * 3`` when a pass finds no OCR boxes *anywhere in the document* --
+# its recovery path for a render too coarse to read -- but only while
+# ``zoomin < 9``, so a parse started at or above that never amplifies.
+#
+# The guard tests the pre-multiplication value, so the ladder overshoots the
+# limit for any zoomin that is not a power-of-three divisor of 9: 2 escalates
+# 2 -> 6 -> 18, and 6 goes straight to 18. That makes the reachable scale
+# non-monotonic in zoomin (2 and 6 reach 18, but 3 stops at 9), which is a
+# defect in deepdoc-lib rather than here; see ``worst_case_parse_zoom``.
 PDF_PARSE_RETRY_ZOOM_FACTOR = 3
 PDF_PARSE_RETRY_ZOOM_LIMIT = 9
 
@@ -61,6 +69,14 @@ def worst_case_parse_zoom(zoomin: int) -> int:
     The retry is recursive: each pass that still finds no boxes triples the
     zoom again, and only the ``< limit`` guard stops it. So zoomin 1 renders
     at 1, 3 and finally 9, not just 3.
+
+    Because deepdoc-lib checks ``zoomin < 9`` *before* multiplying, the last
+    step can overshoot: 2 reaches 18 and 6 reaches 18, while 3 stops at 9.
+    This models that faithfully rather than clamping, so the number used for
+    budgeting is the one the parser can really allocate. The consequence is
+    that this is not monotonic in ``zoomin`` -- which is why the per-page
+    limit, the only budget still enforced at this scale, is checked against
+    every candidate zoom before one is recommended to the caller.
     """
     scale = zoomin
     while scale < PDF_PARSE_RETRY_ZOOM_LIMIT:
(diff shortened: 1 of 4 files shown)


OliverBryant:
## End-to-end verification on GPU hardware

Verified on an RTX 3090 Ti box with real DeepDoc inference, not just the validation layer. Both runs used the same 31-page A4 text PDF (595.28 x 841.89 pt), the same DeepDoc model from ModelScope, and `deepdoc-lib` 0.2.2.

### Before (unpatched, at `e45894915`)

All six `zoomin` values rejected, exactly as reported:

| `zoomin` | reported total | worst case |
|---|---|---|
| 1 | 1,037,233,622 | 9x |
| 2 | 1,082,382,462 | **18x** |
| 3 | 1,037,233,622 | 9x |
| 4 | 1,042,280,369 | 12x |
| 5 | 1,002,251,168 | 15x |
| 6 | 1,082,382,462 | **18x** |

Every message ended in ``lower `zoomin` or split the document``, while `zoomin=2` was budgeted at 18x — four times `zoomin=3`.

(The totals differ from the issue's by ~0.02% only because this document's MediaBox is 595.28 pt rather than an exact 595.)

### After (this PR, same box, same document)

| `zoomin` | result |
|---|---|
| 1 | **parsed**, 1054 elements, 42.0s |
| 2 | 400 → ``retry with `zoomin` 4`` |
| 3 (default) | **parsed**, 1054 elements, 32.1s |
| 4 | **parsed**, 1054 elements, 40.7s |
| 5 | 400 → ``retry with `zoomin` 4`` |
| 6 | 400 → ``retry with `zoomin` 4`` |

Element counts are identical across all three accepted zooms, so the budget change does not alter parse output.

### The advice is actionable

Followed the message mechanically — request at `zoomin=6`, take the value it names, retry:

```
zoomin=6 -> 400 "...exceeding the retry limit of 3000000000; retry with `zoomin` 4"
zoomin=4 -> OK  task=parse  elements=1054
            types: {'text': 1023, 'title': 31}
            first text: "Chapter 1: Evaluation of Document Parsing"
```

1023 text + 31 title matches the document's structure (one heading per page), and the text matches the source. The recommended zoom is *higher* than the request — advice the old "lower `zoomin`" message could not have produced, and which lowering to 5 would not have found either (15x, also rejected).

### Unrelated environment note

Launching DeepDoc on a fresh venv failed with `ImportError: cannot import name 'is_offline_mode' from 'huggingface_hub'` — the venv resolves `huggingface_hub==0.36.2` while the launching environment has 1.26.1, and the venv is rebuilt back to 0.36.2 on each restart. Pinning the venv to 1.26.1 fixes it. This is unrelated to this PR (it reproduces on unpatched `e45894915` too) and looks like the same class of issue as the `deepdoc-lib` hub pin from #5230; worth tracking separately.

xinference/api/pdf_ocr.py
```diff
@@ -97,21 +113,127 @@ def worst_case_parse_peak_pixels(width: float, height: float, zoomin: int) -> in
...
+            )
+        requested_total += int(width * zoomin) * int(height * zoomin)
+        if requested_total > MAX_PDF_PARSE_TOTAL_PIXELS:
+            return (
+                f"The uploaded PDF would rasterize to more than "
+                f"{requested_total} pixels in total at zoomin {zoomin:g}, "
+                f"exceeding the whole-document limit of "
+                f"{MAX_PDF_PARSE_TOTAL_PIXELS}"
+            )
+        retry_total += int(width * worst_case) * int(height * worst_case)
```
OliverBryant: [P1] The retry budget is based on a condition that deepdoc-lib 0.2.2 cannot reach for a non-empty PDF. In `__ocr`, every processed page ends with `self.boxes.append(bxs)`, including when `bxs == []`; therefore, after successful OCR, `len(self.boxes)` is the page count and `len(self.boxes) == 0` does not mean that the document produced no boxes anywhere. As a result, a 74-page A4 PDF at `zoomin=3` is rejected here solely for a hypothetical 9x re-render that will not occur (its requested-scale render is only about 334M pixels). Please remove this unconditional retry budget, or first fix and pin DeepDoc to use an actual all-pages-empty predicate and cover that behavior with an integration test against the real parser.

OliverBryant: [P2] If this retry path is made reachable, this total still understates the actual peak. While evaluating `self.page_images = [...]`, Python fully constructs the new list while the previous full list remains referenced. The aggregate peak is therefore the final render plus the preceding render, exactly as `worst_case_parse_peak_pixels()` already accounts for on a per-page basis. For the 3→9 and 6→18 paths, the real image peak is 10/9 of `retry_total`, so a document admitted just below the 3G ceiling can hold about 3.33G pixels simultaneously. Please sum `worst_case_parse_peak_pixels()` across pages, or lower the ceiling to account for the previous render.

OliverBryant: You're right, and this invalidates the premise I built the retry budget on. Fixed in eae6c3a5a.

I had grepped only for `self.boxes.append(bxs)` and concluded "boxes accumulates across pages, so `len == 0` means nothing was found anywhere". Reading the full function, `__ocr` has two append paths:

```python
bxs = self.ocr.detect(np.array(img), device_id)
if not bxs:
    self.boxes.append([])      # line 317
    return
...
self.boxes.append(bxs)         # line 373
```

so after the OCR pass `len(self.boxes)` is the page count regardless of what was detected, and nothing clears it between there and the `len(self.boxes) == 0` check. The guard is only true when there were no pages to iterate in the first place — a failed load — in which case there is nothing to re-render anyway. The 9x pass is unreachable for any document that renders.

That also explains something I had recorded but not followed up on: across six end-to-end runs of a 31-page document on real hardware, I never observed a 9x re-render.

Removed `MAX_PDF_PARSE_RETRY_TOTAL_PIXELS` and its aggregate check entirely. A 74-page A4 PDF at `zoomin=3` is now accepted, as is the full 200-page ceiling.

I kept the **per-page** ceiling at the worst-case scale. It is cheap, it guards a single outsized MediaBox, and it does not depend on this reasoning holding for every `deepdoc-lib` version — if the predicate is ever fixed to a real all-pages-empty test, the per-page guard is already correct. It is also what keeps the budget non-monotonic in `zoomin`, so the recommendation logic this PR adds is still load-bearing (a 1000x1000 pt page fits at 1, 3 and 4 but not at 2).

On pinning: I have not added a `deepdoc-lib` pin here, since this PR no longer depends on the retry's behaviour in either direction. The upstream predicate still looks wrong to me — `len(self.boxes) == 0` was presumably meant to be something like `not any(self.boxes)` — and I'll raise that separately rather than encoding a workaround in the server.

OliverBryant: Correct as arithmetic, and resolved by removing the budget entirely in eae6c3a5a rather than by adjusting it.

Confirming the reasoning was sound on its own terms: `page_images = [...]` does fully build the new list before rebinding, so the aggregate peak would have been the final render plus the preceding one — 10/9 of `retry_total` on the 3→9 and 6→18 paths, letting a document admitted just under 3 G px hold ~3.33 G px. Summing `worst_case_parse_peak_pixels()` across pages would have been the right fix.

Since P1 establishes the aggregate retry path is unreachable, there is no longer a document-wide retry figure to undercount, so I dropped the check instead of correcting it. The per-page ceiling retains `worst_case_parse_peak_pixels()` unchanged, which already counts the coexisting previous render — that is where the two-renders-alive concern still applies, and it is still enforced.

xinference/api/pdf_ocr.py
```diff
@@ -97,21 +113,127 @@ def worst_case_parse_peak_pixels(width: float, height: float, zoomin: int) -> in
...
+    worst_case = worst_case_parse_zoom(zoomin)
+    requested_total = 0
+    retry_total = 0
+    for index, (width, height) in enumerate(sizes):
+        peak_pixels = worst_case_parse_peak_pixels(width, height, zoomin)
+        if peak_pixels > MAX_PDF_PARSE_PAGE_PIXELS:
+            return (
+                f"Page {index + 1} would rasterize to {peak_pixels} pixels at "
+                f"zoomin {zoomin:g} (the parser re-renders at up to "
+                f"{worst_case:g}x when a document yields no text at all), "
```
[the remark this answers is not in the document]
OliverBryant: The stated reason is not correct — `:g` is valid for `int` and does not raise:

```python
>>> f"{3:g}", f"{9:g}"
('3', '9')
```

`ValueError: Unknown format code 'g' for object of type 'int'` comes from `Decimal`, not `int`. `:g` was also already in use on these values before this PR, so it was not introducing a crash.

That said, the suggested change is worth making for a different reason: `:g` switches to scientific notation past six significant digits, so a large value would render as `1e+06`. `zoomin` is capped at 6 and `worst_case` at 18, so it could not trigger here — but `:g` is meaningless for integers regardless. Applied in eae6c3a5a.

xinference/api/pdf_ocr.py
```diff
@@ -97,21 +113,127 @@ def worst_case_parse_peak_pixels(width: float, height: float, zoomin: int) -> in
...
+                f"exceeding the per-page limit of {MAX_PDF_PARSE_PAGE_PIXELS}"
+            )
+        requested_total += int(width * zoomin) * int(height * zoomin)
+        if requested_total > MAX_PDF_PARSE_TOTAL_PIXELS:
+            return (
+                f"The uploaded PDF would rasterize to more than "
+                f"{requested_total} pixels in total at zoomin {zoomin:g}, "
```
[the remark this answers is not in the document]
OliverBryant: Same as [the first instance](https://github.com/xorbitsai/inference/pull/5308#discussion_r3763387782): `:g` does not raise for `int` (that error is `Decimal`'s), but it is meaningless for integers and would use scientific notation for large values, so the specifiers are dropped in eae6c3a5a. Note this particular line no longer exists — the retry budget it belonged to was removed following the review above.

xinference/api/pdf_ocr.py
```diff
@@ -97,21 +113,127 @@ def worst_case_parse_peak_pixels(width: float, height: float, zoomin: int) -> in
...
+                f"exceeding the whole-document limit of "
+                f"{MAX_PDF_PARSE_TOTAL_PIXELS}"
+            )
+        retry_total += int(width * worst_case) * int(height * worst_case)
+        if retry_total > MAX_PDF_PARSE_RETRY_TOTAL_PIXELS:
+            return (
+                f"The uploaded PDF would rasterize to more than {retry_total} "
+                f"pixels in total if the parser re-rendered it at "
+                f"{worst_case:g}x, which it does when a document yields no "
```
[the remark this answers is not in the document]
OliverBryant: Same as [the first instance](https://github.com/xorbitsai/inference/pull/5308#discussion_r3763387782): `:g` does not raise for `int` (that error is `Decimal`'s), but it is meaningless for integers and would use scientific notation for large values, so the specifiers are dropped in eae6c3a5a. Note this particular line no longer exists — the retry budget it belonged to was removed following the review above.

OliverBryant:
## Updated following review — the retry budget is gone

The review established that the 9x re-render is **unreachable**, which invalidated the premise the retry budget rested on. `eae6c3a5a` removes it.

`__ocr` appends to `self.boxes` on every page — `append([])` when a page detects nothing, `append(bxs)` otherwise — so `len(self.boxes)` equals the page count after the OCR pass, and the `len(self.boxes) == 0` guard the re-render sits behind can only be true when there were no pages to render at all. My original claim ("boxes accumulates, so `len == 0` means nothing was found anywhere") came from grepping one of the two append sites, and was wrong.

This also retroactively explains an observation I had recorded but not chased: across six end-to-end runs on real hardware, a 9x re-render never occurred.

### What changed since the first review

| | before review | now |
|---|---|---|
| Whole-document budget | requested scale + 3 G px retry ceiling | requested scale only |
| 31-page A4 | accepted at 1, 3, 4 | **accepted at every zoomin 1–6** |
| 74-page A4 @ zoomin 3 | rejected | **accepted** |
| 200-page A4 @ zoomin 3 | rejected | **accepted** |
| `MAX_PDF_PARSE_RETRY_TOTAL_PIXELS` | new constant + env var | removed |

The per-page ceiling still runs at the worst-case scale. It is cheap, it guards a single outsized MediaBox, and it does not depend on the reachability argument holding for every `deepdoc-lib` version. It is also what keeps the budget non-monotonic in `zoomin` — a 1000x1000 pt page fits at 1, 3 and 4 but not at 2 — so the recommendation logic remains load-bearing and is still covered by tests.

Also dropped the `:g` format specifiers on integer zoom values.

### Upstream

The predicate still looks wrong upstream: `len(self.boxes) == 0` was presumably intended as something like `not any(self.boxes)`, which would make the recovery path actually reachable for a document that renders but reads as blank. I have deliberately **not** pinned or worked around it here, since this PR no longer depends on the retry behaving one way or the other. Worth raising against `deepdoc-lib` separately.

### Verification

```
396 passed, 17 skipped     # xinference/api/tests/ + xinference/model/image/ocr/tests/
```

pre-commit (black, ruff, isort, mypy, codespell) passes on all changed files.

The GPU end-to-end results posted earlier were produced with the retry budget still in place; they remain valid as a before/after for the original bug, and the change since only *widens* what is accepted, with the accepted set now a superset of what was verified there.

xinference/api/pdf_ocr.py
```diff
@@ -97,31 +113,119 @@ def worst_case_parse_peak_pixels(width: float, height: float, zoomin: int) -> in
...
+# This one is enforced at the *requested* scale. Charging every document for
+# the retry is what stopped an ordinary 31-page A4 text PDF parsing at every
+# zoomin (#5307): it was budgeted at 45 MP per page where it really renders
+# at 4.5.
 #
-# Peak usage is the final render plus the one before it: deepdoc assigns
-# ``self.page_images = [...]``, and the comprehension is fully built before
-# the name is rebound, so the previous render is still referenced while the
-# new one is allocated. ``worst_case_parse_peak_pixels`` accounts for both.
+# There is no separate document-wide ceiling for the retry scale, because for
```
OliverBryant: [P1] This makes whole-document memory safety depend on a bug in one particular DeepDoc release, but the runtime dependency is not pinned to that release: `deepdoc-lib~=0.2.2` accepts every later 0.2.x version. If upstream changes the guard to the likely intended `not any(self.boxes)`, a fresh environment will pick up that fix while this validation still admits 200 A4 pages at `zoomin=3` (~0.902G requested pixels); the retry then renders ~8.116G pixels at 9x (roughly 32GB for page images alone), and every individual page still passes the 200M per-page limit. So the PR does continue to depend on the retry remaining unreachable. Please either pin exactly the known behavior and revisit the budget on upgrades, retain a configurable aggregate retry ceiling, or condition the retry ceiling on the installed DeepDoc behavior/version.

OliverBryant: You're right that the PR still depends on the retry staying unreachable, and I should not have claimed otherwise. But I tried the configurable aggregate ceiling you suggested and it does not work — the bound you are asking for already exists, from a direction I had not spelled out.

**The existing limits already cap the escalated case.** The per-page ceiling is still enforced at the *worst-case* scale, so no page may peak above `MAX_PDF_PARSE_PAGE_PIXELS` (200 M) even after escalating, and at most `MAX_PDF_OCR_PAGES` (200) pages are accepted. An escalated document therefore cannot exceed their product, 40 G px, whatever upstream does to the guard.

That is not just the trivial product — I searched every page geometry and zoom that *both* budgets admit, and the true maximum is 39.96 G px (200 pages of 200x11100 pt at zoomin 1), i.e. the bound is nearly tight. Pinned as a test in 853dd696b so raising either limit has to confront what it does to the escalated worst case.

**Why I did not add the third ceiling.** I implemented it and measured it before discarding it. Any value tight enough to bind is also tight enough to reject ordinary documents:

| ceiling | rejects | catches anything the existing bound misses |
|---|---|---|
| 3 G px (the one this PR removed) | 74-page A4 at zoomin 3 | — |
| 36 G px | 200-page A4 at zoomin 2 and 6 | no |
| 40 G px | nothing | no — unreachable, dead code |

Above ~40 G px it can never fire; below it, it starts rejecting documents the requested-scale budget correctly admits. There is no value that adds coverage, because the per-page × page-count bound already dominates it.

On your specific scenario: 200 A4 pages at `zoomin=3` is 0.902 G px requested and 8.116 G px escalated (~32 GB). That is real, and it is admitted. But it is admitted under a 40 G px worst case that holds regardless of the guard — the retry ceiling would only have moved that number if set low enough to also reject the documents this PR exists to unblock.

**On pinning.** I would rather not pin `deepdoc-lib` exactly. It would freeze the model environment for a reason unrelated to the model, and the 40 G px bound holds without it. If you want the escalated ceiling lower than 40 G px, the honest lever is `MAX_PDF_PARSE_PAGE_PIXELS` or `MAX_PDF_OCR_PAGES` — both already exist, and lowering either tightens the escalated worst case proportionally without depending on upstream behaviour at all. Happy to do that if you want a specific figure; say which and I will set it.

The upstream predicate still looks wrong to me and I will raise it against `deepdoc-lib` separately.

OliverBryant: The 40 Gpx product is a mathematical bound, but it is not an effective memory-safety bound. At roughly 4 bytes per rendered pixel it permits about 160 GB of page images. The concrete 200-page A4 3→9 case is only 8.116 Gpx (~32 GB), so it passes both the 200 M per-page limit and the 200-page limit while still being large enough to OOM an ordinary worker. This is a regression from the previous aggregate protection (~4 GB), not just a different expression of the same bound.

A lower aggregate retry ceiling necessarily rejects some documents if the retry becomes reachable; that rejection is the coverage it adds. It does not have to block the reported case: the original 3 Gpx retry ceiling accepts the 31-page A4 document at the default zoom (about 1.26 Gpx at 9x) while rejecting the 32 GB scenario. A deployment-tunable ceiling would also let larger workers opt in to more headroom.

The referenced `853dd696b` is not currently in the PR and GitHub cannot resolve that SHA, but a test proving the maximum is 39.96 Gpx would reinforce rather than resolve this concern: it demonstrates that the effective fallback permits nearly 160 GB. Please retain an aggregate retry ceiling with a practical default, pin the buggy DeepDoc behavior, or prevent the retry at runtime for documents whose escalated total exceeds that ceiling.

OliverBryant: Agreed on every point, and the correction that matters most is one I got wrong twice. Fixed in 33343e159 (now pushed, along with the earlier commits — the previous push failed on a network drop, which is why `853dd696b` was unresolvable).

**The 3 Gpx ceiling never rejected the reported document.** You are right, and I should have checked this instead of asserting it. The 31-page A4 PDF peaks at 1.398 Gpx at 9x — comfortably inside 3 Gpx. What actually rejected it was budgeting every page at the worst-case scale in the *requested*-scale total, which is what this PR removed. I conflated the two and used that to argue the aggregate ceiling was unusable. It was not.

**And 40 Gpx is a bound, not a protection.** ~160 GB is not a memory-safety limit, and presenting it as "the bound already exists" was wrong in substance, not just in emphasis. Dropping the aggregate ceiling was a regression against the ~4 GB it previously provided, exactly as you say.

**What is in now.** `MAX_PDF_PARSE_RETRY_TOTAL_PIXELS`, default 6 Gpx (~24 GB), summed from `worst_case_parse_peak_pixels` so the render being replaced is counted alongside its replacement — the P2 undercount from the earlier review does not come back. Tunable via `XINFERENCE_MAX_PDF_PARSE_RETRY_TOTAL_PIXELS`.

At 6 Gpx:

| document | escalated peak | result |
|---|---|---|
| 31-page A4 (the reported case) | 1.40 Gpx | accepted at **every** zoomin 1–6 |
| 74-page A4 | 3.34 Gpx | accepted at 1, 3, 4 |
| 100-page A4 | 4.51 Gpx | accepted at 1 and 3 (default works) |
| 200-page A4 @ zoomin 3 | 9.02 Gpx (~36 GB) | **rejected** — your case |

I chose 6 over 3 Gpx so a 100-page document still parses at the default zoom; 3 Gpx would cap that around 65 pages. Both are defensible and it is a one-constant change if you prefer the tighter default — say which and I will set it.

Verified that the advice stays sound under the new ceiling: across A4/A3/Letter at 1, 31, 74, 100 and 200 pages and every zoomin, each of the 50+ rejections either names a zoom that itself validates and is the largest that does, or says to split when nothing fits. No case regressed.

`399 passed, 17 skipped`; pre-commit clean. Docs updated to describe three budgets and both env vars. Still not pinning `deepdoc-lib` — with a real ceiling in place the protection no longer rests on the upstream guard staying broken, which I think addresses the underlying concern better than a pin would. I will raise the predicate upstream separately.

OliverBryant:
## End-to-end re-verified on GPU with the retry ceiling in place

Re-ran on the RTX 3090 Ti against real DeepDoc inference, with all four commits applied (`e45894915` + this branch). Same 31-page A4 document as before, plus 100- and 200-page A4 documents to exercise the new ceiling.

### The reported document — 31 pages

| `zoomin` | before the fix | now |
|---|---|---|
| 1 | 400 | **parsed**, 1054 elements, 41.8s |
| 2 | 400 | **parsed**, 1054 elements, 24.4s |
| 3 (default) | 400 | **parsed**, 1054 elements, 31.5s |
| 4 | 400 | **parsed**, 1054 elements, 42.1s |
| 5 | 400 | **parsed** |
| 6 | 400 | **parsed**, 1054 elements |

Every zoomin now works, where the previous revision of this PR only admitted 1, 3 and 4. Element counts are identical across all of them.

### The ceiling does what it is there for

| document | `zoomin` | result |
|---|---|---|
| 100-page A4 | 3 | **parsed**, 3400 elements (100 title + 3300 text), pages 1–100 |
| 100-page A4 | 4 | 400 → ``retry with `zoomin` 3`` |
| 200-page A4 | 3 | 400 → `peak above 6043013276 ... exceeding the retry limit of 6000000000; split the document` |

The 200-page case is the ~36 GB scenario raised in review, now rejected. The 100-page case confirms the ceiling is not so tight that it costs ordinary long documents the default zoom.

Followed the advice mechanically once more: 100 pages at `zoomin=4` is rejected, the message names `zoomin 3`, and that parses — 3400 elements with page numbers covering 1 through 100 and no gaps.

### One thing worth recording, unrelated to this PR

Running five or six large parses back-to-back against a single model instance eventually fails with `ONNXRuntimeError ... BFCArena ... Available memory of 0`. It is GPU memory fragmentation in the model process, not a validation problem: `zoomin=6` on the 31-page document failed that way at the end of a long run, then parsed cleanly (1054 elements) on a freshly launched instance, as did the 100-page document. It reproduces regardless of these changes and is a separate issue from the size budgets.

### Cleanup

The test instance, repo copy, test home and PDFs were removed; GPU memory and the four pre-existing services are back to their prior state.

qinxuye (commented):
One non-blocking documentation consistency issue remains after the final retry-ceiling change.

xinference/api/pdf_ocr.py
```diff
@@ -97,21 +113,141 @@ def worst_case_parse_peak_pixels(width: float, height: float, zoomin: int) -> in
...
```
qinxuye: The restored aggregate retry ceiling makes this block stale and internally contradictory: this says there is no document-wide retry ceiling and that the retry is not budgeted across the document, while the code immediately below defines and enforces `MAX_PDF_PARSE_RETRY_TOTAL_PIXELS`. The same outdated rationale remains in `_parse_budget_error`'s docstring, `validate_pdf_for_parse`'s docstring, and a few test comments that still claim 200-page A4 documents must be admitted. Please update those comments to describe the final requested-scale + 6 Gpx retry-ceiling policy; the implementation itself looks correct.

OliverBryant: Right on all four locations — thanks. Fixed in 74d1ea455; comments only, no behaviour change.

The block you flagged was the worst of it: it asserted there was no document-wide retry ceiling literally four lines above the one that defines it. That is an artefact of my restoring the ceiling by editing only the constant it sits next to, and leaving every other place that described the interim state.

Updated:

- **The constant block** — now says the retry gets its own looser ceiling *below* rather than that none exists, and the follow-on paragraph reads "the unreachability argued above" instead of implying the argument is what protects us.
- **`_parse_budget_error` docstring** — was "the per-page ceiling … and the document-wide ceiling"; now names all three.
- **`validate_pdf_for_parse` docstring** — same fix, plus the cross-reference that ended "for why the retry is not budgeted for across the document" now points at both constants. Also dropped a stale "in two ways" that had become "three".
- **Two test comments** — one still justified admitting a 200-page A4 document while its own assertion had already moved to 100 pages; the other claimed the requested-scale budget lifts the default-zoom allowance "to the 200-page ceiling". Corrected to ~130, which I measured rather than estimated: 133 A4 pages is the exact cut-off at `zoomin=3` before the retry ceiling binds.

Grepped for any remaining variant of the "no retry ceiling" claim across the module, tests and docs — none left. `399 passed, 17 skipped`; pre-commit clean.

qinxuye (approved):
LGTM

PR #5309, opened 2026-08-12
feat: add Wan2.2 Animate 2 video models

Minamiyama:


xinference/client/restful/async_restful_client.py
```diff
@@ -699,6 +702,18 @@ async def image_to_video(
...
+                with open(video, "rb") as f:
+                    video_data = f.read()
+            else:
+                video_data = video
+            files.append(
+                (
+                    "video",
+                    ("video", video_data, "application/octet-stream"),
+                )
+            )
```
qinxuye, line 718: `aiohttp` does not understand these requests-style `(filename, content, content_type)` tuples. Passing this `files` list as `data` serializes it as `application/x-www-form-urlencoded`, so neither `image` nor the new `video` reaches FastAPI as an `UploadFile` and the async client gets a 422. Please build an `aiohttp.FormData` and add the scalar/file fields with `filename` and `content_type`, as `flf_to_video()` already does.

xinference/api/restful_api.py
```diff
@@ -2493,6 +2494,8 @@ async def create_videos_from_images(
                 parsed_kwargs = json.loads(kwargs)
             else:
                 parsed_kwargs = {}
+            if video is not None:
+                parsed_kwargs["video"] = await video.read()
```
qinxuye, line 2498: These video bytes are forwarded as a `video` keyword to `ModelActor.image_to_video`, but that method's `@log_async` currently stringifies every keyword before truncating it. `str()` on a realistic driving video can allocate hundreds of MB before the 100-character truncation. Please exclude `video` there, e.g. `@log_async(logger=logger, ignore_kwargs=["video"])`, as the audio methods already do for binary inputs.

qinxuye (commented):
Two runtime issues remain in the new driving-video path.

qinxuye (approved):
LGTM

PR #5310, opened 2026-08-12
Feat: enhance download progress details

Minamiyama:
This PR adds detailed model download progress to the launch dialog, allowing users to view each file’s progress, status, download speed, and estimated remaining time while a model is being downloaded.

80b44156d17bba6c495cfb9860d7ad83


xinference/model/utils.py
```diff
@@ -976,6 +977,72 @@ def get_progress(self) -> float:
...
+
+    def get_download_progress_details(self) -> List[Dict[str, Any]]:
+        """Return JSON-serializable snapshots for files downloading right now.
+
+        Repository-level tqdm bars count files and use ``it`` as their unit.
+        File download bars use a byte unit, so only those bars are exposed.
+        Completed byte bars remain visible until the download context resets,
+        allowing clients to show both active and just-finished files.
+        """
+        _, download_progresses = self._progress_snapshots()
```
qinxuye, line 1033: These details are not scoped to this downloader when two models download concurrently on the same worker. `patched_update` iterates `_active_registry` and adds every tqdm bar to every active downloader, so this snapshot for model B can include files from model A. I reproduced this with two `CancellableDownloader` instances: both returned the same file bar. Please associate each tqdm bar with its owning downloader/context before exposing these lists; otherwise filenames, speed, and ETA are misreported across concurrent launches.

qinxuye: The new owner binding prevents cross-attribution, but it still loses file-level progress when the download library creates tqdm bars in its own worker threads. `ContextVar` values are not propagated into a nested `ThreadPoolExecutor`; with two active downloaders, `_get_tqdm_owner()` then has no context and cannot use the single-downloader fallback, so neither downloader tracks the bar. I reproduced this by running two downloaders concurrently and creating/updating each byte-progress bar in a nested executor: both `get_download_progress_details()` calls returned `[]`. Please propagate the downloader ownership into library-created worker threads (or otherwise carry an explicit owner to bar creation) and add a regression test that models this nested-thread case.

qinxuye (approved):
LGTM

PR #5312, opened 2026-08-13
fix(vllm): keep WorkerActor importable without vllm for multi-GPU launch

bluefish-08:
### Problem

Every multi-GPU vLLM launch fails on the official Docker image. Reproduced 100% of the time with 2 GPUs on `xprobe/xinference:latest`, and it is not model- or hardware-specific.

What the user sees is misleading — the surfaced error mentions neither vllm nor an import:

```
RuntimeError: Engine core initialization failed. See root cause above. Failed core proc(s): {}
  ← xoscar.errors.ServerClosed: Remote server unixsocket:///9699328 closed:
```

The real exception is only in `$XINFERENCE_HOME/logs/xinference.log`, never on stderr, because the root logger's `stream_handler` carries `LoggerNameFilter` and drops non-`xinference` loggers:

```
ERROR asyncio  Unhandled exception in client_connected_cb
  File ".../xinference/model/llm/vllm/distributed_executor_v1.py", line 32, in <module>
    from vllm.v1.executor.abstract import Executor
ModuleNotFoundError: No module named 'vllm'
→ xoscar.backends.message.DeserializeMessageFailed
```

### Root cause

1. With more than one GPU, `vllm/core.py` replaces vLLM's own mp executor with `XinferenceDistributedExecutorV1`, so each rank's `WorkerActor` gets registered through xoscar.
2. `xo.create_actor(WorkerActor, ...)` runs in a sub-pool, but the registration message is forwarded to the worker **main pool**, which deserializes it — that means unpickling the `WorkerActor` class ref, which imports its defining module.
3. With per-model virtualenv enabled (the default), the sub-pools run the model venv's python (has vllm) while the main pool runs the base env's python. On the official image the base env has **no vllm at all** (`pip show vllm` → not found; the image drops it because the per-model venv owns the engine, and the base torch is `2.11.0+cu130`, incompatible with the base image's vllm anyway).
4. So the main pool hits `ModuleNotFoundError` while deserializing. The exception escapes `client_connected_cb`, xoscar closes the connection, and the sub-pool reports the confusing `ServerClosed ... 0 bytes read`.

Single GPU never creates a `WorkerActor`, so it is unaffected. Anyone running xinference and vllm in the *same* Python environment is also unaffected — the main pool imports vllm successfully and the latent module-level coupling stays hidden. That is why this survived CI.

### Fix

Move `WorkerActor` into `distributed_worker_actor.py`, a module that does not import vllm at module level (`WorkerWrapperBase` is only needed inside `__init__`, so it becomes a local import). Since pickle resolves a class by `module.qualname`, the main pool now imports only this vllm-free module.

`XinferenceDistributedExecutorV1` itself cannot be made lazy — it subclasses vLLM's `Executor` — which is why splitting the actor out is the fix rather than reordering imports.

Also drop the `vllm_config` argument from the `create_actor` call. It was never used by `WorkerActor.__init__` (`WorkerWrapperBase.__init__` takes only `rpc_rank`/`global_rank`), but it was pickled into the same registration message, so the main pool would need vllm to deserialize a `VllmConfig` even after the class was moved.

`DEBUG_EXECUTOR` moves along with its only consumer; `distributed_executor.py` (the v0 path) keeps its own copy.

### Verification

Applied to a running container on the affected host, both copies of the package (`dist-packages` and `/opt/inference`):

- base env python (no vllm) can now import the actor module — this is exactly what the main pool does:
  `import xinference.model.llm.vllm.distributed_worker_actor` → OK
- model venv python (with vllm) still imports the executor, and `WorkerActor.__module__` is the new module, i.e. the path pickle records:
  `from ...distributed_executor_v1 import XinferenceDistributedExecutorV1, WorkerActor` → OK
- after restart, the previously 100%-failing 2-GPU launch of a 27B FP8 model starts successfully.

`black` and `isort` clean; the one `flake8` E501 in the touched file is pre-existing (only its line number shifted).

### DIFF
--- a/xinference/model/llm/vllm/distributed_executor_v1.py
+++ b/xinference/model/llm/vllm/distributed_executor_v1.py
@@ -30,10 +30,10 @@ from typing import (
 
 import xoscar as xo
 from vllm.v1.executor.abstract import Executor
-from vllm.v1.worker.worker_base import WorkerWrapperBase
 from xoscar.utils import get_next_port
 
 from ....isolation import Isolation
+from .distributed_worker_actor import WorkerActor
 from .utils import get_distributed_init_method
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


qinxuye (approved):
LGTM

PR #5317, opened 2026-08-14
fix(deploy): stop polling requests flooding the access log

bluefish-08:
### Problem

With the Web UI open, the access log is almost entirely polling:

```
INFO uvicorn.access ... "GET /v1/models/Qwen3-Reranker-8B/progress HTTP/1.1" 200
INFO uvicorn.access ... "GET /v1/models/SenseVoiceSmall/progress HTTP/1.1" 200
INFO uvicorn.access ... "GET /v1/models/SenseVoiceSmall/replicas HTTP/1.1" 200
```

The UI polls `/progress` and `/replicas` roughly once a second **per model**, and metrics scrapers hit `/metrics`, so at the default info level everything that actually matters is buried. There is currently no way to turn this off short of raising the whole log level and losing real messages with it.

This is not just cosmetic. While debugging a launch failure on a live deployment, every single log query had to start with `grep -v uvicorn.access` before anything could be read, and the useful lines were a handful among tens of thousands.

### Fix

Add a `PollingAccessFilter` on the `uvicorn.access` logger that drops **successful GETs to polling endpoints** (`/progress`, `/replicas`, `/metrics`, `/status`). Everything else is untouched:

- real API calls (`GET /v1/models`, `POST /v1/models`, `DELETE ...`) still logged
- **failures on polling endpoints still logged** — only 2xx/3xx are dropped, so a broken `/progress` stays visible
- records that are not uvicorn access lines pass through unchanged

Set `XINFERENCE_LOG_POLLING_ACCESS=true` to get the old behaviour back.

### Verification

`xinference/tests/test_logging_filters.py` covers the dropped set, the kept set (including non-2xx polling and non-GET), the opt-in env var, and a non-access record shape.

### DIFF
--- a/xinference/constants.py
+++ b/xinference/constants.py
@@ -271,6 +271,13 @@ XINFERENCE_LOG_CONSOLE = (
     os.environ.get("XINFERENCE_LOG_CONSOLE", "true").lower() == "true"
 )
 
+# The Web UI polls /progress and /replicas about once a second per model and
+# metrics scrapers hit /metrics, so at info level those access lines bury
+# everything else. Set to true to log them anyway.
+XINFERENCE_LOG_POLLING_ACCESS = (
+    os.environ.get("XINFERENCE_LOG_POLLING_ACCESS", "false").lower() == "true"
+)
+
 # Download progress logging level (only effective when XINFERENCE_LOG_CONSOLE=false)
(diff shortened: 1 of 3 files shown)


qinxuye (approved):
LGTM

PR #5318, opened 2026-08-14
fix: normalize replica during autostart recovery

### LINKED ISSUE #5305: BUG: autostart recovery still fails with 'str' object cannot be interpreted as an integer' when replica was set via UI (fix #5220 incomplete - autostart path not covered)
### System Info

- Xinference version: v3.1.0 (Docker image xprobe/xinference:v3.1.0, standalone compose deploy)
- Python: 3.12.13 (in container)
- Model: bge-m3 (embedding, engine=sentence_transformers, replica=2, CPU deployment)

### Bug description

Autostart failed for model bge-m3: 'str' object cannot be interpreted as an integer after container restart.

The model was launched from the Web UI with "replica": "2" (string, this is what the frontend always sends). The API entrypoint converts it correctly via the _validate_replica() guard added in #5220, and the model launches fine. However, the persisted launch record in launch_history.db keeps the **raw string** "2" inside its JSON data column:

`json
{"model_name": "bge-m3", "model_type": "embedding", "replica": "2", "model_engine": "sentence_transformers", ...}
`

On the next restart, the autostart recovery path _launch_autostart_model() (xinference/core/supervisor.py) pops 
eplica straight from that stored payload **without** going through _validate_replica():

`python
replica=payload.pop("replica", 1),
`

so "2" (str) reaches launch_builtin_model -> _build_replica_info() -> list(range("2")) -> TypeError: 'str' object cannot be interpreted as an integer. The autostart silently fails and the model is never recovered.

### Traceback

``
2026-08-10T03:33:58.885Z ERROR xinference.core.supervisor ... Leave launch_builtin_model, error: 'str' object cannot be interpreted as an integer
  File ".../xinference/core/supervisor.py", line 2401, in launch_builtin_model
    self._model_uid_to_replica_info[model_uid] = self._build_replica_info(replica)
  File ".../xinference/core/supervisor.py", line 190, in _build_replica_info
    active_replica_ids = list(range(replica))
TypeError: 'str' object cannot be interpreted as an integer
``

### Root cause

#5220 fixed the two **REST API** launch endpoints (restful_api.py) but the **autostart recovery** path (_launch_autostart_model in supervisor.py) still forwards the stored 
eplica value untouched. Both paths consume the same UI-produced string payload, so the fix is incomplete.

### Repro steps

1. Launch any model with replica >= 2 from the Web UI (replica is sent as a string, e.g. "replica": "2").
2. Restart the xinference container / process.
3. Observe Autostart failed for model <uid>: 'str' object cannot be interpreted as an integer in the logs.

### Expected behavior

Autostart should recover the model using 
eplica=2 (int). Either apply the same _validate_replica() / int() normalization in _launch_autostart_model(), or normalize the stored launch record when writing launch_history.db.

medisean:
## What

Normalize a persisted autostart `replica` value to an integer while loading the launch payload, and cover the UI-produced string form in the autostart regression test.

## Why

The REST launch path accepts values such as `replica="2"`, but launch history can retain that string. On restart, the autostart path previously forwarded it to replica scheduling, where `range("2")` raises a `TypeError` and prevents model recovery.

## Impact

Existing autostart records with integer strings recover correctly. Invalid boolean, float, non-numeric, zero, and negative values are rejected during normalization instead of failing later in model launch.

Fixes #5305.

## Checks

- Isolated normalization regression checks (valid string and invalid values)
- `python3 -m compileall -q xinference/core/autostart.py xinference/core/tests/test_autostart.py`
- `git diff --check`

The focused pytest file could not run in this environment because the optional `xoscar` dependency is not installed.

### DIFF
--- a/xinference/core/autostart.py
+++ b/xinference/core/autostart.py
@@ -65,9 +65,7 @@ def normalize_launch_payload(payload: Dict[str, Any]) -> Dict[str, Any]:
                 "Autostart launch config replica must be an integer."
             ) from None
         if replica < 1:
-            raise ValueError(
-                "Autostart launch config replica must be greater than 0."
-            )
+            raise ValueError("Autostart launch config replica must be greater than 0.")
         launch["replica"] = replica
     launch.pop("wait_ready", None)
     return launch
--- a/xinference/core/autostart.py
+++ b/xinference/core/autostart.py
@@ -54,6 +54,21 @@ def normalize_launch_payload(payload: Dict[str, Any]) -> Dict[str, Any]:
     launch["model_name"] = model_name.strip()
     launch["model_uid"] = model_uid.strip()
     launch.setdefault("model_type", "LLM")
+    if "replica" in launch:
+        replica = launch["replica"]
+        if isinstance(replica, (bool, float)):
+            raise ValueError("Autostart launch config replica must be an integer.")
+        try:
+            replica = int(replica)
+        except (TypeError, ValueError, OverflowError):
+            raise ValueError(
+                "Autostart launch config replica must be an integer."
+            ) from None
+        if replica < 1:
+            raise ValueError(
+                "Autostart launch config replica must be greater than 0."
+            )
+        launch["replica"] = replica
     launch.pop("wait_ready", None)
     return launch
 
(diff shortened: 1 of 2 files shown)


qinxuye:
https://github.com/xorbitsai/inference/actions/runs/31772971486/job/94682941993?pr=5318

lint failed, could you fix it?

medisean:
> https://github.com/xorbitsai/inference/actions/runs/31772971486/job/94682941993?pr=5318
> 
> lint failed, could you fix it?

OK

qinxuye (approved):

PR #5319, opened 2026-08-14
feat: [New Model] indextts 2.5

Minamiyama:


qinxuye:
All current Gemini review comments are under `xinference/thirdparty/indextts`, which is vendored from the upstream IndexTTS project. We intentionally keep this code aligned with upstream and avoid carrying local behavioral changes in this integration PR. I am resolving those threads as out of scope here; applicable fixes should be contributed upstream and picked up through a later vendor sync.

xinference/model/audio/indextts2.py
```diff
@@ -101,6 +135,19 @@ def speech(
...
+            if language is None:
+                language = kwargs.pop("lang", "ZH")
+            else:
+                kwargs.pop("lang", None)
+            language = str(language).upper()
+            duration_factor = kwargs.pop("duration_factor", None)
+            if duration_factor is None:
+                # OpenAI speed > 1 means faster, while IndexTTS duration_factor
+                # > 1 means slower.
+                duration_factor = 1.0 / speed
```
qinxuye: `SpeechRequest.speed` is optional and currently has no bounds, so an explicit `speed: null` or `speed: 0` reaches this reciprocal and fails with `TypeError` or `ZeroDivisionError` before inference. Values outside the documented `0.5`-`2.0` range also produce an unsupported `duration_factor`. Please normalize `None` to `1.0`, validate the supported range before dividing (and validate an explicit `duration_factor` as well), and add focused regression tests for these boundary cases.

qinxuye (approved):
LGTM

PR #5321, opened 2026-08-14
feat(video): add MiniMax-H3 support

qinxuye:
## Summary

- register MiniMax-H3 for text-to-video, image-to-video, and first/last-frame-to-video workflows
- support Hugging Face and ModelScope model sources with an isolated virtual environment
- default to TorchAO INT4 plus group offload so the model can load on a 24 GB consumer GPU
- preserve INT8 and unquantized BF16 loading options, and document their memory trade-offs
- mux the generated 24 FPS video and audio output and report denoising progress to the Xinference progress API

## Why

MiniMax-H3 uses Diffusers' experimental ModularPipeline and has two very large model components. A normal BF16 load does not fit consumer hardware, while a straightforward INT4 load temporarily exceeds 24 GB during quantization and Accelerate redispatch. The implementation streams quantization across CUDA and CPU, keeps selected BF16 modules on CPU, suppresses the unnecessary load-time redispatch, and hands runtime movement to Diffusers group offload.

ModularPipeline's H3 denoiser also updates its own progress bar instead of invoking the standard `callback_on_step_end`. Xinference now wraps that request-local progress bar so `/v1/requests/{request_id}/progress` advances during generation.

## Validation

- `python -m pytest xinference/model/video/tests/test_minimax_h3_video.py -q` (4 passed locally and on the GPU host)
- `pre-commit run --files ...` for all six changed files (Black, Ruff, isort, mypy, codespell passed)
- full real-weight INT4 load from ModelScope on an RTX 3090 Ti completed successfully; peak CUDA allocation was about 19.4 GiB and peak process RSS was about 51.9 GiB

### DIFF
--- a/xinference/model/video/diffusers.py
+++ b/xinference/model/video/diffusers.py
@@ -319,7 +319,7 @@ class DiffusersVideoModel:
 
         manager = ComponentsManager()
         manager.enable_auto_cpu_offload(
-            device=kwargs.pop("offload_device", "cuda"),
+            device=kwargs.pop("onload_device", "cuda"),
             memory_reserve_margin=kwargs.pop("memory_reserve_margin", "12GB"),
         )
         pipeline = ModularPipeline.from_pretrained(
@@ -699,7 +699,7 @@ class DiffusersVideoModel:
         kwargs.pop("num_videos_per_prompt", None)
         kwargs.pop("fps", None)
 
-        generate_kwargs = self._model_spec.default_generate_config.copy()
+        generate_kwargs = (self._model_spec.default_generate_config or {}).copy()
         generate_kwargs.update(kwargs)
         if num_inference_steps is not None:
             generate_kwargs["num_inference_steps"] = num_inference_steps
@@ -766,7 +766,12 @@ class DiffusersVideoModel:
 
         missing = object()
         patched_blocks = []
-        for block in iter_blocks(pipeline._blocks):
+        blocks = getattr(pipeline, "_blocks", None)
+        if blocks is None:
+            yield
+            return
+
+        for block in iter_blocks(blocks):
             if getattr(block, "model_name", None) != "minimax-h3" or not hasattr(
                 block, "loop_step"
             ):
--- a/xinference/model/video/diffusers.py
+++ b/xinference/model/video/diffusers.py
@@ -830,18 +830,24 @@ class DiffusersVideoModel:
 
         sampling_rate = output.get("sampling_rate")
         urls = []
-        for index, video in enumerate(video_outputs):
-            path = os.path.join(XINFERENCE_VIDEO_DIR, uuid.uuid4().hex + ".mp4")
-            audio = audio_outputs[index] if index < len(audio_outputs) else None
-            encode_video(
-                video,
-                fps=24,
-                output_path=path,
-                audio=audio,
-                audio_sample_rate=sampling_rate,
-            )
-            urls.append(path)
-        return urls
+        try:
+            for index, video in enumerate(video_outputs):
+                path = os.path.join(XINFERENCE_VIDEO_DIR, uuid.uuid4().hex + ".mp4")
+                urls.append(path)
+                audio = audio_outputs[index] if index < len(audio_outputs) else None
+                encode_video(
+                    video,
+                    fps=24,
+                    output_path=path,
+                    audio=audio,
+                    audio_sample_rate=sampling_rate,
+                )
+            return urls
+        except BaseException:
+            for video_path in urls:
+                with suppress(OSError):
+                    os.remove(video_path)
+            raise
 
     def _process_image(self, image: PIL.Image.Image, max_area: int) -> PIL.Image.Image:
         assert self._model is not None
--- a/xinference/model/video/diffusers.py
+++ b/xinference/model/video/diffusers.py
@@ -22,7 +22,7 @@ import tempfile
 import time
 import uuid
 from concurrent.futures import ThreadPoolExecutor
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from functools import partial, reduce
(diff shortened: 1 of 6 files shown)


xinference/model/video/diffusers.py
```diff
@@ -375,6 +630,158 @@ def firstlastframe_to_video(
...
+        def iter_blocks(block):
+            yield block
+            sub_blocks = getattr(block, "sub_blocks", None)
+            if sub_blocks:
+                for sub_block in sub_blocks.values():
+                    yield from iter_blocks(sub_block)
+
+        missing = object()
+        patched_blocks = []
+        for block in iter_blocks(pipeline._blocks):
```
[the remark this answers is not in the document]
qinxuye: Fixed in 4a6936aa3: progress tracking now handles a missing `_blocks` attribute gracefully, with a regression test for that fallback.

xinference/model/video/diffusers.py
```diff
@@ -111,6 +112,229 @@ def _register_transformer(pipeline, transformer):
...
+            raise ValueError(
+                f"Unsupported MiniMax-H3 quantization: {quantization}. "
+                "Supported values are int4, int8, none, and bf16."
+            )
+
+        from diffusers import ComponentsManager, ModularPipeline
+
+        manager = ComponentsManager()
+        manager.enable_auto_cpu_offload(
+            device=kwargs.pop("offload_device", "cuda"),
```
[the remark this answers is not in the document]
qinxuye: Fixed in 4a6936aa3: auto CPU offload now uses `onload_device`, with a regression test covering a non-default device.

xinference/model/video/diffusers.py
```diff
@@ -375,6 +630,163 @@ def firstlastframe_to_video(
...
+                call_kwargs["image"] = image
+            if last_image is not None:
+                call_kwargs["last_image"] = last_image
+            with self._track_minimax_h3_progress(pipeline, progressor, video_index, n):
+                output = pipeline(
+                    prompt=prompt,
+                    output=["videos", "audio", "sampling_rate"],
+                    **call_kwargs,
+                )
+            videos.extend(self._encode_minimax_h3_output(output))
```
OliverBryant: **[P2] Clean up videos produced before a later generation fails**

This writes each MP4 immediately inside the `n` loop. If `n > 1` and a later pipeline call is cancelled or raises, the files already appended to `videos` are never returned and never removed, so repeated failed requests can accumulate large, unreachable files under `XINFERENCE_VIDEO_DIR`. Please wrap the loop/response conversion with exception cleanup that removes every path created by this request (while preserving successful `url` responses).

xinference/model/video/model_spec.json
```diff
@@ -418,5 +418,53 @@
...
+      "quantization": "int4",
+      "group_offload": true,
+      "torch_dtype": "bfloat16"
+    },
+    "default_generate_config": {
+      "num_frames": 124
+    },
+    "virtualenv": {
+      "packages": [
+        "git+https://github.com/huggingface/diffusers",
```
OliverBryant: **[P2] Pin Diffusers to the validated commit**

This installs the moving Diffusers `main` branch, while the implementation relies on experimental/private details such as `ModularPipeline._blocks`, `diffusers.models.modeling_utils.dispatch_model`, and the current MiniMax-H3 block layout. A fresh launch of the same Xinference release can therefore pull a different upstream commit and fail even though neither Xinference nor the model revision changed. Please pin this VCS dependency to the exact commit used for the real-weight validation.

OliverBryant (commented):
Two issues worth addressing before merge.

xinference/model/video/diffusers.py
qinxuye: Fixed in 69e330933: MiniMax-H3 generation now removes every output created by the request if a later generation or response conversion fails, while successful `url` responses retain their files. Added regression coverage for partial-output cleanup and successful URL preservation.

xinference/model/video/model_spec.json
qinxuye: Fixed in 69e330933: the H3 virtualenv now pins Diffusers to `2da7040be1a2e5f2fcbc8b985083342a308f5a86`, the exact commit recorded in the GPU host's validated installation.

xinference/model/video/diffusers.py
```diff
@@ -375,6 +630,171 @@ def firstlastframe_to_video(
...
+            )
+        elif not isinstance(audio_outputs, list):
+            audio_outputs = [audio_outputs]
+
+        sampling_rate = output.get("sampling_rate")
+        urls = []
+        for index, video in enumerate(video_outputs):
+            path = os.path.join(XINFERENCE_VIDEO_DIR, uuid.uuid4().hex + ".mp4")
+            audio = audio_outputs[index] if index < len(audio_outputs) else None
+            encode_video(
```
OliverBryant: **[P2] Clean up files when video encoding itself fails**

The outer cleanup only knows paths returned by `_encode_minimax_h3_output`. If `encode_video()` creates a partial file and then raises, the current `path` has not yet been appended to `urls`; if a later item fails, even the earlier successfully encoded paths remain trapped in this local list because the method never returns. None of those files reach the outer `videos` cleanup. Please track the paths inside this method and remove them on any encoding exception, and add a regression test where `encode_video` writes a file before raising.

OliverBryant (commented):
One cleanup path remains uncovered.

xinference/model/video/diffusers.py
qinxuye: Fixed in a395a0392: `_encode_minimax_h3_output` now records each target path before encoding and removes all completed or partial files if any encode raises. The regression test writes one successful file and one partial file before failure and verifies both are removed. The branch was also rebased onto the latest `main`.

OliverBryant (approved):
LGTM

PR #5322, opened 2026-08-14
feat: [New Model] Add OvisOCR2 model support

Minamiyama:


xinference/core/utils.py
```diff
@@ -325,9 +327,15 @@ def get_key(package: str) -> str:
     for pkg in base_packages:
         canonical_key = get_key(pkg)
         pkg_name, separator, marker = pkg.partition(";")
+        normalized_marker = marker.strip().lower()
         key = (
```
qinxuye: This now preserves a marked model override alongside the unmarked dependency expanded from an engine placeholder. For example, Gemma 4's llama.cpp environment ends up with both `xllamacpp>=0.2.6` and `xllamacpp>=2026.6.9713`, which is exactly why the Linux/macOS/Windows CI jobs fail at `test_builtin_gemma_4_llama_cpp_requires_assistant_architecture`. OvisOCR2's SGLang environment likewise keeps both `sglang>=0.5.6` and `sglang>=0.5.9`. Please preserve variants only when both requirements have distinct engine conditions; a marked model-specific requirement must still replace an unmarked default. Please add regression coverage for both cases: the two engine-specific `qwen-vl-utils` entries survive, while Gemma 4 still has only one `xllamacpp` requirement.

qinxuye (commented):
The dependency merge change currently regresses existing engine-specific overrides.

qinxuye (approved):
LGTM

PR #5326, opened 2026-08-15
fix: repair pypiserver release dependencies

qinxuye:
## Summary

- normalize xoscar-only `has_cuda` markers when generating the offline pypiserver manifest, while keeping both CPU and GPU dependency branches mirrored
- keep `decord==0.6.0` on x86_64 and use the API-compatible `decord2==3.4.0` wheel on aarch64 for Wan2.2 Animate 2
- remove an audio CAMPPlus entry that model sync accidentally appended to `llm_family.json`, which broke every Python test during import
- encode ChatTTS voice strings before hashing so both xxhash 3.x and 4.x are supported
- add regression coverage for generated PEP 508 requirements and ChatTTS UTF-8 voice hashing

## Root cause

The v3.2.0 pypiserver release workflow failed in both architecture jobs:

- amd64 passed `deepdoc-lib[gpu]~=0.2.2 ; has_cuda` directly to pip, which rejected xoscar's runtime-only marker as invalid PEP 508 syntax
- arm64 attempted to download `decord==0.6.0`, but that release has no Linux aarch64 distribution

Failing release run: https://github.com/xorbitsai/inference/actions/runs/31828838217

The first PR test run also failed before collection because model sync PR #5324 copied `speech_campplus_sv_zh-cn_16k-common` from the audio registry into `llm_family.json`. That entry is not an LLM family and has no required `model_specs` field.

Failing PR run: https://github.com/xorbitsai/inference/actions/runs/31853031223

After that metadata issue was fixed, the GPU audio group exposed a dependency compatibility regression: xxhash 4.0.0 stopped accepting `str` input, while the Xinference ChatTTS adapter passed `voice` directly to `xxh32_intdigest`. The adapter now hashes explicit UTF-8 bytes, preserving the same digest on xxhash 3.8.1 and 4.0.0.

Failing GPU job: https://github.com/xorbitsai/inference/actions/runs/31853636839/job/94934496656

## Impact

The offline mirror generator now emits pip-compatible DeepDoc requirements for both runtime branches, Wan2.2 Animate keeps the `decord` import namespace on both supported architectures, the built-in LLM registry imports successfully again, and ChatTTS remains compatible with both pre-4.0 and current xxhash releases.

## Validation

- `python -m pytest -q --confcutdir=xinference/deploy/docker/pypiserver/tests xinference/deploy/docker/pypiserver/tests/test_scripts.py` (`13 passed`)
- `python -m pytest -q --confcutdir=xinference/model/audio/tests xinference/model/audio/tests/test_chattts.py::test_voice_seed_hashes_utf8_bytes` (`1 passed`)
- verified the same UTF-8 voice hashes to `1324622125` with xxhash 3.8.1 and 4.0.0
- imported `xinference.model` successfully after validating all 166 LLM entries contain `model_specs`
- verified the CAMPPlus model remains exactly once in `xinference/model/audio/model_spec.json`
- pre-commit on all changed files
- generated the full pypiserver manifest for both amd64 and arm64 and verified the affected pins
- downloaded `deepdoc-lib[gpu]~=0.2.2`
- downloaded the `decord2==3.4.0` CPython 3.12 Linux aarch64 wheel and verified that it provides `decord/__init__.py` and `decord/video_reader.py`
- `git diff --check`


OliverBryant (approved):
LGTM

PR #5329, opened 2026-08-15
fix: mirror sdists with broken metadata builds

qinxuye:
## Summary

- fall back to the PEP 691 Simple API when both constrained and unconstrained `pip download` fail during sdist metadata preparation
- select the newest non-yanked sdist compatible with the requested spec and Python version
- verify the advertised SHA-256 before adding the archive to the offline mirror
- record raw-sdist fallbacks in the build report

## Root cause

The v3.2.0 pypiserver build reaches the new Wan Animate `flash-attn` dependency. PyPI publishes `flash-attn` only as an sdist, and its `setup.py` imports torch while pip prepares metadata. The mirror stage downloads but intentionally does not install the runtime torch wheel, so every available `flash-attn` version fails before pip saves the sdist on both amd64 and arm64.

## Validation

- `pytest -q xinference/deploy/docker/pypiserver/tests/test_scripts.py` (14 passed)
- `pre-commit run --files xinference/deploy/docker/pypiserver/download_packages.py xinference/deploy/docker/pypiserver/tests/test_scripts.py`
- live PyPI fallback fetched and hash-verified `flash_attn-2.8.3.post1.tar.gz`

A no-push run of the exact multi-architecture pypiserver workflow will be dispatched from this branch.

### DIFF
--- a/xinference/deploy/docker/pypiserver/download_packages.py
+++ b/xinference/deploy/docker/pypiserver/download_packages.py
@@ -35,6 +35,7 @@ Writes ``report.json`` with size/version statistics next to the manifest.
 from __future__ import annotations
 
 import argparse
+import hashlib
 import json
 import os
 import subprocess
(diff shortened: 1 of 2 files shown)


OliverBryant (approved):
LGTM

PR #5332, opened 2026-08-15
fix(audio): return playable speech responses

Minamiyama:
## Summary

Fix generated speech responses being displayed as zero-duration or unplayable audio in the Web UI.

This is a generic audio response and media preview fix.

### Backend

- Return the correct audio MIME type for non-streaming `/v1/audio/speech` responses.
- Map supported response formats such as MP3, WAV, FLAC, AAC, OGG, Opus, PCM, M4A, and WebM to their corresponding media types.
- Preserve `application/octet-stream` as the fallback for unknown formats.

### Web UI

- Create and revoke Blob object URLs within the same React effect lifecycle.
- Prevent React Strict Mode effect cleanup from leaving the media element with an already revoked object URL.
- Normalize the Blob MIME type before creating its object URL.

## Scope

The backend change applies to all models served through `/v1/audio/speech`.

The frontend change applies to the shared Blob media preview component used for generated audio, image, and video results.

### DIFF
--- a/frontend/src/components/pages/running-model-detail/panels/result-panels.tsx
+++ b/frontend/src/components/pages/running-model-detail/panels/result-panels.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useEffect, useMemo } from 'react';
+import { useEffect, useState } from 'react';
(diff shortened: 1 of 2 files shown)


qinxuye (approved):
LGTM

PR #5334, opened 2026-08-15
feat(audio): support multiple engines and MLX models

qinxuye:
## Summary

- consolidate audio backends under canonical model names with explicit engine selection while retaining legacy MLX aliases
- add engine discovery, cache isolation, virtual-environment metadata, and documentation support for audio engine variants
- add `mlx-audio` STT/TTS adapters and Apple Silicon registrations for SenseVoice, Fun-ASR, Qwen3-ASR, Qwen3-TTS, MeloTTS, and VoxCPM2
- add Qwen3-TTS reference voice cloning, automatic long-text segmentation and audio concatenation
- add microphone recording to the audio-to-text workspace and reliable audio playback/download handling
- return format-specific MIME types for generated audio

## Fixes included

- decode uploaded MLX STT audio from bytes without relying on an artificial `.audio` suffix
- handle valid MPEG-2/2.5 Layer III streams whose first frame uses headers such as `FF F3`, which `mlx-audio==0.4.6` does not detect from `BytesIO`
- preserve generated Blob URLs through React Strict Mode and expose an explicit download action
- keep old names such as `whisper-small-mlx`, `F5-TTS-MLX`, and `Kokoro-82M-MLX` working as aliases

## Validation

- `pytest -q xinference/model/audio/tests/test_audio_engine.py xinference/model/audio/tests/test_mlx_audio.py` (26 passed)
- `pre-commit run --from-ref upstream/main --to-ref HEAD --all-files`
- frontend ESLint on the changed components
- `npx tsc --noEmit`
- verified Qwen3-TTS MLX reference voice cloning and long-text output locally
- verified the Qwen3-ASR MLX decoder against an actual MPEG-2 Layer III file beginning with `FF F3`

## Local environment note

The existing real-weight Whisper MLX integration tests could not launch in this checkout because a previously created test virtual environment inherited an incompatible `mlx_whisper`/Numba stack from the parent Conda environment. The engine and adapter unit suites above pass; no parent environment packages were modified.


xinference/model/audio/mlx_audio.py
```diff
@@ -0,0 +1,580 @@
...
+                "Streaming mlx-audio output is not yet supported by Xinference"
+            )
+
+        temp_files: List[str] = []
+        try:
+            text_segments = None
+            if (
+                self._model_spec.model_family == "qwen3_tts"
+                and "split_pattern" not in kwargs
+            ):
```
rogercloud, line 539: `split_pattern` is only forwarded here; `_speech` then joins the segments and makes one `_model.generate` call at `:561`. mlx-audio 0.4.6 ignores this kwarg for CustomVoice and VoiceDesign, so long input can exceed the token budget or be truncated. Generate each `text_segments` item separately (or use the upstream segmented API), then pass ordered results through `_join_qwen_results`; make the fake ignore `split_pattern`, assert one call per segment, and add VoiceDesign coverage.

xinference/model/audio/__init__.py
```diff
@@ -83,13 +91,18 @@ def _install():
...
-    # register model description after recording model revision
+    # Register one cache/version entry per engine variant. Hugging Face is the
+    # preferred representative when the same variant has multiple hubs.
     for model_name, model_specs in BUILTIN_AUDIO_MODELS.items():
-        model_spec = (
-            [x for x in model_specs if x.model_hub == "huggingface"] + model_specs
-        )[0]
-        if model_spec.model_name not in AUDIO_MODEL_DESCRIPTIONS:
-            AUDIO_MODEL_DESCRIPTIONS.update(generate_audio_description(model_spec))
+        variants = {}
```
rogercloud, line 131: The downloaded merge entered by `install_models_with_merge(...)` at `:85-92` still reaches a merge keyed only by canonical `model_name` with one global maximum `updated_at`. Because this PR places MLX and PyTorch variants under the same name, an older but valid engine can be dropped before `AUDIO_ENGINES` is rebuilt. Key the merge by variant identity (including engine/model format/cache name/hub), compare timestamps per identity, and add a downloaded-registry fixture with unequal timestamps that preserves both engines.

xinference/model/audio/model_spec.json
```diff
@@ -338,230 +358,260 @@
...
         "model_id": "mlx-community/whisper-tiny"
       }
     },
-    "updated_at": 1783912323,
+    "updated_at": 1786790280,
     "featured": false,
     "virtualenv": {
       "packages": [
         "mlx-whisper",
+        "numba>=0.64.0",
```
rogercloud: The surrounding consolidated Whisper MLX package list at `model_spec.json:379-382` still has no `#engine# == "MLX"` marker. Marker-based virtualenv discovery therefore cannot select MLX when parent libraries are absent and reports it unavailable instead of using the virtualenv-required path. Add the marker to every consolidated Whisper/F5-TTS/Kokoro MLX package list and test clean-parent virtualenv discovery.

xinference/core/supervisor.py
```diff
@@ -2255,6 +2255,13 @@ async def launch_builtin_model(
         envs: Optional[Dict[str, str]] = None,
         **kwargs,
     ) -> str:
+        if (model_type or "").lower() == "audio":
+            from ..model.audio.core import resolve_audio_model_name_and_engine
+
+            model_name, model_engine = resolve_audio_model_name_and_engine(
```
rogercloud, line 2356: This canonicalizes `model_engine`, but the auto/sharded selection that follows still does not filter candidates by worker engine capability. In a mixed Apple/Linux deployment, an explicit MLX request can land on Linux, fail with 400, and never retry a capable worker. Filter every candidate by the requested engine (or select only workers advertising it) and add mixed-worker explicit, auto, and sharded tests.

doc/source/models/model_abilities/audio.rst
```diff
@@ -62,31 +62,31 @@ Audio to text
...
-``pip install 'qwen-asr[vllm]'``, then launch the model with
-``--model-engine vLLM``, for example:
+Audio models with multiple implementations use one model name and select the
+runtime with ``--model-engine``:
+
+* The Whisper models listed above use ``transformers`` by default and also
+  support ``MLX`` on Mac computers with Apple silicon.
+* ``F5-TTS`` and ``Kokoro-82M`` use ``PyTorch`` by default and also support
+  ``MLX`` on Mac computers with Apple silicon.
+* ``Qwen3-ASR-0.6B`` and ``Qwen3-ASR-1.7B`` use ``transformers`` by default. On
```
rogercloud, line 77: The new engine overview still names only Whisper, F5-TTS, Kokoro, and Qwen3-ASR, while generated pages for the new Fun-ASR-Nano, MeloTTS, Qwen3-TTS, SenseVoiceSmall, and VoxCPM2 variants remain base-identical and omit Available Engines/engine-specific launch examples. Regenerate the affected pages, update the overview/backend prose, and add a docs-generation consistency check.

xinference/model/audio/__init__.py
```diff
@@ -83,13 +91,18 @@ def _install():
...
+        for model_spec in model_specs:
+            version = model_spec.cache_name or model_spec.model_name
+            current = variants.get(version)
+            if current is None or model_spec.model_hub == "huggingface":
+                variants[version] = model_spec
+        AUDIO_MODEL_DESCRIPTIONS[model_name] = [
```
rogercloud, line 137: This new per-canonical-name description assignment feeds catalog/cache paths that still collapse same-name registrations. Linux can expose one non-MLX variant while Apple exposes two, so one engine's choices or cache/version state can be silently omitted. Preserve engine/model-format/cache identity in aggregation keys and add mixed-platform catalog and cache-listing coverage; this is metadata visibility, separate from launch placement.

rogercloud (commented):
This PR consolidates legacy audio model names into canonical names with explicit engine, format, cache, and virtual-environment metadata, and adds engine-aware MLX STT/TTS adapters. It also updates launch resolution, worker/catalog registration, model specifications, documentation templates, frontend media handling, and audio tests so multiple implementations can share one model name. The direction is useful, but several registry, long-text, virtual-environment, heterogeneous-worker, and catalog edge cases still affect production behavior.

Blocking: yes — recommended event: COMMENT

## Approach verdict

**acceptable-with-reservations.** Canonical names plus explicit engine selection reuse the existing registry, virtualenv, cache, and catalog mechanisms and are a reasonable root direction. The reservations below are concrete correctness, compatibility, and documentation gaps in those paths; I am not raising a separate architecture-boundary objection.

## Major findings

### 1. `xinference/model/audio/mlx_audio.py:553` — major — Qwen3-TTS CustomVoice/VoiceDesign long-text splitting is bypassed

The adapter computes `text_segments`, joins them with newlines, and forwards `split_pattern` before making one `_model.generate` call (the call is at `:561`). In mlx-audio 0.4.6, the CustomVoice and VoiceDesign paths dispatch directly to `generate_custom_voice`/`generate_voice_design` and ignore `split_pattern`; only the Base path performs that split. Long CustomVoice/VoiceDesign input can therefore exceed the token budget or be truncated.

Please generate each `text_segments` item independently for every Qwen TTS mode (or call an upstream segmented API), then feed the ordered results through the existing `_join_qwen_results` pause/fade logic. Update the fake at `xinference/model/audio/tests/test_mlx_audio.py:230-238` so it ignores `split_pattern` for these modes, assert one generation call per segment and ordered full-text coverage, and add a VoiceDesign long-input case.

### 2. `xinference/model/audio/__init__.py:97` — major — downloaded timestamp merge can discard same-name engine variants

The downloaded-registry path entered by `install_models_with_merge(...)` at `xinference/model/audio/__init__.py:85-92` still reaches `merge_models_by_timestamp`, which groups only by canonical `model_name` and retains entries at one global maximum `updated_at` (`xinference/model/utils.py:2619-2654`). This PR intentionally introduces multiple engines/cache names under one canonical name with unequal timestamps, so a downloaded registry can silently drop a valid older MLX or PyTorch variant before `AUDIO_ENGINES` is rebuilt.

Make the merge identity variant-aware (at least `(model_name, engine, model_format, cache_name, hub)`) and compare timestamps within each identity, then add a downloaded-registry fixture with same-name variants at different timestamps and assert that both engine variants remain discoverable and launchable.

### 3. `xinference/model/audio/model_spec.json:380` — major — consolidated legacy MLX specs lack virtualenv engine markers

The changed line is in the consolidated Whisper MLX block; the surrounding package list at `model_spec.json:379-382` remains `mlx-whisper`, `numba>=0.64.0`, and `#system_numpy#` with no `#engine# == MLX` marker. Marker-based virtualenv discovery therefore cannot select MLX when the parent environment lacks those libraries, and reports the engine as unavailable instead of using the `virtualenv_required` path. The same omission affects the consolidated Whisper, F5-TTS, and Kokoro MLX variants.

Add the engine marker to every consolidated legacy MLX package list (for example, `mlx-whisper ; #engine# == MLX`) and add representative clean-parent/virtualenv discovery coverage for Whisper, F5-TTS, and Kokoro.

### 4. `xinference/core/supervisor.py:2261` — major — explicit MLX launches are not constrained to capable workers

This new canonicalization makes `model_engine=MLX` explicit, but the subsequent auto and sharded candidate selection still filters only model-count/GPU availability. In a heterogeneous Apple/Linux deployment, the first worker's catalog can advertise MLX while placement selects a Linux worker; that worker rejects MLX, the supervisor marks the model as failed, and the REST request returns 400 without trying another capable worker. The previously separate legacy engine names did not make this canonical explicit-engine path reachable in the same way.

Aggregate per-worker engine capabilities and filter every launch candidate by the requested `model_engine` (or retry only after selecting a verified capable worker). Add a mixed Apple/Linux scheduling test covering explicit MLX, auto placement, and sharded placement.

## Minor findings

### 5. `doc/source/models/model_abilities/audio.rst:75` — minor — generated model pages and overview prose omit new engine choices

The new overview text around this line documents only Whisper, F5-TTS, Kokoro, and Qwen3-ASR, while the generated pages for Fun-ASR-Nano, MeloTTS variants, Qwen3-ASR variants, Qwen3-TTS variants, SenseVoiceSmall, and VoxCPM2 remain byte-identical to the base tree. Those pages therefore show one generic model ID and no engine-specific launch or Available Engines information even though the generator/template and model specs now expose MLX variants.

Regenerate all affected audio pages and update the overview/backend prose consistently; add a documentation-generation consistency check or snapshot that fails when a spec exposes multiple engines but its checked-in page does not.

### 6. `xinference/model/audio/__init__.py:103` — minor — canonical-name catalog/cache aggregation hides platform-specific metadata

This new per-canonical-name `AUDIO_MODEL_DESCRIPTIONS` assignment feeds REST/catalog and cache paths that still treat same-name registrations as one identity. Linux can expose one non-MLX variant while Apple exposes two variants, and the existing cache/version aggregation skips the colliding registration when the version-list lengths differ; the result is silently missing engine choices or cache/version state for one platform.

Preserve engine/cache identity in the catalog and cache aggregation keys (for example, include `engine`, `model_format`, and `cache_name` alongside `model_name`) and add mixed-platform catalog and cache-listing tests that assert every variant is retained.

## Prior findings

- **P1 — FIXED.** The body-only pause/fade naming finding is addressed by the named timing constants in `xinference/model/audio/mlx_audio.py:253-257`.
- **P2 — FIXED.** The pause helper now uses those constants at `:379-387` without changing the values or pause-selection math.
- **P3 — FIXED.** The fade helper now uses `_AUDIO_EDGE_FADE_SECONDS` at `:389-395` without changing the fade behavior.

## Simplification opportunities

- L224: delete: `MLXWhisperAudioModel.required_libs` is never consulted because this class overrides `check_lib`. Remove the unused declaration and keep the custom `check_lib`.
- L70: shrink: `MLXAudioSTTModel` and `MLXAudioTTSModel` duplicate the same constructor, `model_ability` property, and load wrapper. Extract those methods into a private `_MLXAudioModel` base and leave only backend-specific `_load` methods.
- L259: shrink: `_OPENAI_VOICES` repeats the same allowlist embedded in `VoxCPMModel._build_generate_kwargs`. Move the allowlist to a shared audio constant and reuse it.

net: -27 lines possible

CI preflight passed with no blocking checks reported; no local tests were run per PR-review policy.

## Blocking status & recommended decision

Blocking: yes — recommended event: COMMENT

Blocking issues:
- `xinference/model/audio/mlx_audio.py:553` — **major** — CustomVoice/VoiceDesign ignore the adapter's long-text split and can truncate long input. **[new]**
- `xinference/model/audio/__init__.py:97` — **major** — timestamp-based downloaded merging can discard valid same-name engine variants. **[new]**
- `xinference/model/audio/model_spec.json:380` — **major** — consolidated legacy MLX virtualenv packages lack engine markers and cannot be discovered through the virtualenv path. **[new]**
- `xinference/core/supervisor.py:2261` — **major** — auto/sharded placement does not filter workers by the requested engine capability. **[new]**

The documentation and catalog findings are minor, and the simplification items are non-blocking. Because the author is qinxuye, the recommended GitHub event is **COMMENT**, not `REQUEST_CHANGES`, despite the blocking major findings.

doc/source/models/model_abilities/audio.rst
qinxuye: Fixed the documentation gap in fa57e0868: all affected Fun-ASR, MeloTTS, Qwen3 ASR and TTS, SenseVoice, and VoxCPM2 pages were regenerated with engine-specific IDs, launch commands, and Available engines sections, and the audio overview now names all of them. I did not add a second docs-consistency framework here; gen_docs.py and the existing auto-doc workflow remain the source of truth, and the checked-in generated output is now synchronized.

xinference/model/audio/model_spec.json
qinxuye: Fixed in fa57e0868. Every consolidated Whisper, F5-TTS, and Kokoro MLX package entry now carries the MLX engine marker. The parameterized clean-parent discovery test verifies that each missing MLX dependency is reported as virtualenv_required.

xinference/core/supervisor.py
qinxuye: I am not expanding this PR into a cluster scheduler redesign. This is not a regression caused by canonical audio names: before this PR, an explicit legacy MLX model name could also be sent to any worker because _choose_worker has never consulted worker model or engine registrations. The requested explicit, auto, and sharded filtering needs a new cross-model worker-capability contract, pinned worker semantics, shard-wide validation, and retry and cleanup behavior. This PR keeps the existing placement contract: callers pin placement with worker_ip or replica_config, and the selected worker validates the engine. Capability-aware heterogeneous scheduling should be a separate core PR.

xinference/model/audio/mlx_audio.py
qinxuye: Fixed in fa57e0868. The adapter now invokes generate once for each Qwen3-TTS segment for every Qwen mode, passes split_pattern=None to each call, and joins the ordered per-segment result lists. The regression fake ignores split_pattern and asserts one call per segment; VoiceDesign coverage is included.

xinference/model/audio/__init__.py
qinxuye: Fixed in fa57e0868 and 53533dfd3. Audio cache descriptions retain engine-specific cache names; CacheTracker now merges by model_version instead of discarding unequal platform lists. Supervisor audio catalog and engine-query results are also unioned by engine, format, cache, hub, and model ID, so Linux-first and Apple-first clusters retain both variants. Mixed-platform catalog, engine-query, and cache tests cover the behavior.

xinference/model/audio/__init__.py
qinxuye: Fixed in fa57e0868. Timestamp merging now accepts a variant identity, and audio uses canonical name, engine, model format, cache name, and hub. The downloaded-registry test gives the transformers entry a newer timestamp and verifies that both transformers and MLX variants remain registered.

qinxuye:
Review follow-up is complete at 53533dfd3 (based on fa57e0868): the Qwen3 per-segment generation, variant-aware downloaded merge, legacy MLX virtualenv markers, generated engine docs, heterogeneous catalog union, engine-query union, and cache-version union are implemented with focused regression tests. The worker-capability scheduling request is intentionally not included: it changes the cross-model placement contract for explicit, auto, pinned, retry, and sharded launches and is not a regression introduced by canonical audio names; it belongs in a separate core scheduling PR. The three non-blocking simplification suggestions are also deliberately left out because they are unrelated cleanup or cross-backend refactoring, not fixes required for this feature.

rogercloud (commented):
This PR consolidates backend-specific audio model names into canonical model identities with explicit `model_engine`, while preserving legacy MLX aliases and adding mlx-audio STT/TTS adapters and Apple registrations. It also updates catalog/cache/virtualenv/worker metadata, generated documentation, and frontend recorder/media paths so multiple engines can coexist under one model name. The current semantic scope is 62 files (+3,290/-515); the direction is sound, but placement, migration, resource, and encoding boundaries still need attention.

Blocking: yes — recommended event: COMMENT

## Update summary

The prior inline comments were anchored to `e5f79d8484766f5e9025d43974cdb93667f14748`, which is not an ancestor of the current head because the PR was rebased; consequently, a formal comparison from that SHA includes unrelated rebase churn. The substantive updates after the equivalent audio tree are `fa57e086` (per-segment Qwen generation, variant-aware timestamp merging, MLX markers, and documentation) and `53533dfd` (heterogeneous worker metadata/catalog/cache-tracker unions); this review treats those as the semantic update rather than claiming that the formal churn was newly authored.

## Approach verdict

**acceptable-with-reservations.** Canonical names plus explicit `model_engine` address duplicate model names and cache collisions while reusing the existing registry, virtualenv, cache, and catalog mechanisms. The reservations are concrete boundaries: placement does not retain per-worker engine capability, pre-PR persisted catalog records need migration, Qwen long-text generation has no server-side resource bound, the advertised response-format set is wider than the guaranteed encoder support, and alias compatibility is only partial outside launch.

## Major findings

### 1. `xinference/core/supervisor.py:1836-1848, 2351-2356` — **major** — heterogeneous worker placement ignores engine capability [prior]

The supervisor now unions engine-query and model-registration results across workers without retaining worker identity, so an Apple-only MLX engine is advertised cluster-wide. Auto and replica candidate selection still uses load/GPU availability rather than the requested engine; sharded registration canonicalizes the name before placement and can therefore admit a non-MLX Linux worker. That worker rejects MLX locally, after which the supervisor returns HTTP 400 and cleans up without retrying another capable worker. Explicit `worker_ip`/`replica_config` pins are intentional, but do not mitigate unpinned auto or sharded launches.

Retain per-worker engine capabilities or affinity and filter auto, replica, and every shard candidate by that capability. Validate impossible explicit pins early, and add mixed Apple/Linux coverage for explicit, auto, replica, and sharded placement.

### 2. `xinference/model/audio/mlx_audio.py:546` — **major** — Qwen long-text generation has unbounded resource use [new]

`_split_qwen_text` creates an unbounded number of segments. Each segment is materialized with `list(generate)`, and all result lists remain in `results_by_segment`; joining then allocates concatenated audio, fade copies, pause buffers, the final audio array, and encoding bytes. The speech input schema has no server-side maximum, the client-side 4096-character value is documentation only, `max_tokens` can be overridden for each segment, streaming is rejected, and ModelActor serialization does not bound one request. This is distinct from the prior `split_pattern`/truncation issue, which is fixed.

Enforce server-side input, segment, and aggregate-output limits before generation; bound `max_tokens`; stream or write incrementally and release completed segment data where possible; and document or reject inputs that exceed the resource budget. Add boundary tests for over-budget text and aggregate output.

### 3. `xinference/model/audio/__init__.py:74-81, 95-103` — **major** — pre-upgrade engine-less catalog records are not migrated [new]

Pre-PR canonical records did not contain `engine`, `model_format`, or `cache_name`. The new identity therefore treats an old `(name, None, None, None, hub)` record as distinct from the current `transformers`/`pytorch` variant. Such records can remain as stale None-engine families, add duplicate worker specifications, overwrite the default description representative, and prevent a newer old-schema downloaded record from overriding the current built-in default because built-ins are considered first. Alias-suffixed old records are normalized, but old canonical records are not.

Map engine-less canonical records to the current default variant before identity and timestamp merging. Add lower-, equal-, and newer-timestamp old-schema migration tests and assert that no None-engine duplicate remains.

### 4. `xinference/model/audio/mlx_audio.py:343` — **major** — MLX response formats are passed to SoundFile as container names [new]

`SpeechRequest` accepts arbitrary strings, while the REST media mapping advertises `aac`, `flac`, `m4a`, `mp3`, `mpeg`, `ogg`, `opus`, `pcm`, `wav`, `wave`, and `webm`. The implementation special-cases only `pcm` and `wav`, then calls `SoundFile(format=response_format.upper())`. SoundFile supports only some major containers; Opus is a subtype of OGG, and aliases or formats such as AAC, M4A, WebM, MPEG, and WAVE are not all valid SoundFile major names. The resulting `ValueError` is converted into HTTP 500 after inference.

Map every advertised format to a guaranteed encoder, including an explicit Ogg Opus path and alias handling, or validate and reject unsupported formats before inference with a 4xx response and an aligned schema/MIME mapping. Add parameterized tests for accepted format signatures and MIME types.

## Minor findings

### 5. `xinference/model/audio/__init__.py:169-182` and `xinference/core/supervisor.py:2351-2356` — **minor** — legacy aliases are launch-only in code but documented broadly [new]

Launch canonicalizes an alias and stores the canonical name/default UID, but registration, version, cache, instance, and virtualenv public lookup paths forward the exact user-supplied name. After launching `whisper-tiny-mlx`, old-name registration/version/cache filters are empty even though base supported lookups remain alias-keyed. The generic engine query was already unavailable for these old names in the base tree, so this is a compatibility-contract gap rather than a new engine-query regression.

Either canonicalize every public `model_name` input or state explicitly that compatibility aliases are launch-only, then add lookup tests for both canonical and legacy names.

### 6. `xinference/core/supervisor.py:1835-1855` — **minor** — empty audio engine discovery performs duplicate fan-out [new]

The new gather calls all workers; when every result is empty, the helper returns `None` and the legacy sequential loop repeats the worker RPCs before local fallback. For audio, use the gathered result and go directly to local fallback; retain the sequential loop only for non-audio paths. Add an empty-discovery regression test that verifies the worker call count.

### 7. `frontend/src/components/ui/media-preview.tsx:43` — **minor** — audio downloads always use an `.mp3` filename [new]

The new download link defaults to `generated-audio.mp3`, and the `BlobMediaPreview` caller supplies no title, while the backend preserves WAV, FLAC, and OGG MIME types. The built-in UI currently selects MP3, but generic API or non-MP3 blobs are mislabeled on download.

Derive the extension and title from the Blob MIME type or `response_format`, retaining `.mp3` only for `audio/mpeg`.

### 8. `doc/source/gen_docs.py:527-549` and `doc/source/models/builtin/audio/qwen3-asr-0.6b.rst:22-26`, `doc/source/models/builtin/audio/qwen3-asr-1.7b.rst:22-26` — **minor** — generated Qwen3-ASR pages omit dynamically registered vLLM [new]

The generator derives Available engines only from physical `model_spec` rows. vLLM shares the Transformers weights and is registered dynamically in `engine.py`, so runtime support exists while these pages list only Transformers/MLX. The broader prior generated-page omission is fixed; this vLLM-specific residual is new.

Derive available engines from runtime declarations or explicit engine metadata, include the vLLM/platform caveat, and regenerate the affected pages.

### 9. `frontend/src/components/pages/running-model-detail/components/audio-recorder-upload.tsx:267` — **minor** — the new recorder lifecycle has no behavioral frontend tests [new]

This file adds a 369-line asynchronous `MediaRecorder` state machine, but the existing frontend test/configuration covers address utilities and does not execute `getUserMedia` or `MediaRecorder`. Add focused mocked component tests for permission denial, successful stop, recorder errors, cancel/reset and late callbacks, unmount track cleanup, object-URL ownership, and form reset/replacement.

The engine/spec fallback/source-of-truth candidate at `xinference/model/audio/core.py:175` was independently dropped: current vLLM intentionally shares the Qwen Transformers artifact, and the existing tests cover selection. It is not a finding. The full-PR and update-only Simplification Lens both returned `Lean already.`; no Simplification opportunities section is included.

## Prior findings

Fixed prior items are not repeated as current findings:

- **P1/P2 — pause timing constants: FIXED.** The named constants preserve the prior values and pause-selection math.
- **P3 — fade duration constant: FIXED.** The named fade constant preserves the prior behavior.
- **P4 — Qwen CustomVoice/VoiceDesign `split_pattern`: FIXED.** Generation is now per segment, with coverage for all Qwen modes.
- **P5 — timestamp-aware same-name variant merge: FIXED** for post-PR variant-shaped records. The separate old-schema migration gap is the new catalog finding above.
- **Prior major #3 — MLX virtualenv markers: FIXED.** Whisper, F5-TTS, and Kokoro have the markers, with parameterized clean-parent discovery coverage.
- **P6 — heterogeneous worker capability placement: NOT FIXED [prior].** This is the same issue as Major finding 1 and remains blocking.
- **P8 — generated audio pages/overview engine omission: FIXED** for model-spec engines. The Qwen3-ASR vLLM omission is the separate new minor finding above.
- **P9 — catalog/cache/platform aggregation: FIXED.** The full variant-tuple union and `CacheTracker` `model_version` merge preserve the platform variants.

## Blocking status & recommended decision

Blocking: yes

Blocking issues:
- `xinference/core/supervisor.py:1836` — **major** — worker placement does not filter auto, replica, and sharded candidates by engine capability. **[prior]**
- `xinference/model/audio/mlx_audio.py:546` — **major** — unbounded Qwen long-text generation can retain multiple large intermediate and final audio buffers. **[new]**
- `xinference/model/audio/__init__.py:74` — **major** — old engine-less persisted catalog records are not normalized into the current default variant. **[new]**
- `xinference/model/audio/mlx_audio.py:343` — **major** — advertised MLX response formats are not mapped to guaranteed encoders and can become HTTP 500 errors. **[new]**

Recommended decision: **COMMENT**.

No concrete PR-introduced security vulnerability was found. CI preflight had no failed checks; the Windows 3.10 and macOS Metal checks remain pending. No local tests were run, per review policy.

qinxuye:
Follow-up on the new top-level review. The current head is f359b87fc, rebased onto upstream a77f3ed5c. There are no unresolved inline review threads.

Accepted and fixed in f359b87fc:

- Finding 3: pre-multi-engine, engine-less downloaded audio records are normalized to the current default variant before identity/timestamp merging. Lower, equal, and newer timestamps are covered; no None-engine duplicate remains.
- Finding 5: the documentation now says explicitly that the former *-mlx names are launch compatibility aliases. Registration, cache, version, and virtual-environment lookups use the canonical name.
- Finding 6: the audio gather path no longer repeats the worker RPCs when every worker returns empty. The regression test asserts one call per worker.

The other findings are not accepted as blockers for this PR:

1. Worker capability placement is the same cross-model scheduler issue already discussed, not a regression created by canonical audio names. Legacy platform-specific audio names were also selected after the supervisor chose a worker without consulting that worker's registration. The requested change would redefine auto, replica, explicit-pin, retry/cleanup, and every-shard placement semantics across model types. That needs its own core scheduling design and PR. Repeating the finding does not turn it into a scoped regression.

2. Qwen generation now splits at a bounded 80 characters per generation unit and preserves the complete ordered output. A request-wide CPU, memory, input, output, and rate budget is an API/runtime policy shared by every TTS backend. Adding an arbitrary Qwen-only hard limit or redesigning the non-stream speech response into an incremental protocol here would create inconsistent behavior and is not a prerequisite for audio engine consolidation. The review itself found no concrete security vulnerability.

4. The advertised media-type map came from already-merged PR #5332 (commit 06b0771f6e), and the SoundFile(format=response_format.upper()) convention is shared by existing TTS backends. Aligning the global SpeechRequest schema, HTTP 4xx validation, MIME declarations, and encoder availability is repository-wide API work. It is not an MLX multi-engine blocker and should be handled consistently in a separate change, not patched only in this adapter.

7. The running-model UI path always sends response_format=mp3 and BlobMediaPreview receives that MP3 blob. There is no reachable built-in UI path passing WAV/FLAC/OGG output to this caller, so generated-audio.mp3 is correct for the implemented path. Generalizing a reusable component for hypothetical callers is not a correctness fix for this PR.

8. Qwen3-ASR vLLM is dynamically registered against the Transformers artifact, and the audio overview already documents the vLLM engine, NVIDIA requirement, and launch example. Reworking generated model pages to import runtime declarations or adding a second engine-metadata source is documentation-generator architecture work, not a blocker for this feature.

9. The frontend currently has no component-test harness for MediaRecorder, getUserMedia, jsdom, or lifecycle mocks. Bootstrapping that infrastructure and the requested lifecycle matrix is a separate frontend testing project. The recorder remains covered by frontend lint/type/build validation and the behavior exercised for this feature.

Validation on the current head:

- 39 focused audio/MLX/cluster/cache tests passed.
- 4 virtual-environment marker tests passed.
- Branch-wide pre-commit passed all configured hooks.
- The prior Windows 3.10 failure was unrelated: the only error was the existing metrics exporter endpoint availability test after 1252 tests passed. The rebased head has a fresh CI run.

This PR is not an open-ended mandate to redesign adjacent scheduling, resource-policy, encoding, documentation-generation, and frontend-test subsystems. If maintainers want those changes, they should be tracked and reviewed independently with repository-wide contracts and tests.

rogercloud:
CI is failing on this PR — holding off on review until checks pass. Failing checks: build_test_job (macos-latest, 3.10, xinference) and build_test_job (windows-latest, 3.10, xinference).

rogercloud (approved):

xinference/core/supervisor.py
```diff
@@ -1754,12 +1832,29 @@ async def query_engines_by_model_name(
     ):
         # search in worker first
         workers = list(self._worker_address_to_worker.values())
-        for worker in workers:
-            res = await worker.query_engines_by_model_name(
-                model_name, model_type=model_type, enable_virtual_env=enable_virtual_env
+        if (model_type or "").lower() == "audio":
+            worker_results = await asyncio.gather(
```
rogercloud, line 1836: **N02 — major [new]:** `asyncio.gather` here propagates a single worker exception, so a stale/unreachable worker prevents healthy results from reaching `_merge_worker_engine_params` and the local fallback. Wrap each RPC with the worker timeout/exception policy, merge successful responses, and add a one-failing/one-healthy regression test.

doc/source/user_guide/backends.rst
```diff
@@ -210,6 +210,10 @@ MLX
 to run LLM on Apple silicon. It's recommended to use for Mac users when running on Apple silicon
 if the model has MLX format support.
 
+MLX can also serve supported audio models on Apple silicon. Whisper models,
```
rogercloud, line 213: **P06 — minor [prior]:** This new backend paragraph still names only Whisper, F5-TTS, and Kokoro even though the PR registers additional MLX audio families. Expand the list or label these names as examples so users can discover the full supported set.

xinference/model/audio/__init__.py
```diff
@@ -70,26 +71,72 @@ def _need_filter(spec: dict):
...
+
+
 def _install():
     # Install models with intelligent merging based on timestamps
     from ..utils import install_models_with_merge
 
+    # Startup and tests may call the installer more than once in one process.
+    # Rebuild these derived registries so model sources and engine variants do
+    # not accumulate duplicate entries.
+    BUILTIN_AUDIO_MODELS.clear()
```
rogercloud, line 113: **N05 — minor [new]:** Clearing the shared registries before parsing/registration exposes empty or partial state to concurrent lookups, and an exception can leave the partial rebuild installed. Build a replacement snapshot off to the side and atomically publish it (or use a coordinated reload/read lock with rollback), with failure/concurrency coverage.

xinference/model/audio/engine.py
```diff
@@ -43,8 +102,242 @@ def match(cls, model_family: "AudioModelFamilyV2") -> bool:
...
+            and _is_engine(model_family, "MLX")
+        )
+
+    @classmethod
+    def is_model_family_supported(cls, model_family: "AudioModelFamilyV2") -> bool:
+        return model_family.model_name in MLX_AUDIO_TTS_MODEL_NAMES
+
 
 def register_builtin_audio_engines() -> None:
     # the first registered engine is the default one for a model
```
rogercloud, line 323: **N06 — minor [new]:** This comment makes dictionary insertion order the public default, while runtime resolution and generated docs independently use `next(iter(...))` and `engine_specs[0]`. Add explicit `default_engine` metadata and have registration, launch resolution, discovery, and docs consume the same validated value.

doc/source/gen_docs.py
```diff
@@ -508,23 +508,62 @@ def main():
...
 
             rendered = env.get_template('audio.rst.jinja').render(model)
             output_file_path = os.path.join(output_dir, f"{model['model_name'].lower()}.rst")
             with open(output_file_path, 'w') as output_file:
-                output_file.write(rendered)
+                output_file.write(rendered.rstrip('\n'))
+            generated_files.add(os.path.basename(output_file_path))
+
+        for filename in os.listdir(output_dir):
+            if filename.endswith('.rst') and filename not in generated_files and filename != 'index.rst':
```
rogercloud, line 565: **N07 — minor [new]:** The generator removes every non-generated `.rst`, while the PR deletes the former `*-mlx.rst` pages; preserving launch aliases does not preserve inbound Sphinx URLs/anchors. Generate compatibility stubs/redirects for retired pages and scope cleanup to an explicit generated manifest.

rogercloud (commented):
## PR summary

PR 5334 consolidates audio model registrations under canonical names with an explicit engine dimension, while retaining retired `*-mlx` names as launch aliases. It adds engine-aware catalog/cache/virtual-environment/discovery handling, new `mlx-audio` STT/TTS adapters and Apple-Silicon specs, and a browser microphone recorder/download path. The central identity refactor fits Xinference's existing abstractions, but the review still finds two major availability/deployment defects plus partial compatibility and documentation gaps.

Blocking: yes — recommended event: COMMENT

## Approach verdict

**acceptable-with-reservations.** Canonical model names plus an explicit engine choice are the right abstraction for this migration and reuse the existing registry, cache, virtual-environment, and catalog machinery better than keeping `*-mlx` as separate model families. The reservations that remain reportable are the implicit default-engine contract (N06), non-atomic registry reload (N05), and documentation/URL compatibility (P06, N07); the author-waived placement, Qwen resource-budget, response-format, and test-infrastructure concerns are intentionally excluded.

## Findings (severity ordered)

### Major

#### N02 — One failed worker aborts audio engine discovery [new] — major

- **Location:** `xinference/core/supervisor.py:1836`
- **Evidence:** The audio path changed to an all-worker `asyncio.gather` before `_merge_worker_engine_params` at `:1846`. It does not use `return_exceptions=True`, a per-worker exception wrapper, or a worker RPC timeout. Therefore an unreachable/stale worker raises the whole gather and the fallback at `:1859-1868` is never reached, even when another worker returned a usable engine result. The base path queried sequentially and could return a healthy result without depending on every worker.
- **Impact:** A transient failure on any worker turns the audio engine query/launch-discovery request into an error and discards healthy worker results. This is a PR-introduced availability regression in clusters where worker liveness is not perfect.
- **Recommendation:** Apply the existing worker timeout/error policy per RPC, merge successful responses while logging and skipping failed workers, and use the local fallback only when no authoritative worker result remains. Add a regression test with one raising worker and one healthy worker.

#### N03 — Browser-recorded containers require undeclared system decoders [new] — major

- **Location:** `frontend/src/components/pages/running-model-detail/components/audio-recorder-upload.tsx:22`
- **Evidence:** The new recorder chooses WebM/Opus, MP4/M4A, or Ogg/Opus (`:22-27`). `MLXAudioSTTModel._decode_audio` forwards the bytes to `mlx_audio.audio_io.read` at `xinference/model/audio/mlx_audio.py:108-124`; the pinned `mlx-audio==0.4.6` path uses external `ffprobe`/`ffmpeg` for these container formats. The MLX STT virtual-environment spec installs only `mlx-audio[stt]==0.4.6` (`xinference/model/audio/model_spec.json:686`), and the Mac MLX documentation does not declare those system binaries.
- **Impact:** On a native Apple-Silicon deployment without separately installed ffmpeg tools, the headline microphone feature records successfully but every WebM/M4A/Ogg transcription fails at decode time, typically surfacing as a backend error after the user has granted microphone permission.
- **Recommendation:** Make `ffmpeg`/`ffprobe` an explicit detected and documented (or packaged) prerequisite, or provide an in-process decoder for the selected formats. Return an actionable missing-dependency error and cover at least WebM and M4A under the declared deployment prerequisites.

### Minor

#### P06 — Backend guide still under-documents the new MLX audio engines [prior] — minor

- **Location:** `doc/source/user_guide/backends.rst:213-215`
- **Evidence:** The canonical generated model pages and audio overview now cover the model-spec engine choices, but this newly added backend paragraph still says MLX audio support is exposed by only Whisper, F5-TTS, and Kokoro-82M. The runtime/spec additions also cover SenseVoice, Fun-ASR, Qwen3-ASR, Qwen3-TTS, MeloTTS, and VoxCPM2.
- **Impact:** Users consulting the backend guide can miss most of the new Apple-Silicon audio functionality and may assume the engine is unsupported for those models.
- **Recommendation:** Expand the list to the complete supported set, or explicitly label the three names as examples and link to the generated audio engine catalog.

#### P09 — Legacy aliases remain only partially compatible outside launch [prior] — minor

- **Location (body-only; no defensible changed-line anchor):** launch canonicalization is at `xinference/core/supervisor.py:2353-2358` and `xinference/core/worker.py:3337-3344`, while public lookup paths still consume the supplied name at `xinference/core/worker.py:2613-2623`, `xinference/core/supervisor.py:2043-2047`, `:3233-3239`, `:4361-4384`, and `:4630-4658`.
- **Evidence:** Launch resolves a retired name such as `whisper-tiny-mlx` to the canonical name plus `MLX`, but registration, version, instance, cache, and virtual-environment lookup methods still forward the exact alias. The user-facing audio documentation says those lookups use the canonical name, so the documented contract and implementation remain inconsistent.
- **Impact:** Clients that retain an old model name can receive empty registration/version/cache/instance/virtual-environment results after a successful alias launch, undermining the compatibility promised for the migration.
- **Recommendation:** Canonicalize every public audio `model_name` lookup before dispatch, or explicitly narrow and test the contract as launch-only compatibility. Keep canonical and legacy-name lookup tests for registration, versions, cache, instances, and virtual environments.

#### N04 — Qwen3-ASR exposes a list in the scalar verbose language field [new] — minor

- **Location:** `xinference/model/audio/mlx_audio.py:228`
- **Evidence:** The adapter copies `result.language` verbatim into the top-level verbose response. The pinned `mlx-audio==0.4.6` Qwen3-ASR `STTOutput` builds `language` as a list of per-segment language values, while the surrounding transcription/API shape expects a scalar top-level `language`. The added fake-based test uses a scalar and does not exercise the dependency's list-shaped result.
- **Impact:** Normal verbose responses can violate typed-client/schema expectations, and multilingual or longer recordings have an undefined top-level value rather than a documented policy.
- **Recommendation:** Normalize the top-level field to a documented scalar (for example the sole or dominant language) and preserve per-segment languages in `segments`. Test the actual list-shaped dependency result, including mixed-language segments.

#### N05 — Audio registry reload is observable before replacement is ready [new] — minor

- **Location:** `xinference/model/audio/__init__.py:113`
- **Evidence:** `_install` clears `BUILTIN_AUDIO_MODELS`, `AUDIO_MODEL_DESCRIPTIONS`, and `AUDIO_ENGINES` before parsing/merging the catalog and rebuilding registrations through `:117-151`. Dynamic `register_builtin_model()` reloads can overlap launches and metadata lookups, and a later parsing/registration exception has no snapshot rollback.
- **Impact:** A reload can transiently make all audio models/engines disappear from concurrent readers; a failed rebuild can leave a partially populated registry until another reload or process restart.
- **Recommendation:** Build replacement registries off to the side and atomically publish one immutable snapshot, or protect all readers and reloaders with a coordinated lock and restore the prior snapshot on failure. Add failure-path and concurrent lookup/reload tests.

#### N06 — Default engine depends on insertion order [new] — minor

- **Location:** `xinference/model/audio/engine.py:323`
- **Evidence:** `register_builtin_audio_engines` declares that the first registered engine is the default. Runtime resolution independently uses `next(iter(AUDIO_ENGINES[model]))` (`xinference/model/audio/core.py:128-143` and `:261-264`), while generated docs choose `engine_specs[0]` (`doc/source/gen_docs.py:536-549`). There is no explicit per-model default invariant tying these orderings together.
- **Impact:** Adding or reordering an engine/spec can silently change launch behavior and make generated examples advertise a different default without an intentional API change.
- **Recommendation:** Add validated `default_engine` metadata per canonical model (or an equivalent engine-definition record) and have registration, launch resolution, discovery ordering, and docs consume that one value rather than dictionary/list order.

#### N07 — Legacy generated documentation URLs and anchors are removed [new] — minor

- **Location:** `doc/source/gen_docs.py:565`
- **Evidence:** The PR deletes the former `whisper-*-mlx.rst`, `f5-tts-mlx.rst`, and `kokoro-82m-mlx.rst` pages while preserving their runtime names only as launch aliases. The generator removes every `.rst` in `models/builtin/audio` that is not generated in the current catalog (`:564-566`), except `index.rst`; it has no redirect or legacy-anchor step.
- **Impact:** Existing inbound links to the published MLX pages and Sphinx anchors break even though the PR intentionally preserves the old launch names. The broad deletion rule can also remove a future hand-authored audio page in that directory.
- **Recommendation:** Generate compatibility stubs/redirects that preserve retired page URLs and anchors, and scope stale-file cleanup to an explicit generated manifest or marker rather than every unrecognized RST file.

### Suggestion

#### N08 — Recorder has no duration or byte budget [new] — suggestion

- **Location:** `frontend/src/components/pages/running-model-detail/components/audio-recorder-upload.tsx:229`
- **Evidence:** `recorder.start()` runs without an automatic stop policy. `ondataavailable` retains every chunk at `:215-218`, the timer only updates the display at `:231-235`, and stopping later creates a Blob and then a File at `:132-143`, retaining/copying the full recording in memory.
- **Impact:** An unattended or very long recording can grow the browser's media buffer without bound and then allocate additional copies during Blob/File creation, risking tab memory exhaustion before upload.
- **Recommendation:** Define and display a product-appropriate maximum duration/size, auto-stop with feedback at the limit, align it with the server-side request bound, and avoid unnecessary copies where browser APIs permit.

## Code Quality

The engine-family abstraction, variant-aware cache identity, and canonical-name migration reuse existing Xinference mechanisms; no standalone style-only or duplication finding survives verification. The maintainability risks are the mutable global registry publication in N05 and order-based API semantics in N06. Both are best addressed with one explicit snapshot/default source of truth rather than additional per-call special cases.

## Potential Defects

N02 and N03 are concrete availability/deployment defects; N04 is a public response-shape defect. P06, P09, and N07 are compatibility or discoverability defects, while N05 is a concurrent lifecycle defect. No additional evidence-backed correctness issue survived verification; the VoxCPM candidate N01 was dropped after checking the pinned upstream contract and is intentionally omitted.

## Security Risks

No concrete PR-introduced authentication, authorization, secret-handling, injection, or data-disclosure vulnerability was found. Microphone access remains user-mediated, temporary reference files have cleanup paths, and browser tracks are stopped on teardown. N02/N03 are availability/deployment issues rather than security findings; the previously waived Qwen resource-budget concern is not re-reported here.

## Performance Issues

N08 is the remaining recorder memory-bound suggestion. N02 can amplify worker failures into cluster-wide audio discovery outages. MLX inference is intentionally serialized on its model thread; no separate performance finding is reported for that design. The author-waived request-wide Qwen long-text resource budget is intentionally excluded from this review.

## Best Practices / Test / Documentation Coverage

- The changed backend tests provide substantial coverage for engine registration, alias normalization, cache/catalog merging, virtual-environment markers, adapter argument mapping, segmentation, and temporary-file cleanup.
- Coverage still needed for the reported contracts includes one failing plus one healthy discovery worker (N02), real ffmpeg-dependent WebM/M4A decode paths (N03), list-shaped Qwen language results (N04), failed/concurrent registry reload (N05), explicit default-engine parity (N06), legacy documentation redirects (N07), recorder limit transitions (N08), and the incomplete backend prose (P06). P09 also needs canonical/legacy lookup behavior tests.
- No local tests were run under the PR-review policy. Precheck observed failures on `macos-latest` Python 3.10 and `windows-latest` Python 3.10; the Windows Python 3.13 check was still pending. This review therefore does not imply that CI is green.
- No linked issue was provided; requirements were compared against the PR body and the existing architecture. The separately waived frontend MediaRecorder test-harness request is not repeated.

## Prior-review disposition

- **Fixed and not repeated:** P01, P02, P03, P04, P05, P07, P08, and P10.
- **Partial and reported above:** P06 and P09.
- **Dropped and omitted:** N01 (VoxCPM reference mapping), after upstream contract verification.
- **Waived and omitted:** worker-capability-aware scheduling, unbounded Qwen long-text budgeting, SoundFile/response-format mapping, generic audio-download extension, the Qwen3-ASR vLLM generated-page residual, frontend MediaRecorder test infrastructure, and the unrelated simplification suggestions. These are not additional findings in this review.

## Blocking status & recommended decision

**Blocking:** yes — N02 and N03 are confirmed major findings. Because the author is qinxuye, the recommended event is **COMMENT**, not `REQUEST_CHANGES`.

**Blocking issues:**

- `xinference/core/supervisor.py:1836`, **major** — one failed worker aborts all-worker audio engine discovery. **[new]**
- `frontend/src/components/pages/running-model-detail/components/audio-recorder-upload.tsx:22`, **major** — browser-recorded containers depend on undeclared `ffmpeg`/`ffprobe` decoders. **[new]**

frontend/src/components/pages/running-model-detail/components/audio-recorder-upload.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+interface AudioRecorderUploadProps {
+  form: FormInstance;
+  value?: FileUploadValue[];
+  onChange?: (value: FileUploadValue[]) => void;
+  error?: boolean;
+  disabled?: boolean;
+}
+
+const RECORDER_MIME_TYPES = [
```
rogercloud, line 22: **N03 — major [new]:** These browser MIME choices include WebM/MP4/Ogg, but pinned `mlx-audio==0.4.6` routes these containers through external `ffmpeg`/`ffprobe`, which the MLX virtualenv does not install or document. Make the binaries an explicit detected/documented prerequisite (or provide an in-process decoder), return an actionable error, and cover WebM and M4A.

xinference/model/audio/mlx_audio.py
```diff
@@ -0,0 +1,580 @@
...
+        audio_data = self._decode_audio(audio)
+        result = self._model.generate(audio_data, **generate_kwargs)
+
+        text = _read_result_field(result, "text", "")
+        if response_format == "json":
+            return {"text": text}
+
+        response = {
+            "task": "transcribe",
+            "language": _read_result_field(result, "language"),
```
rogercloud, line 228: **N04 — minor [new]:** `mlx-audio==0.4.6` returns Qwen3-ASR `STTOutput.language` as a list, so copying it verbatim makes the top-level verbose response violate the scalar API field. Normalize a documented scalar, retain per-segment languages, and test the real list-shaped output.

frontend/src/components/pages/running-model-detail/components/audio-recorder-upload.tsx
```diff
@@ -0,0 +1,369 @@
...
+      recorder.onerror = () => {
+        discardRef.current = true;
+        showRecordingError('Recording failed. Please try again.');
+        if (recorder.state !== 'inactive') {
+          recorder.stop();
+        }
+      };
+      recorder.onstop = () => finishRecording(recorder);
+
+      recorder.start();
```
rogercloud, line 229: **N08 — suggestion [new]:** `recorder.start()` has no duration/byte budget, so `ondataavailable` retains every chunk before Blob/File creation. Define and display a limit, auto-stop with feedback, and align it with the server bound; avoid extra copies where possible.

PR #5335, opened 2026-08-15
feat(webui): add IndexTTS emotion controls

Minamiyama:
## Summary

This PR adds manual emotion control support to the Web UI for IndexTTS2 and IndexTTS-2.5.

Users can enable **Emotion control** in the speech generation panel and adjust the supported emotion weights: happy, angry, sad, afraid, disgusted, melancholic, surprised, and calm.

## Changes

- Show emotion controls only for IndexTTS models that advertise the `Text2audioEmotionControl` capability.
- Provide sliders and numeric inputs for each emotion weight.
- Display and validate the total weight, which must not exceed `0.8`.
- Send `emo_vector` to the speech API only when emotion control is enabled and the vector is valid.
- Keep the existing behavior unchanged for other audio models.

7bba8665d6c6577f456c71a1456ec9af

### DIFF
--- a/frontend/src/components/pages/running-model-detail/capability-config.tsx
+++ b/frontend/src/components/pages/running-model-detail/capability-config.tsx
@@ -32,6 +32,11 @@ import {
   TextToVideoPanel,
 } from './panels/form-panels';
 import { ResultPanels } from './panels/result-panels';
+import {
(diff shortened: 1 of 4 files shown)


qinxuye (approved):
LGTM

PR #5337, opened 2026-08-15
feat(model): support qwen3.8-27B

bluefish-08:
Adds the `qwen3.8` built-in family (27B dense).

### Why this is a data-only change

`Qwen/Qwen3.8-27B` reports `architectures: ["Qwen3_5ForConditionalGeneration"]` and `model_type: qwen3_5`. I diffed its `config.json` against `Qwen/Qwen3.6-27B` key by key (flattened, including `text_config` and `vision_config`) — the key sets are **identical**. So every engine path that already handles the Qwen3.5 architecture handles this one, and no engine code needed touching:

- `vllm/core.py`, `vllm/patches/hybrid_kv_cache_page_size.py`, `sglang/core.py` already list `Qwen3_5ForConditionalGeneration`.
- `transformers/multimodal/qwen2_vl.py` already registers the architecture; only the **name-keyed** `register_batching_multimodal_models` list needed the new entry, which is the one code line here.

### Specs

Every repo below was verified to exist on both hubs (`huggingface.co/api/models/...` and `modelscope.cn/api/v1/models/...` both return 200):

| Format | Model ID |
| --- | --- |
| pytorch | `Qwen/Qwen3.8-27B` |
| fp8 | `Qwen/Qwen3.8-27B-FP8` |
| ggufv2 | `unsloth/Qwen3.8-27B-GGUF` — 21 single-file quants, `BF16` split into 2 parts, `mmproj-BF16/F16` |
| mlx | `mlx-community/Qwen3.8-27B-{4bit,8bit,bf16}` |

`context_length` 262144 (`max_position_embeddings`), `stop_token_ids [248044, 248046]` (from `generation_config.json` `eos_token_id`, mapping to `<|endoftext|>` / `<|im_end|>` in `added_tokens_decoder`), `tool_parser: qwen`, `<think>`/`</think>` reasoning tags, and the `virtualenv` package list mirrors `qwen3.6`.

**No AWQ spec.** `QuantTrio`/`tclf90` have not published a Qwen3.8 AWQ yet. The only Int4 repo present on both hubs (`cyankiwi/Qwen3.8-27B-AWQ-INT4`) is actually compressed-tensors `pack-quantized`, not classic AWQ, so filing it under `model_format: awq` would mislead the vLLM quantization path. Worth adding as a follow-up once a real AWQ build lands.

`mlx-community` has no `5bit`/`6bit` for this model (both 401), so unlike `qwen3.6` those two are omitted rather than copied over.

### Notes

- The chat template contains both the system-first guard and `tool_call.arguments|items`, so `is_strict_system_first_template()` returns True for it and the Coder-style string-arguments path applies. I extended the existing `test_qwen3_family_get_full_context_handles_string_arguments` targets to cover `qwen3.8` — it passes.
- `llm_family.json` was edited programmatically and round-trip verified (`json.dumps(..., indent=2, ensure_ascii=False)` reproduces the original file byte for byte), so the diff is purely the 189 added lines with no reformatting noise.
- Docs regenerated with `doc/source/gen_docs.py`; unrelated drift the generator produced in the rerank pages was reverted to keep the diff scoped.

### Verification

- `pytest xinference/model/llm/tests/test_utils.py -k qwen3_family` — passes.
- `pytest xinference/model/llm/tests/test_llm_family.py xinference/deploy/docker/pypiserver/tests/test_scripts.py` — passes except `test_query_engine_general`, which fails identically on an unmodified checkout in my environment (no `llama.cpp` installed) and is unrelated to this change.
- `pre-commit run --files <changed files>` — all hooks pass.
- Family loads correctly at runtime: 4 specs resolve across both hubs, abilities `chat, vision, tools, reasoning, hybrid`, `strict_system_first` True.
- Deployed to a live server and ran the model end to end: `Qwen/Qwen3.8-27B-FP8` on vLLM, tensor parallel across 2x L20, weights pulled from ModelScope. Launch finished cleanly, and chat completions return correct output.
- Reasoning parsing behaves as expected for this family: the template pre-fills `<think>\n` into the generation prompt, so the model emits only the closing tag; with `reasoning_content=True` the parser splits `reasoning_content` from `content` correctly.

Not run: the gguf, mlx and pytorch specs — only the FP8 spec was exercised on real hardware.

### DIFF
--- a/doc/source/models/builtin/llm/index.rst
+++ b/doc/source/models/builtin/llm/index.rst
@@ -721,6 +721,11 @@ The following is a list of built-in LLM in Xinference:
      - 262144
      - Following the February release of the Qwen3.5 series, we're pleased to share the first open-weight variant of Qwen3.6. Built on direct feedback from the community, Qwen3.6 prioritizes stability and real-world utility, offering developers a more intuitive, responsive, and genuinely productive coding experience.
 
+   * - :ref:`qwen3.8 <models_llm_qwen3.8>`
+     - chat, vision, tools, reasoning, hybrid
+     - 262144
+     - Built on the architectural foundation of Qwen3.5, Qwen3.8 delivers substantial gains across coding, professional work, research, and long-horizon agentic tasks. Qwen3.8-27B brings these advances to a compact, deployment-friendly dense model: a native vision-language model that understands images and videos, with flexible thinking control.
+
    * - :ref:`qwenlong-l1 <models_llm_qwenlong-l1>`
      - chat
      - 32768
@@ -1132,6 +1137,8 @@ The following is a list of built-in LLM in Xinference:
    qwen3.5
   
    qwen3.6
+  
+   qwen3.8
   
    qwenlong-l1
   
--- a/doc/source/models/builtin/llm/qwen3.8.rst
+++ b/doc/source/models/builtin/llm/qwen3.8.rst
@@ -1,0 +1,79 @@
+.. _models_llm_qwen3.8:
+
+========================================
+qwen3.8
+========================================
+
+- **Context Length:** 262144
+- **Model Name:** qwen3.8
+- **Languages:** en, zh
+- **Abilities:** chat, vision, tools, reasoning, hybrid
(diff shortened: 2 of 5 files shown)


qinxuye (approved):
LGTM

PR #5338, opened 2026-08-16
feat: support MiniMax-H3 Lightning LoRA

qinxuye:
## Summary

- add video-model Lightning metadata, caching, and launch options for the official MiniMax-H3 Turbo LoRAs
- support `4step_v0.1`, `8step_v1.0_bf16`, and `4step_v1.0_768p_bf16` with their required LoRA alpha, scheduler shifts, and default evaluation counts
- download the Lightning checkpoint from the same selected hub as the base model, including both Hugging Face and ModelScope
- document the shared video Lightning options and expose the versions through the existing launch UI

## Implementation notes

MiniMax-H3 Turbo checkpoints use PEFT-format LoRA tensors. The adapter is injected before group offload, while the existing INT4 and group-offload defaults remain unchanged. Since TorchAO INT4 weights cannot be dequantized for LoRA fusion, the adapter forward accumulates into the base output in place to avoid the extra full-size tensor that otherwise exceeds 24 GiB VRAM. Mixed-dtype LoRA operands are cast to the base output dtype before this in-place accumulation.

Per-request `num_inference_steps` represents actual transformer evaluations; Xinference adds the scheduler terminal sigma grid point internally. This correction applies to both Lightning and non-Lightning MiniMax-H3: a request for N evaluations now passes N + 1 scheduler grid points, so non-Lightning output may differ from earlier Xinference versions for the same `num_inference_steps` value. If a request omits the value, the selected Lightning version supplies its recommended 4 or 8 evaluations. The Web UI currently pre-fills 25, so users should change **Inference Steps** to match the selected Lightning version.

## Performance

Single-run wall-clock comparison on an NVIDIA GeForce RTX 3090 (24 GiB), using the same MiniMax-H3 defaults (INT4 plus group offload), prompt, and output settings: 1344x768, 124 frames, 24 FPS, one video.

| Configuration | Transformer evaluations | Denoising | End to end |
| --- | ---: | ---: | ---: |
| Baseline, no Lightning | 25 | 2h 12m 49s | 2h 14m 24s (8064s) |
| `4step_v1.0_768p_bf16` | 4 | 15m 23s | 16m 48s (1008s) |
| Measured speedup | — | 8.6x | 8.0x |

Both requests completed successfully with HTTP 200. These are single-run results on one machine; Lightning reduces denoising work but does not reduce the model's memory requirements.

## Validation

- `python -m pytest xinference/model/video/tests/test_minimax_h3_video.py xinference/model/video/tests/test_diffusers_video.py -q` (15 passed, 2 skipped)
- `pre-commit run --files ...`
- Sphinx dummy build succeeded; only existing unrelated reference warnings were reported
- GPU runtime validation passed for both the 25-step baseline and `4step_v1.0_768p_bf16` on the RTX 3090

### DIFF
--- a/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
+++ b/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
@@ -1209,6 +1209,21 @@ export default function LaunchDialog({
         show: !!ggufQuantizations,
       },
       {
+        name: 'lightning_version',
+        label: t('launchModel.lightningVersions'),
+        placeholder: t('launchModel.lightningVersionsPlaceholder'),
+        type: 'select',
+        fieldProps: { options: lightningVersionOptions },
+        show: !!lightningVersionOptions.length,
+      },
+      {
+        name: 'lightning_model_path',
+        label: t('launchModel.lightningModelPath'),
+        placeholder: t('launchModel.lightningModelPathPlaceholder'),
+        type: 'input',
+        show: !!lightningVersionOptions.length,
+      },
+      {
         name: 'worker_ip',
         type: 'multi-select',
         label: t('launchModel.workerIp'),
--- a/xinference/model/video/diffusers.py
+++ b/xinference/model/video/diffusers.py
@@ -150,7 +150,12 @@ class DiffusersVideoModel:
         pipeline.audio_vae.to(onload_device)
 
     def _resolve_minimax_h3_lightning_config(self) -> Optional[dict]:
-        if not self._lightning_model_path:
+        if self._lightning_model_path is None:
+            if self._lightning_version is not None:
+                raise ValueError(
+                    f"Lightning version {self._lightning_version!r} was specified, "
+                    "but no lightning model path was provided."
+                )
             return None
 
         version_configs = self._model_spec.lightning_version_configs or {}
--- a/xinference/model/video/diffusers.py
+++ b/xinference/model/video/diffusers.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import base64
+import gc
 import importlib
 import json
 import logging
@@ -66,6 +67,8 @@ class DiffusersVideoModel:
         model_path: str,
         model_spec: "VideoModelFamilyV2",
         gguf_model_path: Optional[str] = None,
+        lightning_version: Optional[str] = None,
+        lightning_model_path: Optional[str] = None,
         **kwargs,
     ):
         self.model_family = model_spec
@@ -76,6 +79,9 @@ class DiffusersVideoModel:
         self._model = None
         self._kwargs = kwargs
         self._gguf_model_path = gguf_model_path
+        self._lightning_version = lightning_version
+        self._lightning_model_path = lightning_model_path
+        self._minimax_h3_lightning_steps: Optional[int] = None
 
     @property
     def model_spec(self):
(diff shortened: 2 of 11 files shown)


xinference/model/video/diffusers.py
```diff
@@ -41,6 +43,48 @@
...
+            "memory-efficient lightning adapter"
+        )
+
+    for active_adapter in active_adapters:
+        lora_a = self.lora_A[active_adapter]
+        lora_b = self.lora_B[active_adapter]
+        dropout = self.lora_dropout[active_adapter]
+        lora_input = self._cast_input_dtype(x, lora_a.weight.dtype)
+        low_rank_states = lora_a(dropout(lora_input))
+        torch.addmm(
```
OliverBryant, line 81: **Mixed-dtype LoRA weights will raise here.**

`lora_input` is cast to `lora_a.weight.dtype`, so `low_rank_states` and `lora_b.weight` carry the *LoRA* dtype, while `result` carries the *base output* dtype. `torch.addmm` requires them to match:

```
RuntimeError: self and mat2 must have the same dtype, but got BFloat16 and Float
```

(reproduced locally with a bf16 base output and fp32 LoRA weights).

Upstream PEFT `Linear.forward` is immune because it computes `result + lora_B(lora_A(dropout(x))) * scaling` and only casts back via `.to(torch_result_dtype)` afterwards — type promotion absorbs the mismatch. The in-place accumulation has no such slack.

This works today only because the shipped checkpoints happen to match the base dtype (the `_bf16` suffixes hint at this), but `4step_v0.1` is not marked bf16 and nothing validates the dtype at load time. Suggest either asserting the LoRA dtype matches the transformer compute dtype in `_load_minimax_h3_lightning_adapter`, or casting `low_rank_states` (and `lora_b.weight`) to `result.dtype` before the `addmm`.

`test_minimax_h3_memory_efficient_lora_forward_accumulates_in_place` uses uniform fp32 throughout, so it cannot catch this — worth a mixed-dtype case.

Two smaller notes on this function:
- It skips `self._check_forward_args(x, *args, **kwargs)`, which the PEFT `forward` it replaces calls first. The `if kwargs:` guard at the top covers the main thing that protects against (`adapter_names` for mixed-batch inference), so this is largely theoretical — but it is a silent divergence from the replaced method's contract.
- `result_dtype` is captured and returned via `result.to(result_dtype)`, but since `addmm` writes in place into `result`, its dtype never changes — that call is a no-op. Harmless, just slightly misleading about what it guards.

xinference/model/video/diffusers.py
```diff
@@ -701,10 +975,17 @@ def _minimax_h3_generate(
...
+                "num_inference_steps", self._minimax_h3_lightning_steps or 50
+            )
         else:
-            generate_kwargs.setdefault("num_inference_steps", 50)
+            generate_kwargs.pop("num_inference_steps", None)
+        if num_inference_steps < 1:
+            raise ValueError("num_inference_steps must be at least 1")
+        # MiniMaxH3Scheduler includes terminal sigma zero in the requested grid,
+        # so N transformer evaluations require N + 1 grid points.
+        generate_kwargs["num_inference_steps"] = num_inference_steps + 1
```
OliverBryant, line 996: **This `+1` applies unconditionally, so it changes behavior for existing non-Lightning users.**

The `num_inference_steps + 1` adjustment is outside any Lightning branch, so it also affects plain MiniMax-H3. A user who has been requesting 25 steps will now get 26 grid points and therefore different output for an identical request.

The comment explains this as a scheduler-semantics correction (N transformer evaluations need N+1 grid points), and that reasoning is sound — I think it's a legitimate fix. The concern is packaging: a behavior change affecting every existing H3 user is folded into a feature PR without being called out anywhere in the description or docs.

Could you note it explicitly in the PR body (and ideally the video docs), so it's visible that non-Lightning H3 output shifts with this change? Alternatively, scope the `+1` to the Lightning path if the intent was only to fix the distilled-checkpoint step accounting.

OliverBryant (commented):
Nice work overall — the checkpoint validation here is genuinely thorough (rank consistency, paired A/B tensors, shape checks against the instantiated adapter, and a post-`load_state_dict` confirmation that the adapter tensors were actually consumed). `cache_lightning` also mirrors `cache_gguf` closely, which keeps the hub handling consistent.

Three findings below, one of which I'd like to see addressed before merge (the `addmm` dtype issue).

Separately, and not a problem with this PR: this PR triggered a **full four-group GPU CI run** (`llm embedding image audio`, ~33 min on the T4) even though it only touches video code, docs, and the web UI. `xinference/model/video/cache_manager.py` falls through to the `xinference/model/*) full=true` catch-all in `.github/workflows/python.yaml`, whose comment says video has "no dedicated GPU test group; play it safe and run every group." Since there is no video group in the pytest `case` at all, that fallback buys zero coverage of the changed code while paying for every other group. I'll open a separate PR adding a `xinference/model/video/*) ;;` arm (`model/flexible/` has the same issue).

xinference/model/video/diffusers.py
qinxuye: Addressed in 5c2b33a7a. The fast path now calls `_check_forward_args`, casts both the low-rank states and LoRA B weight to the base output dtype before `addmm`, and removes the no-op result cast. The test now uses a BF16 base output with FP32 LoRA weights and asserts that all three `addmm` operands are BF16. The focused suite passes: 15 passed, 2 skipped.

xinference/model/video/diffusers.py
qinxuye: Addressed in 5c2b33a7a while retaining the unified evaluation-count semantics. The video docs and PR body now explicitly state that the N + 1 scheduler grid applies with or without Lightning and that non-Lightning output may differ from earlier Xinference versions for the same `num_inference_steps` value.

frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -1208,6 +1208,21 @@ export default function LaunchDialog({
         type: 'input',
         show: !!ggufQuantizations,
       },
+      {
+        name: 'lightning_version',
+        label: t('launchModel.lightningVersions'),
+        placeholder: t('launchModel.lightningVersionsPlaceholder'),
+        type: 'select',
```
OliverBryant, line 1215: **Selecting the `none` placeholder in this dropdown will fail the launch.**

`lightningVersionOptions` prepends a literal `'none'` option:

```ts
return ['none', ...(model?.lightning_versions || [])].map(...)
```

but `lightning_version` is never stripped before submit. `launch-model/utils.tsx` has an explicit strip list for exactly this pattern:

```ts
if (nextValues.download_hub === 'none') {
  delete nextValues.download_hub;
}
if (nextValues.gguf_quantization === 'none') {
  delete nextValues.gguf_quantization;
}
```

`lightning_version` was not added there, and `utils.tsx` isn't touched by this PR. So `lightning_version: "none"` reaches the backend, where `"none"` is truthy — it clears the `not model_spec.lightning_versions` guard in `create_video_model_instance`, reaches `cache_lightning("none")`, and raises:

```
ValueError: Cannot support lightning version none,
available lightning versions: ['4step_v0.1', '8step_v1.0_bf16', '4step_v1.0_768p_bf16']
```

Since `none` is the first option, this is the natural way a user declines Lightning in the UI, and it hard-fails the launch instead of starting the base model. Fix is one arm in `utils.tsx`:

```ts
if (nextValues.lightning_version === 'none') {
  delete nextValues.lightning_version;
}
```

Worth a look at `lightning_model_path` too — an empty string submitted from that free-text input would hit the same truthiness guard in `create_video_model_instance`.

xinference/model/video/diffusers.py
```diff
@@ -143,6 +197,226 @@ def _enable_minimax_h3_group_offload(pipeline, kwargs: dict):
...
+            if a_tensor.ndim != 2 or b_tensor.ndim != 2:
+                raise ValueError(
+                    f"MiniMax-H3 LoRA tensors for {module_name} must be matrices"
+                )
+            if a_tensor.shape[0] != b_tensor.shape[1]:
+                raise ValueError(
+                    f"MiniMax-H3 LoRA rank mismatch for {module_name}: "
+                    f"A{tuple(a_tensor.shape)} and B{tuple(b_tensor.shape)}"
+                )
+            if not module_name.endswith(target_modules):
```
OliverBryant: **This validation is looser than PEFT's own matching, so it can't reject what it's meant to reject.**

`module_name.endswith(target_modules)` does bare suffix matching, while PEFT's `check_target_module_exists` requires a **dot separator**:

```python
target_module_found = any(key.endswith(f".{target_key}") for target_key in config.target_modules)
```

The two disagree on any name whose final segment merely *ends with* a target:

| module | PEFT wraps it | this check accepts it |
| --- | --- | --- |
| `blocks.0.attn.to_q` | yes | yes |
| `blocks.0.attn.add_to_q` | **no** | **yes** |
| `blocks.0.audio_ff.net.2` | **no** | **yes** |

Those aren't hypothetical shapes for a joint video+audio model like H3 — fused/added projections (`add_to_q`, `add_k_proj`) and a parallel audio branch are exactly what this architecture has.

To be clear on severity: this does **not** produce silently wrong math. A checkpoint targeting `audio_ff.net.2` passes this check, `add_adapter` declines to wrap that module, so the key lands in `unexpected` and the later comparison raises. So it fails loudly — the problem is that it fails with the *wrong* diagnostic ("checkpoint is incompatible with the transformer" plus a key dump) instead of the precise "unsupported target module" error this check exists to produce. The check is effectively dead for the cases it was written to catch.

Suggest matching PEFT's semantics:

```python
if not any(
    module_name == target or module_name.endswith(f".{target}")
    for target in target_modules
):
```

Minor, in `test_minimax_h3_loads_lightning_peft_checkpoint`: the fixture uses `A=(128, 4)`, `B=(4, 128)`, so it asserts `r: 128` on a 4-dimensional linear. The rank extraction (`A.shape[0] == B.shape[1]`) is correct and the test does pass, but a rank-128 adapter on a 4-wide layer is backwards from the real tensor shapes (rank is the small dim). Swapping to something like `A=(4, 128)`, `B=(128, 4)` would make the fixture read the way real checkpoints do and guard against a future transposition mistake.

OliverBryant (commented):
Re-reviewed at `5c2b33a7a`. I verified both earlier findings are genuinely fixed, not just claimed:

- **Mixed dtype** — `low_rank_states` and `lora_b.weight` are now both cast to `result.dtype` (with a guard so matching dtypes skip the copy), `_check_forward_args` is called first, and the no-op `result.to(result_dtype)` is gone. I re-ran the bf16-base/fp32-LoRA case against the new code path: no exception, max deviation from an fp32 reference 0.0079, which is normal bf16 rounding. The updated test genuinely covers it — bf16 `base_layer` with fp32 LoRA weights, asserting `calls["dtypes"] == (torch.bfloat16,) * 3`. That's exactly the case the old fp32-only fixture couldn't catch.
- **`+1` step semantics** — documented in both the PR body and a `.. note::` in `video.rst`, explicitly stating it applies with or without Lightning and that non-Lightning output may differ from earlier versions. Keeping the unified semantics was the right call of the two options I offered.

This pass covered the parts I hadn't yet read closely (config resolution, adapter loading, cache manager, frontend, i18n). Two new findings, one of which is a launch-blocking UI bug.

Nice touches I'll note since they're easy to miss: all four locales (`en`/`zh`/`ja`/`ko`) have the new keys, `cache_lightning` correctly threads `lightning_model_revision` separately from `model_revision` (unlike `cache_gguf`, which reuses the base revision), and the `_skip_peft_gptqmodel_probes` context manager restores the original probes in a `finally`.

frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
qinxuye: Addressed in e804b0eef. `transformFormToFetch` now removes `lightning_version` when the UI value is `none` and also removes a blank or whitespace-only `lightning_model_path`, so declining Lightning launches the base model cleanly. Prettier, targeted ESLint, and the production frontend build all pass.

xinference/model/video/diffusers.py
qinxuye: Addressed in e804b0eef. Target validation now matches PEFT semantics using either an exact module name or a dot-delimited suffix. I added a regression test proving `add_to_q` is rejected by the precise unsupported-target diagnostic, and changed the valid fixture to realistic rank-4 shapes (`A=(4, 128)`, `B=(128, 4)`). The focused video suite passes: 16 passed, 2 skipped.

OliverBryant (commented):
Re-reviewed at `e804b0eef`. Both findings from the previous round are genuinely fixed — verified against the code, not just the replies:

**`none` placeholder / blank path** — `transformFormToFetch` now drops `lightning_version === 'none'` and also drops a whitespace-only `lightning_model_path` via `.trim() === ''`. That covers both the reported bug and the `lightning_model_path` follow-up I raised in the same comment, so declining Lightning in the UI now launches the base model cleanly.

**Target-module matching** — now `module_name == target or module_name.endswith(f".{target}")`. I re-ran my comparison table against PEFT's `check_target_module_exists` semantics; the two now agree on every case I tested, including the previously divergent `add_to_q`, `audio_ff.net.2`, and `add_k_proj`, plus the bare `to_q` exact-match case. The new `test_minimax_h3_rejects_partial_lora_target_match` asserts `add_to_q` hits the precise "Unsupported ... target module" diagnostic rather than the misleading incompatible-checkpoint path — that's a real regression test for the exact defect, not just a green assertion. The valid fixture is also corrected to realistic rank-4 shapes (`A=(4, 128)`, `B=(128, 4)`, `r: 4`).

This pass also covered the pieces I hadn't read before — `gen_docs.py`, the Jinja template, and the CLI path. Nothing further to flag:

- The `gen_docs.py` refactor from `if 'huggingface' in model_src` to `model_src.get('huggingface') or model_src.get('modelscope')` is strictly more robust: an empty `huggingface` dict previously raised `KeyError` on `['model_id']` and now falls back to ModelScope. Same result on all normal inputs.
- The committed `minimax-h3.rst` matches what the updated template renders, so the generated docs are in sync.
- `--lightning_version` is documented in several places; confirmed the CLI's `ignore_unknown_options=True` kwargs passthrough forwards it to `create_video_model_instance`, so the documented commands work as written.

No further findings from me — the implementation, tests, docs, and UI wiring look consistent at this commit. LGTM.

One CI note, unrelated to the code: because this PR touches `xinference/model/video/`, it keeps triggering a full four-group GPU run (`llm embedding image audio`) through the `xinference/model/*) full=true` catch-all in `python.yaml`, even though no GPU group contains a video test. I'll send that gating fix as a separate PR so it doesn't hold this one up.

OliverBryant (approved):
LGTM

PR #5339, opened 2026-08-16
feat(model): complete Qwen3.8 support

qinxuye:
## Summary

- extend Qwen3.8-27B with BF16 GGUF and MLX 4-bit, 8-bit, and BF16 variants, and add the Qwen3.8-2.4T-A95B model as qwen3.8-max
- wire the required MLX, llama.cpp, vLLM, and SGLang architecture/version support and document the new model and engine requirements
- fix Qwen tool-call streaming for multiple calls, propagate model sampling defaults and top_k through MLX, and preserve final and usage chunks
- make automatic model-hub selection avoid Hugging Face endpoints that would use environment proxy traffic, while honoring HF_ENDPOINT, NO_PROXY, and explicit per-launch or service-level source selections

## Tests

- pytest -q (13 focused Qwen3.8, MLX, tool streaming, and model-source regression tests)
- pre-commit run --files (all 18 changed files)


xinference/model/llm/mlx/core.py
```diff
@@ -564,19 +574,29 @@ def _sanitize_generate_config(
...
             generate_config["max_tokens"] = max_tokens  # type: ignore
         return generate_config
 
+    def _update_model_generation_config(self, config: Dict[str, Any]) -> None:
+        nested_config = config.get("generation_config")
+        model_defaults = dict(nested_config) if isinstance(nested_config, dict) else {}
+        for key in ("temperature", "top_p", "top_k"):
+            if key in config:
+                model_defaults[key] = config[key]
+        self._model_generation_config = model_defaults
```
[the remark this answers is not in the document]
qinxuye: Fixed in 57cf93c36. Model generation defaults now discard None values while preserving non-None nested defaults, so nullable checkpoint metadata falls back safely. The regression test covers nullable temperature, top_p, and top_k values.

xinference/model/llm/mlx/core.py
```diff
@@ -161,19 +159,22 @@ def _is_new_mlx_lm() -> bool:
...
+                temperature,
+                top_p,
+                top_k,
             )
             from mlx_lm.generate import BatchGenerator
             from mlx_lm.sample_utils import make_sampler
 
             # Create sampler with specific settings
-            sampler = make_sampler(temp=temperature, top_p=top_p)
+            sampler = make_sampler(temp=temperature, top_p=top_p, top_k=top_k)
```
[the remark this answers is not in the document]
qinxuye: Fixed in 57cf93c36. top_k is normalized to an integer in request sanitization and again at the batch-generator boundary, covering both non-batch sampler calls and stable batch cache keys. Regression tests cover 20.0 and None.

xinference/model/llm/tool_parsers/qwen_tool_parser.py
```diff
@@ -328,30 +359,74 @@ def extract_tool_calls_streaming(
...
+                previous_complete = self.tool_call_complete_regex.findall(
+                    previous_text[-1]
                 )
-                if function_call is None:
-                    return None
-                # Skip if the extracted content is whitespace-only (e.g., the model
-                # generated <tool_call></tool_call> with no JSON inside)
-                if not function_call.strip():
+                current_complete = self.tool_call_complete_regex.findall(current_text)
+                tool_events = []
```
OliverBryant: **[P2] Preserve content before a tool tag in the same delta**

Once `current_text` contains `<tool_call>`, this branch only returns completed or partial tool events. If one detokenizer chunk first introduces the tag after ordinary text—for example, `I should check.\n<tool_call>...</tool_call>`—that prefix was not emitted by an earlier chunk and is silently dropped by `_post_process_completion_chunk`. This is a valid output according to the Qwen template (it explicitly permits natural-language reasoning before the call), and makes streaming differ from the non-streaming parser, which preserves the prefix. Please emit the newly introduced plain-text prefix as a content event before the tool events and add a same-chunk boundary regression test.

OliverBryant (commented):
Found one streaming correctness issue.

xinference/model/llm/tool_parsers/qwen_tool_parser.py
qinxuye: Fixed in b6e0f939a. When a detokenizer delta contains ordinary content before a new `<tool_call>` tag, the parser now emits that newly generated prefix as a content event before the tool events. It avoids replaying prior content and ignores whitespace-only separators between tool calls. Added a same-chunk boundary regression test; the related parser, utility, MLX, and async-client suite passes (48 tests), and pre-commit passes.

xinference/model/llm/tool_parsers/qwen_tool_parser.py
```diff
@@ -328,30 +359,103 @@ def extract_tool_calls_streaming(
...
+                    ) > previous_text[-1].count(self.tool_call_end_token)
+                    if previous_inside_tool:
+                        previous_tool_end = prefix.rfind(self.tool_call_end_token)
+                        prefix = (
+                            prefix[previous_tool_end + len(self.tool_call_end_token) :]
+                            if previous_tool_end >= 0
+                            else ""
+                        )
+                    if prefix.strip():
+                        tool_events.append((prefix, None, None))
```
OliverBryant: **[P2] Keep inter-call content after the call it follows**

This appends the recovered prefix before the loop that emits newly completed calls. When the previous chunk is inside call 0 and the new delta both closes it and later starts call 1—for example, `</tool_call> explanation <tool_call>...`—the returned events are ordered as `[explanation, call_0, call_1]`, although the generated order is `[call_0, explanation, call_1]`. I reproduced this with a delta that completes a JSON call, adds ordinary text, then starts the next call. Please insert the recovered content at its position relative to the newly completed calls (and add this close/content/open boundary case to the regression tests).

OliverBryant (commented):
Found one ordering issue in the follow-up fix.

xinference/model/llm/tool_parsers/qwen_tool_parser.py
qinxuye: Fixed in 296be4421. Recovered content is now placed according to the parser state at the start of the delta: content before a newly opened call remains before its tool event, while content produced after closing an earlier call is emitted after newly completed calls and before the following partial call. Added the requested close/content/open regression case. The related parser, utility, MLX, and async-client suite passes (49 tests), and pre-commit passes.

xinference/model/llm/tool_parsers/qwen_tool_parser.py
```diff
@@ -328,30 +359,113 @@ def extract_tool_calls_streaming(
...
+                            parsed_json["arguments"],
+                            tool_call_index,
+                        )
+                    )
+
+                if inter_call_content is not None:
+                    # The delta closed an earlier call before producing this
+                    # content, so keep it after newly completed calls and before
+                    # any following partial call.
+                    tool_events.append((inter_call_content, None, None))
```
OliverBryant: **[P2] Insert inter-call content before a following completed call too**

Appending `inter_call_content` after the entire `current_complete` loop only preserves order when the following call is still partial. If one delta closes call 0, emits ordinary content, and also fully opens/closes call 1—`</tool_call> explanation <tool_call>...</tool_call>`—the result is `[call_0, call_1, explanation]` instead of `[call_0, explanation, call_1]`. I reproduced that exact JSON-call sequence against this head. A delta can advance multiple complete calls, so the events need to be emitted by their text positions rather than grouping all complete calls ahead of the recovered content; please add the complete-following-call boundary to the regression test as well.

OliverBryant (commented):
Found one remaining multi-call ordering issue.

xinference/model/llm/tool_parsers/qwen_tool_parser.py
qinxuye: Fixed in b0f47155d. Newly completed calls are now associated with their closing-tag positions, and the parser walks the generated text to emit plain content and completed tool events in source order. This covers both a following partial call and a following fully completed call without separate ordering special cases. Added the requested complete-following-call regression case; 62 related parser, utility, and MLX non-runtime tests pass, and pre-commit passes.

xinference/model/llm/mlx/core.py
```diff
@@ -190,11 +219,108 @@ def _get_or_create_generator(self, temperature: float, top_p: float):
...
+            extracted = batch_generator.extract_cache([result.uid]).get(result.uid)
+            if extracted is None:
+                continue
+            prompt_cache, cached_tokens = extracted
+            if len(cached_tokens) != cache_boundary:
+                continue
+
+            self._prompt_cache.insert_cache(
+                self._prompt_cache_model_key,
+                cached_tokens,
```
OliverBryant: **[P1] Batch prompt cache is keyed on a live, generator-mutated token list**

`cached_tokens` here is not a copy. In mlx-lm, `BatchGenerator.extract_cache` returns `self._prompt_batch.tokens[idx]` / `self._generation_batch.tokens[idx]` directly (`mlx_lm/generate.py:1684-1699`), and the generator keeps extending that same list **in place** for the rest of the request — `sti += ti` during prefill (`:1139-1140`) and `sti.append(ti)` per generated token (`:1376-1377`).

`LRUPromptCache.insert_cache` then stores that list **by reference** in the LRU deque (`_lru.push` → `self._lrus[cache_type].append((model, tokens))`), while `PromptTrie.add` walks it and copies the values into trie nodes. So the trie path is frozen at insert time but the LRU key keeps growing. On eviction, `PromptTrie.pop` walks the *grown* key (`path.append(path[-1][tok])`) against a trie path that only has the original prefix → `KeyError`.

Reproduced against upstream 0.31.3's actual trie/LRU logic: with `prompt_cache_size=2` (the new default), store a 3-token prefix `[1,2,3]`, let generation extend the list to `[1,2,3,4,5,6]`, then insert two more entries to trigger eviction:

```
REPRODUCED -> KeyError: 4  (evicting grown key [1, 2, 3, 4, 5, 6])
# same sequence with list(cached_tokens) -> evicted cleanly
```

The exception propagates to `_background_worker`, where the broad `except Exception` at :387-389 swallows it and sleeps 10 ms. The worker then re-hits it on every subsequent stored prefix, so the prompt cache silently stops working and `_n_bytes` drifts permanently because the decrements are skipped. The same aliasing corrupts `pop_prefixes`/`_lru.remove` bookkeeping even before eviction.

Upstream hits this exact spot and defends against it — `mlx_lm/server.py:874-876` passes `cache_key[:]`, and :903 passes `r.all_tokens[:]`. Suggest matching that:

```python
self._prompt_cache.insert_cache(
    self._prompt_cache_model_key,
    list(cached_tokens),
    prompt_cache,
    cache_type="system",
)
```

The VLM path at :1968-1977 is fine — `prompt_tokens[:prompt_cache_prefix_len]` already produces a fresh list. Only the batch path is affected, and `test_mlx_batch_generator_stores_reusable_prefix` can't catch it because its fake `insert_cache` only records the args.

OliverBryant (commented):
Verified the three streaming-order fixes from the earlier rounds hold, and re-checked the two Gemini findings as genuinely fixed. Three new items below: one real streaming regression, one connectivity regression, one doc/impl mismatch.

xinference/model/llm/tool_parsers/qwen_tool_parser.py
```diff
@@ -328,30 +359,128 @@ def extract_tool_calls_streaming(
...
+                    function_call = match.group(1)
+                    if not function_call.strip():
+                        continue
+                    end_index = function_call.find(">")
+                    if end_index != -1:
+                        parsed = self.parse_qwen35_tool_call(function_call)
+                        if parsed:
+                            completed_events[match.end()] = (*parsed, tool_call_index)
+                            continue
+                    parsed_json = json.loads(function_call, strict=False)
```
OliverBryant: **[P2] One malformed call now aborts a valid sibling parsed in the same delta**

The new loop parses *every* newly-completed call, and an unguarded `json.loads` / `parsed_json["name"]` here propagates out through the bare `raise` in the `except` at the bottom of `extract_tool_calls_streaming`. `_post_process_completion_chunk` does not wrap the parser call, so the exception kills the whole stream.

On `main` only `function_calls[-1]` was ever parsed, so a malformed call that was *followed* by a good one could not destroy the good one. Now it does. Verified against both revisions with `<tool_call>{"name": "b"}</tool_call>` (a schema slip: `arguments` missing) immediately followed by a well-formed `<tool_call><function=good><parameter=x>1</parameter></function></tool_call>` in a single delta:

```
main: (None, 'good', {'x': 1})
PR  : RAISED KeyError: 'arguments'
```

Same for a non-JSON body (`<tool_call>oops</tool_call>` + a good call) → `JSONDecodeError`.

Worth noting the split-delta case is *better* on this branch than on main (main drops the good call to `None`; the PR emits it), so the fix is just to make one bad call skippable rather than fatal — wrap the per-match parse in `try/except`, log, and `continue` so the surviving calls still stream. A `<tool_call>` body that reaches `json.loads` with a missing `name`/`arguments` key is exactly the case worth covering in a test; none of the current ones exercise a malformed call alongside a valid one.

xinference/model/utils.py
```diff
@@ -644,20 +644,35 @@ def is_locale_chinese_simplified() -> bool:
...
+
 def _is_hub_endpoint_reachable(url: str, timeout: float) -> bool:
     import requests
 
+    session = requests.Session()
+    # Auto detection must test direct connectivity.  If the endpoint is only
+    # reachable through HTTP(S)_PROXY / ALL_PROXY, selecting Hugging Face would
+    # route large model downloads through the user's proxy and consume its
+    # traffic quota.  Explicitly selecting Hugging Face still remains possible.
+    session.trust_env = False
```
OliverBryant: **[P2] `trust_env = False` also drops the CA bundle, so a private-CA mirror silently falls back to ModelScope**

`trust_env` gates more than proxies — it also controls `REQUESTS_CA_BUNDLE` / `CURL_CA_BUNDLE` and `~/.netrc`. Confirmed with the installed requests:

```
REQUESTS_CA_BUNDLE=/nonexistent/ca.pem
trust_env=True  -> verify='/nonexistent/ca.pem'
trust_env=False -> verify=True          # falls back to certifi
```

Concrete scenario: a self-hosted `HF_ENDPOINT` mirror behind a corporate/private CA, or a TLS-intercepting network, with `REQUESTS_CA_BUNDLE` set. The probe now raises `SSLError`, which the bare `except Exception: return False` below swallows, and detection returns `modelscope`. Before this PR that setup probed fine and selected `huggingface`.

The CA bundle is orthogonal to the proxy concern being addressed here — that mirror is *directly* reachable, which is precisely the case this function is meant to detect. Narrowing the change to proxies only (e.g. keep `trust_env` and pass `proxies={"http": None, "https": None}` on the request, or restore `verify` from the env explicitly) avoids the collateral damage. `~/.netrc` auth for a gated mirror is lost the same way.

The new tests monkeypatch `_is_hub_endpoint_reachable` wholesale, so this path is never exercised.

doc/source/models/sources/sources.rst
```diff
@@ -11,11 +11,12 @@ Automatic Detection
 
 By default, Xinference automatically decides between Hugging Face and ModelScope when
 launching a model: it probes whether the Hugging Face endpoint is reachable (mirrors set
-via ``HF_ENDPOINT`` and proxies set via ``HTTP_PROXY`` / ``HTTPS_PROXY`` are honored).
-If it is reachable, models are downloaded from Hugging Face; otherwise Xinference falls
-back to ModelScope. An HTTP error response (for example ``403``/``407`` from a blocking
-corporate proxy, or a ``5xx`` from a broken mirror) counts as unreachable, since
-downloads would fail anyway. When Hugging Face offline mode is enabled
+via ``HF_ENDPOINT`` are honored). If requests to that endpoint would use a proxy
+configured via ``HTTP_PROXY``, ``HTTPS_PROXY``, or ``ALL_PROXY``, automatic detection
```
OliverBryant: **[P3] `HTTP_PROXY` does not trigger proxy avoidance for the default https endpoint**

The docs say detection picks ModelScope when the endpoint "would use a proxy configured via ``HTTP_PROXY``, ``HTTPS_PROXY``, or ``ALL_PROXY``". `select_proxy` matches on the URL scheme, so for the default `https://huggingface.co` an http-only proxy yields no match:

```
HTTP_PROXY=http://p:8080   -> False   (probes directly, may select huggingface)
HTTPS_PROXY=http://p:8080  -> True
ALL_PROXY=socks5://p:1080  -> True
```

So `ALL_PROXY` is correctly honored, but listing `HTTP_PROXY` alongside it implies an avoidance that does not happen for the default endpoint (it does apply if `HF_ENDPOINT` is an `http://` URL). The `NO_PROXY` claim checks out — verified for `NO_PROXY`, lowercase `no_proxy`, and `NO_PROXY=*`. The code comment at `xinference/model/utils.py:665` ("HTTP(S)_PROXY / ALL_PROXY") carries the same imprecision.

xinference/model/llm/mlx/core.py
qinxuye: Fixed in b026812d2. The batch path now passes list(cached_tokens) to LRUPromptCache, so subsequent in-place generator growth cannot mutate the LRU eviction key. The regression test retains the generator-owned list, mutates it after insertion, and verifies the stored key remains the original prefix and is a distinct object.

xinference/model/llm/tool_parsers/qwen_tool_parser.py
qinxuye: Fixed in b026812d2. Parsing is isolated per completed match: malformed JSON/schema calls are logged and skipped while later valid siblings are still emitted with their absolute index. The regression test covers both a missing-arguments JSON call and a non-JSON body followed by a valid XML call.

xinference/model/utils.py
qinxuye: Fixed in b026812d2. The direct probe keeps trust_env enabled and disables only the http/https/all proxy entries for the request, preserving REQUESTS_CA_BUNDLE/CURL_CA_BUNDLE and netrc handling. The probe test now verifies both that no proxy is selected and that REQUESTS_CA_BUNDLE remains the effective verify setting.

doc/source/models/sources/sources.rst
qinxuye: Fixed in b026812d2. The docs and code comment now describe a proxy applicable to the endpoint URL scheme, explicitly noting that the default HTTPS endpoint uses HTTPS_PROXY or ALL_PROXY, while HTTP_PROXY applies to an HTTP HF_ENDPOINT.

OliverBryant:
Re-checked all four fixes in `b026812d2` against the new head — all confirmed, no regressions. LGTM.

**P1 MLX cache aliasing** — `list(cached_tokens)` matches upstream's own `cache_key[:]` at `mlx_lm/server.py:874`. The test is a real one: it keeps the generator-owned list, appends to it *after* insertion, and asserts the stored key is a distinct object.

**P2 malformed sibling call** — re-ran my original repro against both revisions:

```
BAD then GOOD, same delta:     main: RAISED KeyError   ->  now: (None, 'good', {'x': 1}, 1)
BAD first, GOOD later delta:   main: RAISED KeyError   ->  now: (None, 'good', {'x': 1}, 1)
```

Now strictly better than `main`, which still raises in both. I also checked index bookkeeping after a skip: the malformed call keeps its slot, so with `good / malformed / good` the survivors emit as index 0 and 2 with stable call IDs and a single name emission each. Non-contiguous indexes are correct here — `index` identifies the call, and the alternative would renumber calls the client may already have seen. The partial-name path is regex-only, so it can't raise.

**P2 CA bundle** — verified the narrowed fix does both things it needs to:

```
HTTPS_PROXY + ALL_PROXY + REQUESTS_CA_BUNDLE set, trust_env=True,
proxies={"http": None, "https": None, "all": None}
  -> selected proxy: None
  -> verify: /nonexistent/ca.pem    (CA bundle preserved)
```

**P3 docs** — scheme-dependent wording is accurate now, and the code comment at `model/utils.py:663` was updated to match.

All three new regression tests fail on `b0f47155d` and pass on `b026812d2`, so they're genuinely covering the fixes. Also re-ran the earlier streaming-order cases (prefix content, inter-call content, close/content/open, complete-following-call) — all still correct.

One thing I did not verify: no MLX hardware here, so the prompt-cache path is only checked against upstream mlx-lm's logic and the unit tests, not a live batch run.

OliverBryant (approved):
LGTM

PR #5342, opened 2026-08-16
fix(core): guard missing/invalid usage in non-stream chat metrics

Hyhyhyyy:
## What

In `ModelActor.chat`, the non-stream metrics block in the `finally` clause indexes `record["usage"]` unconditionally:

```python
usage = record["usage"]
completion_tokens = usage["completion_tokens"]
prompt_tokens = usage["prompt_tokens"]
```

If a backend returns a non-stream chat completion **without** a `usage` field (or with `usage: null`), this raises `KeyError` / `TypeError` inside the `finally` block, which masks the otherwise successful response and turns a working request into a 500.

The in-code comment already says *"Some backends may not have a valid usage, we just skip them."* — the intent was to skip, but the code never guarded against a missing/invalid `usage`.

## Why it's reachable

The non-stream path returns `orjson.dumps(dict)` (bytes); the `finally` block does `record = json.loads(response)` and then blindly indexes `usage`. The streaming path (`_to_generator` / `_to_async_gen`) already guards with `if final_usage is not None`, so the non-stream path was simply missing the same protection.

## Fix

Use `.get()` and only record token metrics when `usage` is present and valid. The `time_to_first_token_seconds` metric (non-stream latency) keeps being recorded as before, so existing behavior is unchanged for the normal case.

## Test

Added `test_chat_without_usage_does_not_crash` in `xinference/core/tests/test_model.py`, reusing the existing `MockModelActor` harness: a model whose non-stream chat returns a completion without `usage` must not raise.

### DIFF
--- a/xinference/core/model.py
+++ b/xinference/core/model.py
@@ -881,7 +881,7 @@ class ModelActor(xo.StatelessActor, CancelMixin):
             if record and isinstance(record, dict):
                 usage = record.get("usage")
                 # Some backends may not have a valid usage, we just skip them.
-                if usage:
+                if isinstance(usage, dict):
                     completion_tokens = usage.get("completion_tokens")
                     prompt_tokens = usage.get("prompt_tokens")
                     if completion_tokens is not None and prompt_tokens is not None:
--- a/xinference/core/model.py
+++ b/xinference/core/model.py
@@ -879,15 +879,17 @@ class ModelActor(xo.StatelessActor, CancelMixin):
             elif isinstance(response, bytes):
                 record = json.loads(response)
             if record and isinstance(record, dict):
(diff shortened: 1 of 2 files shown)


Hyhyhyyy:
Good catch, thanks! You're right — `if usage:` only checks truthiness, so a non-dict truthy `usage` (e.g. a string or list) would raise `AttributeError` on `.get()`.

Updated the guard to `isinstance(usage, dict)` in `xinference/core/model.py`, and added a regression test `test_chat_with_non_dict_usage_does_not_crash` that returns `usage` as a string to cover exactly that edge case. Pushed to the same branch.

qinxuye (approved):
LGTM

PR #5343, opened 2026-08-16
fix(ui): auto-size launch model dropdowns

Minamiyama:
## Summary

- Add optional content-aware width calculation to the shared `Select` component.
- Measure all option labels and size the dropdown based on the widest one.
- Preserve the trigger width as the dropdown's minimum width.
- Enable adaptive sizing for the model ability and model update type dropdowns.

## Problem

Some dropdown options, such as audio model abilities and `Embedding`, were truncated because the dropdown width was always fixed to the trigger width.

## before:
47eb2566f32e09195c1378a77363870b

## after:
1de7fd4418f03aeb24352de11c66b655


qinxuye (approved):
LGTM

PR #5345, opened 2026-08-16
feat: [new model] minimax music3

Minamiyama:
## Summary

This PR adds initial MiniMax-Music3 support as a built-in audio model with the new `text2music` ability.

- Adds `MiniMax-Music3` using Diffusers `ModularPipeline`
- Supports Hugging Face and ModelScope model sources
- Adds a dedicated Music Generation Web UI
- Reuses the OpenAI-compatible `POST /v1/audio/speech` endpoint
- Adds user documentation and request examples

## Backend

- Model family: `minimax_music3`
- Model ability: `text2music`
- Engine: Diffusers only
- Accelerator: NVIDIA CUDA only
- Default dtype: BF16
- Supports optional CPU and group offload
- Loads pipeline components from the local Xinference model cache to avoid duplicate downloads
- Pins the Diffusers commit containing MiniMax-Music3 ModularPipeline support
- Preserves the pipeline's native 44.1 kHz, stereo, float32 audio without resampling or quantization
- Encodes the result as a standards-compliant IEEE-float WAV

The model is registered under the MiniMax-Music3 Community License.

## API

Music generation reuses `/v1/audio/speech`:

- `input`: lyrics
- `kwargs.instruct`: music description
- `kwargs.seed`: non-negative random seed, default `0`
- `kwargs.duration`: maximum duration in seconds, default `60`, range `0.04` to `360`
- `response_format`: `wav` only

Unsupported speech or sampling parameters are rejected with model-specific validation errors.

## Web UI

Adds a distinct Music Generation capability with:

- Multiline lyrics input
- Music description input
- Seed input with a random-seed button
- Duration in seconds
- WAV audio playback

Regular TTS and voice-cloning forms retain their existing behavior.

1eec3e0c37196c411e0e52a72dc0bf7e

### DIFF
--- a/frontend/src/components/pages/running-model-detail/capability-config.tsx
+++ b/frontend/src/components/pages/running-model-detail/capability-config.tsx
@@ -748,7 +748,7 @@ export const CAPABILITY_CONFIGS: Partial<Record<ModelAbility, CapabilityConfig>>
           key: 'kwargs',
           required: true,
           value: JSON.stringify({
-            prompt_text: 'Warm acoustic pop with intimate vocals',
+            instruct: 'Warm acoustic pop with intimate vocals',
             seed: 7,
             duration: 60,
           }),
@@ -759,7 +759,7 @@ export const CAPABILITY_CONFIGS: Partial<Record<ModelAbility, CapabilityConfig>>
     },
     initialValues: {
       input: '',
-      prompt_text: '',
+      instruct: '',
       seed: 0,
       duration: 60,
     },
@@ -772,7 +772,7 @@ export const CAPABILITY_CONFIGS: Partial<Record<ModelAbility, CapabilityConfig>>
       voice: 'default',
       response_format: 'wav',
       kwargs: JSON.stringify({
-        prompt_text: stringValue(values.prompt_text),
+        instruct: stringValue(values.instruct),
         seed: numberValue(values.seed, 0),
         duration: numberValue(values.duration, 60),
       }),
--- a/frontend/src/components/pages/running-model-detail/capability-config.tsx
+++ b/frontend/src/components/pages/running-model-detail/capability-config.tsx
@@ -7,6 +7,7 @@ import {
   ImageUp,
   ListFilter,
   Mic,
+  Music2,
   Paintbrush,
   ScanText,
   Video,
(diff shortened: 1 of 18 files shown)


xinference/model/audio/engine.py
```diff
@@ -319,6 +320,20 @@ def is_model_family_supported(cls, model_family: "AudioModelFamilyV2") -> bool:
...
+
+    @classmethod
+    def check_lib(cls):
+        if virtual_env_allows_missing_engine():
+            return True
+        return super().check_lib()
+
+    @classmethod
+    def match(cls, model_family: "AudioModelFamilyV2") -> bool:
+        return model_family.model_family == "minimax_music3"
```
qinxuye: This CUDA-only engine currently matches on every host. On a CPU-only or Apple Silicon worker it is therefore registered as available, and `create_audio_model_instance()` reaches `CacheManager.cache()` before `MiniMaxMusic3Model.load()` rejects the device. That can download the full model only to fail afterward. Please gate `match()` with `has_cuda_device()` (and add a no-CUDA engine-registration test), so unsupported workers reject the launch before caching the weights.

xinference/model/audio/engine.py
```diff
@@ -319,6 +320,20 @@ def is_model_family_supported(cls, model_family: "AudioModelFamilyV2") -> bool:
...
+
+    @classmethod
+    def check_lib(cls):
+        if virtual_env_allows_missing_engine():
+            return True
+        return super().check_lib()
+
+    @classmethod
+    def match(cls, model_family: "AudioModelFamilyV2") -> bool:
+        return has_cuda_device() and model_family.model_family == "minimax_music3"
```
qinxuye, line 334: The CUDA gate here does not yet prevent the unsupported worker from downloading the model. When `has_cuda_device()` is false, MiniMax-Music3 is absent from `AUDIO_ENGINES`, but `create_audio_model_instance()` still reaches `CacheManager.cache()` before the legacy fallback raises `Unsupported audio model family: minimax_music3`. I reproduced this with a mocked cache and observed one cache call. Please reject the unavailable engine before the cache call, and keep a regression test that asserts `CacheManager.cache()` is not called on a no-CUDA worker; the removed test only checked registry state.

frontend/src/components/pages/running-model-detail/capability-config.tsx
```diff
@@ -728,4 +729,53 @@ export const CAPABILITY_CONFIGS: Partial<Record<ModelAbility, CapabilityConfig>>
...
+    resultPanel: ResultPanels.Universal,
+    responseType: 'blob',
+    transformValues: ({ modelUid, values }) => ({
+      model: modelUid,
+      input: stringValue(values.input),
+      voice: 'default',
+      response_format: 'wav',
+      kwargs: JSON.stringify({
+        prompt_text: stringValue(values.prompt_text),
+        seed: values.seed,
```
qinxuye: Please normalize these numeric form values before serializing them. Clearing either input currently sends an empty string in `kwargs`, so the backend rejects the request even though the UI defines valid defaults. Use `numberValue(values.seed, 0)` and `numberValue(values.duration, 60)` here.

qinxuye (approved):
LGTM

PR #5346, opened 2026-08-16
fix(model): calculate download progress per file

Minamiyama:
## Summary

- Calculate model download progress using equal per-file weights.
- Register `tqdm` progress bars when they are created so the total file count is known before the first file completes.
- Add a regression test covering files with different sizes and files that have not started downloading.

## Problem

The previous calculation mixed two weighting strategies:

- Completed files were weighted by file count.
- Active downloads were weighted by their byte sizes.

When a large file completed, its contribution changed from byte-weighted progress to a single completed-file count. This could make the total progress drop significantly and then continue increasing from the lower value.

Additionally, the repository-level progress bar was not registered until its first update. Before the first file completed, the downloader could incorrectly assume that the total file count was `1`.

## Solution

Each file now has an equal weight in the total progress:

```text
progress = (
    completed_file_count
    + sum(active_file_downloaded_bytes / active_file_total_bytes)
) / total_file_count
```

Therefore:

- A completed file contributes `1 / total_file_count`.
- An active file contributes its individual percentage divided by the total file count.
- A file that has not started contributes zero.

This keeps a file's contribution continuous when it transitions from downloading to completed, regardless of its size.

## Validation

Added a regression test with differently sized files and one pending file. The tested progress sequence is:

```text
32.8571% -> 33.3333% -> 50%
```

The progress no longer drops when the large file completes.

### DIFF
--- a/xinference/model/utils.py
+++ b/xinference/model/utils.py
@@ -1029,6 +1029,10 @@ class CancellableDownloader:
         # progress for file downloader
         # mainly when tqdm unit is set
         self._download_progresses: Set[tqdm] = set()
+        # Byte bars and the repository-level bar advance in separate calls.
+        # Keep the reported value monotonic while a completed byte bar waits
+        # for the repository-level bar to account for the same file.
+        self._last_progress = 0.0
         # Instance-specific tqdm tracking
         self._patched_instances: Set[int] = set()
 
@@ -1069,6 +1073,7 @@ class CancellableDownloader:
         with self._progress_lock:
             self._main_progresses.clear()
             self._download_progresses.clear()
+            self._last_progress = 0.0
 
     def _progress_snapshots(self) -> Tuple[List[tqdm], List[tqdm]]:
         """Return stable copies while tqdm callbacks may update the sets."""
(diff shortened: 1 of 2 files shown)


xinference/model/utils.py
```diff
@@ -1092,27 +1092,28 @@ def get_progress(self) -> float:
...
             tasks = 1
 
-        finished_ratio = finished_tasks / tasks
-
-        all_download_progress = finished_download_progress = 0
+        active_file_progress = 0.0
         for download_progress in download_progresses:
-            # we skip finished download
+            # Completed downloads are already represented by finished_tasks.
             if download_progress.n == download_progress.total:
```
qinxuye: A completed byte bar is dropped here before the repository-level bar necessarily advances. There is a real interval between `large_file.update(1)` and `all_bar.update(1)` where the new test returns `0.0`, so progress still falls from about 32.86% to 0% before recovering to 33.33%. Please preserve the completed file contribution until the repository bar catches up, and add an assertion between those two updates to cover this transition.

qinxuye: This is still not fully fixed because `_last_progress` only helps if `get_progress()` happened to be called before the byte bar completed. For example, with a repository bar of total 3, create a 70-byte file bar, call `file_bar.update(70)`, and then call `get_progress()` for the first time: the completed bar is skipped and `_last_progress` is still 0, so the result is 0 instead of 1/3. Please preserve completed-but-not-yet-accounted file contributions rather than relying on the historical maximum, and add a regression assertion for completion without an intermediate poll. Separately, the current head also fails Black on `xinference/model/utils.py`.

qinxuye (approved):
LGTM

PR #5350, opened 2026-08-17
fix(llm): correctly pass multimodal_projector to LlamaCppModel

Ricardo-M-L:
This PR addresses: fix(llm): correctly pass multimodal_projector to LlamaCppModel


qinxuye:
there is conflict, and pls address gemini comment first

PR #5352, opened 2026-08-17
feat(audio): support FireRedTTS3

Minamiyama:
## Summary

Add FireRedTTS3 support as two independent built-in audio model cards:

- `FireRedTTS3-Base`
- `FireRedTTS3-Instruct`

Upstream model:

- https://github.com/FireRedTeam/FireRedTTS3
- https://huggingface.co/FireRedTeam/FireRedTTS3

## Model sources

Both variants share the same model repository and local cache:

- Hugging Face revision: `main`
- ModelScope revision: `master`

### FireRedTTS3-Base

Abilities:

- `text2audio`
- `text2audio_voice_cloning`

Base requires:

- Reference audio through `prompt_speech`
- Reference transcript through `prompt_text`

### FireRedTTS3-Instruct

Abilities:

- `text2audio`
- `text2audio_voice_design`
- `text2audio_voice_cloning`

Instruct supports two modes:

1. Voice design without reference audio
   - Pass a natural-language voice description through `instruct`
   - `prompt_text` is also accepted as a compatible fallback

2. Voice cloning with reference audio
   - Pass reference audio through `prompt_speech`
   - Pass its transcript through `prompt_text`
   - Voice-design instructions are ignored when reference audio is provided


frontend/src/components/pages/running-model-detail/panels/form-panels.tsx
```diff
@@ -525,19 +537,24 @@ export function SpeechPanel({ form, model }: CapabilityFormProps) {
...
+          />
+        </FormField>
+      )}
+      {showPromptText && (
+        <FormField name="prompt_text" label="Prompt Text">
+          <Textarea placeholder="Text spoken in the prompt audio" />
+        </FormField>
+      )}
+      {showInstruct && (
+        <FormField name="instruct" label={instructLabel}>
```
qinxuye: This refactor drops the existing required rule from MiniMax-Music3's music description. The backend rejects an empty instruct, and FireRedTTS3 voice design also requires it, so the shared field should remain required whenever it is shown, for example with rules={[{ required: isMusicGeneration || showVoiceInstruction }]}.

frontend/src/constants/register.ts
```diff
@@ -226,12 +226,14 @@ export const MODEL_ABILITY_IMAGE_OPTIONS = [
...
   Text2audio = 'text2audio',
   Text2music = 'text2music',
+  Text2audioVoiceDesign = 'text2audio_voice_design',
   Audio2text = 'audio2text',
   SpeakerEmbedding = 'speaker_embedding',
 }
 export const MODEL_ABILITY_AUDIO_OPTIONS = [
   { label: 'Text2audio', value: ModelAbilityForAudio.Text2audio },
   { label: 'MusicGeneration', value: ModelAbilityForAudio.Text2music },
+  { label: 'Text2audioVoiceDesign', value: ModelAbilityForAudio.Text2audioVoiceDesign },
```
qinxuye: text2audio_voice_design is a sub-capability, but audio registration uses a radio group and converts the selected value into a one-element model_ability array. Selecting this option therefore registers only text2audio_voice_design, without the primary text2audio; getPrimaryModelAbilities() then filters it out and the running-model page has no usable panel. Please remove this standalone option, or change registration to include text2audio as well (and add a matching FireRedTTS3 family if custom registration is intended).

qinxuye (approved):
LGTM

Issue #5347, opened 2026-08-16
BUG: xinference 3.2.0 wheel is unimportable - KeyError: 'model_specs' (an audio model entry is packaged into llm_family.json)


sfwani:
### System Info / 系統信息

Any. Reproduced on `python:3.12-slim` with a clean `pip install xinference==3.2.0`.

### Running Xinference with Docker? / 是否使用 Docker 运行 Xinfernece？

- [ ] docker / docker
- [x] pip install / 通过 pip install 安装
- [ ] installation from source / 从源码安装

### Version info / 版本信息

xinference 3.2.0 (from PyPI)

### The command used to start Xinference / 用以启动 xinference 的命令

```
xinference-local -H 0.0.0.0 -p 9997
```

Any entrypoint fails identically, because the error happens at import time.

### Reproduction / 复现过程

```
pip install xinference==3.2.0
xinference-local -H 0.0.0.0 -p 9997
```

```
File ".../xinference/model/__init__.py", line 25, in _install
    llm_install()
File ".../xinference/model/llm/__init__.py", line 273, in _install
    load_model_family_from_json("llm_family.json", BUILTIN_LLM_FAMILIES)
File ".../xinference/model/llm/__init__.py", line 187, in load_model_family_from_json
    for spec in json_obj["model_specs"]:
                ~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'model_specs'
```

### Cause

The published wheel's `xinference/model/llm/llm_family.json` contains one entry that is not an LLM
family and has no `model_specs` key:

```python
import json
d = json.load(open('.../site-packages/xinference/model/llm/llm_family.json'))
print(len(d))
print([e['model_name'] for e in d if 'model_specs' not in e])
```

```
167
['speech_campplus_sv_zh-cn_16k-common']
```

That entry is an **audio** model. It is byte-for-byte identical to the entry of the same name in
`xinference/model/audio/model_spec.json`, where it also correctly appears.

This looks like a release-pipeline problem rather than a source problem. The `v3.2.0` git tag is
clean:

| Artifact | entries in `llm_family.json` | entries missing `model_specs` |
|---|---|---|
| `v3.2.0` git tag | 166 | none |
| `xinference-3.2.0-py3-none-any.whl` from PyPI | 167 | 1 (`speech_campplus_sv_zh-cn_16k-common`) |

So the audio entry is being appended to the LLM family JSON somewhere between tagging and
[...]

PR #5353, opened 2026-08-17
fix: validate LLM metadata before packaging

li2631026381-alt:
## Summary

- validate the built-in LLM family metadata before wheel, sdist, or editable builds
- reject non-object entries and families whose `model_specs` field is missing or not a list
- include the offending model name in the build error
- add focused regression coverage to the existing build-backend test suite

## Root cause

The published `xinference==3.2.0` wheel contains the audio model `speech_campplus_sv_zh-cn_16k-common` inside `xinference/model/llm/llm_family.json`. That entry has no `model_specs` field, so every Xinference entry point fails during import with `KeyError: 'model_specs'`.

The wheel records build commit `324c8afebb2ba795e59c965b500580ba777afb01`, where the malformed entry was present. The current `v3.2.0` tag points to a later commit where it has been removed, which explains why checking the current tag alone does not reproduce the artifact.

This change makes malformed built-in LLM metadata a packaging error before the web build and PyPI upload can proceed.

## User impact

Future releases cannot package this class of malformed LLM metadata silently. The error identifies the offending family, for example:

```
Invalid LLM families 'speech_campplus_sv_zh-cn_16k-common' in xinference/model/llm/llm_family.json: each entry must define 'model_specs' as a list
```

The already-published 3.2.0 wheel still requires a new release or replacement version from the maintainers.

## Validation

- `.venv/bin/python -m pytest --noconftest -q xinference/tests/test_build_backend.py` — 8 passed
- `.venv/bin/pre-commit run --files build_backend.py xinference/tests/test_build_backend.py` — all hooks passed
- `NO_WEB_UI=1 .venv/bin/python -m build --wheel ...` — wheel built successfully
- the validator rejects the official 3.2.0 wheel and names `speech_campplus_sv_zh-cn_16k-common`

Addresses #5347.

### DIFF
--- a/build_backend.py
+++ b/build_backend.py
@@ -62,7 +62,7 @@ def _validate_builtin_model_specs(root=None):
         if not isinstance(family, dict):
             invalid_families.append(f"entry {index}")
         elif not isinstance(family.get("model_specs"), list):
-            invalid_families.append(family.get("model_name", f"entry {index}"))
+            invalid_families.append(family.get("model_name") or f"entry {index}")
 
     if invalid_families:
         names = ", ".join(repr(name) for name in invalid_families)
--- a/build_backend.py
+++ b/build_backend.py
@@ -26,6 +26,7 @@ wheels, sdists, and editable installs:
   file is kept as-is.
 """
 
+import json
 import os
 import re
 import subprocess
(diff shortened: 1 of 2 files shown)


qinxuye:
Oh, sorry about that, I just pushed a v3.2.1 tag to address the pypi wheel issue.

qinxuye (approved):
LGTM

PR #5354, opened 2026-08-17
fix(ui): include scrollbar in dropdown width

Minamiyama:
## Summary

Fix auto-width dropdowns truncating long option labels when a vertical scrollbar is present.

## Root Cause

The dropdown width calculation included:

- Widest option width
- Container padding
- Dropdown border width

However, it did not include the width occupied by the vertical scrollbar. When the option list exceeded its maximum height, the scrollbar reduced the available content width and caused labels to remain truncated.

## Changes

- Calculate scrollbar width using:

  ```ts
  optionsContainer.offsetWidth - optionsContainer.clientWidth
  ```

- Include scrollbar width in the final dropdown width.
- Preserve existing behavior when no scrollbar is present, where the calculated scrollbar width is `0`.

### before
bf235fad3f3f0ed9cdb7ae7c73db2c05

### after
a5c1ff7163aad14d06a56e2ca4a9016c

### DIFF
--- a/frontend/src/components/ui/select.tsx
+++ b/frontend/src/components/ui/select.tsx
@@ -200,9 +200,10 @@ export function Select<T extends SelectValue = SelectValue>({
     const optionsContainerPadding =
       Number.parseFloat(optionsContainerStyle.paddingLeft) +
       Number.parseFloat(optionsContainerStyle.paddingRight);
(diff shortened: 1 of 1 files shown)


qinxuye (approved):
LGTM

PR #5355, opened 2026-08-17
Enh: Expose VoiceDesign ability to Qwen3-TTS-Voice-Design

Minamiyama:
## Summary

Add the `text2audio_voice_design` ability to both PyTorch and MLX specifications of `Qwen3-TTS-12Hz-1.7B-VoiceDesign`.

This enables the Web UI voice instruction field and routes its value to the existing `generate_voice_design` implementation through `instruct`.

### DIFF
--- a/xinference/model/audio/model_spec.json
+++ b/xinference/model/audio/model_spec.json
@@ -2281,7 +2281,8 @@
     "model_description": "This is a custom model description.",
     "multilingual": true,
     "model_ability": [
-      "text2audio"
+      "text2audio",
+      "text2audio_voice_design"
     ],
     "model_family": "qwen3_tts",
     "engine": "PyTorch",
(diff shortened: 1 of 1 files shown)


xinference/model/audio/model_spec.json
```diff
@@ -2281,7 +2281,8 @@
     "model_description": "This is a custom model description.",
     "multilingual": true,
     "model_ability": [
-      "text2audio"
+      "text2audio",
+      "text2audio_voice_design"
```
qinxuye, line 2285: This change depends on the frontend support in #5352, which is still open. On the current base, text2audio_voice_design is not defined in ModelAbility, not filtered as a sub-capability, and SpeechPanel does not render or submit the instruct field. As a result, this does not enable voice design; it adds an unknown ability option that opens the empty-state panel. Please merge #5352 first and rebase this PR, or include the required frontend support here.

qinxuye (approved):
LGTM

PR #5364, opened 2026-08-18
fix(core): make worker supervisor reconnection generation-safe

### LINKED ISSUE #5360: BUG: stale worker RPC failures can clear a fresh supervisor connection during reconnect
### System Info / 系統信息

- Xinference: current `main` (`1f3b2d5c` at investigation time)
- Python: 3.11
- OS: Linux
- Reproduced with focused worker unit tests that model supervisor restart and concurrent status/heartbeat activity; no model download or GPU is required.

- [ ] docker
- [ ] pip install
- [x] installation from source

### Version info / 版本信息

Current upstream `main`, including #5250 (worker reconnection after supervisor restart) and #5290 (replica preservation during recovery).

A distributed deployment is affected:

```bash
xinference-supervisor -H <supervisor-host>
xinference-worker -e http://<supervisor-host>:9997 -H <worker-host>
```

### Reproduction / 复现过程

1. Start a worker, then restart the supervisor so the worker has a stale cached supervisor reference.
2. Let a heartbeat or status-report RPC against the stale reference fail while another task reconnects and installs a fresh reference.
3. The failing operation calls an unconditional `_clear_supervisor_refs()`. It can therefore erase the newly installed reference and reset registration state.
4. Concurrent `get_supervisor_ref(add_worker=True)` calls can also run overlapping registration work. A reconnect may register an earlier replica snapshot even if the worker's replica set changed while `add_worker()` was in flight.

### Expected behavior / 期待表现

- A failure may clear only the exact supervisor reference/generation used by that failing RPC; it must not clear a newer connection.
- Supervisor initialization and registration should be single-flight on the actor loop.
- If replica state changes during registration, the latest snapshot should be replayed before the worker is marked registered.
- A non-cancellation failure in the periodic heartbeat/status loop should be retried instead of permanently stopping worker liveness reporting.

This extends #5250 / #5290 with stale-reference race protection and serialized registration.

baka-world:
## Summary

- serialize supervisor reference initialization and registration on the worker actor loop
- associate cached supervisor references with a generation and conditionally clear only the stale reference used by a failed RPC
- replay the latest replica snapshot when the worker's state changes during registration
- make periodic heartbeat/status reporting retry ordinary RPC failures instead of terminating liveness reporting
- add regression coverage for stale-reference cleanup, re-registration after heartbeat failure, concurrent registration, and periodic-report recovery

## Why

#5250 restores reconnection after a supervisor restart, but a concurrent failing heartbeat or status-report RPC can still unconditionally clear a newly installed supervisor reference. Concurrent `get_supervisor_ref(add_worker=True)` calls can also overlap registration and mark the worker registered using an outdated replica snapshot.

This patch makes the recovery path generation-safe and single-flight while preserving the supervisor restart/replica recovery behavior from #5250 and #5290.

Fixes #5360.

## Validation

- `XINFERENCE_HOME=/tmp/xinference-home python -m pytest -q xinference/core/tests/test_worker.py -k 'worker_heartbeat_failure_clears_cached_supervisor_refs or periodical_report_status or clear_supervisor_refs or supervisor_registration' -p no:cacheprovider --timeout=60 --timeout-method=thread` — 8 passed
- `pre-commit run --files xinference/core/worker.py xinference/core/tests/test_worker.py` — passed
- `git diff --check` — passed

### DIFF
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -227,6 +227,7 @@ def _mark_venv_setup_done(
 # and provides a fast uncontended path on Unix as well.
 _venv_locks: Dict[str, threading.Lock] = {}
 _venv_locks_lock = threading.Lock()
+_SUPERVISOR_INIT_TIMEOUT = 60
 
 
 def _wait_for_metrics_export_server(
@@ -694,6 +695,11 @@ class WorkerActor(xo.StatelessActor):
         self._supervisor_address = supervisor_address
         self._supervisor_endpoint = supervisor_endpoint
         self._supervisor_ref: Optional[xo.ActorRefType] = None
+        self._supervisor_ref_lock = threading.Lock()
+        self._supervisor_ref_generation = 0
+        self._supervisor_ref_address: Optional[str] = None
+        self._supervisor_init_lock: Optional[asyncio.Lock] = None
+        self._actor_loop: Optional[asyncio.AbstractEventLoop] = None
         self._main_pool = main_pool
         self._main_pool.recover_sub_pool = self.recover_sub_pool
         self._status_guard_ref: xo.ActorRefType[
@@ -1176,6 +1182,9 @@ class WorkerActor(xo.StatelessActor):
             logger.debug("Startup VRAM poll failed", exc_info=True)
 
     async def __post_create__(self):
+        self._actor_loop = asyncio.get_running_loop()
+        self._supervisor_init_lock = asyncio.Lock()
+
         from ..model.audio import (
             CustomAudioModelFamilyV2,
             generate_audio_description,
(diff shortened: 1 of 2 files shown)


xinference/core/worker.py
```diff
@@ -4271,17 +4412,34 @@ async def report_status(self):
...
             logger.exception("Report status got error.")
+        supervisor_ref = None
+        supervisor_generation = None
         try:
-            supervisor_ref = await self.get_supervisor_ref()
-            await supervisor_ref.report_worker_status(self.address, status)
+            (
+                supervisor_ref,
+                supervisor_generation,
+            ) = await self._get_supervisor_ref_with_generation()
```
qinxuye, line 4430: The full worker test file now fails because this path bypasses the existing `test_worker_report_status_reconnects_and_replays_running_models` monkeypatch of `WorkerActor.get_supervisor_ref`. Please update that test to patch `_get_supervisor_ref_with_generation` instead and return `(ref, generation)` while advancing the generation consistently. Reproduced on this head with `pytest -q xinference/core/tests/test_worker.py`: 1 failed, 74 passed.

baka-world: Addressed in afc782d0: the reconnect test now patches `_get_supervisor_ref_with_generation`, returns `(ref, generation)`, and increments the generation whenever it installs a replacement supervisor reference. The complete worker suite now passes: `XINFERENCE_HOME=/tmp/xinference-home python -m pytest -q xinference/core/tests/test_worker.py -p no:cacheprovider --timeout=60 --timeout-method=thread` — 75 passed. Targeted pre-commit hooks also pass.

baka-world:
Also rebased this branch onto current `main` at `4e250f44` so the reconnect change is tested against the same corrected CI baseline. The rebase was conflict-free; new head is `1a80108f`.

Validation after rebase:

- `XINFERENCE_HOME=/tmp/xinference-home python -m pytest -q xinference/core/tests/test_worker.py -p no:cacheprovider --timeout=60 --timeout-method=thread` — 78 passed
- targeted `pre-commit` hooks — passed
- `git diff --check upstream/main...HEAD` — passed

qinxuye (approved):
LGTM

baka-world:
The rebased GPU T4 job failed before pytest started: the `Install dependencies` step exited 100 after repeated connection timeouts fetching Ubuntu APT packages from `azure.archive.ubuntu.com` (for example `alsa-topology-conf` / `libasound2t64`). The test step was skipped, so this did not execute the reconnect change or its tests.

The same current-main GPU job and the rebased #5366 GPU job completed successfully. I cannot re-run an upstream Actions job with this token because GitHub requires repository-admin permission. Could a maintainer please re-run [this GPU job](https://github.com/xorbitsai/inference/actions/runs/32210541605/job/95942664904)?

qinxuye:
> The rebased GPU T4 job failed before pytest started: the `Install dependencies` step exited 100 after repeated connection timeouts fetching Ubuntu APT packages from `azure.archive.ubuntu.com` (for example `alsa-topology-conf` / `libasound2t64`). The test step was skipped, so this did not execute the reconnect change or its tests.
> 
> The same current-main GPU job and the rebased #5366 GPU job completed successfully. I cannot re-run an upstream Actions job with this token because GitHub requires repository-admin permission. Could a maintainer please re-run [this GPU job](https://github.com/xorbitsai/inference/actions/runs/32210541605/job/95942664904)?

Never mind, just ignore this failure.

PR #5369, opened 2026-08-18
feat(image): sensenova u1.5

Minamiyama:
## Summary

- Add support for `SenseNova-U1.5-8B-MoT-Preview` using the Transformers backend.
- Support both model sources:
  - Hugging Face: `sensenova/SenseNova-U1.5-8B-MoT-Preview`, revision `main`
  - ModelScope: `SenseNova/SenseNova-U1.5-8B-MoT-Preview`, revision `master`
- Support text-to-image and single/multi-reference image editing.
- Vendor the official `src/sensenova_u1` runtime under `xinference/thirdparty`.
- Add virtualenv dependencies without requiring a Git checkout during model launch.

## Motivation

SenseNova-U1.5 uses its own Transformers-based runtime instead of Diffusers.

Installing that runtime through a VCS dependency makes model startup depend on GitHub access and a relatively large repository checkout. This can make virtualenv creation slow or fail in restricted deployment environments.

This change vendors the required upstream runtime from:

- Repository: https://github.com/OpenSenseNova/SenseNova-U1
- Commit: `35fb2fc520869aa09223a819882abdb3dc328ecd`
- License: Apache License 2.0

The vendored upstream source is kept unchanged.

## Implementation

- Add `SenseNovaU1Model` to adapt the official runtime to Xinference image APIs.
- Register SenseNova-U1.5 as a Transformers image engine.
- Exclude the model from Diffusers engine matching.
- Select the model's registered default engine when `model_engine` is omitted, instead of always defaulting to Diffusers.
- Add image size validation, automatic resizing, generation configuration handling, VRAM offloading, and LoRA integration.
- Load the vendored runtime lazily from `xinference/thirdparty`.
- Reuse system Torch, TorchVision, and NumPy packages in the model virtualenv.

### DIFF
--- a/xinference/model/image/core.py
+++ b/xinference/model/image/core.py
@@ -322,12 +322,13 @@ def create_image_model_instance(
         lora_fuse_kwargs = None
 
     from .engine_family import (
+        IMAGE_ENGINES,
         check_engine_by_model_name_and_engine,
         check_engine_by_model_name_and_engine_with_virtual_env,
     )
 
     if model_engine is None:
-        model_engine = "diffusers"
+        model_engine = next(iter(IMAGE_ENGINES.get(model_name, {})), "diffusers")
 
     if enable_virtual_env is None:
         from ...constants import XINFERENCE_ENABLE_VIRTUAL_ENV
--- a/xinference/model/image/core.py
+++ b/xinference/model/image/core.py
@@ -114,6 +114,26 @@ def generate_image_description(
     return res
 
 
+def resolve_image_model_engine(
+    model_name: str, model_engine: Optional[str] = None
+) -> Optional[str]:
+    from .engine_family import IMAGE_ENGINES
+
+    available_engines = IMAGE_ENGINES.get(model_name)
+    if not available_engines:
+        return model_engine
+    if model_engine is None:
+        return next(iter(available_engines))
+    return next(
+        (
+            engine
+            for engine in available_engines
+            if engine.lower() == model_engine.lower()
+        ),
+        model_engine,
+    )
+
+
 def match_diffusion(
     model_name: str,
     download_hub: Optional[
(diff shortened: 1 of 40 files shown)


xinference/model/image/core.py
```diff
@@ -322,12 +322,13 @@ def create_image_model_instance(
         lora_fuse_kwargs = None
 
     from .engine_family import (
+        IMAGE_ENGINES,
         check_engine_by_model_name_and_engine,
         check_engine_by_model_name_and_engine_with_virtual_env,
     )
 
     if model_engine is None:
```
qinxuye: Resolving the default engine here is too late for virtual-environment setup. `WorkerActor.launch_builtin_model` has already built the venv path/environment with the original `model_engine=None`, and later calls `_prepare_virtual_env(..., model_engine)` with that same `None`. Since every dependency in this model spec is guarded by `#engine# == "transformers"`, marker filtering drops the entire package list, so launching without an explicit `--model-engine` fails in a clean default venv. Please resolve and persist the image engine before the worker creates/prepares the venv, and add a regression test for an omitted engine.

xinference/model/image/sensenova_u1.py
```diff
@@ -0,0 +1,415 @@
...
+        if width % _IMAGE_GRID_FACTOR or height % _IMAGE_GRID_FACTOR:
+            raise ValueError(
+                "SenseNova-U1 image width and height must both be multiples of "
+                f"{_IMAGE_GRID_FACTOR}, got {width}x{height}."
+            )
+
+    def _get_generate_config(self, kwargs: Dict[str, Any]) -> Dict[str, Any]:
+        config = (self._model_spec.default_generate_config or {}).copy()
+        config.update(
+            {key: value for key, value in kwargs.items() if value is not None}
```
qinxuye, line 204: Please normalize the seed to Xinference image semantics before calling the upstream generator. When no seed is provided, `seed` is absent and `t2i_generate`/`it2i_generate` silently use their fixed default `0`; when callers pass the conventional `-1`, it is forwarded to `manual_seed(-1)`. Both cases make repeated requests deterministic, unlike the other Xinference image backends, which generate a fresh random seed for missing or `-1`. Preserve explicit non-negative seeds, but replace missing/`-1` with a random seed.

qinxuye (approved):
LGTM

PR #5372, opened 2026-08-18
fix(llm): preserve DeepSeek V4 tool argument types

m199369309:
## Summary

- honor DeepSeek V4's `string` attribute when parsing tool-call parameters
- keep `string="true"` values verbatim while JSON-decoding numbers, booleans, nulls, arrays, and objects for `string="false"`
- apply the same typed behavior to complete and streaming tool-call parsing
- correct the parser-specific streaming error message

## Tests

- `pytest -q xinference/model/llm/tool_parsers/tests/test_deepseek_v4_tool_parser.py`
- `pre-commit run --files xinference/model/llm/tool_parsers/deepseek_v4_tool_parser.py xinference/model/llm/tool_parsers/tests/test_deepseek_v4_tool_parser.py`

### DIFF
--- a/xinference/model/llm/tool_parsers/deepseek_v4_tool_parser.py
+++ b/xinference/model/llm/tool_parsers/deepseek_v4_tool_parser.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import re
 from typing import Any, Dict, List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


qinxuye (approved):
LGTM

PR #5371, opened 2026-08-18
feat(llm): register DeepSeek V4 Flash 0731

m199369309:
## Summary

- register `DeepSeek-V4-Flash-0731` as a separate built-in model from the existing preview entry
- expose the native FP8 checkpoint from Hugging Face and ModelScope with its 1M-token context metadata
- reuse the existing DeepSeek V4 chat/tool integration and document launch, trust-remote-code, reasoning, and opt-in DSpark settings
- add coverage for model matching, hub selection, runtime dependency selection, and preview-entry compatibility

## Tests

- `pytest -q xinference/model/llm/tests/test_llm_family.py -k 'match_deepseek_v4_flash_0731 or match_llm'`
- `pre-commit run --files xinference/model/llm/llm_family.json xinference/model/llm/tests/test_llm_family.py doc/source/models/builtin/llm/index.rst doc/source/models/builtin/llm/deepseek-v4-flash-0731.rst`

PR #5373, opened 2026-08-18
fix(vllm): use block size 256 for DeepSeek V4

m199369309:
Depends on #5371 for the `DeepSeek-V4-Flash-0731` built-in registration.

## Summary

- default vLLM `block_size` to `256` for the `DeepseekV4ForCausalLM` architecture
- preserve any explicit user-provided `block_size`
- keep the existing default of `16` for other model architectures
- use architecture detection so aliases and future DeepSeek V4 entries receive the same compatibility default

## Tests

- `pytest -q xinference/model/llm/vllm/tests/test_speculative_config.py`
- `pre-commit run --files xinference/model/llm/vllm/core.py xinference/model/llm/vllm/tests/test_speculative_config.py`

### DIFF
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -1111,7 +1111,11 @@ class VLLMModel(LLM):
             model_config.setdefault("master_addr", master_addr)  # type: ignore
             model_config.setdefault("master_port", get_next_port())  # type: ignore
         is_deepseek_v4 = "DeepseekV4ForCausalLM" in architectures
-        model_config.setdefault("block_size", 256 if is_deepseek_v4 else 16)
+        if is_deepseek_v4:
+            default_block_size = 128 if is_npu_available() else 256
+        else:
+            default_block_size = 16
+        model_config.setdefault("block_size", default_block_size)
         if VLLM_VERSION < version.parse("0.18.0"):
             model_config.setdefault("swap_space", 4)
         model_config.setdefault("gpu_memory_utilization", 0.90)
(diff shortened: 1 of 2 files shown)


xinference/model/llm/vllm/core.py
```diff
@@ -1110,7 +1110,8 @@ def _sanitize_model_config(
                 master_addr = self._address
             model_config.setdefault("master_addr", master_addr)  # type: ignore
             model_config.setdefault("master_port", get_next_port())  # type: ignore
-        model_config.setdefault("block_size", 16)
+        is_deepseek_v4 = "DeepseekV4ForCausalLM" in architectures
+        model_config.setdefault("block_size", 256 if is_deepseek_v4 else 16)
```
qinxuye: This default also applies on Ascend NPU, where Xinference explicitly supports DeepseekV4ForCausalLM. vLLM-Ascend's DeepSeek V4 cache implementation only defines block sizes 32, 64, and 128, and its deployment recipes use 128 (or 32 for some A3 scenarios), so passing 256 can fail during engine initialization. Please use an NPU-specific default such as 128 if is_npu_available() else 256, while continuing to preserve an explicit user-provided value, and add an NPU-path regression test.

m199369309: Fixed in c3de3eee. DeepSeek V4 now defaults to block size 128 on Ascend NPU and 256 on other platforms, while explicit user-provided values are still preserved via setdefault. Added regression coverage for NPU/non-NPU defaults, explicit values on both paths, and regular-model behavior. Focused tests: 20 passed; pre-commit checks passed.

qinxuye (approved):
LGTM

PR #5374, opened 2026-08-18
fix(vllm): handle engine death during async iteration

m199369309:
## Summary

- handle both the legacy `AsyncEngineDeadError` and the vLLM V1 `EngineDeadError`
- guard exceptions raised while consuming async generators, not only while creating them
- stop and terminate the broken model process before a vLLM-specific exception can cross the xoscar boundary
- close wrapped async iterators when streaming is cancelled, allowing vLLM request cleanup to run
- avoid treating every `RuntimeError` as an engine death when a concrete vLLM exception class is available

## Tests

- `pytest -q xinference/model/llm/vllm/tests/test_utils.py`
- `pre-commit run --files xinference/model/llm/vllm/utils.py xinference/model/llm/vllm/tests/test_utils.py`

### DIFF
--- a/xinference/model/llm/vllm/utils.py
+++ b/xinference/model/llm/vllm/utils.py
@@ -60,9 +60,13 @@ def _stop_after_engine_death(model: Any) -> None:
     os._exit(1)
 
 
-async def _guard_async_iterator(model: Any, iterator: Any) -> AsyncIterator[Any]:
+async def _guard_async_iterator(model: Any, iterable: Any) -> AsyncIterator[Any]:
     completed = False
+    iterator = None
     try:
+        # AsyncIterable objects do not have to implement __anext__ themselves;
+        # close the concrete iterator returned by __aiter__ on cancellation.
+        iterator = aiter(iterable)
         async for item in iterator:
             yield item
         completed = True
@@ -70,10 +74,10 @@ async def _guard_async_iterator(model: Any, iterator: Any) -> AsyncIterator[Any]
         _stop_after_engine_death(model)
     finally:
         # Closing the wrapper (for example after a client disconnects) must
-        # also close the wrapped async generator so that vLLM can release the
+        # also close the wrapped async iterator so that vLLM can release the
         # request and run its cleanup handlers.  Avoid an unnecessary aclose
         # call after normal exhaustion.
-        if not completed:
+        if not completed and iterator is not None:
             aclose = getattr(iterator, "aclose", None)
             if aclose is not None:
                 try:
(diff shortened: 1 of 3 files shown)


xinference/model/llm/vllm/utils.py
```diff
@@ -15,35 +15,90 @@
...
-            # Just kill the process and let xinference auto-recover the model
-            os._exit(1)
-
-    return _async_wrapper
+            result = await fn(self, *args, **kwargs)
+            # async_generate/async_chat return an async generator.  Exceptions
+            # from `async for` happen after the decorated coroutine has already
+            # returned, so the outer try/except cannot see them unless the
+            # iterator itself is wrapped.
+            if hasattr(result, "__aiter__") and hasattr(result, "__anext__"):
```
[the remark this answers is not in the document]
m199369309: Fixed in 7d1d0d0c. The decorator now guards every object implementing `__aiter__`, and `_guard_async_iterator` obtains and closes the concrete iterator returned by `aiter()` so cancellation cleanup also works for `AsyncIterable` wrappers that do not implement `__anext__` or `aclose` themselves. Added regression tests for both engine-death handling and early-close cleanup on an `AsyncIterable`-only wrapper.

Validation:
- `pytest -q xinference/model/llm/vllm/tests/test_utils.py` (6 passed)
- `pre-commit run --files xinference/model/llm/vllm/utils.py xinference/model/llm/vllm/tests/test_utils.py` (all passed)

xinference/model/llm/vllm/utils.py
```diff
@@ -15,35 +15,94 @@
...
+    finally:
+        # Closing the wrapper (for example after a client disconnects) must
+        # also close the wrapped async iterator so that vLLM can release the
+        # request and run its cleanup handlers.  Avoid an unnecessary aclose
+        # call after normal exhaustion.
+        if not completed and iterator is not None:
+            aclose = getattr(iterator, "aclose", None)
+            if aclose is not None:
+                try:
+                    await aclose()
```
qinxuye, line 84: This closes only the directly wrapped iterator. In the real streaming path, `@vllm_check` wraps the `async_chat` conversion generator, which itself consumes the separately guarded `async_generate` iterator. Closing the outer returned stream closes the conversion generator, but Python `async for` does not close that inner iterator, so the vLLM request cleanup runs only after garbage collection. I reproduced this with the same nested decorator chain: immediately after `await stream.aclose()`, the engine stream cleanup count is still 0 and becomes 1 only after forced GC. Please add a nested `async_chat -> async_generate -> engine stream` regression test and explicitly close the upstream iterator when the conversion layer terminates early.

m199369309: Fixed in f70fb712. The chat/tool conversion generators now retain and explicitly close the original upstream model stream in `finally`, so closing the outer `async_chat` stream deterministically propagates cleanup through the nested `async_generate` guard (including when a reasoning-parser wrapper is present). I also added a nested `async_chat -> async_generate -> engine stream` regression test that verifies the engine stream cleanup runs immediately after `await stream.aclose()` without forced GC.

qinxuye (approved):
LGTM

PR #5375, opened 2026-08-18
feat(router): add token-aware virtual model routing

m199369309:
## Summary
- add a native `xinference-router` data-plane process with process-isolated tokenization, admission control, typed first-match routing rules, and HTTP/SSE proxying
- add V1 short/long compatibility plus V2 dynamic 1-16 backend configuration for the `llm_chat` route profile
- add Supervisor-backed Router configuration/runtime registries, virtual model discovery, OpenAI chat dispatch, internal runtime registration, heartbeat, config polling, and acknowledgement APIs
- protect existing deployments with the opt-in `XINFERENCE_TOKEN_ROUTER_ENABLED` feature gate (disabled by default)
- add Router management scopes, audit coverage, API/runtime/configuration tests, and CLI packaging

## Validation
- focused Router suite: 98 passed

## Follow-ups
Tokenizer asset management, bundled assets, Web UI, Agent orchestration, and monitoring are intentionally split into dependent PRs.

### DIFF
--- a/xinference/api/restful_api.py
+++ b/xinference/api/restful_api.py
@@ -24,8 +24,9 @@ import time
 import uuid
 import warnings
 from pathlib import Path
-from typing import Any, List, Optional, Union, get_type_hints
+from typing import Any, AsyncIterator, Dict, List, Optional, Union, get_type_hints
 
+import httpx
 import xoscar as xo
 from aioprometheus import REGISTRY, MetricsMiddleware
 from aioprometheus.asgi.starlette import metrics
@@ -43,7 +44,7 @@ from fastapi import (
 from fastapi.middleware.cors import CORSMiddleware
 from PIL import Image
 from sse_starlette.sse import EventSourceResponse
-from starlette.responses import PlainTextResponse
+from starlette.responses import PlainTextResponse, StreamingResponse
 from uvicorn import Config, Server
 from xoscar.utils import get_next_port
 
@@ -59,6 +60,7 @@ from ..constants import (
     XINFERENCE_LAUNCH_HISTORY_DB_PATH,
     XINFERENCE_MONITOR_CONFIG_DB_PATH,
     XINFERENCE_SSE_PING_ATTEMPTS_SECONDS,
+    XINFERENCE_TOKEN_ROUTER_ENABLED,
     get_auth_encryption_key,
     get_auth_jwt_secret_key,
     is_auth_advanced,
@@ -104,6 +106,46 @@ from .schemas import (
 from .utils import require_model
 
 logger = logging.getLogger(__name__)
+
+_TOKEN_ROUTER_HOP_BY_HOP_HEADERS = {
+    "connection",
+    "keep-alive",
+    "proxy-authenticate",
+    "proxy-authorization",
+    "t" "e",
+    "trailer",
+    "transfer-encoding",
+    "upgrade",
+}
+_TOKEN_ROUTER_REQUEST_HEADERS = {
+    "accept",
+    "authorization",
+    "content-type",
+    "traceparent",
+    "tracestate",
+    "user-agent",
+    "x-request-id",
+}
+
+
+def _token_router_request_headers(request: Request, request_id: str) -> Dict[str, str]:
+    headers = {
+        key.lower(): value
+        for key, value in request.headers.items()
+        if key.lower() in _TOKEN_ROUTER_REQUEST_HEADERS
+    }
+    headers["content-type"] = "application/json"
+    headers["x-request-id"] = request_id
+    return headers
+
+
+def _token_router_response_headers(response: httpx.Response) -> Dict[str, str]:
+    excluded = _TOKEN_ROUTER_HOP_BY_HOP_HEADERS | {"content-length"}
+    return {
+        key: value
+        for key, value in response.headers.items()
+        if key.lower() not in excluded
+    }
 
 
 _AUDIO_RESPONSE_MEDIA_TYPES = {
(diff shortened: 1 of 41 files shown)


xinference/router/app.py
```diff
@@ -0,0 +1,439 @@
...
+                        body_stream(),
+                        status_code=backend_response.status_code,
+                        media_type=backend_response.headers.get(
+                            "content-type", "text/event-stream"
+                        ),
+                        headers={
+                            **response_headers(backend_response),
+                            **router_headers,
+                        },
+                    )
```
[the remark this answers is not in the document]
m199369309: Fixed in 9795c824. The streaming response now uses a Starlette BackgroundTask as an early-disconnect fallback while retaining generator-finally cleanup for stream-time failures. Both paths call the same idempotent, shielded cleanup task, and nested finally blocks guarantee that the backend response, capacity gate, and runtime snapshot are all released. Added a regression test that disconnects before the response body generator starts and verifies each resource is released exactly once.

xinference/api/restful_api.py
```diff
@@ -508,6 +550,93 @@ async def _report_error_event(self, model_uid: Optional[str], content: str) -> N
...
+    async def _proxy_token_router_chat_completion(
+        self,
+        request: Request,
+        raw_body: Dict[str, Any],
+        runtime: Dict[str, Any],
+    ) -> Response:
+        request_id = request.headers.get("x-request-id") or f"xinf-{uuid.uuid4()}"
+        endpoint = runtime["endpoint"]
+        upstream_url = f"{endpoint}/v1/chat/completions"
+        client = httpx.AsyncClient(timeout=None, follow_redirects=False)
```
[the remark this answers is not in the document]
m199369309: Fixed in 9795c824. RESTfulAPI now owns a shared httpx.AsyncClient for token-router proxy requests, reuses its connection pool across requests, and closes it through the FastAPI shutdown handler. Per-response aclose() behavior is preserved. Added tests covering client reuse and connection-error behavior.

xinference/api/restful_api.py
```diff
@@ -105,6 +107,46 @@
 
 logger = logging.getLogger(__name__)
 
+_TOKEN_ROUTER_HOP_BY_HOP_HEADERS = {
+    "connection",
+    "keep-alive",
+    "proxy-authenticate",
+    "proxy-authorization",
+    "t" "e",
```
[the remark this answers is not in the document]
m199369309: Fixed in 9795c824. Replaced the implicitly concatenated literal with the explicit HTTP hop-by-hop header name "te". The valid header token was also added to the codespell allowlist so the clear spelling passes CI.

xinference/router/tokenizer.py
```diff
@@ -0,0 +1,158 @@
...
+        raise TokenizationError(f"Encoding module has no encode_messages(): {path}")
+    return module
+
+
+def _normalize_content(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    normalized: list[dict[str, Any]] = []
+    for original in messages:
+        if not isinstance(original, dict):
+            raise TokenizationError("Each chat message must be an object")
+        message = copy.deepcopy(original)
```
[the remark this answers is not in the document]
m199369309: Fixed in 9795c824. Replaced copy.deepcopy(original) with dict(original), since normalization only replaces the top-level content field. Added a regression test verifying that normalization does not mutate the caller's input messages.

xinference/router/tokenizer.py
```diff
@@ -0,0 +1,157 @@
...
+            raise TokenizationError("messages must be a non-empty list")
+        messages = _normalize_content(raw_messages)
+        tools = payload.get("tools") or []
+        if tools:
+            if not isinstance(tools, list):
+                raise TokenizationError("tools must be a list")
+            messages = _attach_tools(messages, tools)
+
+        template_kwargs = _chat_template_kwargs(payload)
+        enable_thinking = bool(template_kwargs.get("enable_thinking", False))
```
qinxuye, line 131: This does not mirror the existing Chat API normalization. `create_chat_completion` accepts top-level `enable_thinking` and `extra_body.enable_thinking`, but the raw payload is forwarded here unchanged, so both forms are currently classified as non-thinking. The existing API also lets `max_completion_tokens` override `max_tokens`, while this estimator uses the reverse precedence. This can undercount the budget and route a thinking/long request to the wrong backend. Please normalize the same input forms with the same precedence and add regression coverage for the Supervisor dispatch path.

xinference/api/restful_api.py
```diff
@@ -508,6 +552,103 @@ async def _report_error_event(self, model_uid: Optional[str], content: str) -> N
...
+                        runtime.get("router_uid"),
+                        runtime.get("instance_id"),
+                    )
+                    raise
+                finally:
+                    await upstream_response.aclose()
+
+            response_headers.setdefault("cache-control", "no-cache")
+            response_headers.setdefault("x-accel-buffering", "no")
+            return StreamingResponse(
```
qinxuye, line 711: Please apply the same idempotent `BackgroundTask` cleanup that was added in `router/app.py`. If the downstream client disconnects before this response body generator starts, its `finally` block is never entered and the shared httpx response remains open. From the Router's perspective the Supervisor is still connected, so the backend response, capacity gate, and runtime snapshot can all remain held. Add a regression test that disconnects before body iteration and verifies `aclose()` runs exactly once.

xinference/router/tokenization_worker.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def initialize_tokenization_worker(
+    tokenizer_path: str,
+    reserve_tokens: int,
+    default_output_tokens: int,
+) -> None:
+    """Initialize one process-local estimator and remove backend credentials."""
+    global _ESTIMATOR
+    os.environ.pop("XINFERENCE_API_KEY", None)
```
qinxuye, line 20: Control-plane mode normally obtains its credential from `XINFERENCE_TOKEN_ROUTER_INTERNAL_TOKEN`, and spawned tokenizer workers inherit that environment variable. Because `encoding_dsv4.py` is dynamically executed inside this worker, the process-isolation boundary still exposes a credential that authorizes every internal Token Router endpoint. Please remove the internal token before loading the estimator as well, and extend the worker-isolation test to verify it is absent.

xinference/api/restful_api.py
```diff
@@ -781,7 +922,10 @@ async def build_llm_registration_from_config(
 
     async def list_models(self) -> JSONResponse:
         try:
-            models = await (await self._get_supervisor_ref()).list_models()
+            supervisor_ref = await self._get_supervisor_ref()
+            physical_models = await supervisor_ref.list_models()
+            virtual_models = await supervisor_ref.list_virtual_models()
```
qinxuye: This merge also needs to honor `XINFERENCE_TOKEN_ROUTER_ENABLED`. The Router store is persisted even when the feature is later disabled, so restarting with the flag set to false after prior use still advertises these virtual models from `/v1/models`; meanwhile `create_chat_completion` skips Router resolution and treats them as physical models. That leaves advertised but unusable model IDs. Please gate virtual-model discovery here and cover the disable-after-restart case.

xinference/api/tests/test_token_router_audit.py
```diff
@@ -0,0 +1,15 @@
...
+
+def test_internal_token_router_endpoints_skip_audit():
+    assert should_skip_audit("/v1/internal/token-router/instances/register") is True
+    assert (
+        should_skip_audit("/v1/internal/token-router/instances/router-1/config-ack")
+        is True
+    )
+
+
+def test_token_router_management_endpoints_are_admin_audited():
```
qinxuye, line 19: This only verifies the classification helper, not actual audit recording. The current audit middleware invokes `classify_endpoint()` only for paths beginning with `/v1/models` or `/v1/admin`, so `/v1/token_routers/...` never reaches `_record_admin_audit()` despite this test passing. Please extend the middleware prefix and add an ASGI-level regression test that verifies a management request is recorded.

qinxuye (commented):
Found several blocking correctness and lifecycle issues in the current head.

xinference/router/tokenizer.py
m199369309: Addressed in 3cdc7ee3. The Supervisor now normalizes top-level and extra_body enable_thinking before proxying, max_completion_tokens takes precedence over max_tokens, and the estimator independently mirrors the same semantics. Added estimator and Supervisor dispatch regression coverage.

xinference/api/restful_api.py
m199369309: Addressed in 3cdc7ee3. The Supervisor streaming proxy now uses an idempotent cleanup task from both the generator finally block and a BackgroundTask. Added a regression test that never iterates the body and verifies the upstream response is closed exactly once.

xinference/router/tokenization_worker.py
m199369309: Addressed in 3cdc7ee3. Tokenization workers now remove XINFERENCE_TOKEN_ROUTER_INTERNAL_TOKEN before loading the estimator, and worker startup isolation checks cover both Router credentials.

xinference/api/restful_api.py
m199369309: Addressed in 3cdc7ee3. /v1/models only loads and merges persisted virtual models while XINFERENCE_TOKEN_ROUTER_ENABLED is enabled. Added coverage verifying disabled Router mode exposes physical models only and does not query persisted virtual models.

xinference/api/tests/test_token_router_audit.py
m199369309: Addressed in 3cdc7ee3. Token Router management paths are now included in the post-response admin audit middleware. Added an ASGI-level regression test for the final error status and verified internal Token Router endpoints remain excluded.

qinxuye (approved):
LGTM

PR #5376, opened 2026-08-18
feat(router): add tokenizer asset registry

m199369309:
## Summary
- add a managed Tokenizer asset registry for Token-aware Router deployments
- validate manifests, required files, checksums, path traversal and symlink safety
- support external asset roots through explicit configuration and isolated smoke tests
- add focused registry tests

Depends on #5375

## Validation
- `pytest -q xinference/core/tests/test_tokenizer_asset_registry.py`
- `pre-commit run --files xinference/constants.py xinference/core/tokenizer_asset_registry.py xinference/core/tests/test_tokenizer_asset_registry.py`

### DIFF
--- a/xinference/core/tokenizer_asset_registry.py
+++ b/xinference/core/tokenizer_asset_registry.py
@@ -1,0 +1,556 @@
+# Copyright 2022-2026 Xinference Holdings Pte. Ltd
+"""Registered Tokenizer assets for Token-aware Routers."""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import logging
+import multiprocessing
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+import yaml  # type: ignore[import-untyped]
+
+from ..constants import XINFERENCE_TOKENIZER_ASSET_CONFIG
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_REQUIRED_FILES = (
+    "tokenizer.json",
+    "encoding/encoding_dsv4.py",
+)
+_MAX_MANIFEST_BYTES = 1024 * 1024
+_MAX_TOKENIZER_BYTES = 4 * 1024 * 1024 * 1024
+_MAX_ENCODING_BYTES = 16 * 1024 * 1024
+_SMOKE_TEST_TIMEOUT_SECONDS = 30
+
+
+def _execute_smoke_test(path: Path, capabilities: Dict[str, bool]) -> Dict[str, str]:
+    from ..router.tokenizer import DeepSeekV4TokenEstimator
+
+    checks: Dict[str, str] = {}
+    estimator = DeepSeekV4TokenEstimator(
+        path, reserve_tokens=64, default_output_tokens=512
+    )
+    payloads: Dict[str, Dict[str, Any]] = {
+        "chat_smoke_test": {
+            "messages": [{"role": "user", "content": "hello"}],
+            "max_tokens": 8,
+        }
+    }
+    if capabilities.get("tools", False):
+        payloads["tools_smoke_test"] = {
+            "messages": [{"role": "user", "content": "hello"}],
+            "tools": [
+                {
+                    "type": "function",
+                    "function": {
+                        "name": "ping",
+                        "description": "ping",
+                        "parameters": {"type": "object", "properties": {}},
+                    },
+                }
+            ],
+            "max_tokens": 8,
+        }
+    if capabilities.get("thinking", False):
+        payloads["thinking_smoke_test"] = {
+            "messages": [{"role": "user", "content": "hello"}],
+            "chat_template_kwargs": {"enable_thinking": True},
+            "max_tokens": 8,
+        }
+
+    for name, payload in payloads.items():
+        first = estimator.estimate(payload)
+        second = estimator.estimate(payload)
+        if first.total_tokens < 1:
+            raise TokenizerAssetError(f"{name} returned an invalid Token budget")
+        if first != second:
+            raise TokenizerAssetError(f"{name} returned an unstable Token budget")
+        checks[name] = "ok"
+    return checks
+
+
+def _smoke_test_worker(
+    path: str, capabilities: Dict[str, bool], connection: Any
+) -> None:
+    try:
+        connection.send(("ok", _execute_smoke_test(Path(path), capabilities)))
+    except BaseException as exc:  # noqa: BLE001 - isolate untrusted asset code.
+        connection.send(("error", f"{type(exc).__name__}: {exc}"))
+    finally:
+        connection.close()
+
+
+class TokenizerAssetError(ValueError):
+    """Raised when a registered Tokenizer asset cannot be used safely."""
+
+
+class TokenizerAssetRegistry:
+    """Load and validate the Supervisor-side Tokenizer asset whitelist."""
+
+    def __init__(self, config_path: Optional[str] = None) -> None:
+        configured_path = config_path
+        if configured_path is None:
+            configured_path = XINFERENCE_TOKENIZER_ASSET_CONFIG
+        self._config_path = (
+            Path(configured_path).expanduser() if configured_path else None
+        )
+        self._asset_roots: List[Path] = []
+        self._entries: Dict[str, Dict[str, Any]] = {}
+        # Preserve the pre-registry API when no production registry is configured.
+        self._allow_custom_path = True
+        self._config_error = ""
+        self._last_validated_at: Dict[str, str] = {}
+        self.reload()
+
+    @property
+    def allow_custom_path(self) -> bool:
+        return self._allow_custom_path
+
+    @property
+    def config_error(self) -> str:
+        return self._config_error
+
+    def reload(self) -> None:
+        self._asset_roots = []
+        self._entries = {}
+        self._allow_custom_path = True
+        self._config_error = ""
+        if self._config_path is None:
+            return
+        if not self._config_path.is_file():
+            self._allow_custom_path = False
+            self._config_error = "Tokenizer asset config does not exist"
+            logger.error("%s: %s", self._config_error, self._config_path)
+            return
+
+        try:
+            raw = yaml.safe_load(self._config_path.read_text(encoding="utf-8")) or {}
+            if not isinstance(raw, dict):
+                raise TokenizerAssetError("Tokenizer asset config must be an object")
+            if int(raw.get("schema_version", 1)) != 1:
+                raise TokenizerAssetError("Unsupported Tokenizer asset schema_version")
+
+            roots = raw.get("asset_roots", [])
+            if not isinstance(roots, list):
+                raise TokenizerAssetError("asset_roots must be a list")
+            self._asset_roots = [
+                Path(str(root)).expanduser().resolve() for root in roots
+            ]
+            self._allow_custom_path = bool(raw.get("allow_custom_path", False))
+
+            assets = raw.get("assets", [])
+            if not isinstance(assets, list):
+                raise TokenizerAssetError("assets must be a list")
+            for entry in assets:
+                if not isinstance(entry, dict):
+                    raise TokenizerAssetError("Each Tokenizer asset must be an object")
+                asset_id = str(entry.get("asset_id", "")).strip()
+                if not asset_id:
+                    raise TokenizerAssetError("Tokenizer asset_id is required")
+                if asset_id in self._entries:
+                    raise TokenizerAssetError(
+                        f"Duplicate Tokenizer asset_id: {asset_id}"
+                    )
+                self._entries[asset_id] = dict(entry)
+        except Exception as exc:
+            self._config_error = str(exc)
+            self._asset_roots = []
+            self._entries = {}
+            self._allow_custom_path = False
+            logger.error(
+                "Failed to load Tokenizer asset config %s: %s",
+                self._config_path,
+                exc,
+            )
+
+    def _entry_path(self, entry: Dict[str, Any]) -> Path:
+        raw_path = str(entry.get("path", "")).strip()
+        if not raw_path:
+            raise TokenizerAssetError("Tokenizer asset path is required")
+        path = Path(raw_path).expanduser()
+        if not path.is_absolute():
+            if not self._asset_roots:
+                raise TokenizerAssetError(
+                    "Relative Tokenizer asset path requires asset_roots"
+                )
+            path = self._asset_roots[0] / path
+        return path.resolve()
+
+    def _ensure_allowed_root(self, path: Path) -> None:
+        if not self._asset_roots:
+            return
+        if not any(
+            path == root or path.is_relative_to(root) for root in self._asset_roots
+        ):
+            raise TokenizerAssetError(
+                "Tokenizer asset path escapes configured asset_roots"
+            )
+
+    @staticmethod
+    def _read_manifest(path: Path) -> Dict[str, Any]:
+        manifest_path = path / "asset.json"
+        if not manifest_path.is_file():
+            raise TokenizerAssetError("Missing Tokenizer asset manifest")
+        resolved_manifest_path = manifest_path.resolve()
+        if (
+            resolved_manifest_path != path
+            and not resolved_manifest_path.is_relative_to(path)
+        ):
+            raise TokenizerAssetError(
+                "Tokenizer asset manifest escapes asset directory"
+            )
+        if resolved_manifest_path.stat().st_size > _MAX_MANIFEST_BYTES:
+            raise TokenizerAssetError("Tokenizer asset manifest is too large")
+        try:
+            manifest = json.loads(resolved_manifest_path.read_text(encoding="utf-8"))
+        except OSError as exc:
+            raise TokenizerAssetError(
+                f"Cannot read Tokenizer asset manifest: {exc.strerror or exc}"
+            ) from exc
+        except json.JSONDecodeError as exc:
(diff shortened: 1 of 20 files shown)


xinference/router/control_plane.py
```diff
@@ -0,0 +1,253 @@
...
+                    )
+                except Exception:
+                    logger.exception(
+                        "Token Router control-plane synchronization failed"
+                    )
+
+            try:
+                await asyncio.wait_for(stop.wait(), timeout=poll_interval_seconds)
+            except TimeoutError:
+                pass
```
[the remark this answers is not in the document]
m199369309: Fixed upstream in c2e6a605 and synchronized into this PR via ed81953d. The control-plane poll loop now catches asyncio.TimeoutError, which is required on supported Python 3.10 where it is distinct from the built-in TimeoutError. The control-plane test suite passes after the dependency sync.

xinference/router/admission.py
```diff
@@ -0,0 +1,79 @@
...
+            try:
+                await asyncio.wait_for(
+                    self._condition.wait_for(lambda: self._active < self.max_active),
+                    timeout=self.queue_timeout_seconds,
+                )
+                self._active += 1
+            except TimeoutError as exc:
+                raise AdmissionRejected(
+                    self.pool, "queue_timeout", self.retry_after_seconds
+                ) from exc
```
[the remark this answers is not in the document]
m199369309: Fixed upstream in c2e6a605 and synchronized into this PR via ed81953d. CapacityGate now catches asyncio.TimeoutError and consistently converts queue timeouts into AdmissionRejected on Python 3.10. Verified with Python 3.10.16 and the admission/control-plane tests.

xinference/api/restful_api.py
```diff
@@ -508,6 +550,93 @@ async def _report_error_event(self, model_uid: Optional[str], content: str) -> N
...
+    async def _proxy_token_router_chat_completion(
+        self,
+        request: Request,
+        raw_body: Dict[str, Any],
+        runtime: Dict[str, Any],
+    ) -> Response:
+        request_id = request.headers.get("x-request-id") or f"xinf-{uuid.uuid4()}"
+        endpoint = runtime["endpoint"]
+        upstream_url = f"{endpoint}/v1/chat/completions"
+        client = httpx.AsyncClient(timeout=None, follow_redirects=False)
```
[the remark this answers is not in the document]
m199369309: Fixed in the dependency PR by 9795c824 and synchronized into this PR via ed81953d. RESTfulAPI now reuses a shared httpx.AsyncClient, closes it on application shutdown, and only closes each upstream response in the request lifecycle. Reuse and connection-error tests are included.

xinference/router/tokenizer.py
```diff
@@ -0,0 +1,158 @@
...
+        raise TokenizationError(f"Encoding module has no encode_messages(): {path}")
+    return module
+
+
+def _normalize_content(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    normalized: list[dict[str, Any]] = []
+    for original in messages:
+        if not isinstance(original, dict):
+            raise TokenizationError("Each chat message must be an object")
+        message = copy.deepcopy(original)
```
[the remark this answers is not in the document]
m199369309: Fixed in the dependency PR by 9795c824 and synchronized into this PR via ed81953d. Message normalization now uses dict(original) instead of copy.deepcopy(original), with a regression test confirming the caller input is not mutated.

xinference/core/tokenizer_asset_registry.py
```diff
@@ -0,0 +1,556 @@
...
+            raise TokenizerAssetError(f"{name} returned an unstable Token budget")
+        checks[name] = "ok"
+    return checks
+
+
+def _smoke_test_worker(
+    path: str, capabilities: Dict[str, bool], connection: Any
+) -> None:
+    try:
+        connection.send(("ok", _execute_smoke_test(Path(path), capabilities)))
```
qinxuye: **[P1] Do not execute asset Python with Supervisor privileges**

This calls `DeepSeekV4TokenEstimator`, which imports and executes the asset's `encoding/encoding_dsv4.py`. A spawned daemon process is not a security sandbox: it still inherits the Supervisor's environment, filesystem identity, and network access. Unlike the Router tokenization worker, this child does not even remove the API/internal tokens, and the checksum is declared by the same asset manifest rather than an independent trust root. A compromised asset can therefore read control-plane secrets or modify Supervisor-owned state during create/validate. Please run this in a genuinely least-privilege sandbox (or remove executable asset code from Supervisor validation), with secrets, filesystem/network access, and the trust anchor isolated from the asset.

qinxuye (commented):
Three findings from the current-head review:

xinference/router/runtime.py
```diff
@@ -170,6 +170,11 @@ async def summary(self) -> Dict[str, Any]:
             "active_requests": snapshot.active_requests,
             "pools": pools,
             "tokenization": tokenization,
+            "tokenizer_asset": {
+                "asset_id": snapshot.config.tokenizer_asset_id,
+                "revision": snapshot.config.tokenizer_asset_revision,
+                "fingerprint": snapshot.config.tokenizer_asset_fingerprint,
```
qinxuye: **[P1] Report the asset that was actually loaded**

These values are copied from the control-plane config, and `validate_token_router()` compares them with that same stored config. The Router never hashes the files loaded by `TokenizationService`, so a different asset at the same path (for example on another host/mount, or after a TOCTOU replacement) still reports the expected fingerprint and passes validation. Please compute the fingerprint/revision from the files actually loaded by the Router/tokenization worker and publish that measured value in the heartbeat; add a test where the configured fingerprint differs from the local asset contents.

xinference/core/tokenizer_asset_registry.py
```diff
@@ -0,0 +1,556 @@
...
+            supplied = str(Path(tokenizer_path).expanduser().resolve())
+            if supplied != resolved_path:
+                raise TokenizerAssetError(
+                    "tokenizer_asset_id and tokenizer_path resolve to different directories"
+                )
+        return {
+            "tokenizer_asset_id": asset_id,
+            "tokenizer_path": resolved_path,
+            "tokenizer_asset_revision": str(asset.get("revision", "")),
+            "tokenizer_asset_fingerprint": str(asset.get("fingerprint", "")),
```
qinxuye, line 421: **[P2] Enforce the declared asset capabilities at runtime**

`tools` and `thinking` currently only decide which smoke tests run. `resolve()` drops the capability map, `RouterConfig` has no corresponding field, and the request path still accepts tool/thinking payloads unconditionally. An asset declaring `tools: false` or `thinking: false` can therefore be deployed and used for an unsupported request, causing a tokenization failure or incorrect budget/routing. Please propagate the validated capabilities into the Router config and reject unsupported requests/configured rules, with regression coverage for both false flags.

xinference/core/tokenizer_asset_registry.py
m199369309: Fixed in 462cd8e2. The Supervisor no longer executes any asset Python.

- Removed `_execute_smoke_test` / `_smoke_test_worker` / `_smoke_test` and `_SMOKE_TEST_TIMEOUT_SECONDS` from `xinference/core/tokenizer_asset_registry.py`. `DeepSeekV4TokenEstimator` is never instantiated in the Supervisor or any child process it spawns.
- Validation is now purely static: manifest structure/field checks, SHA-256 checksum verification, path-safety checks (absolute paths, `..` traversal, escape from the asset directory), and file hashing for the fingerprint.
- `validate_path()` and `_inspect_entry()` keep the `smoke_test` parameter only for API compatibility; it is unused.
- Tests updated to assert the static results (`required_files`, `checksums`, `manifest`, `capabilities`) instead of smoke-test checks, including `test_registered_asset_list_resolve_and_validate` and `test_declared_capabilities_are_reported`.

xinference/router/runtime.py
m199369309: Fixed in 462cd8e2. The Router now measures the asset it actually loaded and publishes the measured values.

- New `xinference/router/tokenizer_asset.py`: `compute_tokenizer_asset_fingerprint()` hashes the files the worker uses (`tokenizer.json`, `encoding/encoding_dsv4.py`) with a path-safety check, and `read_tokenizer_asset_revision()` reads `asset.json` revision statically (JSON only, no code execution). The same aggregation function is shared with the registry so both sides compute identical fingerprints.
- `initialize_tokenization_worker()` now computes the fingerprint and revision before and after constructing `DeepSeekV4TokenEstimator`; if they differ (TOCTOU replacement during load) initialization raises. It also scrubs all three Router credential env vars.
- `ping_worker()` returns the measured fingerprint/revision; `TokenizationService.start()` verifies every worker loaded the same asset (single non-empty fingerprint and single revision) and exposes `asset_fingerprint` / `asset_revision`.
- The heartbeat `process.tokenizer_asset` and `RouterRuntime.summary()` now report the measured values from the snapshot instead of copying the control-plane config, and Supervisor validation compares against the measured values.
- Regression coverage: `test_start_measures_tokenizer_asset_fingerprint` and `test_validation_detects_loaded_fingerprint_mismatch` (configured fingerprint differs from local asset contents).

xinference/core/tokenizer_asset_registry.py
m199369309: Fixed in 462cd8e2. Validated capabilities now flow into the Router config and are enforced on both the request path and rule validation.

- `resolve()` returns `tokenizer_asset_capabilities` and `tokenizer_asset_files`; `RouterConfig` gains both fields and `config_from_control_plane()` reads them (manifest may declare capabilities as a dict or a list).
- Request path in `xinference/router/app.py`: tool requests are pre-checked before tokenization (400 `tools_not_allowed`), and thinking-mode requests are rejected after budget computation (400 `thinking_not_allowed`).
- Rule validation rejects rules that require tools/thinking when the asset does not support them, in both `_validate_config()` and Supervisor v2 validation in `validate_token_router()`.
- Custom paths keep the phase-1 behavior (full capabilities + default required files) for backward compatibility.
- Regression coverage: `test_tools_request_rejected_when_asset_lacks_tools_capability`, `test_thinking_request_rejected_when_asset_lacks_thinking_capability`, v2 rule-rejection cases, and `test_v2_validation_rejects_rules_asset_cannot_support`.

xinference/router/tokenization.py
```diff
@@ -94,13 +100,25 @@ async def start(self) -> None:
...
+                raise TokenizationWorkerUnavailable(
+                    "Tokenization workers loaded different Tokenizer assets"
+                )
+            if len(revisions) != 1:
+                raise TokenizationWorkerUnavailable(
+                    "Tokenization workers loaded different Tokenizer asset revisions"
+                )
+            worker_pids = {pid for pid, _, _, _, _ in results}
             if len(worker_pids) >= self._max_workers:
+                self._asset_fingerprint = fingerprints.pop()
```
qinxuye: **[P1] Fail closed when the measured asset differs from the configuration**

This only proves that the tokenization workers agree with one another. `TokenizationService` is never given the expected fingerprint/revision, so if every worker on this host loads the same wrong asset, startup succeeds, the runtime ACKs the revision and reports `ready`. `resolve_token_router_runtime()` also selects ready/ACKed instances without checking the measured heartbeat metadata; the manual validation endpoint therefore does not prevent traffic from reaching the mismatched tokenizer. I reproduced a configured `sha256:expected` plus heartbeat `sha256:swapped` resolving as `available=True`. Please compare the measured values with `RouterConfig` before the snapshot becomes ready (or exclude mismatched heartbeats in Supervisor runtime selection), and add a regression proving such an instance cannot be resolved/routed.

xinference/router/app.py
```diff
@@ -237,6 +266,15 @@ async def chat_completions(request: Request) -> Response:
                     tools_present=bool(payload.get("tools")),
                     stream=bool(payload.get("stream", False)),
                 )
+                if budget.enable_thinking and "thinking" not in capabilities:
```
qinxuye, line 297: **[P2] Reject unsupported thinking before invoking the asset tokenizer**

This capability check runs only after `estimate()`. An asset declaring `thinking: false` may fail while rendering the thinking payload, so control never reaches this branch and the client receives a generic `invalid_request_error` instead of `thinking_not_allowed`. The new `_payload_thinking()` helper is currently unused. Please perform equivalent normalized detection before tokenization (including the existing top-level / `extra_body` / JSON-string `chat_template_kwargs` precedence), and cover the case where the estimator would raise if it were invoked.

qinxuye (commented):
Two remaining findings after reviewing the fixes in 462cd8e2:

m199369309:
## CI failure: click 8.4 compatibility for cmdline tests (fixed in eb3a4d6a)

The failing `xinference/deploy/test/test_cmdline.py` tests are **not caused by this PR** — the same two failures appeared on earlier commits (e.g. `140df640`, run 32388623931) before the recent review fixes, and the file is untouched by the feature work.

**Root cause: dependency drift.** `pyproject.toml` pins `click<8.2.0`, but the CI optional deps (`diffusers` → newly released `huggingface-hub>=1.23`) resolve `click` to 8.4.2. Pip installs it and warns `xinference ... requires click<8.2.0, but you have click 8.4.2 which is incompatible`. click 8.4 removed `CliRunner(mix_stderr=...)` (replaced by `capture=sys`) and captures stdout/stderr separately by default.

Two failing tests:
- `test_cmdline_model_path_error`: `CliRunner(mix_stderr=False)` raises `TypeError` on click 8.4.
- `test_cmdline_of_custom_model`: `list_model_registrations` writes its table to **stderr** (`print(..., file=sys.stderr)`); click 8.1 default `mix_stderr=True` mixed stderr into `result.stdout`, click 8.4 does not.

**Fix (eb3a4d6a):** make the assertions version-agnostic without changing CLI behavior:
- `CliRunner(mix_stderr=False)` → `CliRunner()` (the test only inspects `exit_code`/`exc_info`).
- Assert on `result.output` instead of `result.stdout` — `result.output` is the combined stream on both click <8.2 (`mix_stderr=True` default) and click >=8.4 (`capture=sys`).

Verified with minimal repros on both click 8.1.8 and 8.4.2; pre-commit passes and the fast unit test in the file passes. New CI run (32428725269) is in progress on eb3a4d6a.

xinference/router/app.py
```diff
@@ -106,7 +124,15 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
             control_stop.set()
             try:
                 if control_task is not None:
-                    await control_task
+                    # The control-plane task may be blocked in an HTTP call or
+                    # while applying a new runtime snapshot. Setting the stop
+                    # event alone cannot interrupt either operation, so cancel
+                    # the task explicitly before waiting for shutdown.
+                    control_task.cancel()
```
qinxuye, line 142: **[P1] Clean up a replacement snapshot when shutdown cancels `apply()`**

This cancellation can land while `control_plane.run()` is awaiting `runtime.apply()`. `RouterRuntime.apply()` only catches `Exception` around `replacement.tokenization.start()`, but `asyncio.CancelledError` is a `BaseException`, so the replacement snapshot is abandoned without closing its HTTP client, tokenization service, or process pool; the subsequent `runtime.aclose()` only closes the current snapshot. I reproduced cancellation during replacement startup with all three replacement resources still open, which can make the shutdown hang on orphaned worker processes. Please make the pre-swap portion of `apply()` cancellation-safe (clean up on `CancelledError`/`BaseException`, then re-raise; use bounded/cancellation-safe executor cleanup as needed), and add a regression that blocks replacement startup, cancels `apply()`, and asserts the replacement snapshot is closed.

qinxuye (commented):
One additional finding in the shutdown follow-up:

xinference/router/app.py
qinxuye: Still not fully fixed on `cffb82c7e`. The cleanup `try` ends after `replacement.tokenization.start()`, while `async with self._lock` remains outside it. If `apply()` is cancelled while waiting for that lock, the replacement has already started but is neither installed nor closed. I reproduced this by holding `_lock`, starting `apply()`, waiting for replacement startup to finish, and then cancelling: the replacement snapshot, HTTP client, and tokenization service all remained open. The new regression only covers cancellation during `start()`. Please keep cleanup ownership through the pre-swap lock acquisition and add a regression that holds the lock before cancelling `apply()`.

xinference/router/runtime.py
```diff
@@ -129,17 +135,23 @@ async def release(self, snapshot: RuntimeSnapshot) -> None:
...
             raise
 
-        old: RuntimeSnapshot
-        async with self._lock:
-            old = self._current
-            old.draining = True
-            self._current = replacement
         self._notify_swap(replacement)
         if old.active_requests == 0:
             await old.close()
```
qinxuye: **[P1] Make post-swap cleanup cancellation-safe**

The new ownership guard fixes cancellation before the swap, but after the replacement becomes current, `apply()` awaits `old.close()` here. If shutdown cancels the control-plane task at this await, `RuntimeSnapshot.close()` has already set `closed=True` while its HTTP client and tokenization process pool may still be open. `runtime.aclose()` only visits the current replacement, and a retry on the old snapshot returns immediately because it is marked closed. I reproduced this with a blocking old client: after cancelling `apply()` and then calling `runtime.aclose()`, both the old client and tokenization service remained open. Please make post-swap old-snapshot cleanup cancellation-safe/retryable (for example, track it as a drain task before awaiting it) and add a regression that cancels while `old.close()` is blocked.

qinxuye (approved):
LGTM

PR #5380, opened 2026-08-18
feat(ui): add Router Agent and cluster management

m199369309:
## Summary

- add Token Router Agent/node management to the Web UI
- show managed deployment replicas, assignments, desired/observed generations, runtime state, and node resources
- add tokenizer asset inventory and asset-binding status views
- integrate Router Agent/runtime information into the cluster information page
- extend permissions, service types, navigation, and all supported locale resources

## Dependency

Depends on #5378
Depends on #5379

This PR intentionally targets `main`; maintainers can merge it after the dependencies above are merged.

## Validation

- `npm ci`
- `npm run lint` (passes with existing repository warnings only)
- `npm run build` (passes; `/token-router` exported successfully)
- focused pre-commit hooks for all changed frontend files


frontend/src/components/pages/cluster-information.tsx
```diff
@@ -129,122 +113,172 @@ export default function ClusterInfo() {
...
-    fetchClusterInfo();
+    void fetchClusterInfo();
     return () => {
-      if (timerRef.current) {
-        clearTimeout(timerRef.current);
-      }
+      if (timerRef.current) clearTimeout(timerRef.current);
     };
-  }, []);
+  }, [fetchClusterInfo]);
```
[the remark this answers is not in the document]
m199369309: Fixed in e2893258. The polling lifecycle now lives inside the effect with a local cancellation flag and a single timer. Cleanup prevents an in-flight request from scheduling another poll, and stale responses are ignored before updating state.

xinference/core/router_assignment_store.py
```diff
@@ -0,0 +1,270 @@
...
+                (
+                    observed_state,
+                    pid,
+                    instance_id,
+                    last_error,
+                    json.dumps(observed or {}, ensure_ascii=False),
+                    now,
+                    now,
+                    assignment_id,
+                ),
```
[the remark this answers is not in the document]
m199369309: Fixed in f68e0275 and propagated to this branch. Missing observed metadata now preserves the stored value, while an explicit empty object can still clear it. The agent client and orchestration layer also preserve the distinction between omitted and explicit observed values. Regression coverage verifies heartbeat metadata survives later status updates.

xinference/api/restful_api.py
```diff
@@ -508,6 +550,93 @@ async def _report_error_event(self, model_uid: Optional[str], content: str) -> N
...
+                [chunk async for chunk in upstream_response.aiter_raw()]
+            )
+        finally:
+            await upstream_response.aclose()
+            await client.aclose()
+        return Response(
+            content=response_body,
+            status_code=upstream_response.status_code,
+            headers=response_headers,
+        )
```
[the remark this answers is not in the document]
m199369309: Fixed by the parent-stack change 9795c824, now propagated to this PR via 0e6d6a70. RESTfulAPI reuses an instance-level AsyncClient, closes it on application shutdown, and only closes individual upstream responses per request. A regression test verifies reuse across requests.

xinference/router/agent/process_manager.py
```diff
@@ -0,0 +1,499 @@
...
+                ),
+            )
+            process.terminate()
+            try:
+                await asyncio.wait_for(
+                    process.wait(), timeout=self.drain_timeout_seconds
+                )
+            except asyncio.TimeoutError:
+                process.kill()
+                await process.wait()
```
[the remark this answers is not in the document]
m199369309: Fixed in f68e0275 and propagated to this branch. Both terminate() and kill() are guarded against OSError/ProcessLookupError races, with regression tests covering a process exiting immediately before either signal.

frontend/src/components/pages/token-router/index.tsx
```diff
@@ -150,199 +181,244 @@ export default function TokenRouterPage() {
...
-                    <Loader2 className="size-4 animate-spin" />
-                    {t('tokenRouter.loading')}
-                  </div>
-                </TableCell>
-              </TableRow>
-            ) : routers.length === 0 ? (
+      <RouterNodeSection
+        nodes={routerNodes}
+        initialLoading={initialLoading}
+        canOperate={auth.canOperateRouters}
```
qinxuye: [P1] Preserve the new management controls when advanced auth is disabled

`useMenuAuth()` returns no scopes for the `NO_AUTH` token, so this prop (and the asset-management prop below) is always false in the normal `auth_advanced=false` mode. `PermissionGuard` and the API deliberately allow anonymous access in that mode, but the newly added Router Agent management UI then exposes no node or asset operations. Please derive effective capabilities such as `!clusterUIConfig?.auth_advanced || auth.canOperateRouters` after the global config is ready, and apply the same rule to write-scoped controls.

frontend/src/components/pages/token-router/tokenizer-asset-section.tsx
```diff
@@ -0,0 +1,260 @@
...
+                                    t,
+                                    binding.desired_state,
+                                    binding.observed_state
+                                  )}
+                                </Badge>
+                              </div>
+                              {binding.last_error && (
+                                <div className="mt-1 text-destructive">{binding.last_error}</div>
+                              )}
+                              {canOperate && (
```
qinxuye: [P2] Gate binding mutations with `routers:write`

This `canOperate` block exposes the PATCH/DELETE binding actions, and the same prop also controls the bind action below. The backend requires `routers:write` for binding POST/PATCH/DELETE; only revalidation requires `routers:operate`. Consequently an operate-only user sees controls that fail with 403, while a write-only user cannot perform mutations the API permits. Please pass separate write and operate capabilities and gate each action with its actual backend scope.

frontend/src/components/pages/token-router/index.tsx
m199369309: Fixed in 766387c9. Router management capabilities now use the global auth configuration: controls stay hidden until configuration is ready, are fully enabled when advanced auth is disabled, and preserve distinct JWT scopes when advanced auth is enabled.

frontend/src/components/pages/token-router/tokenizer-asset-section.tsx
m199369309: Fixed in 766387c9. Tokenizer asset binding POST/PATCH/DELETE controls are now gated by routers:write, while revalidation remains gated by routers:operate.

frontend/src/components/pages/token-router/router-capabilities.mjs
```diff
@@ -0,0 +1,29 @@
...
+export const resolveRouterCapabilities = ({
+  globalReady,
+  authAdvanced,
+  canWriteRouters,
+  canOperateRouters,
+}) => {
+  if (!globalReady) {
+    return { canWriteRouters: false, canOperateRouters: false };
+  }
+  if (!authAdvanced) {
```
qinxuye: [P2] Fail closed when UI configuration is unavailable

`globalReady` only means both global requests have settled: `fetchGlobalAfterAuth()` sets it to true even when `/v1/cluster/ui_config` rejects, leaving `authAdvanced` undefined. This `!authAdvanced` branch then treats that failure as no-auth mode and enables both write and operate controls for advanced-auth users, so the page exposes actions that only return 401/403. Please grant all controls only when `authAdvanced === false`; keep them hidden while the value is undefined (or track successful config loading separately), and cover the undefined case in the capability test.

qinxuye (approved):

PR #5381, opened 2026-08-18
feat(monitoring): add Token Router observability

m199369309:
## Summary

- add Prometheus metrics for Token Router Runtime requests, routing decisions, backend latency/errors, concurrency pools, tokenization, and process resources
- export Supervisor snapshots for Router Agents, Assignments, Runtime health/config state, Tokenizer Asset Bindings, and logical Router availability
- add authenticated Prometheus HTTP service discovery for dynamically assigned Router Runtime ports
- add localized Token Router Grafana dashboards, synchronized alert rules, Alertmanager inhibition examples, and Filebeat Router log inputs
- expose Token Router monitoring metadata in Monitor Center and document the monitoring setup

## Scope

This PR is intentionally limited to Token Router observability. It does not include the separate generic model-replica lifecycle, placement, recovery, or model-load dashboard changes from the local customization history.

## Dependencies

Depends on #5375
Depends on #5379

## Validation

- `pytest` focused monitoring/router/API suite: 42 passed
- `npm ci`
- `npm run lint` (existing repository warnings only)
- `npm run build`
- `pre-commit run --files <all changed files>`
- Python `compileall` for changed runtime, API, metrics, and test modules

A broader combined async test invocation reproduced the repository's existing asyncio teardown hang after its completed tests had passed; the focused changed-code suite above completes successfully.

### DIFF
--- a/xinference/router/app.py
+++ b/xinference/router/app.py
@@ -321,8 +321,6 @@ def create_app(config: RouterConfig) -> FastAPI:
         backend_started: float | None = None
         backend_model_uid = ""
 
-        await metrics.request_started(metric_config.router_uid, current_pool)
-
         async def finish_request(
             result: str, pool: str, *, backend_result: str | None = None
         ) -> None:
@@ -375,6 +373,8 @@ def create_app(config: RouterConfig) -> FastAPI:
         metric_config = config
 
         try:
+            await metrics.request_started(metric_config.router_uid, current_pool)
+
             if not _authorized(request, config):
                 await finish_request("auth_rejected", current_pool)
                 logger.warning(
@@ -464,7 +464,7 @@ def create_app(config: RouterConfig) -> FastAPI:
                 )
             capabilities = config.tokenizer_asset_capabilities
             if bool(payload.get("tools")) and "tools" not in capabilities:
-                await metrics.increment("tools_not_allowed", "none")
+                await finish_request("tools_not_allowed", current_pool)
                 return _error(
                     400,
                     "Tool requests are not supported by this Tokenizer asset",
@@ -472,7 +472,7 @@ def create_app(config: RouterConfig) -> FastAPI:
                     headers={"x-request-id": request_id},
                 )
             if "thinking" not in capabilities and _payload_thinking(payload):
-                await metrics.increment("thinking_not_allowed", "none")
+                await finish_request("thinking_not_allowed", current_pool)
                 return _error(
                     400,
                     "Thinking-mode requests are not supported by this "
@@ -490,7 +490,7 @@ def create_app(config: RouterConfig) -> FastAPI:
                     stream=stream,
                 )
                 if budget.enable_thinking and "thinking" not in capabilities:
-                    await metrics.increment("thinking_not_allowed", "none")
+                    await finish_request("thinking_not_allowed", current_pool)
                     return _error(
                         400,
                         "Thinking-mode requests are not supported by this "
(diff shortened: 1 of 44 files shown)


xinference/api/restful_api.py
```diff
@@ -508,6 +550,93 @@ async def _report_error_event(self, model_uid: Optional[str], content: str) -> N
...
+        try:
+            upstream_request = client.build_request(
+                "POST",
+                upstream_url,
+                headers=_token_router_request_headers(request, request_id),
+                json=raw_body,
+            )
+            upstream_response = await client.send(upstream_request, stream=True)
+        except httpx.HTTPError as exc:
+            await client.aclose()
```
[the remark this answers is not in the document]
m199369309: Fixed in a6a134d4 by syncing the latest dependency PR #5379 changes. RESTfulAPI now reuses an instance-level httpx.AsyncClient and closes it during application shutdown, so cancellation during client.send no longer abandons a per-request client or connection pool. Individual upstream responses remain request-owned and are closed by the streaming/non-streaming paths. The existing regression coverage verifies shared-client reuse and lifecycle cleanup.

xinference/router/app.py
```diff
@@ -820,6 +941,8 @@ async def body_stream() -> AsyncIterator[bytes]:
         finally:
             if runtime_release_owned:
                 await runtime.release(snapshot)
+            if not stream_handed_off and not request_metric_finished:
+                await finish_request("router_error", current_pool)
```
qinxuye, line 945: [P2] Preserve policy-rejection outcomes instead of classifying them as router errors

The earlier `tools_not_allowed` and `thinking_not_allowed` return paths only increment the legacy counter and never call `finish_request()`, so this fallback records those expected policy rejections in `xinference_token_router_route_requests_total` as `result="router_error"`. The dashboard route-outcome/error breakdown will therefore be incorrect. Please route all three capability-rejection exits through `finish_request(...)` with their actual result and add an integration assertion on `/metrics`.

xinference/router/app.py
```diff
@@ -292,11 +314,44 @@ async def chat_completions(request: Request) -> Response:
...
         requested_model: str | None = None
         stream = False
+        metric_config = runtime.current.config
+        current_pool = "none"
+        request_metric_finished = False
+        stream_handed_off = False
+        backend_started: float | None = None
+        backend_model_uid = ""
+
+        await metrics.request_started(metric_config.router_uid, current_pool)
```
qinxuye: [P2] Balance in-flight accounting when runtime acquisition is cancelled

`request_started()` increments the in-flight gauge before `await runtime.acquire()`, but the cleanup `finally` is entered only after acquisition succeeds. If the request task is cancelled while waiting for the runtime lock, or acquisition raises anything other than `RouterDisabled`, the `pool="none"` series remains positive permanently. Please either start accounting after successful acquisition or wrap acquisition so every started request is finalized on all exit paths, with cancellation coverage.

qinxuye (commented):
Found two monitoring correctness issues that should be addressed before approval.

xinference/router/app.py
m199369309: Fixed in 3f794363. All three tokenizer capability-rejection exits now use finish_request() with tools_not_allowed or thinking_not_allowed. The integration tests assert the /metrics route result, legacy result, absence of router_error, and a zero in-flight gauge.

xinference/router/app.py
m199369309: Fixed in 3f794363. In-flight accounting now starts only after runtime.acquire() succeeds and after entering the snapshot cleanup scope. A cancellation test blocks on the real runtime lock, cancels the request, and verifies both the in-flight metric and active snapshot count remain balanced.

qinxuye (approved):

PR #5386, opened 2026-08-19
fix(venv): honor configured sources in post-install hooks

m199369309:
## Summary

- reuse the virtualenv `index_url`, `extra_index_url`, `find_links`, `trusted_host`, and `index_strategy` settings in FlashInfer post-install commands
- keep hook-specific public indexes as online fallbacks without overriding configured package sources
- allow explicit offline installs to use configured private indexes or local find-links while continuing to block public fallback indexes
- pass the resolved virtualenv install configuration to both FlashInfer post-install hooks

## Motivation

The main virtualenv installation already honors configured package sources, but the FlashInfer AOT and cubin synchronization hooks invoke `uv` directly. Those commands previously ignored part or all of the resolved source configuration, which could break private-mirror and offline installations after the main package install had succeeded.

## Tests

- `pytest -q xinference/core/tests/test_virtual_env_manager.py`
- `pytest -q xinference/core/tests/test_utils.py`
- `pre-commit run --files xinference/core/virtual_env_manager.py xinference/core/worker.py xinference/core/tests/test_virtual_env_manager.py`

### DIFF
--- a/xinference/core/virtual_env_manager.py
+++ b/xinference/core/virtual_env_manager.py
@@ -1034,10 +1034,61 @@ def ensure_sglang_inherited_packages_compatible_post_install(
         )
 
 
+def _as_uv_option_values(value: Any) -> List[str]:
+    if not value:
+        return []
+    return [value] if isinstance(value, str) else list(value)
+
+
+def build_uv_source_options(
+    conf: Dict[str, Any],
+    *,
+    public_index_urls: Optional[List[str]] = None,
+    allow_public_install: bool = True,
+) -> List[str]:
+    """Build uv source options from the virtualenv install configuration.
+
+    Post-install hooks invoke ``uv`` directly, so they must explicitly reuse the
+    same package sources as the main virtualenv installation. Administrator-
+    supplied public indexes are appended as fallbacks only when public installs
+    are allowed.
+    """
+    options: List[str] = []
+    index_url = conf.get("index_url")
+    if index_url:
+        options += ["--index-url", index_url]
+
+    extra_index_urls = _as_uv_option_values(conf.get("extra_index_url"))
+    if allow_public_install:
+        configured_urls = {index_url, *extra_index_urls}
+        extra_index_urls.extend(
+            url
+            for url in (public_index_urls or [])
+            if url and url not in configured_urls
+        )
+    for url in extra_index_urls:
+        options += ["--extra-index-url", url]
+
+    for link in _as_uv_option_values(conf.get("find_links")):
+        options += ["--find-links", link]
+    for host in _as_uv_option_values(conf.get("trusted_host")):
+        options += ["--trusted-host", host]
+
+    index_strategy = conf.get("index_strategy")
+    if index_strategy:
+        options += ["--index-strategy", index_strategy]
+    return options
+
+
+def _has_configured_package_source(conf: Dict[str, Any]) -> bool:
+    return any(conf.get(key) for key in ("index_url", "extra_index_url", "find_links"))
+
+
 def ensure_flashinfer_cubin_matches_post_install(
     model_engine: Optional[str],
     virtual_env_manager: Any,
     allow_public_install: bool = True,
+    conf: Optional[Dict[str, Any]] = None,
 ) -> None:
     """Keep FlashInfer's Python package and cubin package on the same version.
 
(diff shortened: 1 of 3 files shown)


xinference/core/virtual_env_manager.py
```diff
@@ -1034,10 +1034,61 @@ def ensure_sglang_inherited_packages_compatible_post_install(
...
+    extra_index_urls = _as_uv_option_values(conf.get("extra_index_url"))
+    if allow_public_install:
+        configured_urls = {index_url, *extra_index_urls}
+        extra_index_urls.extend(
+            url
+            for url in (public_index_urls or [])
+            if url and url not in configured_urls
+        )
+    for url in extra_index_urls:
+        options += ["--extra-index-url", url]
```
qinxuye: [P1] Keep configured package sources ahead of public fallbacks

When `conf` contains only a private `index_url` (a common corporate-mirror setup), this emits `--index-url <private> --extra-index-url <flashinfer.ai>`. `uv` gives every extra index higher priority than `--index-url`, and an unreachable extra index can fail resolution instead of falling back, so the hook can still bypass or be blocked by the public source even when the configured mirror contains these packages. Please encode the configured sources at higher priority and the hook-specific public URL as the lower-priority fallback (for example with `--index` / `--default-index`), and add a regression test for effective priority rather than only flag presence.

qinxuye (commented):
The source propagation still has one blocking priority issue.

xinference/core/virtual_env_manager.py
qinxuye: Thanks for reworking the flags, but the blocking behavior remains in the actual vLLM path. `apply_engine_virtualenv_settings()` supplies `index_strategy="unsafe-best-match"`, and `build_uv_source_options()` forwards it unchanged. Under uv, that strategy searches every index, so a private `--index` that already contains the pinned package plus an unreachable public `--default-index` still fails resolution; I reproduced this with uv 0.11.26. Please make these post-install commands use first-index semantics when appending a public fallback (or otherwise stop consulting the public index after a configured source supplies the package), and add a resolver-level regression with an unreachable fallback. The current command-order assertions do not cover the effective behavior.

xinference/core/virtual_env_manager.py
```diff
@@ -1034,10 +1034,92 @@ def ensure_sglang_inherited_packages_compatible_post_install(
...
+        options += ["--trusted-host", host]
+
+    # ``unsafe-best-match`` queries every index even after a configured source
+    # provides the pinned package. An unavailable hook-specific public fallback
+    # would therefore still fail the whole install. Use first-index semantics
+    # whenever a public fallback is appended so configured sources remain
+    # authoritative; otherwise preserve the caller's existing strategy.
+    index_strategy = (
+        "first-index"
+        if configured_index_urls and public_fallback_urls
```
qinxuye, line 1118: [P1] Do not let a public fallback defeat configured find-links

This condition only forces first-index when configured index URLs are present. With `conf={"find_links": "/srv/wheels"}` and no index URL, the hook still emits the public `--default-index`; uv 0.11.26 queries that index and fails if it is unreachable even when `/srv/wheels` contains the exact wheel. I reproduced this with `uv pip install --dry-run --no-deps`: both the default strategy and explicit `first-index` returned exit 2 after hitting the unavailable fallback. Please cover find-links-only configurations, likely by trying configured sources without the public fallback first and using the public index only on a genuine miss, and add a resolver-level regression.

xinference/core/virtual_env_manager.py
```diff
@@ -1034,10 +1034,147 @@ def ensure_sglang_inherited_packages_compatible_post_install(
...
+            text=True,
+        )
+        if result.returncode == 0:
+            return result
+        logger.info(
+            "Post-install requirements were not satisfied by configured package "
+            "sources; retrying with the allowed public fallback"
+        )
+
+    cmd = base_cmd + build_uv_source_options(
```
qinxuye, line 1166: [P1] Let the second pass actually reach the public fallback

After the configured-only attempt fails because a private index has the package name but not the requested pinned version, this retry combines that private index with the public fallback. `build_uv_source_options()` then forces `first-index`, so uv stops at the private index and never considers the compatible public version. I reproduced this with uv 0.11.26: the private index exposed only 0.9.0, the public index exposed 1.0.0, and a request for `==1.0.0` queried the private index twice, never queried public, and returned 1 with uv's compatible-version-on-a-subsequent-index hint. Please make the retry use the fallback independently or otherwise allow a satisfying version from the later index, and add a resolver-level version-miss regression.

qinxuye: Fixed in a660c846. The retry now drops the configured-source options and queries the hook-specific public fallback independently, so a private index containing only a mismatched version cannot mask the compatible public version. The new uv 0.11.26 resolver-level regression exercises private 0.9.0 versus public 1.0.0 and passes; the focused virtual-env-manager suite is 55/55 passing.

qinxuye (approved):
LGTM

PR #5387, opened 2026-08-19
feat(venv): add secure per-launch find-links support

m199369309:
## Summary

- add a typed `virtual_env_find_links` launch parameter across REST, Supervisor, Worker, sync/async clients, and CLI
- validate request-level paths on each target Worker before invoking `uv`
- restrict paths to administrator-configured roots via `XINFERENCE_VIRTUAL_ENV_FIND_LINKS_ALLOWED_ROOTS` (default: `${XINFERENCE_HOME}/wheels`)
- canonicalize paths, reject URLs/relative paths/files/symlink escapes, check readability, and deduplicate while preserving order
- preserve configured model/pip find-links and append validated per-launch directories
- document CLI and Python usage plus distributed-deployment and trust requirements

## Dependency

Depends on #5386 so direct FlashInfer post-install commands honor the same configured find-links sources as the main virtualenv installation.

This PR is intentionally stacked on #5386. The shared commit will disappear from this PR after #5386 is merged.

## Security

Validation happens on each selected Worker because the Supervisor cannot safely validate Worker-local filesystems. Allowed roots must be controlled by administrators and must not be writable by untrusted users, since installing wheels executes package code.

Set `XINFERENCE_VIRTUAL_ENV_FIND_LINKS_ALLOWED_ROOTS` to an empty value to disable request-level local paths.

## Tests

- `pytest -q xinference/core/tests/test_virtual_env_manager.py xinference/core/tests/test_worker.py xinference/core/tests/test_launch_strategy.py xinference/core/tests/test_model_replica_scale.py` (163 passed)
- CLI help smoke test for `--virtual-env-find-link`


xinference/core/worker.py
```diff
@@ -3095,6 +3098,14 @@ def _prepare_virtual_env(
                 if hasattr(settings, k) and not getattr(settings, k):
                     setattr(settings, k, v)
 
+        if virtual_env_find_links is not None:
+            requested_find_links = validate_virtual_env_find_links(
+                virtual_env_find_links
+            )
+            settings.find_links = merge_virtual_env_find_links(
```
qinxuye, line 3102: **[P1] Keep request-level find-links out of shared settings.** `match_llm()` makes a shallow family copy, so its `virtualenv` object is still shared with the global family; for vLLM, SGLang, and Diffusers this method also skips the deep copy above. Assigning here therefore persists this launch's directory into shared settings. I reproduced a second vLLM launch with `virtual_env_find_links=None` still passing the first launch's path to `install_packages`, so an ephemeral or stale request path can affect unrelated later launches. Please deep-copy `settings` before merging request-scoped sources for every engine and add a two-launch vLLM regression test.

qinxuye (approved):
LGTM

PR #5388, opened 2026-08-19
feat(ui): add per-launch virtualenv find-links

m199369309:
## Summary

- add a repeatable Local Wheel Directories field to the model launch runtime-environment form
- serialize `virtual_env_find_links` into launch requests and restore it into form state
- include the field in launch parameter filtering
- generate repeated `--virtual-env-find-link` CLI arguments and parse them back from launch commands
- add English, Chinese, Japanese, and Korean labels

## Dependency

Depends on #5387, which adds the secure backend, client, and CLI support for per-launch find-links.

This PR is intentionally stacked on #5387 (and transitively #5386). Shared commits will disappear from this PR after the dependencies are merged.

## Validation

- `npx prettier --check` on all modified frontend files
- `npx eslint .` (passed with 56 pre-existing warnings and no errors)
- `npm run build`

### DIFF
--- a/frontend/src/components/pages/launch-model/launch-dialog/advanced-config.tsx
+++ b/frontend/src/components/pages/launch-model/launch-dialog/advanced-config.tsx
@@ -133,6 +133,14 @@ const AdvancedConfig: FC<AdvancedConfigProps> = ({
         </div>
 
         <div className="p-2">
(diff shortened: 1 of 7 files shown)


qinxuye (approved):
LGTM

PR #5393, opened 2026-08-20
fix(ui): clarify Token Router identifiers and asset layout

m199369309:
## Summary

- clarify the distinction between the routing policy identifier and the virtual model UID used by clients
- use the same policy label in Router runtime details and deletion confirmation
- improve the Tokenizer Asset catalog layout for long asset IDs, revisions, node IDs, errors, and action controls
- update English, Chinese, Japanese, and Korean translations

## Dependency

Depends on #5380.

This PR is intentionally stacked on top of #5380 because the affected Token Router UI components are introduced there. Please merge #5380 first; after that merge, this PR can be rebased onto `main`.

## Validation

- `git diff --check`
- `pre-commit run --files <changed-files>`
- `npx prettier --check <changed-files>`
- `npx eslint <changed-files>`
- `npm run build`


qinxuye (dismissed):
LGTM

qinxuye (commented):
Withdrawing my previous approval. The current head still leaves seven review items unaddressed: defensive handling of the cluster-info response, empty-string fallback in `toNumber`, null-safe tokenizer form values, offloading synchronous `sweep_nodes` work from the event loop, and bounding process termination during Agent shutdown (three related threads). Please address the existing unresolved threads before re-requesting review.

qinxuye (approved):
LGTM

PR #5395, opened 2026-08-20
fix(cache): fully remove downloaded models

Minamiyama:
- Resolve nested cache symlinks recursively so ModelScope and Hugging Face payload files are removed instead of leaving real weights behind.

- Preserve files referenced by another cached model and unlink local model_uri entries without touching user-managed source directories.

- Restrict recursive directory deletion to Xinference-managed cache and tensorizer roots to prevent arbitrary path removal.

- Prune empty Hub directories up to, but never including, each managed download root; stop when sibling models or residual files remain.

- Delete linked files before cache trees and tolerate concurrent cleanup races while still reporting real deletion failures.

- Treat a false cache-delete API result as an error in WebUI and add localized messages for English, Chinese, Japanese, and Korean.

- Cover nested targets, shared blobs, local sources, unmanaged paths, and ModelScope parent pruning with focused worker tests.

## Summary

- Remove actual downloaded model files referenced by nested Xinference cache
  symlinks.
- Preserve files still referenced by another cached model.
- Prune empty model directories without deleting Hub root directories.
- Prevent local `model_uri` sources and unmanaged directories from being
  removed.
- Report backend deletion failures correctly in WebUI.

## Root Cause

Xinference recursively mirrors downloaded Hub snapshots into its cache using
symlinks. However, `WorkerActor.list_deletable_models()` only inspected the
first directory level.

For models containing nested directories, deleting the cache removed the
Xinference symlinks but left most ModelScope or Hugging Face payload files on
disk. Disk space was therefore not reclaimed.

The WebUI also ignored a backend response containing `result: false` and always
displayed a successful deletion notification.

This behavior was not Docker-specific. It affected Docker, local, and virtual
environment deployments; bind-mounted Docker storage only made the remaining
files easier to observe.

### Backend

- Recursively collect nested symlink file targets without following directory
  symlinks.
- Delete external payload files before removing the Xinference cache tree.
- Preserve targets still referenced by another Xinference cache entry.
- Treat a root cache symlink as a user-provided `model_uri` and unlink only the
  cache entry.
- Reject cache paths outside `XINFERENCE_CACHE_DIR`.
- Restrict recursive directory deletion to Xinference cache and tensorizer
  roots.
- Tolerate files already removed by concurrent cleanup.
- Remove empty parent directories with `os.rmdir` after deleting downloaded
  files.
- Stop parent cleanup when a directory contains another model, revision,
  lock file, incomplete download, or other residual content.
- Preserve the managed download roots:
  - Hugging Face cache root
  - ModelScope `models` root
  - OpenMind cache root

For example:

```text
/root/.xinference/modelscope/models/
└── ZhipuAI--GLM-Image/
    └── snapshots/
        └── master/
```

After all files for this model are deleted, empty directories are removed up
through `ZhipuAI--GLM-Image`, while the `modelscope/models` root remains.

### WebUI

- Check the `result` field returned by the cache deletion API.
- Display an error instead of a false success notification when deletion fails.
- Add localized failure messages for English, Chinese, Japanese, and Korean.

## Safety

- Shared payload files remain when another cached model references them.
- User-managed `model_uri` source directories are never recursively deleted.
- Empty parent cleanup uses `os.rmdir`, not `shutil.rmtree`.
- Parent cleanup never removes a configured Hub root.
- Non-empty directories stop cleanup naturally.


xinference/core/worker.py
```diff
@@ -4578,48 +4580,165 @@ async def list_cached_models(
...
+        except ValueError:
+            # Different drives on Windows, for example.
+            return False
+
+    @classmethod
+    def _collect_symlink_file_targets(
+        cls, root: str, excluded_root: Optional[str] = None
+    ) -> Set[str]:
+        """Collect existing file targets without following directory links."""
+        if not os.path.isdir(root) or os.path.islink(root):
```
qinxuye: **[P1] Account for references through other cache-directory symlinks**

This early return also applies when this helper scans `XINFERENCE_CACHE_DIR` for `shared_targets`. A user-provided `model_uri` is represented by a root directory symlink, so none of the files reachable through that other cached model are counted as shared. I reproduced this with one normal cache entry linking to an HF blob and a second `model_uri` entry linking to the same snapshot: deleting the normal entry returned `True` but removed the blob and left the second model’s `weights` link broken. Please include targets reachable through *other* directory-symlink cache entries during shared-use detection (read-only), while still avoiding traversal of the entry being deleted.

xinference/core/worker.py
```diff
@@ -4578,48 +4580,165 @@ async def list_cached_models(
...
+            paths.add(path)
+        elif os.path.isdir(path):
+            paths.add(path)
 
-        if os.path.isfile(path):
-            path = os.path.dirname(path)
+            # ``create_symlink`` mirrors the whole Hub snapshot recursively, so
+            # resolving only the first level leaves nearly all nested weights in
+            # ModelScope/Hugging Face caches. Resolve every linked file instead.
+            targets = WorkerActor._collect_symlink_file_targets(path)
```
qinxuye, line 4806: **[P2] Handle the non-symlink copy fallback**

This assumes `create_symlink` always leaves file symlinks to collect. On Windows without symlink privileges, `huggingface_hub._create_symlink(..., new_blob=False)` [copies the file instead](https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/file_download.py). In that case `targets` is empty, deletion removes only the Xinference copy, then returns success and updates the tracker while the full Hub payload remains. Please retain enough source-cache information to clean up the copy-fallback path as well, and add a regression test that forces symlink creation to fall back to copying.

qinxuye (commented):
Two additional cache-deletion findings:

xinference/core/worker.py
```diff
@@ -4578,48 +4582,240 @@ async def list_cached_models(
...
+            targets.update(
+                target
+                for target in get_cache_source_paths(path)
```
qinxuye: **[P1] Restrict manifest paths to managed download roots**

`get_cache_source_paths(path)` accepts every absolute path from `__cache_source_paths.json`, and those paths are later passed to `os.remove()`. Because `create_symlink()` mirrors this manifest from a downloaded source repository, a source-controlled manifest can name an unrelated local file. On this exact head I reproduced `unrelated_marked_for_deletion=True`, followed by a successful deletion with `unrelated_survives=False`. Please reject symlinked or externally supplied manifests and validate every recorded source path against the configured managed Hub roots before adding it to `targets`; also add a malicious-manifest regression test.

qinxuye (commented):
One cache-deletion safety issue remains:

qinxuye (approved):
LGTM

PR #5398, opened 2026-08-20
fix(router): stabilize managed runtime instance identity

m199369309:
## Summary
- stabilize managed Router Runtime instance IDs across Agent reconciliation and Runtime restarts
- derive the Agent-scoped prefix from the assigned listen host while preserving the UUID suffix
- pass the normalized ID through the Runtime environment and every assignment lifecycle status update
- preserve persisted instance IDs when status reports omit the field
- avoid using wildcard listen addresses directly as standalone instance prefixes

## Dependency
- Depends on #5379 (Router Agent orchestration)

## Validation
- `pytest -q xinference/router/tests/test_agent.py xinference/router/tests/test_control_plane.py xinference/core/tests/test_router_assignment_store.py`
- pre-commit checks for all changed files

`deploy/systemd/` is not included.

### DIFF
--- a/xinference/api/routers/token_routers.py
+++ b/xinference/api/routers/token_routers.py
@@ -542,7 +542,7 @@ async def register_router_node(
 ) -> JSONResponse:
     try:
         result = await (await _supervisor(api)).register_token_router_node(
-            payload.dict()
+            payload.dict(exclude_unset=True)
         )
     except ValueError as exc:
         raise HTTPException(status_code=409, detail=str(exc)) from exc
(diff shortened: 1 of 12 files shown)


m199369309:
Addressed the review feedback in commit `33598105`:

- Added a `BackgroundTask` cleanup fallback for streamed upstream responses while retaining generator `finally` cleanup.
- Preserved an existing assignment PID when a status report omits `pid`.
- Added finite, warning-backed parsing for Token Router float environment settings.
- Preserved the distinction between omitted and explicitly empty `reported_labels` at the API and store layers.
- Added regression coverage for all four behaviors.

Validation: targeted Token Router tests pass (70 passed); pre-commit hooks pass.

qinxuye (approved):
LGTM

PR #5399, opened 2026-08-20
fix(ui): harden Token Router running model and detail views

m199369309:
## Summary

- support Token Router virtual models in the Running Model page and detail view
- filter disabled or unavailable Token Router entries from physical-model operations
- show virtual-model, Router, Runtime, deployment, and node details with safe fallbacks
- load Runtime and assignment information independently so partial API failures do not hide Router configuration
- normalize sparse and legacy Router configuration before editing or saving
- preserve backend defaults for managed backends and avoid duplicate error notifications
- support virtual-model requests in the Try To API flow
- add English, Chinese, Japanese, and Korean strings for the new Runtime and virtual-model states

## Dependency

- Depends on #5380 for the Token Router and Router Agent management UI.
- Depends on #5397 for the virtual-model running-details API contract.

The identifier-label and Tokenizer Asset layout changes from #5393 are intentionally excluded from this PR to avoid duplicate submission.

## Validation

- `git diff --check`
- `npx prettier --check <changed frontend files>`
- `npx eslint <changed frontend files>`
- `npm run build`
- `pre-commit run --files <changed frontend files>` (Python hooks skipped because the change is frontend-only)

The build completed successfully. Next.js reported only existing repository warnings in unrelated files.


frontend/src/components/pages/running-model-detail/utils.ts
```diff
@@ -185,15 +185,30 @@ export function transformFileInfoForResult(message?: ChatChoicesMessage) {
...
+      ? detail.router_uid
+      : undefined;
 
   return {
     ...detail,
-    // fix model_ability was not returned when model_type was Rerank or Embedding.
-    model_ability: Array.isArray(detail.model_ability)
-      ? detail.model_ability
-      : (modelType && MODEL_TYPE_ABILITY_MAP[modelType]) || [],
+    ...(tokenRouterModelName ? { model_name: tokenRouterModelName } : {}),
```
qinxuye: [P2] Preserve the virtual model name

The backend descriptor deliberately returns model_name as the virtual_model_uid, but this shared transform replaces it with router_uid for every Token Router model. The detail view then shows the Router UID in both Model Name and Router UID, and other callers of the transform lose the actual virtual model name. Please keep model_name unchanged here and use router_uid only in the dedicated Router field or card subtitle.

frontend/src/components/pages/token-router/router-form-dialog.tsx
```diff
@@ -533,6 +591,15 @@ export function RouterFormDialog({ open, router, onOpenChange, onSaved }: Props)
       toast.success(t(router ? 'tokenRouter.updateSuccess' : 'tokenRouter.createSuccess'));
       onOpenChange(false);
       onSaved();
+    } catch (error) {
```
qinxuye, line 595: [P2] Avoid duplicate network-error notifications

The shared Axios response interceptor already emits SERVER_ERROR when a request has no response, including network and timeout failures. This catch classifies those Axios errors as non-HTTP and adds another toast, so one failed save produces two notifications despite the comment above. Please suppress all Axios errors here, for example with axios.isAxiosError, or reserve the local toast for genuinely non-Axios exceptions.

frontend/src/components/pages/running-model/index.tsx
```diff
@@ -281,11 +354,6 @@ const RunningModel = () => {
   }, [activeModel, replicaLogs]);
 
   const fetchClusterWorkers = useCallback(() => {
-    if (!clusterAuth || (clusterAuth.auth && !isAdmin)) {
-      setClusterWorkerInfo(new Map());
-      return;
-    }
-
     request
```
qinxuye, line 362: [P2] Keep the admin guard around cluster info

This callback now calls `/v1/cluster/info` for every Running Models visitor, but that route requires the `admin` scope whenever authentication is enabled. The shared response interceptor turns the resulting 403 into a global `Server error: 403 No auth` toast, so ordinary users who can access Running Models see a false authorization error on every page load. Please restore the `clusterAuth.auth && !isAdmin` guard (and its hook dependencies), keeping the model-derived addresses as the fallback.

qinxuye (approved):
LGTM

PR #5400, opened 2026-08-21
feat: add media seed controls and dimension swap

Minamiyama:
## Summary

Add consistent seed controls to image, video, and audio generation WebUIs, and add a width/height swap button wherever both dimensions are configurable.

## Changes

- Add `Seed(s)` support for image generation.
  - Accept comma-separated seeds.
  - Default to `-1`.
  - Pad missing seeds to the requested image count on the backend.
  - Treat missing values and `-1` as random seeds.
  - Regenerate exactly one fresh seed per requested image whenever 🎲 is clicked.
- Add scalar seed controls to video and audio generation.
- Pass seeds through the existing `kwargs` API field.
- Normalize and forward seeds across supported image, video, and audio backends.
- Safely ignore seeds when the underlying model does not support them.
- Add width/height swap controls to image and video forms.
- Add frontend and backend seed utility tests.

## API Examples

Image generation:

```json
{
  "n": 4,
  "kwargs": "{\"seed\":[11,22]}"
}
```

This is handled as `[11, 22, -1, -1]`.

Video and audio generation:

```json
{
  "kwargs": "{\"seed\":11}"
}
```

## Compatibility

No top-level API schema changes. Seed values continue to use the existing `kwargs` passthrough mechanism.

d65cbed45276fd63f20e5afd71c322d9

### DIFF
--- a/xinference/model/audio/chattts.py
+++ b/xinference/model/audio/chattts.py
@@ -74,13 +74,15 @@ class ChatTTSModel:
         response_format: str = "mp3",
         speed: float = 1.0,
         stream: bool = False,
+        **kwargs,
     ):
         import ChatTTS
         import numpy as np
         import torch
 
-        from .utils import audio_stream_generator, audio_to_bytes
+        from .utils import apply_audio_seed, audio_stream_generator, audio_to_bytes
 
+        requested_seed = apply_audio_seed(kwargs)
         rnd_spk_emb = None
 
         if len(voice) > 400:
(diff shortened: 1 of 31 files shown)


xinference/model/audio/chattts.py
```diff
@@ -95,7 +97,7 @@ def speech(
                 logger.info("Fallback to random speaker due to %s", e)
 
         if rnd_spk_emb is None:
-            seed = _voice_seed(voice)
+            seed = requested_seed if requested_seed is not None else _voice_seed(voice)
```
qinxuye: The WebUI now submits `seed=-1` by default, and `ModelActor.speech` resolves that to a fresh random value before this method runs. For short ChatTTS voice IDs, using `requested_seed` here therefore replaces the stable `_voice_seed(voice)` mapping, so the same Voice ID selects a different speaker on every request. Please keep speaker selection derived from `voice`, then apply the requested seed before inference (or omit the default seed from generic audio requests).

qinxuye (approved):

PR #5403, opened 2026-08-21
fix(llm): support Hugging Face cache roots in auto-fill

### LINKED ISSUE #5402: Fix LLM auto-fill for cache roots and nested text configs
### Issue Description

LLM auto-fill cannot reliably build registration data from common local cache layouts and newer model configs:

- A Hugging Face repository cache root does not contain `config.json` directly; the file lives under `snapshots/<revision>`.
- Multimodal configs may keep text-model metadata such as context length and architecture dimensions under `text_config`.
- Model directory names can contain several `*B` fragments for versions, total parameters, activated parameters, training steps, or quantization variants, so deriving model size from a path is ambiguous.
- Integral and fractional model sizes currently produce different response types, which can break clients that normalize the field as a string.

This can make auto-fill fail for a Hugging Face cache root or populate incorrect model metadata even though the underlying model files are valid.

### Expected Behavior

Auto-fill should:

1. Resolve a Hugging Face cache root through `refs/main` or `refs/master`, with a safe single-snapshot fallback.
2. Read inference metadata from nested `text_config` while preserving top-level precedence.
3. Determine model size from configuration metadata, falling back to the architecture estimate without parsing directory names.
4. Return a consistently string-typed `model_size_in_billions` field.
5. Reject ambiguous cache roots containing multiple snapshots when no usable branch ref is present.

### Validation

Add focused unit tests covering Hugging Face cache roots, nested config precedence, ambiguous model directory names, and ambiguous snapshots.

amumu96:
## Summary

- resolve Hugging Face repository cache roots through `refs/main` or `refs/master`, with a unique-snapshot fallback and an explicit ambiguity error
- read context and architecture metadata from nested `text_config` while preserving top-level precedence
- derive model size only from configuration metadata, falling back to the existing architecture estimate instead of parsing ambiguous path names
- normalize auto-fill model-size output to a string so integral and fractional sizes have a consistent response type
- add focused regression coverage for Hugging Face cache layouts, nested configs, and ambiguous model directory names

## Release note

- LLM auto-fill now supports Hugging Face cache root directories and newer nested model configs, and returns consistent model-size values without guessing from directory names.

Closes #5402

## Validation

- `pytest -q xinference/model/llm/tests/test_config_parser.py` (6 passed)
- `pre-commit run --files xinference/model/llm/config_parser.py xinference/model/llm/tests/test_config_parser.py`

### DIFF
--- a/xinference/model/llm/config_parser.py
+++ b/xinference/model/llm/config_parser.py
@@ -267,10 +267,15 @@ def _infer_model_size_from_path(model_path: str) -> Optional[Union[int, str]]:
             r"[0-9a-f]{40}|[0-9a-f]{64}", basename
         ):
             matches = re.findall(
-                r"(?<![a-z0-9])(\d+(?:[._]+\d+)?)b(?=$|[^a-z0-9])", basename
+                r"(?<![a-z0-9.])(a?)(\d+(?:[._]+\d+)?)b(?=$|[^a-z0-9.])",
+                basename,
             )
             if matches:
-                size_text = re.sub(r"[._]+", ".", matches[-1])
+                # Prefer a total-size token over an activated-size token like
+                # A3B, but accept the latter when it is the only size present.
+                total_sizes = [size for prefix, size in matches if not prefix]
+                size_text = total_sizes[-1] if total_sizes else matches[-1][1]
+                size_text = re.sub(r"[._]+", ".", size_text)
                 size_in_billions = _extract_numeric_size(size_text)
                 if size_in_billions:
                     return _format_size_in_billions(size_in_billions)
--- a/xinference/model/llm/config_parser.py
+++ b/xinference/model/llm/config_parser.py
@@ -1,8 +1,39 @@
 import json
 import os
+import re
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 _BUILTIN_FAMILY_CACHE: Optional[List[Dict[str, Any]]] = None
+
+
+def _resolve_huggingface_snapshot(model_dir: str) -> Optional[str]:
+    """Resolve a Hugging Face repository cache root to one snapshot."""
+    snapshots_dir = os.path.join(model_dir, "snapshots")
+    if not os.path.isdir(snapshots_dir):
+        return None
+
+    main_ref = os.path.join(model_dir, "refs", "main")
+    if os.path.isfile(main_ref):
+        with open(main_ref, "r") as file:
+            revision = file.read().strip()
+        if revision:
+            main_snapshot = os.path.join(snapshots_dir, revision)
+            if os.path.isfile(os.path.join(main_snapshot, "config.json")):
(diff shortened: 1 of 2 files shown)


xinference/model/llm/config_parser.py
```diff
@@ -203,6 +256,28 @@ def _infer_model_size_in_billions(config: Dict[str, Any]) -> Optional[Union[int,
...
+    path = os.path.abspath(model_path)
+    for _ in range(4):
+        basename = os.path.basename(path).lower()
+        if not basename:
+            break
+        if basename not in {"snapshots", "refs"} and not re.fullmatch(
+            r"[0-9a-f]{40}|[0-9a-f]{64}", basename
+        ):
+            matches = re.findall(
+                r"(?<![a-z0-9])(\d+(?:[._]+\d+)?)b(?=$|[^a-z0-9])", basename
```
qinxuye: **[P1] Do not parse version or activated-size fragments as the model size.** Because `.` and `_` are accepted both inside the capture and as boundaries, this regex can start in the middle of a token: `Qwen1.5-MoE-A2.7B-Chat` returns `7` instead of `2_7`, while `qwen1.5_7b_chat` returns `5_7` instead of `7`. Since `build_llm_registration_from_local_config()` prefers this path result over the architecture estimate, auto-fill silently writes incorrect registration metadata for these common cache names. Please tighten the token boundaries and add regression cases for both forms.

qinxuye (commented):
One blocking model-size parsing issue remains.

xinference/model/llm/config_parser.py
```diff
@@ -203,6 +256,33 @@ def _infer_model_size_in_billions(config: Dict[str, Any]) -> Optional[Union[int,
...
+        ):
+            matches = re.findall(
+                r"(?<![a-z0-9.])(a?)(\d+(?:[._]+\d+)?)b(?=$|[^a-z0-9.])",
+                basename,
+            )
+            if matches:
+                # Prefer a total-size token over an activated-size token like
+                # A3B, but accept the latter when it is the only size present.
+                total_sizes = [size for prefix, size in matches if not prefix]
+                size_text = total_sizes[-1] if total_sizes else matches[-1][1]
```
qinxuye: [P1] Do not choose the last non-activated `*B` token blindly. Real model IDs already present in this catalog contain multiple such tokens: `mlx-community/ERNIE-4.5-300B-47B-PT-*` is inferred as 47 instead of 300, while `Xorbits/TinyLlama-1.1B-step-50K-105b-GGUF` is inferred as 105 instead of 1.1. Since this path result takes precedence over the architecture estimate, auto-fill silently writes incorrect `model_size_in_billions` for those cache roots. Please disambiguate multiple candidates (or fall back to config/architecture when the path is ambiguous) and add regression cases for both names.

qinxuye (approved):
LGTM

PR #5404, opened 2026-08-21
FEAT: [model] glm-5.2 support

llyycchhee:
Automated sync for model "glm-5.2" (llm) by user llyycchhee.


xinference/model/llm/llm_family.json
```diff
@@ -34921,5 +34921,721 @@
...
+        }
+      },
+      {
+        "model_size_in_billions": 753,
+        "model_format": "ggufv2",
+        "model_src": {
+          "huggingface": {
+            "model_id": "unsloth/GLM-5.2-GGUF",
+            "model_file_name_template": "",
+            "model_file_name_split_template": "GLM-5.2-{quantization}-{part}.gguf",
```
qinxuye: **[P1] Fix the GGUF quantization and file-path metadata.** These specs flatten to only `quantization="none"`, while `quantization_parts` is keyed by `BF16`, `Q8_0`, and the `UD-*` variants. `generate_model_file_names_with_quantization_parts()` therefore misses the parts map and returns the empty `model_file_name_template`, so the GGUF download cannot work. The upstream files also live under per-quantization directories, for example `BF16/GLM-5.2-BF16-00001-of-00033.gguf`. Please enumerate the actual quantizations, include the `{quantization}/` directory in the split template, and provide a non-empty final filename for both hub entries.

xinference/model/llm/llm_family.json
```diff
@@ -34921,5 +34921,721 @@
     },
     "featured": false,
     "updated_at": 1786731484
+  },
+  {
+    "model_name": "glm-5.2",
+    "model_description": "We're introducing GLM-5.2, our latest flagship model for long-horizon tasks",
+    "context_length": 202752,
```
qinxuye: **[P1] Use the checkpoint's 1M context length.** The official config sets `max_position_embeddings` to `1048576`, and the model card advertises a 1M-token context. This family value becomes llama.cpp's default `n_ctx` and is also reported through the model APIs, so `202752` silently limits and misreports the supported context.

qinxuye (commented):
Blocking metadata issues found in the generated GLM-5.2 entry.

xinference/model/llm/llm_family.json
```diff
@@ -34921,5 +34921,721 @@
...
+    "model_lang": [
+      "en",
+      "zh"
+    ],
+    "model_ability": [
+      "chat",
+      "vision",
```
qinxuye: **[P2] Remove the unsupported `vision` ability.** The checkpoint is a text-generation `GlmMoeDsaForCausalLM`, and this entry's own chat template converts image/video/audio inputs into an `unable to process` reminder. Advertising `vision` makes clients and engine selection treat unsupported multimodal input as valid.

xinference/model/llm/llm_family.json
qinxuye: Fixed in 600985622. The GGUF entries now enumerate all 22 upstream quantizations, use {quantization}/GLM-5.2-{quantization}-{part}.gguf for split files, and produce GLM-5.2-{quantization}.gguf as the merged filename for both hubs. Both hubs were checked against the upstream directory list, and local validation covered all 44 flattened hub/quantization specs and their generated part and final filenames.

xinference/model/llm/llm_family.json
qinxuye: Fixed in 600985622. context_length is now 1048576, matching the official checkpoint configuration.

xinference/model/llm/llm_family.json
qinxuye: Fixed in 600985622. The unsupported vision ability has been removed; the family now advertises chat, tools, reasoning, and hybrid only.

qinxuye (approved):
LGTM

PR #5409, opened 2026-08-22
fix(supervisor): reschedule autostart when the last replica dies

### LINKED ISSUE #5408: Autostart-configured model is never relaunched after mark_replica_dead evicts its last replica
## System Info / 系統信息

Read from source, current main (39c1ad40), not a specific deployed version.

## Version info / 版本信息

main branch, HEAD 39c1ad40

## Reproduction / 复现过程

`SupervisorActor.mark_replica_dead()` (`xinference/core/supervisor.py`) is
called by a worker once `XINFERENCE_MODEL_ACTOR_AUTO_RECOVER_LIMIT` bounded
local recovery is exhausted for a replica. When it is the last replica, the
model is advanced to `LaunchStatus.TERMINATED` and the function returns.

Autostart (`xinference/core/autostart.py`) exists specifically to relaunch a
model that goes down, with its own independent retry budget, but its
background task only gets woken from three places: supervisor startup,
`upsert_autostart_model`, and worker `add_worker`. `mark_replica_dead` is a
fourth place a model can go to `TERMINATED`, and it never calls
`self._schedule_autostart()`.

Grepping the full body of `mark_replica_dead` (`git show
39c1ad40:xinference/core/supervisor.py` at the relevant lines) for
`_schedule_autostart` returns no hits, confirmed against the three sites
that do call it.

## Expected behavior / 期待表现

An Autostart-launched model whose last replica dies after exhausting local
recovery should be relaunched by Autostart, the same as if it went down any
other way. Today it is silently left `TERMINATED` until an operator restarts
the supervisor process or manually relaunches it.

This only affects deployments that set
`XINFERENCE_MODEL_ACTOR_AUTO_RECOVER_LIMIT` to a finite value (default is
unbounded local retry, so this path is never reached at default settings),
which is an intended, supported configuration rather than an edge case.

PR with a fix and a regression test to follow.

AmirF194:
`SupervisorActor.mark_replica_dead()` advances a model to `TERMINATED` when
its last replica dies (worker exhausted
`XINFERENCE_MODEL_ACTOR_AUTO_RECOVER_LIMIT`), but never wakes Autostart.
Autostart's background task is only re-armed from three places: supervisor
startup, `upsert_autostart_model`, and worker `add_worker`
(`git log -G '_schedule_autostart' --follow` shows a single introducing
commit, #5076, and no other touches). `mark_replica_dead` is a fourth path
to `TERMINATED` and was never wired in when Autostart was added, so an
Autostart-launched model that exhausts local recovery is left dead until an
operator restarts the supervisor or relaunches it by hand.

Fix: call `self._schedule_autostart()` after the `TERMINATED` transition,
matching the existing three call sites. It is a no-op when there is nothing
to autostart, same as those.

Verification:
- New regression test in `test_autostart.py` drives the real
  `mark_replica_dead` through a minimal dummy supervisor (same pattern the
  file already uses for `_load_autostart_entries`/`_autostart_one_model`):
  it fails on main (Autostart never woken) and passes on this branch. A
  second test confirms a degraded-but-not-dead model (a healthy replica
  remains) correctly does not wake Autostart.
- `xinference/core/tests/test_autostart.py` and `test_worker.py` (which
  covers `mark_replica_dead`'s rank0/Xavier teardown path): 101 passed.
- `pre-commit run --files xinference/core/supervisor.py
  xinference/core/tests/test_autostart.py`: black, ruff, isort, mypy,
  codespell all clean.
- Not run: the full GPU-backed model test groups, and an end-to-end run
  through a real multi-actor supervisor/worker pair with a live crash-loop;
  this is a targeted unit-level regression test against the actual method,
  not a full cluster reproduction.

Fixes #5408

### DIFF
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -4255,6 +4255,10 @@ class SupervisorActor(xo.StatelessActor):
         except Exception:
             logger.warning("Failed to mark %s TERMINATED in status guard", base_uid)
 
+        # Autostart owns relaunching a dead model; this is its 4th trigger,
+        # alongside startup, upsert_autostart_model and add_worker.
+        self._schedule_autostart()
+
     @log_async(logger=logger)
     async def get_model(self, model_uid: str) -> xo.ActorRefType["ModelActor"]:
         replica_info = self._model_uid_to_replica_info.get(model_uid, None)
--- a/xinference/core/tests/test_autostart.py
+++ b/xinference/core/tests/test_autostart.py
@@ -13,11 +13,11 @@
 # limitations under the License.
 
 import asyncio
-from unittest.mock import MagicMock
+from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 
-from xinference.core.supervisor import SupervisorActor
+from xinference.core.supervisor import ReplicaInfo, SupervisorActor
 
 
 class _DummySupervisor:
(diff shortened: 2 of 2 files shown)


qinxuye (approved):

AmirF194:
Thanks for the quick review and merge.

PR #5417, opened 2026-08-24
FEAT: [model] Kimi-K3 support

llyycchhee:
Automated sync for model "Kimi-K3" (llm) by user llyycchhee.


xinference/model/llm/llm_family.json
```diff
@@ -35159,5 +35159,454 @@
...
+  },
+  {
+    "model_name": "Kimi-K3",
+    "model_description": "Kimi K3 is an open-weight, native multimodal agentic model and our most capable model to date.",
+    "context_length": 1048576,
+    "model_lang": [
+      "en",
+      "zh"
+    ],
+    "model_ability": [
```
qinxuye, line 35171: Kimi K3 always uses thinking and emits XTML think/response/tool channels, but this family exposes only chat and vision and provides neither reasoning metadata nor a Kimi-K3 parser. The current adapters therefore cannot split reasoning_content from content or parse its tool calls correctly; merely registering the vLLM architecture leaves the OpenAI chat response semantics incomplete. Please add Kimi-K3 reasoning/tool parsing and the matching abilities, or explicitly normalize the XTML output before advertising chat support.

xinference/model/llm/llm_family.json
```diff
@@ -35159,5 +35159,454 @@
...
+      },
+      {
+        "model_size_in_billions": "2779_93",
+        "activated_size_in_billions": 104,
+        "model_format": "ggufv2",
+        "model_src": {
+          "huggingface": {
+            "model_id": "unsloth/Kimi-K3-GGUF",
+            "model_file_name_template": "",
+            "model_file_name_split_template": "Kimi-K3-{quantization}-{part}.gguf",
```
qinxuye: The GGUF entry cannot resolve any downloadable file as written. It advertises only the quantization "none", while quantization_parts is keyed by UD-* values; the llama.cpp loader only uses model_file_name_split_template when the selected quantization exists in quantization_parts, so "none" falls back to the empty model_file_name_template. The source repository also stores every shard under a quantization subdirectory, for example UD-IQ1_M/Kimi-K3-UD-IQ1_M-00001-of-00015.gguf, but this template omits that directory. Please expose the actual UD-* quantizations and use a template such as "{quantization}/Kimi-K3-{quantization}-{part}.gguf", or remove the GGUF spec until it is launchable.

xinference/model/llm/llm_family.json
```diff
@@ -35159,5 +35159,454 @@
...
+    ],
+    "model_ability": [
+      "chat",
+      "vision"
+    ],
+    "model_specs": [
+      {
+        "model_size_in_billions": "2779_93",
+        "activated_size_in_billions": 104,
+        "model_format": "pytorch",
```
qinxuye, line 35181: This pytorch spec is currently exposed as Transformers-capable, but KimiK3ForConditionalGeneration falls through to the generic PytorchChatModel. Since the family declares vision and Kimi-K3 is not in XINFERENCE_BATCHING_ALLOWED_VISION_MODELS, PytorchChatModel._should_use_batching() returns false and chat() dispatches to the unimplemented _direct_chat(), which raises NotImplementedError. Please either add a dedicated Transformers multimodal implementation or exclude this architecture from the generic Transformers matcher so only working engines are advertised.

xinference/model/llm/llm_family.json
```diff
@@ -35159,5 +35159,475 @@
...
+        }
+      }
+    ],
+    "architectures": [
+      "KimiK3ForConditionalGeneration"
+    ],
+    "reasoning_start_tag": "<|open|>think<|sep|>",
+    "reasoning_end_tag": "<|close|>think<|sep|>",
+    "tool_parser": "kimi-k3",
+    "chat_template": "",
```
qinxuye, line 35615: This empty template still makes Kimi-K3 chat fail before inference. `VLLMMultiModel.async_chat()` falls back to `tokenizer.chat_template` and raises `ValueError` when it is falsy, but the official Kimi-K3 tokenizer does not define a Jinja `chat_template`; it implements a custom Python `apply_chat_template()` instead. As a result, the tokenizer-native renderer is never reached. Please route Kimi-K3 through `tokenizer.apply_chat_template()` when no template string is available (or provide an equivalent supported renderer), and add a regression test covering a tokenizer with `chat_template=None`.

xinference/model/llm/vllm/core.py
```diff
@@ -429,6 +430,11 @@ def _update_vllm_supported_lists() -> None:
             VLLM_SUPPORTED_MULTI_MODEL_LIST, "MiniMaxM3SparseForConditionalGeneration"
         )
 
+    if effective_version >= version.parse("0.27.0"):
```
qinxuye: This version gate leaves Kimi-K3 unavailable in the default virtualenv flow. `_get_effective_vllm_version()` returns `DEFAULT_VLLM_VERSION` (`0.21.0`) whenever virtualenv is enabled, so this branch never registers `KimiK3ForConditionalGeneration`. On this head, `get_engine_params_by_name_with_virtual_env("LLM", "Kimi-K3", True)` returns an unavailable string for vLLM instead of launch parameters, while the Kimi virtualenv dependency expands only to `vllm>=0.11.2`. Please make the Kimi virtualenv select and advertise a vLLM version that satisfies this `>=0.27.0` gate, and add a regression test for engine discovery with the default virtualenv setting; otherwise the only supported PyTorch engine cannot be selected or launched.

doc/source/models/builtin/llm/kimi-k3.rst
```diff
@@ -0,0 +1,47 @@
+.. _models_llm_kimi-k3:
+
+========================================
+Kimi-K3
+========================================
+
+- **Context Length:** 1048576
+- **Model Name:** Kimi-K3
+- **Languages:** en, zh
+- **Abilities:** chat, vision
```
qinxuye: These generated docs are stale after the follow-up fixes. They still list only `chat, vision` here, still advertise Transformers for the PyTorch spec, and still show the GGUF quantization as `none`. The current catalog declares `reasoning` and `tools`, explicitly rejects Kimi-K3 from the generic Transformers adapter, and exposes the UD-* GGUF quantizations. Please regenerate this file from the current head, or update it equivalently, so the user-facing engine and launch matrix matches the implementation.

qinxuye (approved):
LGTM

PR #5422, opened 2026-08-26
FEAT: support replica config in launch CLI and Web UI

m199369309:
## Summary

- Add formal `--replica-config` support to `xinference launch`, including the compatible `--replica_config` alias.
- Validate JSON shape, replica count, and conflicts with global worker/GPU placement options before sending the launch request.
- Preserve per-replica Worker/GPU placement when copying a Web UI launch form to a command line and restore it when parsing the command.
- Hide and clean up `n_worker` and legacy global placement fields in per-replica placement mode.
- Share per-replica validation between model launch and command-line copy so incomplete configurations cannot produce invalid commands.
- Add CLI tests and update the launch documentation.

## Compatibility

- The existing REST API and Python Client `replica_config` structure is unchanged.
- Automatic scheduling and existing global Worker/GPU placement behavior are unchanged.
- Each replica remains restricted to one device, consistent with the existing backend validation.

## Tests

- `npm run lint` — passed, 0 errors (existing warnings remain).
- `npm run build` — passed.
- `python3 -m compileall -q xinference/deploy/cmdline.py xinference/deploy/test/test_cmdline.py` — passed.
- `pytest -q xinference/deploy/test/test_cmdline.py -k replica_config` — 11 passed.
- `git diff --check` — passed.

### DIFF
--- a/frontend/src/components/pages/launch-model/utils.tsx
+++ b/frontend/src/components/pages/launch-model/utils.tsx
@@ -30,6 +30,11 @@ import type {
   WorkerOption,
 } from './types';
 import type { TFunc } from '@/contexts/i18n-context';
+import {
+  parseReplicaGpuIndexes,
+  transformReplicaConfigToFormRows,
+  transformReplicaFormRowsToConfig,
+} from './replica-config-utils.mjs';
 
 export const MODEL_ENGINE_TYPES: RequestModelType[] = [
   ModelType.LLM,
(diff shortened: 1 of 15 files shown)


frontend/src/components/pages/launch-model/utils.tsx
```diff
@@ -572,8 +573,15 @@ export function transformFetchToForm(values: FormValues) {
       };
     });
     nextValues.replica_placement_mode = 'custom';
+    nextValues.n_worker = undefined;
+    nextValues.worker_ip = undefined;
+    nextValues.n_gpu = undefined;
```
qinxuye, line 550: [P1] Preserve n_gpu during command import

The replica-config conversion above drops deviceRecord.n_gpu whenever a command is imported. A valid config with n_gpu: 1 and no explicit gpu_idx therefore becomes an empty GPU-index field, and transformFormToFetch later emits n_gpu: "auto" when the user launches the model or copies the command again. The backend accepts and preserves a numeric n_gpu without gpu_idx, so this silently changes the requested resource allocation. Please retain n_gpu in the form representation, or reject configurations the editor cannot represent, and add a round-trip test for this case.

m199369309: Fixed in `3608fb93`. The imported per-replica `n_gpu` value is now retained as form metadata when `gpu_idx` is absent, while an explicit/edited `gpu_idx` still determines `n_gpu`. Added round-trip coverage for numeric `n_gpu`, `"auto"`, and explicit GPU indexes. Local frontend build, focused ESLint, Node tests, and related Python tests pass.

qinxuye (approved):
LGTM

PR #5423, opened 2026-08-26
FIX: stabilize MLX runtime dependencies and locks

qinxuye:
## Summary

- pin F5-TTS MLX child virtual environments to MLX 0.32.1, the known-good release for f5-tts-mlx 0.2.6
- run the F5-TTS Metal integration test on a real local actor pool, so `start_python` actually launches the model actor from its virtual environment instead of the in-process `test://` interpreter
- keep f5-tts-mlx out of the Metal host environment, preventing host and child MLX dependency mixing
- require filelock 3.30+ and preserve MLX conversion lock files so producer/deleter synchronization remains stable on Windows
- verify the lock configuration directly so the regression is meaningful on Linux, macOS, and Windows
- add regression coverage for the isolated F5-TTS dependency and persistent conversion locks

## Tests

- real F5-TTS MLX integration passed locally on Apple Silicon (`1 passed`), covering both direct speech generation and the OpenAI-compatible speech API through the child virtual environment
- MLX video tests passed (`13 passed`), including the cross-platform lock configuration assertion
- 43 focused audio/video/worker tests passed
- 27 audio engine tests passed
- pre-commit passed on all modified files (Black, Ruff, isort, mypy, codespell, whitespace checks)
- workflow YAML parsed successfully


OliverBryant (approved):
Verified both fixes independently. No blockers — LGTM.

## 1. `filelock>=3.30.0` + `preserve_lock_file=True`

Checked each claim against the actual filelock sources:

- **The version floor is exact.** Diffed the 3.29.7 and 3.30.0 wheels: `preserve_lock_file` is absent in 3.29.7 and present in 3.30.0 (`filelock/_api.py:354`). `>=3.30.0` lands on precisely the right boundary.
- **The floor is load-bearing, not cosmetic.** `BaseFileLock.__call__` in 3.29.7 accepts `**kwargs` for subclass forwarding (`_api.py:128`), so passing `preserve_lock_file=True` on an older filelock raises nothing and is **silently ignored** (confirmed: `hasattr(lock, "preserve_lock_file")` is `False`, no exception). Without the pin users would get a quietly inert fix, so declaring the dependency is the right call rather than relying on it transitively.
- **The mechanism being fixed is real.** `WindowsFileLock._release()` unlinks the lock file by default, and `preserve_lock_file` is exactly what skips that unlink (upstream #605). The comment in `worker.py` about old and new acquirers locking different inodes after an unlink holds up.
- **Sibling call sites are consistent.** All three MLX-conversion `FileLock(` sites are updated; the unrelated one in `xinference/model/world/model.py:85` is correctly left alone.
- **No dependency conflict.** `virtualenv`'s `filelock<4` is the only upper bound in the surrounding ecosystem, so the new floor resolves cleanly.

Updating the `convert()` helper in `test_worker.py` to match is right — it stands in for the producer, so it has to construct the lock the same way production does.

## 2. `mlx==0.32.1`

`f5-tts-mlx` 0.2.6 declares only `mlx>=0.18.1` with no upper bound, and 0.32.2 is the current latest on PyPI (there is no 0.32.3). So without this pin a fresh child venv resolves straight to the broken version — pinning at the venv-spec layer is the appropriate workaround given the loose upstream constraint. Scoping it to the F5-TTS spec is precise (it is the only audio spec using a bare `f5-tts-mlx`), and an exact pin matches the convention already used by neighboring specs (`mlx-audio==0.4.6`, `mlx-lm==0.24.1`).

## One non-blocking note

In `test_mlx_video.py`, the new `assert Path(f"{converted_path}.lock").exists()` passes even with the fix reverted on non-Windows platforms. I mutation-tested it: the unlink only happens in the Windows `_release()` path, while the Unix implementation retains the lock file regardless of `preserve_lock_file`, so the assertion is vacuous on Linux and macOS runners.

It is not dead weight — `windows-latest` is in the CI matrix and the video tests are not in the `--ignore` list, so the Windows job does exercise it. Just noting that the guard is platform-limited rather than general. Asserting on the constructed lock's configuration instead of on file existence would make it meaningful everywhere, but the current form is acceptable.

PR #5426, opened 2026-08-27
FEAT: support configurable dynamic replica scale-up

### LINKED ISSUE #5283: [Feature request]Dynamic scaling model replicas.
### Feature request / 功能建议

Is it possible to support the dynamic addition of model replicas? Dynamic deletion of deployed replicas is already supported.

### Motivation / 动机

For instance, sometimes when we restart a specific worker for some reason and wish to add more replicas without stopping the existing ones,we do not want to stop the inference service. This approach offers greater flexibility and facilitates operations and maintenance.

### Your contribution / 您的贡献

-

leslie2046:
## Summary

Enhance dynamic replica scale-up for running models with configurable:

close https://github.com/xorbitsai/inference/issues/5283

- Model engine
- Number of replicas to add
- Device allocation (CPU/GPU)
- Worker and GPU placement

|before|after|
|---|---|
|图片|图片|

## Changes

- Add a redesigned replica scale-up dialog in the Web UI.
- Support adding multiple replicas in a single operation.
- Allow overriding the model engine for newly added replicas.
- Support CPU, GPU, automatic device selection, GPU indexes, and target Workers.
- Roll back replicas created by the current operation if any launch fails.
- Preserve backward compatibility for existing single-replica API calls.
- Normalize `n_gpu=0` to `None` before Worker launch, fixing scale-up on CPU-only servers.
- Add Python client and async client support for the new parameters.
- Update English, Chinese, Japanese, and Korean translations.
- Add API, client, and supervisor regression tests.
- Update the launch documentation.

## Validation

- Frontend production build passed.
- ESLint and TypeScript checks passed.
- Black, Ruff, isort, mypy, codespell, and repository commit hooks passed.
- Desktop and narrow-screen UI layouts were visually verified.
- Python syntax checks passed.

Full pytest execution was unavailable locally because the environment was missing the compiled `xoscar._utils` extension.


frontend/src/components/pages/running-model/add-replica-dialog.tsx
```diff
@@ -40,110 +49,226 @@ const AddReplicaDialog: FC<AddReplicaDialogProps> = ({
...
+        }
+      });
+
+    return () => {
+      active = false;
+    };
+  }, [defaultDevice, modelEngine, modelName, modelType, open]);
+
+  const engineOptions = useMemo<SelectOption<string>[]>(() => {
+    const options: SelectOption<string>[] = Object.entries(engineMap).map(([engine, metadata]) => ({
```
qinxuye: Please filter these options against the currently running model's exact spec (`model_format`, model size, and quantization), or have the backend select a compatible spec when changing engines. The engine-discovery response contains all compatible specs for each engine, but this mapping discards that metadata and enables every engine whose value is an array. For example, a PyTorch `qwen2.5-instruct` replica is offered `llama.cpp` because the model family also has a separate GGUF spec. The supervisor only replaces `model_engine` and keeps `model_format=pytorch`, so llama.cpp then rejects the launch because it only supports `ggufv2`. As written, the dialog advertises engine choices that deterministically fail.

leslie2046: Fixed. The dialog now filters engine options against the running model’s exact model_format, normalized model size, and quantization. Incompatible engines, such as llama.cpp for a PyTorch spec, are no longer offered. The currently running engine is preserved as a fallback, and regression tests were added for format, size, and quantization matching.

qinxuye (approved):
LGTM

PR #5429, opened 2026-08-28
fix(supervisor): reschedule autostart on whole-worker death

### LINKED ISSUE #5428: Autostart-configured model is never relaunched after a whole worker dies (heartbeat timeout / reverse-channel dead)
## System Info / 系統信息

Read from source, current main (66c7ed44), not a specific deployed version.

## Version info / 版本信息

main branch, HEAD 66c7ed44

## Reproduction / 复现过程

`SupervisorActor._check_dead_nodes()` (`xinference/core/supervisor.py`) is the
background task that detects a dead worker, either via heartbeat timeout or
the reverse-channel `ping()` probe. Both branches call the shared
`_handle_dead_worker()` helper, which advances every model on that worker to
`LaunchStatus.TERMINATED` when the last replica is gone, exactly the same
transition `mark_replica_dead()` performs when a replica exhausts local
auto-recovery (fixed for that path in #5409).

Autostart exists specifically to relaunch a model that goes down, but its
background task only gets woken from four places: supervisor startup,
`upsert_autostart_model`, `add_worker`, and `mark_replica_dead`. Neither call
site of `_handle_dead_worker()` inside `_check_dead_nodes()` calls
`self._schedule_autostart()`, so a whole-worker death (the box crashing, or
its reverse channel going unresponsive) never triggers a relaunch. This is
reachable under default settings, unlike #5409's fix, which needed a finite
`XINFERENCE_MODEL_ACTOR_AUTO_RECOVER_LIMIT`.

Grepping `_check_dead_nodes` for `_schedule_autostart` (`git show
66c7ed44:xinference/core/supervisor.py`) returns no hits, confirmed against
the four sites that do call it.

## Expected behavior / 期待表现

An Autostart-launched model whose last replica goes TERMINATED because its
worker died should be relaunched by Autostart, the same as when a replica
exhausts local auto-recovery. Today it is silently left TERMINATED until an
operator notices and manually relaunches it or restarts the supervisor.

PR with a fix and a regression test to follow.

AmirF194:
Fixes #5428

`_check_dead_nodes()` detects a worker dying two ways (heartbeat timeout, and
the reverse-channel `ping()` probe) and both routes go through the shared
`_handle_dead_worker()` helper, which advances any fully-gone model to
`LaunchStatus.TERMINATED`. That is the same transition `mark_replica_dead()`
performs on last-replica death, which #5409 already taught Autostart to
relaunch from. Neither call site inside `_check_dead_nodes()` calls
`self._schedule_autostart()`, so a whole-worker death leaves an
autostart-configured model TERMINATED with nothing to bring it back.

The fix adds the same `self._schedule_autostart()` call already used by
`mark_replica_dead`, `add_worker`'s re-registration path, and supervisor
startup, at both `_check_dead_nodes()` call sites. `remove_worker()` also
calls `_handle_dead_worker()` but is intentionally left out: it is a graceful
admin shutdown, and its own comment already documents that models are left
for redeploy rather than auto-recovered.

Verification:
- New regression test `test_check_dead_nodes_reschedules_autostart_on_heartbeat_timeout`
  in `xinference/core/tests/test_autostart.py`, mirroring the
  `_DummyReplicaDeathSupervisor` pattern #5409 added for `mark_replica_dead`:
  drives `_check_dead_nodes()` through one heartbeat-timeout iteration and
  asserts `_schedule_autostart` gets called. Confirmed it fails on unpatched
  `main` and passes on the branch.
- Full `xinference/core/tests/` (493 passed, excluding
  `test_continuous_batching.py` per this repo's own CI ignore list) still
  passes.
- `ruff check`, `black --check`, `isort --check`, `codespell`, and `mypy
  --ignore-missing-imports --follow-imports skip` all clean on both changed
  files.
- Not run: `test_restful_api.py` (timed out under this session's time budget
  before finishing, unrelated to the changed code path), the GPU test legs,
  and the Windows/macOS matrix; this change has no GPU or platform-specific
  code path.

### DIFF
--- a/xinference/core/supervisor.py
+++ b/xinference/core/supervisor.py
@@ -3669,6 +3669,10 @@ class SupervisorActor(xo.StatelessActor):
                     self._worker_status.pop(address, None)
                     self._worker_address_to_worker.pop(address, None)
                     self._worker_model_gpu_memory.pop(address, None)
+                if dead_nodes:
+                    # Autostart owns relaunching a model TERMINATED above, same
+                    # trigger as mark_replica_dead's last-replica-death path.
+                    self._schedule_autostart()
 
                 # ---- Reverse-channel probe ----
                 # Heartbeat only covers worker->supervisor; this probes
@@ -3743,6 +3747,7 @@ class SupervisorActor(xo.StatelessActor):
                             self._worker_status.pop(address, None)
                             self._worker_address_to_worker.pop(address, None)
                             self._worker_model_gpu_memory.pop(address, None)
+                            self._schedule_autostart()
                             dead_nodes.append(address)
                             self._reverse_ping_failures.pop(address, None)
                         else:
--- a/xinference/core/tests/test_autostart.py
+++ b/xinference/core/tests/test_autostart.py
@@ -13,11 +13,11 @@
 # limitations under the License.
 
 import asyncio
-from unittest.mock import AsyncMock, MagicMock
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 
-from xinference.core.supervisor import ReplicaInfo, SupervisorActor
+from xinference.core.supervisor import ReplicaInfo, SupervisorActor, WorkerStatus
 
 
 class _DummySupervisor:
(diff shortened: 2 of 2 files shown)


qinxuye (approved):
LGTM

AmirF194:
The one failing check here, build_test_job (ubuntu-latest, 3.12), fails inside test_log_rotation.py, a file this diff does not touch. This PR only changes supervisor.py and its own new test in test_autostart.py.

qinxuye:
> The one failing check here, build_test_job (ubuntu-latest, 3.12), fails inside test_log_rotation.py, a file this diff does not touch. This PR only changes supervisor.py and its own new test in test_autostart.py.

yeah, never mind, merge now.

AmirF194:
Thanks for the quick look and merge. Appreciate it, and glad the sibling fix from #5409 held up here too.

PR #5430, opened 2026-08-28
feat: Add Gemma-4 Transformers backend with batching support

maoyuehui:
## Summary

- add the missing legacy and layer-based DynamicCache compatibility helpers used by continuous batching
- preserve sliding-cache state and shared layers while merging and reducing batches
- move processor-based direct chat generation into a shared mixin instead of duplicating it in `gemma4.py`
- support Gemma 4 Unified 12B with Transformers 5.10 and pass tools to the direct chat template
- add focused cache compatibility, direct chat, usage, tool-template, and virtualenv dependency tests

## Release note

- Add the Gemma 4 Transformers backend with complete continuous-batching support across legacy and layer-based DynamicCache layouts.

## Validation

- `pre-commit run --files <changed files>`
- `pytest -q xinference/model/llm/transformers/tests/test_gemma4.py xinference/model/llm/transformers/tests/test_cache_compatibility.py xinference/model/llm/tests/test_utils.py` (49 passed)
- OPT integration launch was attempted separately; its model setup could not complete locally because the host ran out of disk space, while the unrelated FP4 case is not present in the current catalog.

### DIFF
--- a/xinference/model/llm/transformers/gemma4.py
+++ b/xinference/model/llm/transformers/gemma4.py
@@ -1,0 +1,339 @@
+# Copyright 2022-2026 Xinference Holdings Pte. Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import uuid
+from threading import Thread
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Union, cast
+
+from ....core.model import register_batching_multimodal_models
+from ....types import ChatCompletion, ChatCompletionChunk, CompletionChunk
+from ...utils import cache_clean
+from ...scheduler.request import InferenceRequest
+from ..utils import generate_completion, generate_completion_chunk
+from ..llm_family import LLMFamilyV2, LLMSpecV1, register_transformer
+from .core import PytorchChatModel, register_non_default_model
+
+
+@register_batching_multimodal_models("gemma-4")
+@register_transformer
+@register_non_default_model(
+    "Gemma4ForConditionalGeneration", "Gemma4UnifiedForConditionalGeneration"
+)
+class Gemma4ChatModel(PytorchChatModel):
+    GEMMA4_ARCHITECTURES = {"Gemma4ForConditionalGeneration"}
+    GEMMA4_UNIFIED_ARCHITECTURES = {"Gemma4UnifiedForConditionalGeneration"}
+    GEMMA4_MIN_TRANSFORMERS_VERSION = "5.5.0"
+    GEMMA4_UNIFIED_MIN_TRANSFORMERS_VERSION = "5.10.0"
+
+    @classmethod
+    def check_lib(cls) -> Union[bool, Tuple[bool, str]]:
+        result = super().check_lib()
+        if result is not True:
+            return result
+
+        import transformers
+        from packaging.version import Version
+
+        if Version(transformers.__version__) < Version(
+            cls.GEMMA4_MIN_TRANSFORMERS_VERSION
+        ):
+            return (
+                False,
+                f"Gemma-4 requires transformers>={cls.GEMMA4_MIN_TRANSFORMERS_VERSION}",
+            )
+        return True
+
+    @classmethod
+    def match_json(
+        cls, model_family: "LLMFamilyV2", model_spec: "LLMSpecV1", quantization: str
+    ) -> Union[bool, Tuple[bool, str]]:
+        if model_spec.model_format not in ["pytorch", "gptq", "awq", "bnb", "fp4"]:
+            return (
+                False,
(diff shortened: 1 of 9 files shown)


xinference/model/llm/transformers/gemma4.py
```diff
@@ -0,0 +1,114 @@
...
+from ....core.model import register_batching_multimodal_models
+from ...scheduler.request import InferenceRequest
+from ..llm_family import LLMFamilyV2, LLMSpecV1, register_transformer
+from .core import PytorchChatModel, register_non_default_model
+
+
+@register_batching_multimodal_models("gemma-4")
+@register_transformer
+@register_non_default_model("Gemma4ForConditionalGeneration")
+class Gemma4ChatModel(PytorchChatModel):
```
qinxuye: Please implement the non-batching chat path before registering this backend. `Gemma4ChatModel` inherits `PytorchChatModel._direct_chat()`, which always raises `NotImplementedError`. This is reachable whenever `max_num_seqs=1` disables the batch scheduler, and the regression test added in this PR currently fails on this exact call; the Linux, macOS, and Windows main test jobs all fail for the same reason. Please implement `_direct_chat` for streaming and non-streaming requests (including multimodal processor inputs) and keep the regression test exercising the real implementation.

xinference/model/llm/transformers/gemma4.py
```diff
@@ -0,0 +1,114 @@
...
+    @classmethod
+    def check_lib(cls) -> Union[bool, Tuple[bool, str]]:
+        result = super().check_lib()
+        if result is not True:
+            return result
+
+        import transformers
+        from packaging.version import Version
+
+        if Version(transformers.__version__) < Version("5.5.0"):
```
qinxuye: Please account for the Unified 12B checkpoint instead of allowing Transformers 5.5 for every built-in Gemma-4 PyTorch spec. The catalog includes `google/gemma-4-12B-it`, whose official config uses `Gemma4UnifiedForConditionalGeneration` / `model_type=gemma4_unified`; that implementation is absent in Transformers 5.5 and first appears in 5.10. Here `check_lib()` returns true on 5.5, while the family-level architecture match still selects this backend for the 12B spec, so `AutoModelForCausalLM.from_pretrained()` fails during launch. Please support the Unified architecture with the appropriate minimum version, or explicitly exclude that spec/version combination, and add a regression test. Official config: https://huggingface.co/google/gemma-4-12B-it/blob/main/config.json

qinxuye (commented):
Found two blocking issues on the current head.

xinference/model/llm/transformers/gemma4.py
```diff
@@ -0,0 +1,214 @@
...
+                "Gemma4 transformer supports pytorch/gptq/awq/bnb/fp4 formats only",
+            )
+        if not model_family.has_architecture(
+            *cls.GEMMA4_ARCHITECTURES, *cls.GEMMA4_UNIFIED_ARCHITECTURES
+        ):
+            return (
+                False,
+                f"Model architectures {model_family.architectures} are not Gemma-4-it",
+            )
+        if cls._is_unified_model_spec(model_spec):
```
qinxuye: Please make the virtualenv engine selection resolve this spec to `Gemma4ChatModel` independently of the host Transformers version. With the supported host Transformers 5.5.0 and the default `XINFERENCE_ENABLE_VIRTUAL_ENV=1`, this version-dependent `match_json()` result prevents the 12B spec from being registered under the Gemma-4 backend. I reproduced `check_engine_by_spec_parameters_with_virtual_env("Transformers", "gemma-4", "pytorch", 12, "none", ...)` selecting `QwenOmniChatModel`, so installing Transformers >=5.10 inside the model virtualenv never fixes the launch—the wrong class has already been chosen. Please add a regression test asserting that the 12B virtualenv path selects `Gemma4ChatModel` on a pre-5.10 host.

xinference/model/llm/transformers/gemma4.py
```diff
@@ -0,0 +1,214 @@
...
+        model = model_cls.from_pretrained(
+            self.model_path,
+            **kwargs,
+        )
+        self._processor = processor
+        self._device = model.device
+        return model, tokenizer
+
+    def _get_full_prompt(self, messages: List[Dict], tools, generate_config: dict):
+        return self._transform_messages(messages)
```
qinxuye, line 130: Please preserve request-level chat-template arguments in the batching path. `PytorchChatModel.prepare_batch_inference()` passes `tools` and `generate_config` into this override, but this return value drops both; `build_prefill_kwargs()` then calls `apply_chat_template()` without either `tools` or `enable_thinking`. I reproduced the default batching path (`max_num_seqs=16`) receiving only the tokenize/padding kwargs, so the advertised tools and hybrid reasoning controls work in direct chat but not in the normal batched path. Please thread these per-request template arguments into prefill and cover batching with tools and `enable_thinking` in a regression test.

amumu96: Fixed in ed875860. Gemma 4 batching now forwards tools and chat_template_kwargs (including enable_thinking) to the processor. The scheduler keeps prefill requests with different template inputs in separate batches so request-level tools/reasoning settings cannot leak across requests. Added forwarding and grouping regression tests.

xinference/model/llm/transformers/gemma4.py
amumu96: Fixed in ed875860. Gemma4ChatModel.match_json() now selects the Unified 12B backend independently of the host Transformers version; the >=5.10 check runs immediately before model loading instead. Added a regression test that exercises check_engine_by_spec_parameters_with_virtual_env() on a 5.5 host and verifies Gemma4ChatModel is selected.

qinxuye (approved):
LGTM

PR #5432, opened 2026-08-28
FIX: stabilize HY-WorldPlay runtime and progress reporting

qinxuye:
## Summary

- reuse the host PyTorch and torchvision stack for HY-WorldPlay virtual environments, avoiding incompatible CUDA wheel resolution on GB10
- make World model instances serializable when they cross the actor-process boundary
- expose HY-WorldPlay progress in the Web UI and report weight loading, chunk generation, decoding, and video saving stages

## Validation

- pytest -q xinference/model/world/tests/test_world_model.py (39 passed)
- pre-commit run on all modified files
- ESLint on capability-task-panel.tsx

### DIFF
--- a/doc/source/models/model_abilities/flexible.rst
+++ b/doc/source/models/model_abilities/flexible.rst
@@ -1,8 +1,8 @@
 .. _flexible:
 
-====================================
-Traditional ML models (Experimental)
-====================================
+=====================
+Traditional ML models
+=====================
 
 Learn how to inference traditional machine learning models with Xinference.
 These flexibly extensible models are referred to as **Flexible Models** within Xinference.
--- a/doc/source/models/model_abilities/flexible.rst
+++ b/doc/source/models/model_abilities/flexible.rst
@@ -8,7 +8,6 @@ Learn how to inference traditional machine learning models with Xinference.
 These flexibly extensible models are referred to as **Flexible Models** within Xinference.
 
 .. versionadded:: v1.7.1
-  This ability is public since v1.7.1, now the API is not stable and may change during evolving.
 
 
 Introduction
--- a/frontend/src/components/pages/running-model-detail/panels/capability-task-panel.tsx
+++ b/frontend/src/components/pages/running-model-detail/panels/capability-task-panel.tsx
@@ -75,7 +75,9 @@ const CapabilityTaskPanel = forwardRef<CapabilityTaskPanelMethod, CapabilityTask
       config.showProgress &&
       (model.model_type !== ModelType.World ||
         model.model_family === 'Astra' ||
-        model.model_name === 'Astra')
+        model.model_name === 'Astra' ||
+        model.model_family === 'HY-WorldPlay' ||
+        model.model_name === 'HY-WorldPlay-5B')
     );
 
     const showCopyResult = useMemo(() => {
--- a/xinference/model/world/model.py
+++ b/xinference/model/world/model.py
@@ -213,6 +213,24 @@ class WorldModel:
         self._running_processes: Dict[str, subprocess.Popen] = {}
         self._request_cancellations: Dict[str, threading.Event] = {}
 
+    def __getstate__(self):
+        # Model instances are serialized when handed to the actor subprocess.
+        # Synchronization primitives and live process state cannot cross that
+        # boundary; the instance has not started serving requests at this point.
+        state = self.__dict__.copy()
+        state["_process_lock"] = None
+        state["_runner_lock"] = None
+        state["_running_processes"] = {}
+        state["_request_cancellations"] = {}
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self._process_lock = threading.Lock()
+        self._runner_lock = threading.Lock()
+        self._running_processes = {}
+        self._request_cancellations = {}
+
     @property
     def model_ability(self):
         return self._model_spec.model_ability
--- a/xinference/model/world/model.py
+++ b/xinference/model/world/model.py
@@ -823,22 +823,33 @@ class HYWorldPlayModel(WorldModel):
                     nonlocal current_chunk
                     if progressor is None:
                         return
-                    if match := re.search(r"XINFERENCE_PROGRESS:([0-9.]+):(.+)", line):
-                        progressor.set_progress(float(match.group(1)), match.group(2))
-                    elif match := re.search(
-                        r"Generate time for chunk\s+(\d+)\s+is", line
-                    ):
-                        current_chunk = int(match.group(1))
-                        completed = (current_chunk + 0.7) / num_chunks
-                        progressor.set_progress(
-                            0.18 + 0.72 * completed,
-                            f"Generated chunk {current_chunk + 1}/{num_chunks}",
-                        )
-                    elif "Decode latent 0:" in line and current_chunk >= 0:
-                        completed = (current_chunk + 1) / num_chunks
-                        progressor.set_progress(
-                            0.18 + 0.72 * completed,
-                            f"Decoded chunk {current_chunk + 1}/{num_chunks}",
+                    try:
+                        if match := re.search(
+                            r"XINFERENCE_PROGRESS:([0-9.]+):(.+)", line
+                        ):
+                            progressor.set_progress(
+                                float(match.group(1)), match.group(2)
+                            )
+                        elif match := re.search(
+                            r"Generate time for chunk\s+(\d+)\s+is", line
+                        ):
+                            current_chunk = int(match.group(1))
+                            completed = (current_chunk + 0.7) / num_chunks
+                            progressor.set_progress(
+                                0.18 + 0.72 * completed,
+                                f"Generated chunk {current_chunk + 1}/{num_chunks}",
+                            )
+                        elif "Decode latent 0:" in line and current_chunk >= 0:
+                            completed = (current_chunk + 1) / num_chunks
+                            progressor.set_progress(
+                                0.18 + 0.72 * completed,
+                                f"Decoded chunk {current_chunk + 1}/{num_chunks}",
+                            )
+                    except Exception:
+                        logger.warning(
+                            "Failed to parse HY-WorldPlay progress output: %r",
+                            line[:200],
+                            exc_info=True,
                         )
 
                 if progressor:
--- a/xinference/model/world/model.py
+++ b/xinference/model/world/model.py
@@ -815,6 +815,32 @@ class HYWorldPlayModel(WorldModel):
                         env.get("PYTHONPATH", ""),
                     ]
                 ).rstrip(os.pathsep)
+
+                num_chunks = int(config["num_chunk"])
+                current_chunk = -1
+
+                def update_progress(line: str) -> None:
+                    nonlocal current_chunk
+                    if progressor is None:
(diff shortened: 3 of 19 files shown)


xinference/model/world/model.py
```diff
@@ -797,13 +815,40 @@ def world_generate(
...
+                        progressor.set_progress(
+                            0.18 + 0.72 * completed,
+                            f"Generated chunk {current_chunk + 1}/{num_chunks}",
+                        )
+                    elif "Decode latent 0:" in line and current_chunk >= 0:
+                        completed = (current_chunk + 1) / num_chunks
+                        progressor.set_progress(
+                            0.18 + 0.72 * completed,
+                            f"Decoded chunk {current_chunk + 1}/{num_chunks}",
+                        )
```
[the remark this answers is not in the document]
qinxuye: Fixed in c2f3e7d16. HY-WorldPlay progress parsing is now best-effort: parser or reporter exceptions are logged and ignored, and the test covers a malformed progress marker followed by valid updates.

rogercloud (commented):
## Summary

This PR extends Xinference's World-model support around HY-WorldPlay by replacing child-environment PyTorch pins with host-stack markers, restoring fresh runtime state when `WorldModel` crosses into the model actor, and emitting best-effort generation progress. It connects the existing request-ID progress tracker to task-panel polling and updates World navigation, ability titles, and Chinese World documentation. Together, these runtime and documentation changes aim to make WorldPlay launch and run through the existing worker/actor infrastructure while exposing progress in the UI.
Blocking: yes — recommended event: COMMENT

## Round 0 approach verdict

`acceptable-with-reservations`

The approach leverages existing virtual-environment markers, worker-to-actor serialization hooks, Progressor/stdout callbacks, and documentation/navigation seams; parser telemetry is appropriately best-effort so malformed output does not fail inference. The host dependency contract is under-specified and is represented by the blocking line-level finding below. The stage metadata/UI coupling reservation was dropped after verifying that the public progress contract is pre-existing and numeric-only, so there is no separate stage/API finding.

## Update summary

Since the last substantive review at `5ac3efdc`, three commits were added: `c2f3e7d1` (`fix: keep WorldPlay progress parsing best-effort`) makes progress parsing errors non-fatal; `511dbf4d` (`docs: add World navigation and refresh ability titles`) adds World navigation and refreshes ability titles; and `5192c68b` (`docs: add Chinese world model translations`) adds Chinese World-model translations. The first update independently resolves the parser-exception root, while the documentation commits expand navigation/localization and the current review reports the fresh findings below.

## Prior-findings checklist

| Canonical root and preserved history | Status | Independent current-code evidence |
|---|---|---|
| Parser exceptions could crash WorldPlay generation — review-body ID `5050767000`; inline root ID `3880312651` at `xinference/model/world/model.py:853`; reply ID `3880344361` (`in_reply_to_id=3880312651`) | **FIXED** | `xinference/model/world/model.py:822-853` now catches regex, conversion, arithmetic, and reporter errors; current tests cover a malformed marker. The qinxuye reply is a fix claim independently verified against this code, not a waiver. No duplicate follow-up comment is needed for this thread. The complete root/reply chain is represented by these source IDs; no linked issue or PR discussion was present. |

## Confirmed fresh findings

### 1. Host placeholders erase the HY-WorldPlay torch floor — `xinference/model/world/model_spec.json:98-99`

- **Severity:** major
- **Blocking:** yes
- **Trigger:** A supported NVIDIA host passes `PyTorchWorldEngineModel.check_host` with CUDA available but has torch below the upstream HY requirement (for example, torch 2.5 with torchvision 0.20), or lacks torchvision. The new `#system_torch#` and `#system_torchvision#` entries reuse the host versions instead of enforcing the former child-environment floor and companion requirement.
- **Observable impact:** The worker accepts and inherits that host stack, so package preparation can fail for a missing companion or model loading/generation can fail with the old or incompatible stack. Device reservation and model download/cache work occur before virtual-environment package preparation, making the failure late rather than an early launch rejection.
- **Evidence:** `xinference/model/world/model_spec.json:98-99` changed explicit `torch>=2.6.0` and pinned companion requirements to system markers; `xinference/model/world/engine.py:25-32` checks only CUDA/NVML and declares only torch; `xinference/core/worker.py:3834-3848,3941-3972,3999-4017` reserves devices, creates/downloads the model, and prepares the child environment with inherited system packages; the pinned upstream requirements still require `torch>=2.6.0`.
- **Suggestion:** Keep host reuse but preserve an enforceable torch minimum and compatible torchvision pair, or preflight host distribution presence, versions, and ABI compatibility and reject/fallback before downloads and device reservation. Add regression cases for old torch, missing torchvision, and incompatible pairs.

### 2. Add regression coverage for the HY-WorldPlay polling gate — `frontend/src/components/pages/running-model-detail/panels/capability-task-panel.tsx:79`

- **Severity:** minor
- **Blocking:** no
- **Trigger:** The new exact family/name predicate at lines 74-81 is changed, misspelled, or no longer matches a future model identity/response shape, and no component test catches the branch change.
- **Observable impact:** Request-ID progress polling is silently skipped for HY-WorldPlay, so generation can still succeed while the UI shows indeterminate progress.
- **Evidence:** `frontend/src/components/pages/running-model-detail/panels/capability-task-panel.tsx:74-81` controls polling at `:179-180` and `:231-232`; the backend identity is exactly `HY-WorldPlay`/`HY-WorldPlay-5B`; no existing component/panel tests cover this branch, and backend parser tests do not exercise it.
- **Suggestion:** Add a focused component or pure-helper test with HY-WorldPlay true and Matrix false, asserting request-ID forwarding, polling, 0..1-to-percent rendering, completion termination, and reset cancellation.

### 3. Restore Flexible's stability qualifier — `doc/source/models/model_abilities/flexible.rst:4`

- **Severity:** minor
- **Blocking:** no
- **Trigger:** The English title changed from `Traditional ML models (Experimental)` to unqualified `Traditional ML models`, while unchanged lines 10-11 say that the API is not stable and may change. The corresponding Chinese PO title likewise lost the qualifier while its warning at lines 34-39 remains.
- **Observable impact:** User-facing documentation presents contradictory stability signals for Flexible.
- **Evidence:** `doc/source/models/model_abilities/flexible.rst:4` removes `Experimental`, while the page's instability warning remains; the Chinese title has the same change. The PR contains no Flexible graduation or API change, and World remains labeled Experimental.
- **Suggestion:** Restore the qualifier in both locales, or explicitly update/remove the warning and document an intentional graduation if that change is intended.

## Simplification lens

No confirmed simplification opportunities were identified. The full-PR Simplification Lens returned `Lean already.`; the mandatory update-only lens over `5ac3efdc...HEAD` was unavailable because the model hit its usage limit.

## Blocking status & recommended decision

**Blocking: yes.**

- `xinference/model/world/model_spec.json:98`, major — host placeholders can accept a too-old or incomplete torch/torchvision stack and lead to late HY-WorldPlay model-load/generation failure. `[new]`

Recommended event: `COMMENT` because the PR author is qinxuye.

xinference/model/world/model_spec.json
```diff
@@ -95,9 +95,8 @@
     },
     "virtualenv": {
       "packages": [
-        "torch>=2.6.0 ; #engine# == \"pytorch\"",
-        "torchvision==0.21.0",
-        "torchaudio==2.6.0",
+        "#system_torch# ; #engine# == \"pytorch\"",
```
rogercloud, line 98: **Major; Blocking: yes.** The new system markers reuse whatever host torch/torchvision versions are present, while `PyTorchWorldEngineModel.check_host` only verifies CUDA and HY-WorldPlay's pinned requirements still require `torch>=2.6.0` plus a compatible torchvision. A CUDA host with torch 2.5/torchvision 0.20 or missing torchvision can pass the host check and fail during preparation or load/generation after reservation/download side effects. Please keep an enforceable minimum/pair check or preflight and reject/fallback before those side effects; add old/missing/incompatible-host cases.

frontend/src/components/pages/running-model-detail/panels/capability-task-panel.tsx
```diff
@@ -75,7 +75,9 @@ const CapabilityTaskPanel = forwardRef<CapabilityTaskPanelMethod, CapabilityTask
       config.showProgress &&
       (model.model_type !== ModelType.World ||
         model.model_family === 'Astra' ||
-        model.model_name === 'Astra')
+        model.model_name === 'Astra' ||
+        model.model_family === 'HY-WorldPlay' ||
```
rogercloud, line 79: **Minor; Blocking: no.** The exact family/name gate here controls whether request-ID progress polling runs. A future identity or response-shape typo could silently leave HY-WorldPlay with indeterminate UI while generation still succeeds. Please add a focused panel/helper test for HY true, Matrix false, request-ID forwarding, 0..1-to-percent rendering, polling completion, and reset cancellation.

doc/source/models/model_abilities/flexible.rst
```diff
@@ -1,8 +1,8 @@
 .. _flexible:
 
-====================================
-Traditional ML models (Experimental)
-====================================
+=====================
+Traditional ML models
```
rogercloud, line 4: **Minor; Blocking: no.** `Traditional ML models` now drops `Experimental`, but the page still says the API is unstable and may change; the corresponding Chinese title drops the qualifier while its warning remains. This gives users contradictory stability signals. Please restore the qualifier in both locales, or intentionally update/remove the warning if Flexible has graduated.

xinference/model/world/model_spec.json
qinxuye: Fixed in cb99efb7f. The preflight is deliberately scoped to HY-WorldPlay because Matrix and Astra install their own pinned runtime stacks. HY-WorldPlay now rejects host torch below 2.6, missing torchvision, and a torch/torchvision pair that violates the installed torchvision Requires-Dist constraint before virtualenv or model preparation. Compatibility is derived from package metadata rather than a hard-coded version table, so vendor and local CUDA versions such as 2.9.0+cu130 remain supported. Focused tests cover a valid local-suffix pair plus old, missing, and incompatible hosts.

frontend/src/components/pages/running-model-detail/panels/capability-task-panel.tsx
qinxuye: No code change here. This request conflates the four-line HY-WorldPlay allowlist delta with the existing generic CapabilityTaskPanel lifecycle. Request-ID forwarding, progress normalization, polling termination, and reset/abort behavior are unchanged shared code paths; this PR only adds the exact built-in identifiers emitted by the current model spec, HY-WorldPlay and HY-WorldPlay-5B, while Matrix remains excluded. A hypothetical future rename or response typo is not a current defect. This frontend has no component-test harness, so introducing one and retrofitting the full pre-existing polling lifecycle is outside this targeted runtime fix and should not gate it.

doc/source/models/model_abilities/flexible.rst
qinxuye: Fixed in cb99efb7f. Dropping Experimental was intentional, so restoring the qualifier would directly undo the requested documentation change. I removed the stale instability sentence from the English source and the zh_CN catalog instead, while retaining the v1.7.1 version-added history.

OliverBryant (approved):
LGTM

PR #5437, opened 2026-08-29
FEAT(model): Add Breeze-TTS-2 support

Minamiyama:
## Summary

Add Breeze-TTS-2 as a built-in audio model using its native PyTorch/CUDA streaming runtime.

- Register `BreezeBlue/Breeze-TTS-2` from Hugging Face (`main`) and ModelScope (`master`).
- Support text-to-speech, natural-language voice design, zero-shot voice cloning, and streaming/non-streaming output.
- Vendor the required upstream inference runtime with package-relative imports.
- Expose optional CUDA Graph acceleration through `fast_all` and per-stage settings.
- Add model documentation, ability listings, and focused tests.

### DIFF
--- a/xinference/model/audio/breeze_tts.py
+++ b/xinference/model/audio/breeze_tts.py
@@ -1,0 +1,391 @@
+# Copyright 2022-2026 Xinference Holdings Pte. Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import math
+import os
+import tempfile
+import uuid
+from dataclasses import replace
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Iterator, Optional
+
+if TYPE_CHECKING:
+    import torch
+
+    from .core import AudioModelFamilyV2
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_INSTRUCTION = "Speak clearly and naturally."
+_DEFAULT_MAX_NEW_TOKENS = 1500
+_DEFAULT_MAX_SEQ_LEN = 2048
+_DEFAULT_REPETITION_PENALTY = 1.1
+_FAST_CONFIG = (
+    Path(__file__).resolve().parents[2]
+    / "thirdparty"
+    / "breeze_tts"
+    / "configs"
+    / "fast.json"
+)
+
+
+def _load_breeze_runtime_components():
+    from ...thirdparty.breeze_tts.breeze_infer.runtime import (
+        load_runtime,
+        resolve_device,
+        set_all_seeds,
+        update_generation_config_for_breeze,
+    )
+    from ...thirdparty.breeze_tts.breeze_infer.templates import (
+        get_template,
+        prepare_inputs,
+    )
+    from ...thirdparty.breeze_tts.models.fast_streaming import (
+        FastBreezeStreamingRuntime,
+        FastStreamingConfig,
+    )
+    from ...thirdparty.breeze_tts.models.warmup_profile import load_warmup_profile
+
+    return (
+        load_runtime,
+        resolve_device,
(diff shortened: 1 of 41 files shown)


xinference/model/audio/breeze_tts.py
```diff
@@ -0,0 +1,391 @@
...
+            voice=voice,
+            prompt_speech=prompt_speech,
+            prompt_text=prompt_text,
+            instruction=instruction,
+            cfg_scale=cfg_scale,
+            speaker=speaker,
+        )
+        response_format = response_format or "mp3"
+        if stream:
+            return self._stream_audio(inputs, request_id, response_format, seed)
```
qinxuye: [P1] Serialize the runtime for the entire streaming lifetime. `ModelActor._call_wrapper` releases its per-model lock as soon as this generator is enqueued, so a concurrent non-streaming `speech()` call can run `_generate_audio()` while this generator is still advancing. Both paths mutate the same `FastBreezeStreamingRuntime`, whose codec is configured with `max_active_reqs=1`; opening the second request can evict or overwrite the first request state. I reproduced this by pausing a stream after its first chunk and starting a non-streaming request, which entered the runtime concurrently. The upstream Breeze server holds a single-concurrency lock until the stream closes. Please hold a model-owned lock across both the full generator lifetime and the non-streaming path, and add a mixed stream/non-stream regression test.

xinference/model/audio/model_spec.json
```diff
@@ -2629,6 +2629,43 @@
...
+      "text2audio",
+      "text2audio_voice_design",
+      "text2audio_voice_cloning"
+    ],
+    "multilingual": true,
+    "virtualenv": {
+      "packages": [
+        "transformers==4.57.3",
+        "qwen-tts==0.1.1",
+        "flash-attn==2.8.3",
```
qinxuye: [P2] Avoid making FlashAttention a mandatory source build for the default eager path. `flash-attn==2.8.3` is published on PyPI as an sdist only, and its build needs a CUDA development toolchain. As a result, a supported Linux/NVIDIA runtime-only host can fail while creating the model virtualenv before Breeze is loaded, even though `BreezeTTS2Model.load()` defaults to `attn_implementation="eager"` and does not use FlashAttention. Please install this only when `flash_attention_2` is requested, provide a compatible prebuilt wheel, or explicitly enforce/document the required build toolchain.

qinxuye (commented):
Two issues need to be addressed before approval.

xinference/model/audio/breeze_tts.py
qinxuye: The new lock still starts inside `_stream_audio` / `_generate_audio`, after `speech()` has already called `_prepare_request`. I reproduced this by pausing a stream after its first chunk and starting a non-streaming call: the second request entered `_prepare_inputs` while the first still held `_inference_lock` (`second_prepare_entered_while_stream_lock_held=True`). For voice cloning, that means `audio_tokenizer.encode` can overlap the active stream's decode on the same tokenizer. The upstream Breeze server acquires `_request_lock` before `prepare_inputs` and releases it only after the streaming body closes. Please move preparation under the same lock and update the mixed test to assert that the second preparation cannot enter, rather than checking only the second `iter_audio_chunks` call.

xinference/model/audio/breeze_tts.py
```diff
@@ -0,0 +1,420 @@
...
+    ):
+        self._attn_implementation = kwargs.get("attn_implementation") or "eager"
+        if self._attn_implementation == "flash_attention_2":
+            virtualenv = getattr(model_spec, "virtualenv", None)
+            if (
+                virtualenv is not None
+                and _FLASH_ATTN_PACKAGE not in virtualenv.packages
+            ):
+                model_spec = deepcopy(model_spec)
+                model_spec.virtualenv.packages.append(_FLASH_ATTN_PACKAGE)
```
qinxuye, line 131: [P1] Fix the current mypy failure before this can merge. CI reports `Item "None" of "VirtualEnvSettings | None" has no attribute "packages"` here. The earlier `virtualenv is not None` check narrows the local variable, but that narrowing does not apply to `deepcopy(model_spec).virtualenv`. Please narrow or assert the copied value before appending. Because lint fails at this step, all Python test jobs on the current head are skipped.

xinference/model/audio/breeze_tts.py
qinxuye: Rechecked the latest head `f329b312`. The only new change adds the `virtualenv` type assertion, so the mypy failure is fixed, but request preparation still happens before `_inference_lock` is acquired. This concurrency issue therefore remains unresolved.

qinxuye (approved):

PR #5438, opened 2026-08-29
FIX: preserve available model download hubs

Minamiyama:
## Summary

Fix the download hub selector incorrectly hiding supported model sources.

Some catalog responses contain a complete model-level `download_hubs` list but only a partial or deduplicated `model_specs` list. For example, FireRedTTS3-Instruct reports both Hugging Face and ModelScope in `download_hubs`, while `model_specs` may contain only the Hugging Face entry. The previous frontend logic treated the partial specification list as authoritative and hid ModelScope.

## Changes

- Use model-level `download_hubs` when no model engine is selected.
- Preserve engine-specific hub filtering after an engine is selected.
- Fall back to model-level hubs when engine metadata is missing or incomplete.
- Fall back to specification hubs when `download_hubs` is unavailable.
- Deduplicate generated download hub options.
- Keep the implementation inside `launch-dialog.tsx`.

### DIFF
--- a/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
+++ b/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
@@ -344,19 +344,29 @@ export default function LaunchDialog({
   }, [gpuAvailable, modelType, form]);
 
   const downloadHubOptions = useMemo(() => {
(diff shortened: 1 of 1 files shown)


qinxuye (approved):
LGTM

PR #5446, opened 2026-08-30
fix(llm): emit streaming tool call metadata once

m199369309:
## Summary
- Emit tool-call `id` and `type` metadata only in the first streaming chunk for each tool call.
- Keep subsequent chunks focused on argument deltas.
- Preserve compatibility for stateless tool-call generation.

## Tests
- `pytest -q xinference/model/llm/tests/test_utils.py` (36 passed)
- Pre-commit checks passed.

### DIFF
--- a/xinference/model/llm/utils.py
+++ b/xinference/model/llm/utils.py
@@ -55,6 +55,7 @@ from ...types import (
     CompletionChunk,
     CompletionLogprobs,
     CompletionUsage,
+    ToolCallDelta,
 )
 from .core import chat_context_var
 from .reasoning_parser import ReasoningParser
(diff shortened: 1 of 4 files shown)


m199369309:
The review feedback has been addressed.

I changed `dict.setdefault("sent_metadata", set())` to lazy initialization: a `set()` is created only when `sent_metadata` is missing. This avoids allocating a temporary object on every invocation in the streaming tool-call hot path.

Validation:
- `pytest -q xinference/model/llm/tests/test_utils.py`: 36 passed
- Pre-commit checks for `xinference/model/llm/utils.py`: passed

Fixing commit: `cf493bb9`

xinference/model/llm/utils.py
```diff
@@ -992,14 +1002,14 @@ def _post_process_completion_chunk(
...
-                        "id": call_id,
-                        "type": "function",
-                        "function": function_delta,
-                    }
-                )
+                tool_call_delta: Dict[str, Any] = {
+                    "index": tool_call_index,
+                    "function": function_delta,
+                }
+                if include_metadata:
```
qinxuye, line 1009: [P1] Update the MLX regression test together with this stream-shape change. The current Metal CI fails in `test_mlx_streaming_parses_multiple_qwen_tool_calls` with `KeyError: 'id'` because that test still reads `id` from every delta. Please assert that each tool-call index emits `id`/`type` exactly once on its first delta, compare those first IDs across indices, and continue aggregating argument deltas from every chunk.

qinxuye (approved):

PR #5447, opened 2026-08-30
feat(vllm): support native multiprocessing executor routing

m199369309:
## Summary
- Add explicit `auto`, `native_mp`, and `xoscar` executor routing for vLLM.
- Use vLLM native multiprocessing for eligible single-worker, multi-GPU deployments.
- Validate vLLM V1, GPU allocation, and tensor/pipeline parallelism before selecting native MP.
- Force the `spawn` multiprocessing method for the native route and retain the existing xoscar route as the fallback.

## Tests
- `pytest -q xinference/model/llm/vllm/tests/test_executor_routing.py` (24 passed)
- Pre-commit checks passed.

### DIFF
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -92,6 +92,7 @@ if TYPE_CHECKING:
 
 
 class VLLMModelConfig(TypedDict, total=False):
+    xinference_vllm_executor_backend: str
     tokenizer_mode: Optional[str]
     trust_remote_code: bool
     tensor_parallel_size: int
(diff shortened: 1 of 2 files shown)


m199369309:
The two issues identified by Gemini Code Assist have been fixed and pushed to this pull request.

- `_get_allocated_device_count()`: falls back to `self._get_cuda_count()` when the accelerator list, the initialized `_device_count`, and explicit TP/PP settings are unavailable. This prevents multi-GPU configurations from being incorrectly treated as single-device launches before `load()` and avoids creating an unnecessary xoscar pool.
- `_get_native_mp_parallelism()`: explicitly handles TP/PP settings set to `None` by applying the defaults, while preserving validation for non-positive values.

Validation:
- `pytest -q xinference/model/llm/vllm/tests/test_executor_routing.py`: 28 passed
- Pre-commit checks (Black, Ruff, isort, mypy, and codespell): passed

Latest commit: `2de5a037 fix(vllm): harden native executor routing fallbacks`

qinxuye (approved):
LGTM

PR #5449, opened 2026-08-30
fix(venv): protect virtual environments in use

m199369309:
## Summary
- Persist deterministic virtual-environment setup fingerprints in restart-safe markers.
- Track preparing and active model users per virtual environment.
- Prevent dependency mutation or deletion while models are using an environment.
- Return HTTP 409 when virtual-environment removal conflicts with active models.

## Tests
- `pytest -q xinference/core/tests/test_utils.py xinference/core/tests/test_virtual_env_manager.py xinference/core/tests/test_virtual_env_deletion.py` (144 passed)
- Admin conflict test passed (4 selected tests).
- Pre-commit checks passed.

### DIFF
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -14,6 +14,7 @@
 
 import asyncio
 import errno
+import hashlib
 import json
 import logging
 import os
@@ -33,6 +34,7 @@ from logging import getLogger
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     Dict,
     List,
     Literal,
@@ -114,6 +116,7 @@ from .utils import (
     purge_dir,
     rewrite_direct_url_packages_for_index,
 )
+from .virtual_env_manager import VirtualEnvConflictError
 from .virtual_env_manager import VirtualEnvManager as XinferenceVirtualEnvManager
 from .virtual_env_manager import (
     ensure_system_torch_pin,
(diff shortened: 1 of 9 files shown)


m199369309:
Thank you for the review. I verified the suggestions as follows:

- `xinference/core/worker.py` explicitly imports `threading`, `time`, and `pathlib`, so the uses in `_mark_venv_setup_done()`, `_reserve_virtual_env_usage()`, and the two `pathlib.Path(venv_path).name` calls do not cause `NameError` exceptions.
- `xinference/core/virtual_env_manager.py` explicitly imports `Path`, so `Path(real_env_path).relative_to(model_root).parts` is also valid.
- The relevant code passed `python -m compileall`, and the related tests passed with `169 passed, 1 skipped` at the time of this review.

Therefore, the reported missing imports are false positives. The suggestions to replace these calls with `os.path.basename` and `os.path.relpath` are style changes rather than fixes for an actual runtime issue. To avoid unrelated behavioral changes, the current implementation is retained.

xinference/core/worker.py
```diff
@@ -5232,8 +5501,17 @@ async def remove_virtual_env(
         python_version: Optional[str] = None,
     ) -> bool:
         """Remove a virtual environment for a specific model."""
+        with self._virtual_env_usage_lock:
+            active_model_uids_by_path = {
+                path: sorted(usage.active_model_uids | usage.preparing_model_uids)
+                for path, usage in self._virtual_env_usages.items()
+                if usage.active_model_uids or usage.preparing_model_uids
+            }
         return self._virtual_env_manager.remove_virtual_env(
```
qinxuye: [P1] Keep the usage check atomic with virtualenv deletion

This snapshots `_virtual_env_usages` under the lock but releases the lock before `remove_virtual_env()` deletes the directory. A concurrent launch can reserve the same environment after this snapshot and before `shutil.rmtree()`, so the deletion still sees the stale empty snapshot and removes a venv that is now being prepared or used. I reproduced that interleaving on this head: the delete path captured `{}`, a launch then added the model to `preparing_model_uids`, and deletion still proceeded. Please serialize deletion with reservation/setup using the same per-venv lock and re-check usage while that lock is held, with a concurrent delete-vs-launch regression test.

qinxuye (approved):

PR #5458, opened 2026-08-30
ENH: support JoyAI image editing models

Minamiyama:
## Summary

Add built-in support for the following JoyAI image editing models:

- `joyai-image-edit`
- `joyai-image-edit-plus`

Both Hugging Face and ModelScope sources are supported.

## Implementation

- Register model metadata, default generation settings, and virtualenv dependencies.
- Use `DiffusionPipeline` because JoyAI edit pipelines are not registered in `AutoPipelineForText2Image`.
- Support JoyAI Image Edit Plus multi-image input:
  - Merge the primary image and `reference_images`.
  - Pass images through the pipeline's `images` argument.
  - Always provide a list, including single-image requests.
  - Avoid passing unsupported `width` and `height` arguments.
- Add built-in model documentation.
- Add focused metadata, loading, and multi-image tests.

## Model Sources

| Model | Hugging Face | Revision | ModelScope | Revision |
| --- | --- | --- | --- | --- |
| `joyai-image-edit` | `jdopensource/JoyAI-Image-Edit-Diffusers` | `main` | `jd-opensource/JoyAI-Image-Edit-Diffusers` | `master` |
| `joyai-image-edit-plus` | `jdopensource/JoyAI-Image-Edit-Plus-Diffusers` | `main` | `jd-opensource/JoyAI-Image-Edit-Plus-Diffusers` | `master` |

Official project: https://github.com/jd-opensource/JoyAI-Image

### DIFF
--- a/xinference/model/image/stable_diffusion/core.py
+++ b/xinference/model/image/stable_diffusion/core.py
@@ -769,10 +769,17 @@ class DiffusionModel(SDAPIDiffusionModelMixin):
                 self._deepcache_helper.pipe = None
 
     @staticmethod
-    def _process_progressor(kwargs: dict):
+    def _process_progressor(
+        kwargs: dict,
+        *,
+        progressor: Optional["Progressor"] = None,
+        pipeline_call_index: int = 0,
+        pipeline_call_count: int = 1,
+    ):
         import diffusers
 
-        progressor: Progressor = kwargs.pop("progressor", None)
+        if progressor is None:
+            progressor = kwargs.pop("progressor", None)
 
         def report_status_callback(
             pipe: diffusers.DiffusionPipeline,
@@ -781,7 +788,10 @@ class DiffusionModel(SDAPIDiffusionModelMixin):
             callback_kwargs: dict,
         ):
             num_steps = pipe.num_timesteps
-            progressor.set_progress((step + 1) / num_steps)
+            local_progress = (step + 1) / num_steps
+            progressor.set_progress(
+                (pipeline_call_index + local_progress) / pipeline_call_count
+            )
 
             return callback_kwargs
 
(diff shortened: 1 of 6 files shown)


xinference/model/image/stable_diffusion/core.py
```diff
@@ -1052,6 +1074,10 @@ def image_to_image(
         # generate config for lightning
         self._gen_config_for_lightning(kwargs)
 
+        if is_joyai_image_edit_plus:
+            kwargs["images"] = image
```
qinxuye, line 1110: `image_to_image` exposes `n` as the requested output count, but `JoyImageEditPlusPipeline.__call__` has no `num_images_per_prompt` parameter. `_call_model()` therefore filters it out, so `n > 1` silently returns one image (reproduced with `n=2`). Please either implement repeated/batched generation for this pipeline or reject unsupported values instead of returning the wrong number of results, and add a regression test for `n > 1`.

xinference/model/image/stable_diffusion/core.py
```diff
@@ -813,7 +833,19 @@ def _call_model(
             if kwargs.get("guidance_scale", "unset") is None:
                 kwargs.pop("guidance_scale", None)
             self._filter_kwargs(model, kwargs)
-            images = model(**kwargs).images
+            if _num_pipeline_calls == 1:
+                images = model(**kwargs).images
+            else:
+                images = []
+                generators = kwargs.get("generator")
+                for call_index in range(_num_pipeline_calls):
```
qinxuye, line 853: The progress callback is installed once before this loop and reports only the fraction within one pipeline invocation. When _num_pipeline_calls is greater than 1, every invocation therefore restarts the same callback: with two calls and two steps I reproduced 0.5, 1.0, 0.5, 1.0. Progressor uploads whenever the local value reaches 1.0, so the request progress endpoint reports 100% after the first image while later outputs are still generating. Please scale each callback value by call_index and _num_pipeline_calls (or use equivalent sub-stages), and add a regression test asserting monotonic overall progress that reaches 1.0 only after the final pipeline call.

qinxuye (approved):
LGTM

PR #5460, opened 2026-08-31
FIX: apply download worker environment to snapshots

### LINKED ISSUE #5459: FIX: apply download worker environment to snapshots
## Context

PR #5427 introduced model_download_workers in system settings and the worker exposes that value as HF_HUB_DOWNLOAD_WORKERS while creating a model. Supported Hugging Face Hub versions do not automatically consume that environment variable, and ModelScope snapshot downloads require an explicit max_workers argument.

## Proposed change

- Add retry_snapshot_download as the dedicated wrapper for HF/ModelScope snapshot APIs.
- Read HF_HUB_DOWNLOAD_WORKERS and pass it as max_workers.
- Preserve an explicit max_workers supplied by a caller.
- Keep single-file, OpenMind, and CSGHub downloads on retry_download.
- Leave upstream defaults unchanged when the environment variable is absent.
- Require ModelScope 1.21 or newer, where snapshot_download supports max_workers.

## Validation

Cover environment injection, explicit override, single-file isolation, and the no-environment fallback.

amumu96:
## Summary

- add a dedicated `retry_snapshot_download` wrapper that passes `HF_HUB_DOWNLOAD_WORKERS` as `snapshot_download(max_workers=...)`
- route Hugging Face and ModelScope snapshot call sites through the dedicated wrapper while keeping single-file, OpenMind, and CSGHub downloads on `retry_download`
- propagate the effective system download-worker setting into model subpools so auxiliary downloads during `ModelActor.load()` use the same concurrency
- preserve an explicit `max_workers` argument and leave library defaults unchanged when the environment variable is absent
- require ModelScope 1.21, where `snapshot_download` supports `max_workers`

## Release note

- Fixed the existing model download workers setting so it controls Hugging Face and ModelScope snapshot download concurrency, including auxiliary model downloads during model loading.

Closes #5459

## Validation

- 6 focused tests passed for subpool propagation, load-stage auxiliary download, environment injection, explicit argument priority, file-download isolation, and default preservation
- pre-commit passed on all modified files, including mypy


qinxuye (commented):
The primary cache path now consumes the download-worker setting, but the auxiliary snapshot paths added here still do not. `HF_HUB_DOWNLOAD_WORKERS` is set only around `create_model_instance` in the worker and is restored before the model subpool is spawned and `ModelActor.load()` runs. Qwen3-ASR forced-aligner, MLX LTX text-encoder, and World auxiliary downloads happen during that later load phase. The subpool inherits the persistent `XINFERENCE_MODEL_DOWNLOAD_WORKERS` setting, while `retry_snapshot_download()` reads only `HF_HUB_DOWNLOAD_WORKERS`; in a normal subpool-style environment with `XINFERENCE_MODEL_DOWNLOAD_WORKERS=2`, the wrapper still leaves the snapshot default at 8.

Please either propagate the effective worker count into the subpool as `HF_HUB_DOWNLOAD_WORKERS`, or have the wrapper consume the validated Xinference setting as a fallback, and add a load-stage test covering one of these auxiliary downloads.

Separately, the current Python CI lint failure is introduced by this head: `xinference/model/world/model.py:274` passes `str | None` as the `model_name` argument to `retry_snapshot_download`, which requires `str`. Please narrow/assert `auxiliary_model_id` before the call so the all-files mypy check passes.

amumu96:
Addressed the review in c23a301d:

- propagate the effective `XINFERENCE_MODEL_DOWNLOAD_WORKERS` value into each model subpool as `HF_HUB_DOWNLOAD_WORKERS`, so auxiliary downloads in `ModelActor.load()` are covered
- add a worker launch assertion for the propagated subpool environment
- extend the LTX external text-encoder load-stage test to verify `max_workers=2`
- narrow `auxiliary_model_id` before the World snapshot call, fixing the mypy failure

The 6 focused tests pass locally, and pre-commit (including mypy) passes on all modified files. The refreshed CI run is in progress.

qinxuye (approved):
LGTM

PR #5461, opened 2026-08-31
FIX: Restore supported vLLM and SGLang model engines

qinxuye:
## Summary

- register Qwen3-Next for the SGLang chat engine so regenerated docs keep SGLang for pytorch, FP8, and AWQ specs
- correct the vLLM version gates for Hy-MT2, MiniCPM-V-4.6 (including Thinking), and MiniMax-M3
- register MiniMax-M3 for SGLang and declare verified per-model minimum engine versions
- regenerate the affected model and backend documentation from the corrected registry

## Context

A full engine-line scan of #5442 found three incorrect removals:

- SGLang from Qwen3-Next Instruct and Thinking
- vLLM from MiniCPM-V-4.6
- vLLM from MiniMax-M3, plus Hy-MT2 from the vLLM backend lists

The same registry audit also found two already-stale entries outside #5442: vLLM was missing from MiniCPM-V-4.6-Thinking and SGLang was missing from MiniMax-M3. They are fixed here as well. The Kimi-K3 Transformers addition and Ornith-1.5-397B SGLang addition in #5442 are valid and are left unchanged.

This PR is based independently on `main`. After it merges, #5442 can be refreshed and regenerated without dropping supported engines. Qwen3-Next has no generated-page diff in this PR because `main` already shows SGLang; the registry fix makes that existing documentation reproducible.

Upstream support used for the version gates:

- Qwen3-Next: SGLang >= 0.5.2
- Hy-MT2: vLLM >= 0.21.0
- MiniCPM-V-4.6: vLLM >= 0.22.0 and SGLang >= 0.5.12
- MiniMax-M3: vLLM >= 0.24.0 and SGLang >= 0.5.16

## Validation

- `pytest -q xinference/model/llm/tests/test_llm_family.py` (62 passed, 2 skipped)
- `pre-commit run --files <all changed files>`
- `python doc/source/gen_docs.py` followed by an idempotency check
- Sphinx dummy build completed successfully (17 pre-existing unresolved-label warnings)

### DIFF
--- a/doc/source/getting_started/installation.rst
+++ b/doc/source/getting_started/installation.rst
@@ -123,7 +123,8 @@ Currently, supported models include:
 - ``GLM-4.7-Flash``
 - ``glm-5``, ``glm-5.1``, ``glm-5.2``
 - ``DeepSeek-V4-Flash``, ``DeepSeek-V4-Flash-0731``, ``DeepSeek-V4-Pro``
-- ``Hy-MT2-1.8B``, ``Hy-MT2-7B``, ``Hy-MT2-30B-A3B``
+- ``Hy-MT2-1.8B``, ``Hy-MT2-7B``
+- ``Hy-MT2-30B-A3B``
 
 .. vllm_end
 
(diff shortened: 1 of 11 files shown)


OliverBryant (approved):
LGTM

PR #5463, opened 2026-08-31
feat(model): add download-only flow

Minamiyama:
## Summary

Add a **Download Only** action to the existing model launch dialog. Users can pre-download model artifacts without deploying the model, reserving devices, creating model actors, or loading the runtime.

The implementation reuses the existing deployment download path instead of introducing separate model-download logic or a new page.

## Changes

- Add a **Download Only** button to the built-in model launch dialog.
- Extract the model artifact download process into a shared worker helper used by both deployment and download-only operations.
- Add download-only APIs:
  - `POST /v1/cache/models`
  - `GET /v1/cache/models/{cache_uid}/progress`
  - `POST /v1/cache/models/{cache_uid}/cancel`
- Report per-file download progress, downloaded size, total size, speed, status, and estimated remaining time.
- Support cancellation during active downloads and while waiting in the launch queue.
- Update cache metadata after a successful download.
- Add English, Chinese, Japanese, and Korean UI translations.
- Document the download-only workflow and API behavior.

## Progress fixes

- Recognize ModelScope repository progress bars using `unit="file"`.
- Prevent active downloads from reporting terminal `100%`.
- Keep frontend polling active through pending states and transient query failures.
- Stop polling only after the backend reports `completed`, `failed`, or `cancelled`.

## Permissions

Download-only operations reuse the existing deployment permissions:

- Starting or cancelling a download requires `models:write`.
- Querying download progress requires `models:read`.
- Users who cannot deploy models cannot start model downloads.
- Authentication-disabled deployments keep both operations available.

## Compatibility

- Existing `POST /v1/models` behavior remains unchanged.
- Download-only is limited to built-in models.
- No additional page or model-card action is introduced.
- Download-only stops before virtual environment setup, subpool creation, model actor creation, and model loading.

162905bc99cd3c28b820e820f4d7384e


qinxuye (approved):
LGTM

PR #5468, opened 2026-09-01
feat(model): support dots ocr model

llyycchhee:
Support dots ocr model.


xinference/model/image/ocr/dots_ocr.py
```diff
@@ -0,0 +1,125 @@
...
+            trust_remote_code=True,
+        )
+        self._processor = AutoProcessor.from_pretrained(
+            self._model_path, trust_remote_code=True
+        )
+
+    def ocr(
+        self,
+        image: PIL.Image,
+        prompt: str = "",
```
qinxuye: The public OCR client calls this method without a prompt. dots.ocr is prompt-driven, and upstream defines a non-empty OCR prompt; with the current empty default, a normal `handle.ocr(image)` request sends the image with blank text and does not actually request OCR. Please use the upstream OCR instruction as the default and cover the no-prompt path with a focused test.

xinference/model/image/ocr/dots_ocr.py
```diff
@@ -0,0 +1,125 @@
...
+        inputs = inputs.to("cuda")
+
+        # Inference: Generation of the output
+        assert self._model is not None
+        generated_ids = self._model.generate(**inputs, max_new_tokens=24000)
+        generated_ids_trimmed = [
+            out_ids[len(in_ids) :]
+            for in_ids, out_ids in zip(inputs.input_ids, generated_ids)
+        ]
+        output_text = self._processor.batch_decode(
```
qinxuye: `batch_decode()` returns a list even for this single-image request. Returning it directly makes the REST OCR endpoint produce a JSON array, while the plain-text OCR response contract and the other VLM OCR adapters return a JSON string. Please return the first decoded result (with empty-output handling) and add a focused test for the response shape.

qinxuye (commented):
Please also run `doc/source/gen_docs.py` and include the generated `dots.ocr.rst` plus image index update. The auto-generation job is skipped for this feature branch, so these catalog docs will not be added automatically.

xinference/model/image/ocr/dots_ocr.py
llyycchhee: Fixed.

xinference/model/image/ocr/dots_ocr.py
llyycchhee: Fixed. The adapter now returns the first decoded result as a string, and returns an empty string when batch_decode() produces no output. I also added focused tests covering both the response shape and empty-output handling.

qinxuye (approved):
LGTM

PR #5471, opened 2026-09-01
feat(cache): add resumable cache management

Minamiyama:
## Background

Model download progress, cached models, and virtual environments were previously managed through separate entry points. Interrupted downloads also lacked persistent pause, resume, and cleanup capabilities.

### Cache Management Page

- Added a new **Cache Management** page under **Model Management**.
- Unified active, incomplete, and completed model downloads in the **Model Cache** tab.
- Displayed model format, parameter size, quantization, status, progress, cache paths, and Worker information.
- Added collapsible download-file details with animation.
- Standardized action buttons as icons with tooltips.
- Added a separate tab for virtual environment management.

### Download Task Management

- Added support for downloading model artifacts without launching a model instance.
- Added pause, resume, cancel, and delete operations.
- Persisted unfinished tasks so interrupted downloads remain resumable after a Supervisor or Worker restart.
- Stored task state in SQLite using WAL mode.
- Added `XINFERENCE_DOWNLOAD_TASK_DB_PATH` for configuring the database path.

### Safe Cache Cleanup

- Tracked primary, draft, ControlNet, GGUF, and Lightning repositories used by each task.
- Protected shared repositories used by multiple Workers on the same host.
- Preserved task state when cancellation times out instead of reporting success or deleting the record.
- Prevented failed repository resolution or database writes from leaving stale running task IDs.

### API

Added or extended:

- `POST /v1/cache/models`
- `GET /v1/cache/models/{cache_uid}/progress`
- `POST /v1/cache/models/{cache_uid}/cancel`
- `GET /v1/downloads`
- `POST /v1/downloads/{cache_uid}/pause`
- `POST /v1/downloads/{cache_uid}/resume`
- `DELETE /v1/downloads/{cache_uid}`

## Compatibility

- Existing model launch APIs remain unchanged.
- Pause and resume apply only to download-only tasks.
- The default database path is `$XINFERENCE_HOME/download_tasks.db`.
- Byte-range continuation depends on the selected model Hub SDK. If unsupported, the incomplete file is downloaded again.

d55867cdd2e63ebf0d754b6a8c0c2b73

6354675584a8ffabaad169b657123f16


xinference/core/supervisor.py
```diff
@@ -2612,9 +2692,11 @@ async def cache_builtin_model(
         model_path: Optional[str] = None,
         enable_virtual_env: Optional[bool] = None,
         virtual_env_packages: Optional[List[str]] = None,
+        _resume: bool = False,
         **kwargs,
     ) -> Dict[str, Any]:
         """Cache one model on one worker without creating a running instance."""
+        download_repositories = kwargs.pop("_download_repositories", None)
```
qinxuye, line 2699: [P1] Do not trust client-controlled cleanup metadata

The public cache-model route forwards unknown JSON fields through `kwargs`, so a caller can supply `_resume` and `_download_repositories` here. This bypasses server-side repository resolution, persists the supplied list, and later lets DELETE choose those Hub repository directories for cleanup even when they are unrelated to the requested model. Please reject or strip these internal keys at the API boundary (or use an allow-list), and add a regression test proving public requests cannot control them.

xinference/core/worker.py
```diff
@@ -4250,7 +4267,10 @@ async def launch_builtin_model(
 
         _was_queued = False
         try:
-            self._model_uid_launching_guard[model_uid] = launch_info = LaunchInfo()
+            async with self._download_artifact_cleanup_lock:
```
qinxuye: [P1] Keep launch cleanup compatible and preserve the original exception

This new lock access breaks the existing worker test doubles that exercise `launch_builtin_model` without running `WorkerActor.__init__`. All eight ordinary Linux, macOS, and Windows Python CI jobs currently fail the same three tests: the missing lock raises before the launch guard is inserted, then the unconditional `del` in `finally` masks it as a `KeyError`. Please initialize or access the lock compatibly (and update the supported test fixtures as needed), and remove the guard with `pop(model_uid, None)` so early failures retain their real exception.

qinxuye (approved):
LGTM

PR #5475, opened 2026-09-02
feat(model): support MonkeyOCR

llyycchhee:
Support MonkeyOCR.

### DIFF
--- a/doc/source/models/builtin/image/monkeyocr.rst
+++ b/doc/source/models/builtin/image/monkeyocr.rst
@@ -1,0 +1,22 @@
+.. _models_builtin_monkeyocr:
+
+=========
+MonkeyOCR
+=========
+
+- **Model Name:** MonkeyOCR
+- **Model Family:** ocr
+- **Abilities:** ocr
+- **Available ControlNet:** None
+
+Specifications
+^^^^^^^^^^^^^^
+
+- **Model ID:** echo840/MonkeyOCR
+
+Execute the following command to launch the model::
+
+   xinference launch --model-name MonkeyOCR --model-type image
+
+
+
--- a/xinference/model/image/ocr/monkeyocr.py
+++ b/xinference/model/image/ocr/monkeyocr.py
@@ -14,20 +14,21 @@
 import logging
 
 # import fitz
-from typing import TYPE_CHECKING, List, Optional, Union
+from typing import TYPE_CHECKING, Optional, Union
 
 from PIL import Image
 
 if TYPE_CHECKING:
     from ..core import ImageModelFamilyV2
 
+from ...utils import allow_trust_remote_code
 from .ocr_family import OCRModel
 
 logger = logging.getLogger(__name__)
 
 
 class MonkeyOCRModel(OCRModel):
-    required_libs = ("transformers",)
+    required_libs = ("transformers", "qwen_vl_utils")
 
     @classmethod
     def match(cls, model_family: "ImageModelFamilyV2") -> bool:
@@ -41,6 +42,9 @@ class MonkeyOCRModel(OCRModel):
         model_spec: Optional["ImageModelFamilyV2"] = None,
         **kwargs,
     ):
+        if model_path is None:
+            raise ValueError("model_path is required for MonkeyOCR")
+
         self.model_family = model_spec
         self._model_uid = model_uid
         self._model_path = f"{model_path}/Recognition"
@@ -62,13 +66,17 @@ class MonkeyOCRModel(OCRModel):
             MonkeyChat_transformers,
         )
 
-        self._model = MonkeyChat_transformers(self._model_path, device=self._device)
+        self._model = MonkeyChat_transformers(
+            self._model_path,
+            device=self._device,
+            trust_remote_code=allow_trust_remote_code(self.model_family),
+        )
 
     def ocr(
         self,
         image: Union[str, Image.Image],
         **kwargs,
-    ) -> List[str]:
+    ) -> str:
         if self._model is None:
             raise RuntimeError("Model must be loaded.")
 
@@ -84,4 +92,4 @@ class MonkeyOCRModel(OCRModel):
             question = pre_question
 
         result = self._model.batch_inference([image], [question])
-        return result
+        return result[0] if result else ""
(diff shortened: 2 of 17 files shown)


xinference/thirdparty/monkeyocr/magic_pdf/model/custom_model.py
```diff
@@ -0,0 +1,210 @@
...
+            self.model = Qwen2_5_VLForConditionalGeneration.from_pretrained(
+                        model_path,
+                        torch_dtype=torch.bfloat16 if bf16_supported else torch.float16,
+                        attn_implementation=attn_implementation,
+                        device_map=self.device,
+                    )
+                
+            self.processor = AutoProcessor.from_pretrained(
+                model_path,
+                trust_remote_code=True
```
qinxuye: [P1] Preserve the custom-model remote-code trust boundary

`MonkeyOCRModel.match()` accepts a custom OCR family named `monkeyocr` because the built-in-name conflict check is case-sensitive, while this loader unconditionally sets `trust_remote_code=True`. A user-registered repository can therefore execute processor code without the required `XINFERENCE_TRUST_REMOTE_CODE` opt-in. Please pass `allow_trust_remote_code(model_family)` into this wrapper and add coverage for built-in versus custom families.

xinference/model/image/ocr/monkeyocr.py
```diff
@@ -0,0 +1,87 @@
...
+            question = kwargs.pop("question")
+            if isinstance(question, str):
+                question = pre_question + question
+            else:
+                raise ValueError("The parameter question type must be str.")
+        else:
+            question = pre_question
+
+        result = self._model.batch_inference([image], [question])
+        return result
```
qinxuye: [P1] Return the OCR value instead of the batch container

This adapter always submits one image but returns the complete list produced by `batch_inference`. `ModelActor.ocr` serializes that value directly, so REST clients receive `["recognized text"]` instead of the established OCR string response. I reproduced this with the adapter itself. Please return the single result (with a safe empty-result policy) and add an OCR response-contract test.

doc/source/models/builtin/image/monkeyocr.rst
```diff
@@ -0,0 +1,22 @@
+.. _models_builtin_monkeyocr:
```
qinxuye, line 1: [P2] Add gettext updates for every maintained locale

This is a new English documentation page, but there is no corresponding gettext catalog update under any of the nine existing `doc/source/locale/*/LC_MESSAGES/` trees. Please update the affected catalogs, validate them with `msgfmt --check --check-format`, run `python doc/build_i18n.py --all`, and build the changed page in each maintained language as required by the repository documentation policy.

qinxuye (commented):
Please also correct the PR title from `monkeycor` to `MonkeyOCR`. `git diff --check` currently reports trailing whitespace throughout the newly vendored `custom_model.py`; please clean that up as part of the revision.

xinference/model/image/ocr/monkeyocr.py
```diff
@@ -0,0 +1,87 @@
...
+        self,
+        model_uid: str,
+        model_path: Optional[str] = None,
+        device: Optional[str] = None,
+        model_spec: Optional["ImageModelFamilyV2"] = None,
+        **kwargs,
+    ):
+        self.model_family = model_spec
+        self._model_uid = model_uid
+        self._model_path = f"{model_path}/Recognition"
```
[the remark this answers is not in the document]
qinxuye: The runtime launch path resolves a concrete local model path before constructing the OCR adapter. If no path can be resolved, model creation already fails before this constructor; changing `"None/Recognition"` to `None` would only defer the same invalid state to `from_pretrained`. This is not an actionable runtime issue; the annotation could instead be tightened to `str` in a cleanup.

xinference/thirdparty/monkeyocr/magic_pdf/model/custom_model.py
```diff
@@ -0,0 +1,210 @@
...
+            
+            logger.info(f"Processing batch {i//self.max_batch_size + 1}/{(total_items-1)//self.max_batch_size + 1} "
+                       f"(items {i+1}-{batch_end})")
+            
+            try:
+                batch_results = self._process_batch(batch_images, batch_questions)
+                results.extend(batch_results)
+            except Exception as e:
+                logger.error(f"Batch processing failed for items {i+1}-{batch_end}: {e}")
+                logger.info("Falling back to single processing...")
```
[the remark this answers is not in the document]
qinxuye: The fallback concern is valid, but `torch.cuda.empty_cache()` alone is not sufficient here. While fallback runs inside this `except` block, the exception retains the `_process_batch` traceback and potentially its live CUDA tensors; `empty_cache()` cannot free live allocations. Please move fallback outside the exception scope or explicitly clear the retained exception/tensor references first, then optionally call `empty_cache()`, and cover the OOM fallback path with a focused test.

xinference/model/image/ocr/monkeyocr.py
llyycchhee: Fixed.

xinference/thirdparty/monkeyocr/magic_pdf/model/custom_model.py
llyycchhee: Fixed.

doc/source/models/builtin/image/monkeyocr.rst
llyycchhee: Fixed.

xinference/thirdparty/monkeyocr/magic_pdf/utils/load_image.py
```diff
@@ -0,0 +1,116 @@
...
+import base64
+import os
+from io import BytesIO
+from typing import Union
+
+import requests
+# import fitz
+from typing import List
+from PIL import Image, ImageFile
+from loguru import logger
```
qinxuye: [P1] Please remove the undeclared loguru dependency here (for example, use the standard-library logging module), or declare it in every relevant dependency set. loguru is absent from both the base dependencies and MonkeyOCR virtualenv packages, so importing custom_model.py fails through this module in a clean environment. The current Ubuntu 3.10-3.13 and macOS 3.10 jobs all fail the three new MonkeyOCR tests with ModuleNotFoundError: No module named loguru.

qinxuye (approved):

PR #5476, opened 2026-09-02
feat(cache): open cached models for launch

Minamiyama:
## Summary

- Add a launch shortcut to the operation column for completed model caches.
- Resolve the model type by comparing `model_name` against existing model registration list APIs.
- Navigate to the corresponding launch-model tab and move the target model card to the first position.
- Open the existing deployment dialog for the target model.
- Use one-time in-memory navigation state to keep the URL clean and prevent the dialog from reopening after refresh.

89c8b5e583f467b98b7a435cf9c16714

### DIFF
--- a/frontend/src/components/pages/cache-management/index.tsx
+++ b/frontend/src/components/pages/cache-management/index.tsx
@@ -21,6 +21,7 @@ import { toast } from 'sonner';
 
 import DownloadProgressDetails from '@/components/pages/launch-model/launch-dialog/download-progress-details';
 import {
+  createLatestRequestGuard,
   findModelRegistration,
   getLaunchModelHref,
   getSuccessfulRegistrationGroups,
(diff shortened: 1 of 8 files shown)


frontend/src/components/pages/cache-management/index.tsx
```diff
@@ -259,6 +285,36 @@ export default function CacheManagement() {
...
+        }))
+      );
+      const registrationGroups = getSuccessfulRegistrationGroups(registrationResults);
+      const registration = findModelRegistration(registrationGroups, item.model_name);
+      const href = registration
+        ? getLaunchModelHref(registration.modelType, item.model_name, registration.isBuiltin)
+        : null;
+
+      if (href && registration) {
+        setPendingLaunchModelTarget(registration.modelType, item.model_name);
```
qinxuye, line 313: [P2] Ignore stale launch lookups

Only the clicked row is disabled, so while model A is resolving the user can still click model B. Both requests then independently call `setPendingLaunchModelTarget()` and `router.push()`, and whichever finishes last wins; an older request can overwrite the newer target or open the wrong model dialog. Please either disable all launch actions while a lookup is active, or use a request token so only the latest request may update the target, navigate, and clear the loading state. Add an overlapping-request regression test.

qinxuye (approved):
LGTM.

PR #5480, opened 2026-09-02
feat(ui): select sole available model engine

Minamiyama:
## What do these changes do?

Automatically select the model engine when the launch dialog opens and exactly one engine is available.

Unavailable engines are excluded from the available engine count. When multiple engines are available, the existing selection behavior remains unchanged.

### DIFF
--- a/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
+++ b/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
@@ -181,10 +181,18 @@ export default function LaunchDialog({
 
     setModelEngineMap({});
 
-    const res = await request.get(url);
+    const res = await request.get<ModelEngine>(url);
 
-    setModelEngineMap(res || {});
-  }, [isLLM, model?.model_name, modelType]);
+    const engineMap = res || {};
+    setModelEngineMap(engineMap);
+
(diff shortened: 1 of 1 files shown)


frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -181,10 +181,21 @@ export default function LaunchDialog({
...
+    const engineMap = res || {};
+    setModelEngineMap(engineMap);
+
+    const availableEngines = Object.entries(engineMap).filter(([, engineData]) =>
+      Array.isArray(engineData)
+    );
+    if (availableEngines.length === 1) {
+      const soleEngine = availableEngines[0][0];
+      if (form.getFieldValue('model_engine') !== soleEngine) {
+        form.setFieldValue('model_engine', soleEngine);
```
qinxuye, line 200: This async callback can apply a response for a stale model. `LaunchDialog` remains mounted, so if the user closes model A and opens model B before the A request finishes, the late A response can overwrite the B engine map and set the form to the sole engine from A. Please add request-generation or cancellation guarding and check it before both `setModelEngineMap` and `setFieldValue`, so only the latest request for the currently open model may mutate state. A rapid model-switch regression test would cover this.

qinxuye (approved):
LGTM

PR #5481, opened 2026-09-02
fix(model): make register_builtin_model idempotent across refreshes

### LINKED ISSUE #5466: Make runtime model registry refresh idempotent and concurrency-safe
## Context

The worker update_model_type path can call register_builtin_model repeatedly while model launches continue on executor threads.

## Current problems

- Built-in engine class lists use unconditional extend calls and grow on every refresh.
- Some live global model tables are cleared and rebuilt during refresh.
- Embedding and rerank derived engine entries can accumulate across repeated installs.
- Readers can observe partially rebuilt state or race with dictionary-size changes.

This behavior predates the startup engine-hook work in #5465; that PR now guarantees its callbacks run only during initial bootstrap and therefore does not address the host registry refresh itself.

## Expected behavior

- Repeated registry refreshes are idempotent.
- Built-in class and derived engine entries do not grow across refreshes.
- Live readers observe a consistent registry snapshot.
- Add regression coverage for repeated LLM, embedding, and rerank refreshes.

AmirF194:
Fixed the growth part. `_install()` in embedding/rerank/llm unconditionally `.extend()`s the shared `SUPPORTED_ENGINES` class lists, and `load_model_family_from_json()` unconditionally appends a parsed family, so every `register_builtin_model()` refresh (worker.py's `update_model_type` calls it repeatedly during a hub refresh) reappends the same entries. Added `extend_classes_once()` and `family_identity_key()` in `model/utils.py` and used them at each `_install()` / `load_model_family_from_json()` call site in the three model types.

One wrinkle worth flagging: `family_identity_key()` excludes `is_builtin`. `_install()` marks every freshly loaded family `is_builtin=True` right after loading it, so comparing the raw parsed candidate against the already-marked entry would never see them as equal, and dedup would silently do nothing.

Didn't touch the `BUILTIN_LLM_FAMILIES.clear()` + `.extend()` path in `llm/__init__.py` (only reached when there are user-defined models to merge in). It already dedups by model name so it isn't part of the growth bug, but the clear and the extend are two separate statements, so in principle a reader could see an empty list between them. Closing that needs an atomic swap or a lock, it's a different problem from the one this PR fixes, and I didn't want to bundle an untested concurrency change in here.

Added one regression test per model type (`test_register_builtin_model_is_idempotent`), plus two unit tests for the new helpers. The new tests break without the change and pass with it. Ran black/ruff/isort/mypy/codespell clean on the changed files, and the rest of the three test modules passes too, except a handful that were already red on main in this environment for unrelated reasons (missing `sentence-transformers`/`FlagEmbedding`/`xllamacpp`, and one that hits the network for a tokenizer lookup).

Fixes #5466

### DIFF
--- a/xinference/model/embedding/tests/test_embedding_models.py
+++ b/xinference/model/embedding/tests/test_embedding_models.py
@@ -426,3 +426,39 @@ def test_register_builtin_model_is_idempotent():
         sum(len(specs) for specs in BUILTIN_EMBEDDING_MODELS.values())
         == baseline_model_table
     )
+
+
+def test_register_builtin_model_downloaded_catalog_merge_is_idempotent():
+    # Worker.update_model_type() re-parses a downloaded catalog file and
+    # merges it into the built-in table on every refresh. A downloaded entry
+    # that is value-identical to the built-in one (same content, same
+    # updated_at) must not keep padding the family list on repeat refreshes.
+    from ....constants import XINFERENCE_MODEL_DIR
+    from .. import register_builtin_model
+
+    spec_path = os.path.join(os.path.dirname(__file__), "..", "model_spec.json")
+    with open(spec_path) as f:
+        raw_entry = json.load(f)[0]
+    model_name = raw_entry["model_name"]
+
+    register_builtin_model()
+
+    builtin_dir = os.path.join(XINFERENCE_MODEL_DIR, "v2", "builtin", "embedding")
+    os.makedirs(builtin_dir, exist_ok=True)
+    catalog_path = os.path.join(builtin_dir, "embedding_models.json")
+    with open(catalog_path, "w") as f:
+        json.dump([raw_entry], f)
+
+    try:
+        register_builtin_model()
+        baseline_count = len(BUILTIN_EMBEDDING_MODELS[model_name])
+
+        for _ in range(3):
+            register_builtin_model()
+
+        assert len(BUILTIN_EMBEDDING_MODELS[model_name]) == baseline_count
+        # the vetted built-in entry must still be present, not shadowed out
+        # by the freshly re-parsed downloaded duplicate.
+        assert any(f.is_builtin for f in BUILTIN_EMBEDDING_MODELS[model_name])
+    finally:
+        os.remove(catalog_path)
--- a/xinference/model/embedding/tests/test_embedding_models.py
+++ b/xinference/model/embedding/tests/test_embedding_models.py
@@ -388,3 +388,41 @@ def test_convert_ids_to_tokens():
     assert tokens == [["ｘ", "ｉ", "ｎ", "ｆ"], ["b", "e", "r", "r", "p"]]
 
     shutil.rmtree(model_path, ignore_errors=True)
+
+
+def test_register_builtin_model_is_idempotent():
+    # Worker.update_model_type() calls register_builtin_model() again on
+    # every runtime hub-refresh, on the same already-imported process. It
+    # must leave the engine registry as if it had only run once.
+    from .. import register_builtin_model
+    from ..embed_family import SUPPORTED_ENGINES
+
+    register_builtin_model()
+    model_name = next(iter(EMBEDDING_ENGINES))
+    baseline_classes = {
+        engine: list(classes) for engine, classes in SUPPORTED_ENGINES.items()
+    }
+    baseline_engine_entries = sum(
+        len(specs) for specs in EMBEDDING_ENGINES[model_name].values()
+    )
+    baseline_model_table = sum(
+        len(specs) for specs in BUILTIN_EMBEDDING_MODELS.values()
+    )
+
+    for _ in range(3):
+        register_builtin_model()
+
+    assert {
+        engine: list(classes) for engine, classes in SUPPORTED_ENGINES.items()
+    } == baseline_classes
+    assert (
+        sum(len(specs) for specs in EMBEDDING_ENGINES[model_name].values())
+        == baseline_engine_entries
+    )
+    # BUILTIN_EMBEDDING_MODELS itself must not grow either: load_model_family_from_json
+    # unconditionally appended a fresh spec per model name on every refresh, independent
+    # of the engine-class and engine-entry guards above.
+    assert (
+        sum(len(specs) for specs in BUILTIN_EMBEDDING_MODELS.values())
+        == baseline_model_table
+    )
--- a/xinference/model/embedding/tests/test_embedding_models.py
+++ b/xinference/model/embedding/tests/test_embedding_models.py
@@ -20,7 +20,11 @@ import tempfile
 import pytest
 
 from ..cache_manager import EmbeddingCacheManager as CacheManager
-from ..core import EmbeddingModelFamilyV2, TransformersEmbeddingSpecV1
+from ..core import (
+    EMBEDDING_MODEL_DESCRIPTIONS,
+    EmbeddingModelFamilyV2,
+    TransformersEmbeddingSpecV1,
+)
 from ..embed_family import BUILTIN_EMBEDDING_MODELS, EMBEDDING_ENGINES
 
 TEST_MODEL_SPEC = EmbeddingModelFamilyV2(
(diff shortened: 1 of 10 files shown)


xinference/model/embedding/tests/test_embedding_models.py
```diff
@@ -388,3 +388,41 @@ def test_convert_ids_to_tokens():
...
+
+
+def test_register_builtin_model_is_idempotent():
+    # Worker.update_model_type() calls register_builtin_model() again on
+    # every runtime hub-refresh, on the same already-imported process. It
+    # must leave the engine registry as if it had only run once.
+    from .. import register_builtin_model
+    from ..embed_family import SUPPORTED_ENGINES
+
+    register_builtin_model()
```
qinxuye, line 404: [P1] Cover the downloaded-catalog refresh path

This test calls `register_builtin_model()` without enabling the downloaded catalog, so it misses the path used by `Worker.update_model_type()`: the worker stores the downloaded JSON first, then `install_models_with_merge()` merges the existing live registry with freshly parsed downloaded entries. With the same downloaded JSON and `updated_at`, repeated refreshes still grow the family list (`1 -> 2 -> 3 -> 4`) because the final merge retains every equal-timestamp candidate. Please add a regression test with `has_downloaded_models()` active and make the final merge/rebuild idempotent (while preserving the correct `is_builtin` provenance).

AmirF194:
Good catch, that path is real. install_models_with_merge's fallback merge (used here since embedding/rerank don't pass model_identity_func) kept every candidate sharing the max updated_at instead of collapsing value-identical ones, so a downloaded entry matching an existing built-in on content and timestamp padded the list on every refresh.

Fixed in fcf56662 -> 1300282a: the merge now skips a user-model candidate whose family_identity_key already matches one already collected, same value-based dedup load_model_family_from_json uses on its own loading path. Added a regression test for both embedding and rerank that enables has_downloaded_models() and writes a downloaded catalog entry copied from model_spec.json, confirming the family list stays flat across repeated refreshes and the built-in entry (is_builtin=True) is what survives.

xinference/model/embedding/tests/test_embedding_models.py
```diff
@@ -388,3 +388,77 @@ def test_convert_ids_to_tokens():
...
+    with open(spec_path) as f:
+        raw_entry = json.load(f)[0]
+    model_name = raw_entry["model_name"]
+
+    register_builtin_model()
+
+    builtin_dir = os.path.join(XINFERENCE_MODEL_DIR, "v2", "builtin", "embedding")
+    os.makedirs(builtin_dir, exist_ok=True)
+    catalog_path = os.path.join(builtin_dir, "embedding_models.json")
+    with open(catalog_path, "w") as f:
```
qinxuye, line 457: [P1] Keep regression tests out of the real model directory

This test opens `XINFERENCE_MODEL_DIR/v2/builtin/embedding/embedding_models.json` with `"w"` and unconditionally removes it in `finally`; the new rerank test does the same for `rerank_models.json`. Running either test on a developer machine with an existing downloaded catalog therefore truncates and deletes real local state. Please use `tmp_path` and patch the model-directory lookup (including the rerank module-level constant as needed) so these tests cannot touch the user catalog.

AmirF194: Good catch, fixed. Both tests now route through tmp_path via monkeypatch; rerank/__init__.py binds XINFERENCE_MODEL_DIR at import time, so that module's own attribute needed patching too, not just constants.py. Pushed in 6a38dac3.

xinference/model/utils.py
```diff
@@ -3090,8 +3117,17 @@ def merge_models_by_timestamp(
...
-            # Add user models
+            # Add user models, skipping any value-identical to a model already
+            # present. install_models_with_merge reruns on every runtime refresh
+            # against the same downloaded JSON, so without this guard a
+            # downloaded entry sharing content and updated_at with an existing
+            # one keeps re-appending as a duplicate on every call.
+            existing_keys = {family_identity_key(model) for _, model in all_models}
             for model in user_model_list:
+                key = family_identity_key(model)
+                if key in existing_keys:
```
qinxuye, line 3145: [P1] Preserve downloaded-model provenance across refreshes

This deduplication turns a downloaded model into a trusted built-in after the second refresh. On the first call, a newer downloaded family correctly survives with `is_builtin=False`. On the next call, `install_models_with_merge()` reloads the packaged specs into the existing live dictionary and then marks every family already present there as built-in, including that downloaded family. Because `family_identity_key()` excludes `is_builtin`, this branch then treats the freshly parsed `is_builtin=False` copy as a duplicate and discards it, leaving only the incorrectly promoted `is_builtin=True` entry. I reproduced the transition as `False -> True` on two calls to `register_builtin_model()`. Since `allow_trust_remote_code()` trusts `is_builtin`, this bypasses the operator opt-in for downloaded model code. Please rebuild the packaged built-in state separately or otherwise preserve provenance, and add a regression assertion that a newer/downloaded-only family remains `is_builtin=False` after repeated refreshes.

AmirF194: Pushed 8e32fb62. Loading and marking into a fresh dict every call instead of built_in_dict directly, so a downloaded family from a prior merge never gets swept up in the is_builtin promotion. New test in test_embedding_models.py, fails on the old code (promotes to True) and passes now. Also fixed the isort failure the previous commit left in test_rerank.py.

xinference/model/utils.py
```diff
@@ -3158,21 +3199,22 @@ def install_models_with_merge(
...
+                    model_normalize_func(user_model, freshly_loaded_builtins)
 
         # Merge models, keeping the latest version based on updated_at
         merged_models = merge_models_by_timestamp(
-            built_in_models_copy, user_models, model_identity_func
+            freshly_loaded_builtins, user_models, model_identity_func
         )
 
         # Update the dictionary with merged results
         built_in_dict.clear()
```
qinxuye, line 3230: [P1] Rebuild the derived registries when replacing the model table

Loading into a fresh dict means a downloaded-only model disappears from `built_in_dict` when a later complete catalog no longer contains it, but the embedding/rerank installers never clear their engine or description registries. I reproduced a refresh after removing such a model: it was absent from `BUILTIN_EMBEDDING_MODELS` while still present in both `EMBEDDING_ENGINES` and `EMBEDDING_MODEL_DESCRIPTIONS`. The model therefore remains advertised even though launch fails in `match_embedding()`. Please rebuild or prune the derived registries together with the primary table, and cover catalog removal in a regression test.

AmirF194: Right, loading into a fresh dict every call means a dropped model vanishes from BUILTIN_EMBEDDING_MODELS/BUILTIN_RERANK_MODELS but EMBEDDING_ENGINES/EMBEDDING_MODEL_DESCRIPTIONS (and the rerank pair) never got the same treatment, so they kept the stale entry. Added prune_stale_derived_registries() in utils.py and call it at the end of both _install() functions, scoped to names not in the current builtin table or the live custom registry so a registered custom model is never swept up. Added a regression test in each module: register a downloaded-only model, confirm it's live everywhere, drop it from the catalog, confirm all three (builtin table, engines, descriptions) agree it's gone. Confirmed both fail on the pre-fix code and pass with it. Pushed e8e6252.

xinference/model/utils.py
```diff
@@ -3158,21 +3216,22 @@ def install_models_with_merge(
...
```
qinxuye, line 3230: [P1] Keep image derived registries in sync with the fresh model table

`install_models_with_merge()` is shared by image models too, so this fresh `clear()`/`update()` now removes a downloaded-only image from `BUILTIN_IMAGE_MODELS` when a later catalog drops it. Unlike audio/world/video and the embedding/rerank fix in this commit, `image._install()` never rebuilds or prunes `IMAGE_MODEL_DESCRIPTIONS` and `IMAGE_ENGINES`. I reproduced the refresh as `(builtin, description, engine) = (True, True, True) -> (False, True, True)`: the removed model remains advertised while `match_diffusion()` can no longer find it, so launch fails. Please keep the image derived registries synchronized as well and add a catalog-removal regression test.

AmirF194: Right, image models share the same fresh-clear/update path and never got the same pruning. Added the same prune_stale_derived_registries() call at the end of image._install(), scoped to IMAGE_ENGINES and IMAGE_MODEL_DESCRIPTIONS against the live builtin + user-defined model names. Added a regression test mirroring the embedding/rerank ones (register a downloaded-only image model, confirm it lands in all three registries, drop it from the catalog, confirm all three agree it is gone); confirmed it fails on the pre-fix code and passes with it. Pushed 35bffbdd.

qinxuye: This is still incomplete for OCR models. `image._install()` populates a separate `OCR_ENGINES`, while the new prune call only covers `IMAGE_ENGINES` and `IMAGE_MODEL_DESCRIPTIONS`. I reproduced a downloaded-only `PaddleOCR-VL-catalog-removal-test` refresh as `(builtin, description, OCR engine) = (True, True, True) -> (False, False, True)`, so the removed OCR model still retains a stale launch engine. Please import `OCR_ENGINES`, include it in `prune_stale_derived_registries(...)`, and extend the catalog-removal regression test with an OCR entry that asserts removal from `OCR_ENGINES`. Also run isort on `xinference/model/image/__init__.py`; the current lint job fails because the `engine_family` imports are split.

qinxuye: Fixed in eecc29118. `OCR_ENGINES` is now pruned together with the image engine and description registries, and the catalog-removal regression test includes a downloaded-only PaddleOCR-VL variant and verifies that all three derived entries are removed.

AmirF194: OCR_ENGINES needed the same treatment as the other three derived tables, missed it there. Fix and test look right to me.

qinxuye (approved):
LGTM

PR #5483, opened 2026-09-03
fix(security): report bans lifted, not records dropped, from bulk unban

NgoQuocViet2001:
## Problem

The bulk-unban endpoints answer with `len()` of the record map:

```python
def unban_all_ips(self) -> int:
    count = len(self._ip_records)
    self._ip_records.clear()
```

But `_ip_records` is not a map of bans. `record_invalid_key` creates an entry on the **first** failure, so the map holds every IP that ever mistyped a key, plus expired bans whose records are only dropped when `is_ip_banned` happens to read them back.

So `POST /unban_all_ips` reports a number far above what `GET /banned_ips` just listed:

```
get_banned_ips()  -> 1 banned IP
_ip_records       -> 2 records          (one IP, one failure short of the threshold)
unban_all_ips()   -> "unbanned": 2      ← should be 1
```

`unban_all_keys` has the same defect. `unban_key_all(key_id)` is the worst of the three: it reports a count scoped to one key id, so the number reads as specific and verifiable when it is neither —

```
two IPs banned on key 7, one more with a single failure
unban_key_all(7) -> 3                   ← should be 2
```

An operator reading "unbanned: 340" after a burst of bad requests cannot tell it from a real ban count, which is the number that matters when judging whether a lockout was working.

## Fix

Count records under an active ban — the same `banned_until > now` predicate `get_banned_ips`, `get_banned_keys` and `_update_ban_gauges` already use — via one small helper. The unban itself is unchanged: every record for the target is still cleared, expired or not.

## Scope

`rate_limiter.py` only: one helper and the three count expressions. No route, no config, no ban decision changes.

## Test plan

- Added: `xinference/api/tests/test_rate_limiter_unban_counts.py` — bulk IP unban, bulk key unban, per-key unban, and an expired ban, each asserting the count matches what `get_banned_*` reports.
- Not run: `pytest` (`xoscar` has no wheel for this host's interpreter and fails to build from source, so `xinference/conftest.py` cannot import).
- Checked: imported `rate_limiter` directly with the same scenarios as the new tests. Before the change `unban_all_ips()` returns 2 against 1 banned IP and `unban_key_all(7)` returns 3 against 2 bans; after it, 1 and 2, with the other key's ban left in place.
- Checked: `black --check` and `isort` clean on both files.

### DIFF
--- a/xinference/api/oauth2/advanced/rate_limiter.py
+++ b/xinference/api/oauth2/advanced/rate_limiter.py
@@ -84,6 +84,12 @@ class RateLimiter:
             for k in stale_keys:
                 del records[k]
 
+    @staticmethod
+    def _count_banned(records: Dict) -> int:
+        """How many of these records are under an active ban. Must hold self._lock."""
+        now = time.time()
+        return sum(1 for r in records.values() if r.banned_until > now)
+
     def _update_ban_gauges(self) -> None:
         try:
             from ....core.metrics import banned_ips_total, banned_keys_total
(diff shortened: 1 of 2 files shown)


qinxuye (approved):
LGTM.

PR #5488, opened 2026-09-04
FIX: Restore audio streaming with TorchCodec

Minamiyama:
## Summary

- Fall back to TorchCodec when `torchaudio.io.StreamWriter` is unavailable.
- Preserve the existing StreamWriter path for older torchaudio versions.
- Convert audio chunks to the layout and dtype required by TorchCodec.
- Flush buffered encoder packets when streaming completes.
- Normalize AAC, M4A, PCM, and WAVE container formats.
- Delay MP4 reads until container metadata has been finalized.
- Add regression coverage for format mapping and final packet delivery.

## Motivation

torchaudio 2.9 and later removed `torchaudio.io.StreamWriter`. Any TTS
implementation using the shared `audio_stream_generator` could therefore return
HTTP 200 and then fail while consuming the streamed response with:

`AttributeError: module 'torchaudio' has no attribute 'io'`

The issue was first reproduced with Breeze TTS, but it belongs to the shared
audio encoding layer and can affect other streaming TTS implementations.

### DIFF
--- a/xinference/model/audio/utils.py
+++ b/xinference/model/audio/utils.py
@@ -110,7 +110,14 @@ def _audio_stream_generator_with_torchcodec(
     output_generator: typing.Iterator[typing.Any],
     output_chunk_transformer: Callable,
 ):
-    from torchcodec.encoders import Encoder
+    try:
+        from torchcodec.encoders import Encoder
+    except ImportError as exc:
+        raise ImportError(
+            "Failed to import 'torchcodec'. It is required for audio streaming "
+            "when 'torchaudio.io.StreamWriter' is unavailable. Install a "
+            "TorchCodec build compatible with the installed PyTorch version."
+        ) from exc
 
     normalized_format = response_format.lower().lstrip(".")
     response_pcm = normalized_format == "pcm"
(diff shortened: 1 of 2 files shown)


xinference/model/audio/utils.py
```diff
@@ -104,6 +104,81 @@ def ensure_sample_rate(
...
 
 
+def _audio_stream_generator_with_torchcodec(
+    response_format: str,
+    sample_rate: int,
+    output_generator: typing.Iterator[typing.Any],
+    output_chunk_transformer: Callable,
+):
+    try:
+        from torchcodec.encoders import Encoder
```
qinxuye: [P1] This fallback uses torchcodec.encoders.Encoder, but the incremental Encoder API was introduced in TorchCodec 0.13. The matching releases for PyTorch and torchaudio 2.9 and 2.10 are TorchCodec 0.9 and 0.10, which expose AudioEncoder but not Encoder, so the import still fails in the exact environments this PR claims to restore. Please either support the older matching API, explicitly constrain support to torch and torchcodec versions that provide Encoder, or add a compatible fallback, and cover the chosen compatibility range with a real-version test rather than only mocking Encoder.

qinxuye (approved):

PR #5489, opened 2026-09-04
DOC: fix model type card layout

qinxuye:
## Summary

- keep the model type cards in balanced two-column rows
- pair `world` and `flexible` in the same grid

## Validation

- `git diff --check`
- `pre-commit run --files doc/source/models/index.rst`
- Sphinx dummy builder (target page parsed without warnings)


OliverBryant (approved):
LGTM

PR #5491, opened 2026-09-05
feat: support multi-worker replica scaling

leslie2046:
## Summary

Support scaling model replicas across multiple workers from the WebUI, and make replica state reconciliation robust when workers reconnect or replay their model snapshots during scale-up / scale-down.

## Changes

- Allow selecting multiple workers when adding replicas from the running-model dialog.
- Distribute new replicas across selected workers in selection order, reusing workers round-robin when needed.
- Filter CPU-only workers for GPU replicas while keeping workers without GPU metadata selectable.
- Preserve explicit GPU index assignment across replicas.
- Reconcile supervisor routing and replica metadata when a worker reconnects/replays state during replica launch or termination.
- Clean up replayed routes and replica metadata after failed scale-up.
- Allow removal of retained `TERMINATED` replica records even when no live route remains.
- Keep instance replica counts/statuses based on active replicas instead of retained terminal rows.
- Add frontend utility tests and supervisor regression tests for multi-worker placement, replay during launch/delete, failed load cleanup, and terminal-record deletion.

## Notes

This keeps the existing automatic worker selection behavior when no workers are explicitly selected, while enabling explicit multi-worker placement from the WebUI.


xinference/core/supervisor.py
```diff
@@ -4896,19 +4936,40 @@ async def _terminate_model_replica(
...
+                index
+                for index in replica_info.active_replica_ids
+                if index != replica_id
+            ]
+            self._refresh_replica_scheduler(replica_info)
 
-        remaining_replica_count = await self._status_guard_ref.remove_replica_status(
-            model_uid, replica_id
+        await self._status_guard_ref.remove_replica_status(model_uid, replica_id)
+        # StatusGuard may retain other terminal rows; these are not live replicas.
```
qinxuye, line 4954: Please apply this active-replica accounting to `mark_replica_dead` as well. That path still trusts the count returned by `remove_replica_status()`, which includes retained `TERMINATED` rows. A reachable sequence is: worker B goes down and leaves a terminal row, then the last active replica on worker A exhausts auto-recovery. `mark_replica_dead` removes the active row, sees the retained row as `remaining == 1`, and writes `READY` / `replica=1` even though `active_replica_ids` is empty. I reproduced this locally. Please derive the remaining count from `ReplicaInfo.active_replica_ids` there too and add a regression test.

qinxuye (commented):
Two blocking issues found on the current head.

xinference/core/supervisor.py
```diff
@@ -4896,19 +4936,40 @@ async def _terminate_model_replica(
                 raise
 
         self._replica_model_uid_to_worker.pop(replica_model_uid, None)
-        replica_info.replica_to_worker_refs.pop(replica_id, None)
-        replica_info.active_replica_ids.remove(replica_id)
-        self._refresh_replica_scheduler(replica_info)
+        self._replica_model_uid_to_worker_shards.pop(replica_model_uid, None)
```
qinxuye, line 4939: This new map access breaks the existing `test_terminate_model_replica_updates_active_replica_set`: its `DummySupervisor` does not define `_replica_model_uid_to_worker_shards`, so the focused replica suite fails with `AttributeError` at this line. Please update the fixture (or make this cleanup tolerant) so the existing test suite remains green.

qinxuye (approved):
LGTM

PR #5492, opened 2026-09-05
  FIX: Improve model experience actions and latency feedback

leslie2046:
## Summary

  - Move Generate and Reset actions directly below the primary input.
  - Enable progressive frontend echo for text completion responses.
  - Display frontend-measured request latency in the Results panel for all capabilities, including Embedding.
  - Cancel active streaming requests when resetting or leaving the page.
  - Keep backend APIs unchanged.

  ## Validation

  - TypeScript check passed.
  - ESLint passed.
  - `git diff --check` passed.

### DIFF
--- a/frontend/src/components/pages/running-model-detail/panels/capability-task-panel.tsx
+++ b/frontend/src/components/pages/running-model-detail/panels/capability-task-panel.tsx
@@ -16,6 +16,7 @@ import { Form } from '@/components/ui/form';
 import { ModelAbility, ModelType, RequestEvents } from '@/constants';
 import { createForm } from '@/hooks/use-form';
 import request from '@/lib/request';
+import { EventStreamController, postEventStreamFetcher } from '@/lib/eventStream';
 import { eventBus } from '@/lib/event-bus';
 import { cn, copyToClipboard, sleep } from '@/lib/utils';
 import { isNumber } from '@/lib/is';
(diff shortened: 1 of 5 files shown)


frontend/src/components/pages/running-model-detail/panels/capability-task-panel.tsx
```diff
@@ -303,24 +367,46 @@ const CapabilityTaskPanel = forwardRef<CapabilityTaskPanelMethod, CapabilityTask
...
+        <Button type="submit" className="h-11 flex-1 rounded-full" loading={loading}>
+          <Sparkles className={cn('size-4', loading && 'hidden')} />
+          {config.submitLabel || 'Generate'}
+        </Button>
+        <Button
+          type="button"
+          variant="secondary"
+          size="icon"
+          className="size-11 rounded-full"
+          disabled={loading && !config.showProgress && audioStreamRef.current === undefined}
```
qinxuye: [P1] Keep reset available for completion streams

Generate now uses a live `EventStreamController`, but this disabled condition still only permits reset while progress tracking or an audio stream is active. For Generate, `config.showProgress` is unset and `audioStreamRef.current` is undefined, so Reset stays disabled for the entire completion stream and the user cannot invoke `reset()` / `disposeCompletionStream()` to abort a long request. Please make an active completion stream cancellable here, ensure the controller state enables the button even before the first token arrives, and add regression coverage.

frontend/src/components/pages/running-model-detail/panels/result-panels.tsx
```diff
@@ -503,7 +503,7 @@ export function UniversalResultPanel({
   progress,
   ability,
 }: CapabilityResultProps) {
-  if (loading && !isAudioStreamResult(result)) {
+  if (loading && !isAudioStreamResult(result) && !(ability === ModelAbility.Generate && (result as CompletionResponse)?.choices?.[0]?.text)) {
```
qinxuye: [P2] Format the updated condition

The edited file does not pass the repository formatter: `npx prettier --check .../result-panels.tsx` reports this file as needing changes. Please run Prettier on it before merging.

qinxuye (commented):
Two additional issues found on the current head.

qinxuye (approved):
LGTM

PR #5496, opened 2026-09-06
feat(models): add embedding and rerank multimodal abilities

Minamiyama:
## Summary

- Add vision, video, and audio ability metadata for supported embedding and rerank models.
- Expose abilities in model descriptions and runtime status.
- Add model-ability filtering for embedding and rerank model lists.
- Add localized `video` ability labels.

### DIFF
--- a/xinference/api/schemas/requests.py
+++ b/xinference/api/schemas/requests.py
@@ -50,8 +50,8 @@ class CreateEmbeddingRequest(BaseModel):
 
 class RerankRequest(BaseModel):
     model: str
-    query: str
-    documents: List[str]
+    query: Union[str, Dict[str, Any]]
+    documents: List[Union[str, Dict[str, Any]]]
     top_n: Optional[int] = None
     return_documents: Optional[bool] = False
     return_len: Optional[bool] = False
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -3039,8 +3039,14 @@ class WorkerActor(xo.StatelessActor):
         from ..model.llm.core import LLM
 
         ability_map = {
-            "embedding": ["embed"],
-            "rerank": ["rerank"],
+            "embedding": [
+                "embed",
+                *getattr(model.model_family, "model_ability", []),
+            ],
+            "rerank": [
+                "rerank",
+                *getattr(model.model_family, "model_ability", []),
+            ],
             "flexible": ["flexible"],
         }
         if model_type in ability_map:
--- a/xinference/core/worker.py
+++ b/xinference/core/worker.py
@@ -3038,14 +3038,16 @@ class WorkerActor(xo.StatelessActor):
     async def _get_model_ability(self, model: Any, model_type: str) -> List[str]:
         from ..model.llm.core import LLM
 
+        model_family = getattr(model, "model_family", None)
+        model_ability = getattr(model_family, "model_ability", None) or []
         ability_map = {
             "embedding": [
                 "embed",
-                *getattr(model.model_family, "model_ability", []),
+                *model_ability,
             ],
             "rerank": [
                 "rerank",
-                *getattr(model.model_family, "model_ability", []),
+                *model_ability,
             ],
             "flexible": ["flexible"],
         }
--- a/xinference/model/embedding/tests/test_embedding_models.py
+++ b/xinference/model/embedding/tests/test_embedding_models.py
@@ -80,6 +80,33 @@ def test_engine_supported():
     assert "sentence_transformers" in EMBEDDING_ENGINES[model_name]
 
 
+def test_multimodal_model_abilities_are_exposed():
+    expected = {
+        "jina-clip-v2": ["vision"],
+        "jina-embeddings-v4": ["vision"],
+        "jina-embeddings-v5-omni-nano": [
+            "vision",
+            "video",
+            "audio",
+        ],
+        "jina-embeddings-v5-omni-small": [
+            "vision",
+            "video",
+            "audio",
+        ],
+        "Qwen3-VL-Embedding-2B": ["vision", "video"],
+        "Qwen3-VL-Embedding-8B": ["vision", "video"],
+        "WeMM-Embedding-2B": ["vision", "video"],
+        "WeMM-Embedding-4B": ["vision", "video"],
+        "WeMM-Embedding-9B": ["vision", "video"],
+    }
+
+    for model_name, abilities in expected.items():
+        family = BUILTIN_EMBEDDING_MODELS[model_name][0]
+        assert family.model_ability == abilities
+        assert family.to_description()["model_ability"] == ["embed", *abilities]
+
+
 def test_jina_v5_requires_transformers_5_compatible_sentence_transformers():
     from packaging.requirements import InvalidRequirement, Requirement
     from packaging.utils import canonicalize_name
--- a/xinference/model/embedding/tests/test_embedding_models.py
+++ b/xinference/model/embedding/tests/test_embedding_models.py
@@ -94,6 +94,8 @@ def test_multimodal_model_abilities_are_exposed():
             "video",
             "audio",
         ],
+        "gme-Qwen2-VL-2B-Instruct": ["vision"],
+        "gme-Qwen2-VL-7B-Instruct": ["vision"],
         "Qwen3-VL-Embedding-2B": ["vision", "video"],
         "Qwen3-VL-Embedding-8B": ["vision", "video"],
         "WeMM-Embedding-2B": ["vision", "video"],
--- a/xinference/model/rerank/model_spec.json
+++ b/xinference/model/rerank/model_spec.json
@@ -709,6 +709,10 @@
     "version": 2,
     "model_name": "Qwen3-VL-Reranker-8B",
     "type": "normal",
+    "model_ability": [
+      "vision",
+      "video"
+    ],
     "language": [
       "en",
       "zh"
@@ -750,6 +754,10 @@
     "version": 2,
     "model_name": "Qwen3-VL-Reranker-2B",
     "type": "normal",
+    "model_ability": [
+      "vision",
+      "video"
+    ],
     "language": [
       "en",
       "zh"
--- a/xinference/model/rerank/model_spec.json
+++ b/xinference/model/rerank/model_spec.json
@@ -799,6 +799,9 @@
     "version": 2,
     "model_name": "jina-reranker-m0",
     "type": "normal",
+    "model_ability": [
+      "vision"
+    ],
     "language": [
       "en",
       "zh",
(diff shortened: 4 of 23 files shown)


xinference/model/rerank/model_spec.json
```diff
@@ -709,6 +709,10 @@
     "version": 2,
     "model_name": "Qwen3-VL-Reranker-8B",
     "type": "normal",
+    "model_ability": [
```
qinxuye, line 712: These abilities are advertised through the catalog and runtime status, but the public rerank API cannot pass them through: `RerankRequest.query` is still `str` and `documents` is `List[str]`, so multimodal objects fail request validation. Passing an image or video URL as a string does not work for Qwen3-VL either, because `_normalize_vl_text()` wraps every string as `{"text": value}`. Please extend the request schema plus sync/async client signatures to accept multimodal objects and add route-level coverage, or do not expose `vision` / `video` yet.

xinference/model/embedding/tests/test_embedding_models.py
```diff
@@ -80,6 +80,33 @@ def test_engine_supported():
...
+            "vision",
+            "video",
+            "audio",
+        ],
+        "Qwen3-VL-Embedding-2B": ["vision", "video"],
+        "Qwen3-VL-Embedding-8B": ["vision", "video"],
+        "WeMM-Embedding-2B": ["vision", "video"],
+        "WeMM-Embedding-4B": ["vision", "video"],
+        "WeMM-Embedding-9B": ["vision", "video"],
+    }
```
qinxuye, line 104: The ability list is incomplete for models that are already integrated with multimodal input. `gme-Qwen2-VL-2B-Instruct` and `gme-Qwen2-VL-7B-Instruct` accept image inputs through SentenceTransformer, and `jina-reranker-m0` supports visual reranking, but all three keep the empty default. The new Vision filter therefore hides supported models and launched instances report only `embed` / `rerank`. Please add `vision` to those registry entries and include them in the ability tests.

xinference/core/worker.py
```diff
@@ -3039,8 +3039,14 @@ async def _get_model_ability(self, model: Any, model_type: str) -> List[str]:
...
+            "embedding": [
+                "embed",
+                *getattr(model.model_family, "model_ability", []),
+            ],
+            "rerank": [
+                "rerank",
+                *getattr(model.model_family, "model_ability", []),
+            ],
             "flexible": ["flexible"],
         }
```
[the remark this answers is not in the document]
qinxuye: Fixed in `48e4890e`: the lookup now uses nested `getattr(..., None) or []`, with regression coverage for a missing family and a `None` ability.

xinference/model/embedding/tests/test_embedding_models.py
qinxuye: Fixed in `556aab897`: both GME-Qwen2-VL entries now declare `vision`, `jina-reranker-m0` declares `vision`, and the registry tests cover all three.

xinference/model/rerank/model_spec.json
qinxuye: Thanks, the request schema and client signatures are fixed, but this is still incomplete for the selectable vLLM engine. Qwen3-VL-Reranker is registered for vLLM>=0.14.0, while VLLMRerankModel currently (1) constructs LLM without runner="pooling" or the Qwen3VLForSequenceClassification hf_overrides, (2) passes Xinference-style {"image": ...} / {"video": ...} objects straight to LLM.score instead of translating them to vLLM ScoreMultiModalParam {"content": [{"type": "image_url" / "video_url", ...}]}, and (3) does not pass the required Qwen3-VL reranker chat template. The new route test stops at an AsyncMock, so none of this execution path is exercised. Please implement the vLLM initialization, input conversion, and template path with coverage against VLLMRerankModel, or stop advertising vLLM support for these multimodal rerankers.

xinference/api/schemas/requests.py
```diff
@@ -50,8 +50,8 @@ class Config:
 
 class RerankRequest(BaseModel):
     model: str
-    query: str
-    documents: List[str]
+    query: Union[str, Dict[str, Any]]
+    documents: List[Union[str, Dict[str, Any]]]
```
qinxuye, line 54: Blocking: the API now accepts arbitrary multimodal objects, but Qwen3-VL-Reranker is also selectable with the vLLM engine and that execution path is incompatible with the accepted Xinference payloads. VLLMRerankModel passes {"image": ...} / {"video": ...} directly to LLM.score(), while vLLM 0.14 expects ScoreMultiModalParam in the form {"content": [{"type": "image_url" / "video_url", ...}]}; its loader also lacks runner="pooling", the Qwen3VLForSequenceClassification overrides, and the required reranker chat template. The new AsyncMock route test cannot catch this. Please implement the vLLM initialization, input conversion, and template path with coverage against VLLMRerankModel, or remove vLLM support for these Qwen3-VL rerankers.

qinxuye: The loader, overrides, and template are now present, but the input conversion still does not produce a valid vLLM 0.14 call. _rerank() passes List[{"content": [...]}] to LLM.score(). vLLM unwraps "content" only when the top-level argument is that dictionary; list elements are treated as individual content parts and passed to get_score_prompt(). A {"content": [...]} element has neither "type" nor a direct media key, so the real parser raises ValueError: Missing "type" field in multimodal part. The new MagicMock test only asserts this invalid call shape, so it cannot detect the failure. Please call score with each multimodal query/document wrapper as a top-level argument (looping and collecting outputs for multiple documents), or use another representation supported by vLLM 0.14, and add coverage that executes the vLLM score parsing path rather than only mocking score().

qinxuye: Still blocking at `9a32443c`. I did a full pass over both Qwen3-VL engines; there are two independent correctness bugs to fix together:

1. **vLLM loses the one-document/one-score mapping.** vLLM 0.14 unwraps each top-level `ScoreMultiModalParam["content"]` and feeds its entries through 1:N/N:N scoring. Therefore a Xinference document such as `{"text": "d1", "image": "..."}` produces two outputs. With documents `[mixed d1, video d2]`, the current `outputs.extend()` produces three outputs, while the later slicing/`zip` still assumes two; the image score from d1 is assigned to d2 and the real d2 score is dropped. A media query versus a media document also raises `ValueError: Only one multi-modal item is supported` because vLLM uses one multimodal tracker for the pair. This includes the image-query/video-document case in the route test and the video-query/image-document case in the new MagicMock test. The mock forces one output and masks both failures; the parser test covers only one text-versus-image part and is skipped when vLLM is absent. Please guarantee exactly one score per Xinference document, or reject unsupported combinations before invoking vLLM, and add coverage for a mixed text+image document, multiple documents, and media-versus-media using real vLLM 0.14 parsing semantics.

2. **sentence-transformers dynamic batching leaks the first query into later requests.** `BatchMixin` defaults to `batch_size=32`; the batch handler concatenates documents and queries from concurrent calls, but `_rerank_vl()` discards every query except `query[0]` and calls `process()` once. If `q1/docs1` and `q2/docs2` land in the same batch, every document is scored against q1. Please preserve the request offsets and invoke the VL reranker per original request, as the Jina v3.5 branch already does, with a two-request regression test.

Please address both in one update; the current passing CI and MagicMock tests do not exercise either failure.

xinference/model/rerank/vllm/core.py
```diff
@@ -137,11 +159,20 @@ def _rerank(
...
+            query_list = [
+                self._to_score_multimodal_param(query) for query in query_list
+            ]
+            documents = [
+                self._to_score_multimodal_param(document) for document in documents
+            ]
+            score_kwargs["chat_template"] = self._qwen3_vl_reranker_template
+            outputs = []
+            for query, document in zip(query_list, documents):
+                outputs.extend(self._model.score(query, document, **score_kwargs))
```
qinxuye: Blocking: a full pass over both Qwen3-VL engines found two independent correctness bugs that need to be fixed together.

1. **vLLM breaks the one-document/one-score mapping.** vLLM 0.14 unwraps `ScoreMultiModalParam["content"]` and treats its entries as 1:N/N:N score inputs. A single Xinference document such as `{"text": "d1", "image": "..."}` therefore yields two outputs. For `[mixed d1, video d2]`, this `outputs.extend()` produces three outputs, but the later slicing and `zip` still assume two documents: the image score from d1 is assigned to d2 and the actual d2 score is dropped. Media-query versus media-document pairs also raise `ValueError: Only one multi-modal item is supported`; that includes the image/video pair in the route test and the video/image pair in the MagicMock test. The mock forces one output and hides both failures, while the parser test only covers one text-versus-image part and is skipped without vLLM. Please guarantee exactly one score per Xinference document, or reject unsupported combinations before calling vLLM, and cover mixed text+image, multiple documents, and media-versus-media with real vLLM 0.14 parsing semantics.

2. **sentence-transformers dynamic batching leaks the first query into later requests.** `BatchMixin` defaults to `batch_size=32`; its batch handler concatenates concurrent requests, but `_rerank_vl()` discards every query except `query[0]` and calls `process()` once. If `q1/docs1` and `q2/docs2` share a batch, all documents are scored against q1. Preserve the original request offsets and invoke the VL reranker per request, as the Jina v3.5 branch does, with a two-request regression test.

Please address both in one update; the current passing CI and MagicMock tests do not exercise either failure.

xinference/model/rerank/sentence_transformers/core.py
```diff
@@ -541,35 +543,48 @@ def _normalize_vl_text(self, val: Any) -> Dict[str, Any]:
...
+                    payload[key] = kwargs[key]
+            return [float(score) for score in self._vl_reranker.process(payload)]
+
+        if batch_offsets is None:
+            return process_request(query[0], documents)
+
+        scores = []
+        for offset, size in batch_offsets:
+            scores.extend(
+                process_request(query[offset], documents[offset : offset + size])
```
qinxuye, line 581: [P2] Skip empty request slices before indexing the query

The previous score-mapping and query-isolation issues are fixed in `8624c262`, but the new offset loop introduces an empty-request regression. The API accepts `documents=[]`. If a normal request followed by an empty request is coalesced into one batch, the offsets are `[(0, 1), (1, 0)]` and the flattened query list has length 1. The second iteration evaluates `query[1]` and raises `IndexError`, causing the entire batch—including the normal request—to fail. I reproduced this through `model.rerank.batch(...)`, not just by calling the helper directly.

Please skip slices with `size == 0` before accessing `query[offset]`, preserving an empty result for that request. Add a regression through the batch handler with a normal request followed by an empty request, asserting that the normal request succeeds, the empty request returns no results, and the model is not called for the empty slice.

qinxuye (approved):
LGTM

PR #5497, opened 2026-09-06
fix: handle runtime and device options in flexible Transformers launcher

leslie2046:
Flexible Transformers models fail to load when `device` is specified because it is passed both explicitly and through the configuration. The worker also injects `enable_virtual_env`, which reaches the tokenizer through pipeline preprocessing defaults and causes `PreTrainedTokenizerFast._batch_encode_plus()` to fail on inference.

Build a copied pipeline configuration with a single device argument and remove the worker-only runtime option. Defer the Transformers import until loading and share the existing pipeline implementation between classification and other tasks. Preserve the pipeline output format.

Add regression coverage for device/runtime arguments and local BERT single-label and multi-label inference without downloading weights. Add a short local classification model format note with corresponding updates for all nine maintained locales.

Validation:
- 22 focused flexible-model tests passed on CPU (Python 3.12, Transformers 4.57.6; `--noconftest`, with torch imported first for the local Windows DLL loading order).
- All pre-commit checks passed for the changed Python files.
- Modified catalogs passed `msgfmt --check --check-format`; all catalogs compiled with `python doc/build_i18n.py --all`.
- The changed page built and its added content was checked in English and all nine maintained languages. Isolated page builds retained existing cross-reference and example/translation warnings.
- GPU execution and full distributed/API regression tests were not run.


qinxuye (approved):
LGTM

PR #5498, opened 2026-09-06
feat(ui): prefer system download source

Minamiyama:
## Summary

- Read system `download_source` when opening model launch dialog.
- Set `download_hub` only when source exists in current filtered hub options.
- Preserve current selection when source does not match, settings request fails, or caller lacks permission.

### DIFF
--- a/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
+++ b/frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
@@ -87,6 +87,10 @@ interface LaunchProgressResponse {
   replicas?: LaunchProgressReplica[];
 }
 
+interface SystemSettingsResponse {
+  download_source: string;
+}
+
 const DOWNLOAD_TERMINAL_STAGES = new Set(['completed', 'failed', 'cancelled']);
 
 const DOWNLOAD_ONLY_EXCLUDED_FIELDS = new Set([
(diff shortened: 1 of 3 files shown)


frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -1848,6 +1857,34 @@ export default function LaunchDialog({
...
+    const applyPreferredDownloadSource = async () => {
+      try {
+        const { download_source } = await request.get<SystemSettingsResponse>(
+          '/v1/cluster/system_settings'
+        );
+        if (
+          active &&
+          downloadHubOptionsRef.current.some((option) => option.value === download_source)
+        ) {
+          form.setFieldValue('download_hub', download_source);
```
qinxuye, line 1890: [P2] Preserve user selections while the settings request is pending

This callback unconditionally applies the system download source once the request completes. If the user opens the dialog, selects `huggingface`, and a delayed settings response returns `download_source: "modelscope"`, their explicit selection is silently replaced with `modelscope` (provided both hubs are available). The subsequent launch/download then uses a source the user did not choose. The `active` flag only protects against closing or switching dialogs; it does not detect edits within the current dialog. I reproduced this by executing the current callback with a deferred settings response.

Track whether the download-hub field has been changed by the user during the current dialog session and only apply the default if it remains untouched. Reset that tracking on a new session. Add a delayed-response regression that selects a hub before resolving the settings request and verifies that the selection is preserved.

qinxuye (approved):
LGTM.

PR #5500, opened 2026-09-07
feat(audio): add multilingual TTS language controls

Minamiyama:
## Summary
- Expose `model_lang` for TTS models that directly accept a language parameter.
- Render model-specific language options from the model spec in Web UI.
- Omit `kwargs.language` when no language is selected.
- Refine the Language and Streaming field layout.
- Add coverage for language capabilities in audio model descriptions.

20c5255270677f1418076fa643a03a2d

### DIFF
--- a/frontend/src/components/pages/running-model-detail/panels/form-panels.tsx
+++ b/frontend/src/components/pages/running-model-detail/panels/form-panels.tsx
@@ -713,6 +713,7 @@ export function SpeechPanel({ form, model }: CapabilityFormProps) {
       : SPEECH_RESPONSE_FORMAT_OPTIONS;
   const supportsVoiceCloning = model.model_ability.includes(ModelAbility.Text2audioVoiceCloning);
   const supportsVoiceDesign = model.model_ability.includes(ModelAbility.Text2audioVoiceDesign);
+  const languageOptions = (model.model_lang ?? []).map((value) => ({ label: value, value }));
   const promptSpeech = useWatch('prompt_speech', form);
   const hasPromptSpeech =
     supportsVoiceCloning && Array.isArray(promptSpeech) && promptSpeech.length > 0;
@@ -746,6 +747,11 @@ export function SpeechPanel({ form, model }: CapabilityFormProps) {
           <FormField name="speed" label="Speed" normalize={normalizeNumberInput}>
             <Input type="number" min={0.5} max={2} step={0.1} />
           </FormField>
+          {languageOptions.length > 0 && (
+            <FormField name="language" label="Language">
+              <Select options={languageOptions} placeholder="Optional" />
+            </FormField>
+          )}
         </div>
       )}
       <div className="grid grid-cols-2 gap-3">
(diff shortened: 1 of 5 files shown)


xinference/model/audio/model_spec.json
```diff
@@ -2204,6 +2256,21 @@
...
+      "english",
+      "french",
+      "german",
+      "italian",
+      "japanese",
+      "korean",
+      "portuguese",
+      "russian",
+      "spanish",
+      "beijing_dialect",
```
qinxuye: [P2] Only expose language values accepted by the Qwen3 backend

Selecting `beijing_dialect` or `sichuan_dialect` for either the 0.6B or 1.7B PyTorch CustomVoice model makes speech generation fail before inference. `Qwen3TTSModel.speech()` checks the submitted language against `self._model.get_supported_languages()`, while upstream Qwen3-TTS explicitly excludes codec language IDs containing `"dialect"` from that supported-language list. These internal dialect IDs are therefore not valid values for this dropdown, even though they exist in the model configuration.

Please remove these two entries from the affected PyTorch CustomVoice language lists and update the new test, which currently asserts the invalid values. Add a regression ensuring the advertised choices are accepted by the backend's language validation rather than only matching another hard-coded registry list.

xinference/model/audio/tests/test_audio_engine.py
```diff
@@ -70,6 +72,87 @@ def _get_spec(model_name: str):
...
+            "german",
+            "italian",
+            "japanese",
+            "korean",
+            "portuguese",
+            "russian",
+            "spanish",
+        ]
+
+    def generate_custom_voice(self, **kwargs):
```
qinxuye, line 131: [P2] Implement get_supported_speakers on the fake backend

Both new parameterized regression cases fail before reaching `generate_custom_voice`: the CustomVoice branch of `speech()` unconditionally calls `self._model.get_supported_speakers()`, even when `voice=""`. This fake does not implement that method, so the tests raise `AttributeError` instead of the expected `RuntimeError("generation reached")`.

Please add `get_supported_speakers()` returning `[]` to this fake. On this exact head, `python -m pytest -q xinference/model/audio/tests/test_audio_engine.py` gives **28 passed, 2 failed**; supplying just that missing method in memory gives **30 passed**. The production language-list fix itself is correct.

frontend/src/components/pages/running-model-detail/panels/form-panels.tsx
```diff
@@ -746,6 +748,21 @@ export function SpeechPanel({ form, model }: CapabilityFormProps) {
...
+            valuePropName="checked"
+            layout="horizontal"
+            className="flex h-full items-center"
+            tooltip="Play supported audio formats while they are generated."
+          >
+            <Switch />
+          </FormField>
+          {supportedLanguages.length > 0 && (
+            <FormField name="language" label="Language">
+              <Select options={languageOptions} placeholder="Optional" />
```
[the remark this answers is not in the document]
qinxuye: The shared Select component already defaults allowClear to true (frontend/src/components/ui/select.tsx), so this field is clearable without an explicit prop. Resolving as already supported.

xinference/model/audio/model_spec.json
qinxuye: Verified on f257d649: the invalid dialect options have been removed from the PyTorch CustomVoice language lists, and both models now have backend-validation regression coverage. The focused audio engine suite passes all 30 tests.

xinference/model/audio/tests/test_audio_engine.py
qinxuye: Fixed in f257d649 by adding get_supported_speakers() returning []. Re-ran python -m pytest -q xinference/model/audio/tests/test_audio_engine.py on this exact head: 30 passed.

qinxuye (approved):
LGTM

PR #5503, opened 2026-09-07
fix(rerank): launch GGUF rerank models without explicit engine/quantization

### LINKED ISSUE #5502: BUG: Custom GGUF rerank models fail to launch due to hardcoded default engine and missing quantization parsing
## Describe the bug

When registering and launching a custom GGUF-format rerank model (e.g. `bge-reranker-v2-m3-gguf`), the model fails to start due to two separate bugs:

### Bug 1: Default engine hardcoded to sentence_transformers

**File**: `xinference/model/rerank/core.py` → `create_rerank_model_instance()`

```python
if model_engine is None:
    # unlike LLM and for compatibility,
    # we use sentence_transformers as the default engine for all models
    model_engine = "sentence_transformers"
```

Unlike LLM models (which auto-select engine based on registered engines), rerank models always default to `sentence_transformers`. This causes:

```
ValueError: Model bge-reranker-v2-m3-gguf cannot be run on engine sentence_transformers
```

**Workaround**: Must explicitly pass `--model-engine llama.cpp` when launching.

### Bug 2: Quantization always None for non-LLM models

**File**: `xinference/core/supervisor.py` → `launch_builtin_model()`

```python
quantization=parse_results[3] if model_type == "LLM" else None,
```

The `quantization` parameter is only parsed for LLM types, always `None` for rerank. This causes the GGUF filename template `bge-reranker-v2-m3-{quantization}.gguf` to render as `bge-reranker-v2-m3-None.gguf`, which doesn't match the actual file `bge-reranker-v2-m3-Q4_K_M.gguf`.

**Result**: `RuntimeError: Failed to init server`

## To Reproduce

1. Register a custom GGUF rerank model via JSON config
2. Run `xinference launch --model-type rerank --model-name bge-reranker-v2-m3-gguf`
3. Observe the errors above

## Expected behavior

Custom GGUF rerank models should launch without requiring explicit `--model-engine` flag, similar to how LLM models auto-select their engine.

## Suggested fix

1. In `rerank/core.py`, auto-select engine from `RERANK_ENGINES` dict instead of hardcoding `sentence_transformers`
2. In `supervisor.py`, parse quantization for rerank models too, OR in `llama_cpp/core.py` fall back to `self._model_spec.quantization` when `self._quantization` is None

## Environment

- Xinference version: 3.3.0
- OS: Windows 11
- Model: bge-reranker-v2-m3-gguf (Q4_K_M)

7487:
Fixes #5502

A GGUF-only rerank model (e.g. a custom `bge-reranker-v2-m3-gguf`) could not be launched with plain `xinference launch --model-type rerank --model-name <name>`:

1. `create_rerank_model_instance` always defaulted `model_engine` to `sentence_transformers`, so the launch failed with `Model <name> cannot be run on engine sentence_transformers` unless `--model-engine llama.cpp` was passed.
2. Even with the engine given, `RerankModel` stored the raw launch `quantization` (`None` when `--quantization` is omitted), so the llama.cpp backend rendered `bge-reranker-v2-m3-None.gguf` and failed to start.

### Changes

- `create_rerank_model_instance`: when no engine is given, pick `llama.cpp` if the spec matched by `match_rerank` is `ggufv2` (that is the only engine `check_format_with_engine` allows for GGUF); every other format keeps the `sentence_transformers` default, so existing behaviour is unchanged.
- `RerankModel.__init__`: fall back to the quantization already resolved by `match_rerank` when none was passed. This is the engine-side option from the issue rather than the `supervisor.py` one: `parse_model_version` only yields the model name for rerank, so there is no quantization to parse on that path, and the default `xinference launch` path never goes through it anyway.
- Test: registers a custom GGUF rerank family and launches it with neither engine nor quantization; asserts the llama.cpp class, `model_engine == "llama.cpp"` and `_quantization == "Q4_K_M"`. On `main` it fails with the exact error from the issue.

### Verification

```
pytest xinference/model/rerank/llama_cpp/tests/test_llama_cpp.py -k "defaults_engine or raises_xllamacpp"   # both with and without XINFERENCE_ENABLE_VIRTUAL_ENV
pytest xinference/model/rerank/tests/test_rerank.py -k "not restful and not from_local_uri and not auto_detect"
black / ruff / isort / mypy / codespell on the touched files
```

Not run: `test_rerank_model_with_xllamacpp` (downloads the real GGUF). The embedding family has the same two hard-codes (`embedding/core.py`); left untouched here to keep this scoped to the issue, happy to follow up if wanted.


qinxuye (approved):
LGTM

PR #5506, opened 2026-09-08
feat(llm): add MiniCPM5-2B support

Minamiyama:
## Summary

Add built-in support for `openbmb/MiniCPM5-2B`.

- Register PyTorch, GGUF, MLX, and AWQ model specifications.
- Add Transformers, vLLM, SGLang, and MLX backend support.
- Add the MiniCPM5 XML tool-call parser.
- Add model registration and tool parser tests.

### DIFF
--- a/xinference/model/llm/llm_family.json
+++ b/xinference/model/llm/llm_family.json
@@ -11091,6 +11091,130 @@
   },
   {
     "version": 2,
+    "context_length": 131072,
+    "model_name": "minicpm5-2b",
+    "model_lang": [
+      "en",
+      "zh"
+    ],
+    "model_ability": [
+      "chat",
+      "reasoning",
+      "hybrid",
+      "tools"
+    ],
+    "model_description": "MiniCPM5-2B is the second model in the MiniCPM5 series. It is a dense 2B Transformer built for on-device, local deployment, and resource-constrained scenarios. Supports hybrid thinking via enable_thinking and native XML-style tool calling (MCP-compatible).",
+    "model_specs": [
+      {
+        "model_format": "pytorch",
+        "model_size_in_billions": 2,
+        "model_src": {
+          "huggingface": {
+            "quantizations": [
+              "none"
+            ],
+            "model_id": "openbmb/MiniCPM5-2B"
+          },
+          "modelscope": {
+            "quantizations": [
+              "none"
+            ],
+            "model_id": "OpenBMB/MiniCPM5-2B"
+          }
+        }
+      },
+      {
+        "model_format": "ggufv2",
+        "model_size_in_billions": 2,
+        "model_src": {
+          "huggingface": {
+            "quantizations": [
+              "F16",
+              "Q4_K_M",
+              "Q8_0"
+            ],
+            "model_id": "openbmb/MiniCPM5-2B-GGUF",
+            "model_file_name_template": "MiniCPM5-2B-{quantization}.gguf"
+          },
+          "modelscope": {
+            "quantizations": [
+              "F16",
+              "Q4_K_M",
+              "Q8_0"
+            ],
+            "model_id": "OpenBMB/MiniCPM5-2B-GGUF",
+            "model_file_name_template": "MiniCPM5-2B-{quantization}.gguf"
+          }
+        }
+      },
+      {
+        "model_format": "mlx",
+        "model_size_in_billions": 2,
+        "model_src": {
+          "huggingface": {
+            "quantizations": [
+              "4bit"
+            ],
+            "model_id": "openbmb/MiniCPM5-2B-MLX"
+          },
+          "modelscope": {
+            "quantizations": [
+              "4bit"
+            ],
+            "model_id": "OpenBMB/MiniCPM5-2B-MLX"
+          }
+        }
+      },
+      {
+        "model_format": "awq",
+        "model_size_in_billions": 2,
+        "model_src": {
+          "huggingface": {
+            "quantizations": [
+              "Int4"
+            ],
+            "model_id": "openbmb/MiniCPM5-2B-GPTQ"
+          },
+          "modelscope": {
+            "quantizations": [
+              "Int4"
+            ],
+            "model_id": "OpenBMB/MiniCPM5-2B-GPTQ"
+          }
+        }
+      }
+    ],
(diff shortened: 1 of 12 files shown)


xinference/model/llm/llm_family.json
```diff
@@ -11089,6 +11089,130 @@
...
...         {%- endif %}\n                        {{- '</param>' }}\n                    {%- endfor %}\n                {%- endif %}\n                {{- '</function>' }}\n            {%- endfor %}\n        {%- endif %}\n        {{- '<|im_end|>\\n' }}\n    {%- elif message.role == \"tool\" %}\n        {%- if loop.first or (messages[loop.index0 - 1].role != \"tool\") %}\n            {{- '<|im_start|>user' }}\n        {%- endif %}\n        {{- '\\n<tool_response>\\n' }}\n        {%- if message.content is string %}\n            {{- content }}\n        {%- else %}\n            {{- message.content | tojson(ensure_ascii=False) }}\n        {%- endif %}\n        {{- '\\n</tool_response>' }}\n        {%- if loop.last or (messages[loop.index0 + 1].role != \"tool\") %}\n            {{- '<|im_end|>\\n' }}\n        {%- endif %}\n    {%- endif %}\n{%- endfor %}\n{%- if add_generation_prompt %}\n    {{- '<|im_start|>assistant\\n' }}\n    {%- if enable_thinking is defined %}\n        {%- if enable_thinking is false %}\n            {{- '<think>\\n\\n</think>\\n\\n' }}\n        {%- elif enable_thinking is true %}\n            {{- '<think>\\n' }}\n        {%- endif %}\n    {%- endif %}\n{%- endif %}\n",
```
qinxuye, line 11187: [P2] Normalize OpenAI tool-call arguments before rendering this template

A normal tool-result follow-up fails here: OpenAI-compatible assistant history contains `tool_calls[].function.arguments` as a JSON string, but this template assigns it to `args_dict` and calls `.items()`. `ChatModelMixin._chat_template_needs_dict_arguments()` only detects the literal `tool_call.arguments|items`, so it does not normalize arguments for this template. Passing a user message, an assistant tool call with `arguments='{"city":"Beijing"}'`, and its tool response through `get_full_context()` raises `UndefinedError: 'str object' has no attribute 'items'` before inference.

Please extend the existing normalization to cover this template (or use the supported template form), and add a multi-turn rendering regression using JSON-string arguments as returned by the OpenAI API.

xinference/model/llm/tool_parsers/minicpm5_tool_parser.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+@register_tool_parser("minicpm5")
+class MiniCPM5ToolParser(ToolParser):
+    """Parse MiniCPM5 XML-style function calls."""
+
+    _FUNCTION_START = "<function"
+    _FUNCTION_END = "</function>"
+    _FUNCTION_RE = re.compile(
+        r"<function\s+name=['\"]([^'\"]+)['\"]\s*>(.*?)</function>", re.DOTALL
```
qinxuye: [P2] Ignore closing-tag text inside CDATA when parsing calls

The function and parameter regexes treat closing tags inside CDATA as structural delimiters. This corrupts valid XML/code passed to tools such as `write_file`: `<function name="write_file"><param name="content"><![CDATA[example </param> end]]></param></function>` produces `{"content": "<![CDATA[example"}` instead of preserving the value. Replacing the embedded tag with `</function>` produces an empty argument object and leaks the remainder as ordinary content. `_completed_output()` uses the same delimiter assumptions, so a streaming chunk can also appear complete before the CDATA value ends.

Please make both complete-call parsing and streaming boundary detection CDATA-aware. Add regressions for literal `</param>` and `</function>` inside CDATA, including chunk boundaries within those sequences, and verify that the full original argument reaches the tool unchanged.

xinference/model/llm/tool_parsers/minicpm5_tool_parser.py
```diff
@@ -0,0 +1,146 @@
...
+        return self._parse_calls(model_output)
+
+    @classmethod
+    def _completed_output(cls, text: str) -> str:
+        for length in range(len(cls._FUNCTION_START) - 1, 0, -1):
+            if text.endswith(cls._FUNCTION_START[:length]):
+                text = text[:-length]
+                break
+        position = 0
+        while match := cls._FUNCTION_OPEN_RE.search(text, position):
```
qinxuye: [P2] Buffer incomplete function opening tags during streaming

_FUNCTION_OPEN_RE only matches after the closing > arrives. Therefore _completed_output returns an unfinished opening tag as ordinary content when a chunk ends at <function or <function name="get_weather". On this head, feeding chunks ["<function", " name=\"get_weather\"", "><param name=\"city\">Beijing</param></function>"] emits the first two chunks as assistant content before emitting the tool call. This also corrupts the content-offset accounting: with the unfinished opening tag as previous_text and the completed call followed by "Done" as current_text, the parser emits the call but drops "Done" entirely.

Please buffer unfinished opening tags as well as unfinished function bodies, while retaining the CDATA-aware closing-tag handling. Add regressions across every split position of the opening tag (including immediately after <function and within its name attribute), character-by-character streaming, and normal text before/after one or multiple calls. The final streamed content and calls should match non-streaming parsing without leaked XML, duplicated text, or missing trailing content.

xinference/model/llm/llm_family.json
qinxuye: Verified fixed on 7e548535: template detection now normalizes JSON-string tool arguments for this template. The multi-turn regression verifies rendered arguments and preserves the original input messages.

xinference/model/llm/tool_parsers/minicpm5_tool_parser.py
qinxuye: Verified fixed on 7e548535: complete-call parsing and streaming boundaries now skip closing-tag text inside CDATA. Regression tests preserve literal </param> and </function> in tool arguments.

xinference/model/llm/tool_parsers/minicpm5_tool_parser.py
qinxuye: Verified fixed in 7e548535: incomplete function opening tags are buffered. Tests cover every opening-tag split, character-by-character streaming, and text surrounding multiple calls without leaked tags or dropped trailing content. The full MiniCPM5 parser and LLM utils test files pass: 43 passed.

qinxuye (approved):
LGTM

PR #5510, opened 2026-09-09
fix(chatglm): emit a usage-only final streaming chunk

m199369309:
## Summary

- preserve the terminal ChatGLM streaming chunk and its finish reason
- append an OpenAI-compatible usage-only chunk when `stream_options.include_usage` is enabled
- add focused regression coverage for the final chunk sequence and usage values

## Testing

- `pytest -q xinference/model/llm/transformers/tests/test_chatglm.py` (1 passed)
- `pre-commit run --files xinference/model/llm/transformers/chatglm.py xinference/model/llm/transformers/tests/test_chatglm.py`


qinxuye (approved):
LGTM

PR #5511, opened 2026-09-09
fix(transformers): handle batched logits when selecting next token

m199369309:
## Summary

- select the target batch row without retaining an extra singleton dimension
- keep greedy and top-k token selection scalar-safe for batched logits
- add regression coverage that selects the second request from a batch

## Testing

- `pytest -q xinference/model/llm/transformers/tests/test_batch_inference.py xinference/model/llm/transformers/tests/test_cache_compatibility.py` (10 passed, 1 skipped)
- `pre-commit run --files xinference/model/llm/transformers/utils.py xinference/model/llm/transformers/tests/test_batch_inference.py`


qinxuye (approved):
LGTM

PR #5512, opened 2026-09-09
fix(transformers): enforce token limits across batched decode steps

m199369309:
## Summary

- persist the effective `max_tokens` value across batched scheduler invocations
- separate physical cache-advancement tokens from user-visible completion tokens
- prevent repeated terminal chunks and cap streaming/non-streaming output and usage at the requested limit
- emit the optional usage-only streaming chunk after the final physical decode round

## Testing

- `pytest -q xinference/model/llm/transformers/tests/test_batch_inference.py xinference/model/llm/transformers/tests/test_cache_compatibility.py` (10 passed, 1 skipped)
- `pre-commit run --files xinference/model/llm/transformers/utils.py xinference/model/scheduler/request.py xinference/model/llm/transformers/tests/test_batch_inference.py`


xinference/model/llm/transformers/utils.py
```diff
@@ -305,6 +305,16 @@ def _batch_inference_one_step_internal(
         )
         for r in valid_req_list
     }
+    for r, generate_config in generate_config_mapping.items():
+        max_new_tokens = generate_config[0]
+        if r.effective_max_new_tokens is None:
+            if max_new_tokens == 0:
+                max_new_tokens = xinf_model_obj.get_context_len() - len(r.prompt_tokens)
```
qinxuye: [P1] Resolve the implicit token limit after prompt tokenization

A fresh InferenceRequest starts with prompt_tokens=None. build_prefill_kwargs(), called below this block, tokenizes the prompt and fills that field. For max_tokens=None (or another configuration resolved to zero), this new len(r.prompt_tokens) therefore raises TypeError before prefill. This is also happening in the current GPU CI: test_continuous_batching and the Qwen3 thinking-parameter tests fail with "object of type 'NoneType' has no len()". The new test hides the problem by assigning prompt_tokens before calling the scheduler.

Please compute and persist the implicit limit only after prefill has established the actual prompt length, then reuse it for subsequent decode-only invocations. Add a fresh-request regression that leaves prompt_tokens unset and populates it through the prefill path, plus the existing multi-step limit checks.

xinference/model/llm/transformers/utils.py
```diff
@@ -445,10 +450,12 @@ def _batch_inference_one_step_internal(
                 So the implementation here is to decode all the tokens that have been generated each time,
                 and then take the slice.
                 """
-                if r.stopped or len(r.new_tokens) % stream_interval == 0:
+                if not was_stopped and (
+                    r.stopped or r.visible_new_tokens_count % stream_interval == 0
+                ):
                     if output is None:
```
qinxuye, line 464: [P2] Bound the stop-string decoding path as well

The slice below only applies when output is None. When stop_str is configured, the earlier stop-string branch already decoded all of r.new_tokens into output, so streaming bypasses this cap. Using the new test fixtures with max_tokens=1, stream=True, and stop="ZZ" emits "52" (two generated tokens), while visible_new_tokens_count and completion_tokens both report 1. This occurs because the first decode round follows the prefill token even when the visible limit is already reached.

Please apply the visible-token bound before stop-string detection/decoding too, so physically generated tokens beyond the limit cannot appear in output or change the finish reason. Cover max_tokens=1 with matching and nonmatching stop strings in streaming and non-streaming modes, checking text, finish reason, and usage together.

xinference/model/llm/transformers/tests/test_batch_inference.py
```diff
@@ -0,0 +1,155 @@
...
+        "prompt_tokens": 14,
+        "completion_tokens": 18,
+        "total_tokens": 32,
+    }
+
+
+def test_stream_output_and_usage_do_not_exceed_max_tokens(monkeypatch):
+    request = _make_request(3, stream=True, include_usage=True)
+
+    _run_step(monkeypatch, request, decode_round=5)
```
qinxuye, line 185: [P2] Remove the new regression test's dependency on wall-clock resolution

This test already fails in the current Windows 3.10 CI with ZeroDivisionError: the fake model finishes quickly enough that the two time.time() readings in _batch_inference_one_step_internal can be equal, and its throughput logging divides by their difference. The failure is reproducible locally by making those clock readings equal; it is separate from the unrelated Windows registry-encoding failure.

Please make the focused test deterministic with a controlled advancing clock, or make the throughput calculation safely handle a zero elapsed interval. Do not rely on a sleep or retry to make this test pass.

qinxuye (commented):
Three issues to address together: fresh-prefill initialization, stop-string output bounds, and the new regression test's clock dependency. The existing focused suite passes locally (10 passed, 1 skipped), but the additional cases below reproduce the failures.

xinference/model/llm/transformers/utils.py
qinxuye: Verified fixed on e3618b68: the implicit limit is resolved after prefill builds prompt tokens and persists across subsequent decode steps. The fresh-request regression now starts with prompt_tokens unset. Current GPU CI also passes.

xinference/model/llm/transformers/utils.py
qinxuye: Verified fixed on e3618b68: stop-string and stop-token checks use only visible tokens. Regressions cover max_tokens=1 with both streaming modes and matching/nonmatching stop strings beyond the visible limit, including output, finish reason, and usage.

xinference/model/llm/transformers/tests/test_batch_inference.py
qinxuye: Verified fixed on e3618b68: throughput logging now guards against nonpositive elapsed time, with a constant-clock regression. Local batch inference, cache compatibility, and batch scheduler suites: 22 passed, 1 skipped.

qinxuye (approved):
LGTM

PR #5513, opened 2026-09-09
fix(transformers): propagate batched generate request IDs

m199369309:
## Summary

- propagate actor-level request IDs into the generation config consumed by the Transformers batch scheduler
- avoid passing `request_id` as an unsupported keyword to the concrete Transformers `generate` method
- copy the generation config before injection so the caller's dictionary is not mutated
- add regression coverage with the real `(prompt, generate_config=None)` method shape

## Testing

- `pytest -q xinference/core/tests/test_model.py -k moves_request_id_into_batch_config` (1 passed)
- `pre-commit run --files xinference/core/model.py xinference/core/tests/test_model.py`


qinxuye (approved):
LGTM

PR #5515, opened 2026-09-10
BLD: support Python 3.14 and require xoscar 0.10.0

qinxuye:
## Summary

- retain Xinference's Python >=3.10 floor and add the Python 3.14 classifier
- require xoscar>=0.10.0 in package metadata and the CPU Docker dependency set
- test every supported Python version on Linux, while testing only 3.10 and 3.14 on macOS and Windows
- keep the legacy NumPy 1.x pin off Python 3.14 and cover the supported-version metadata with a regression test

## Validation

- `pre-commit run --files .github/workflows/python.yaml AGENTS.md pyproject.toml xinference/deploy/docker/requirements_cpu/requirements_cpu-base.txt xinference/tests/test_build_backend.py`
- Python 3.14.6 editable install with xoscar 0.10.0
- Python 3.14.6 CI dependency install and import smoke test, including `transformers<4.49`, bitsandbytes, and xllamacpp
- `python3.14 -m pytest -q xinference/tests/test_build_backend.py xinference/core/tests/test_utils.py` (75 passed)


OliverBryant (approved):
Reviewed the full diff against `main`. No blockers.

What I verified:

- **xoscar floor**: `xoscar` 0.10.0 is on PyPI (released 2026-09-09), `requires_python >= 3.10`, and ships `cp310`–`cp314` wheels, so the new floor is satisfiable on every version in the matrix, including 3.14. 0.9.10 only went up to `cp313`, so the bump is actually required for the new job, not just cosmetic.
- **No stale pins left behind**: `pyproject.toml` and `requirements_cpu-base.txt` were the only two `xoscar` version pins in the tree (`requirements-runtime.txt`, the GPU `Dockerfile`, and `release.yaml` carry none), so nothing else drifts.
- **Matrix / excludes**: quoting the versions in `exclude` is a correctness improvement (unquoted `3.10` would parse as the float `3.1`). The metal `include` on macOS 3.10 is untouched.
- **NumPy step narrowing**: the old condition covered `(macos && 3.13) || windows`. NumPy 1.x has no `cp313`/`cp314` wheels, and macOS 3.13 plus Windows 3.11–3.13 are now excluded, so `windows && 3.10` is the only cell where the `numpy<2` pin still applies and is installable. The narrowing is consistent rather than a behavior change.
- **PyTorch / transformers gating**: the `!= 3.13` shell branch and the two `if:` conditions were updated together, so 3.14 takes the same path 3.13 already does.
- **Regression test**: `test_project_declares_supported_python_versions_and_xoscar_floor` reuses the existing `_REPO_ROOT` / `tomllib` fixtures in the file and asserts the exact strings this PR introduces, so it would catch an accidental revert of either pin.

Non-blocking note: the `build_test_job` matrix (including the three new 3.14 cells) was still pending when I looked, so the final proof that the 3.14 dependency set resolves on all three OSes is the CI run itself. Approving on the strength of the diff; please just make sure the 3.14 jobs go green before merging.

PR #5519, opened 2026-09-10
fix(core): preserve model GPU memory across transient failures

m199369309:
## What does this PR do?

Makes per-model GPU-memory telemetry resilient to transient NVML and worker status collection failures.

The previous protocol treated a missing/failed sample the same as a successful empty sample, so the supervisor immediately discarded the last valid GPU-memory snapshot and `/v1/models` temporarily lost `gpu_memory`.

This change introduces explicit three-state semantics:

- field omitted: collection was skipped or failed; preserve the last valid snapshot;
- `model_gpu_memory: {}`: collection succeeded with no model-owned GPU process; clear immediately;
- non-empty mapping: replace the cached snapshot and refresh its timestamp.

Additional changes:

- adds strict NVML collection while preserving the existing non-strict API default;
- logs collection failures with throttled tracebacks and recovery messages;
- expires retained snapshots after `XINFERENCE_MODEL_GPU_MEMORY_CACHE_TTL` (90 seconds by default);
- clears snapshots immediately when a worker, model, or replica is removed;
- handles `psutil.ZombieProcess` during recursive PID expansion;
- documents the new environment variable in English and all maintained locales.

The `/v1/models` response shape is unchanged.

## Verification

- `pytest -q xinference/core/tests/test_autostart.py xinference/core/tests/test_list_models.py xinference/core/tests/test_worker.py` (147 passed)
- `pre-commit run --files <all modified files>` (passed)
- `msgfmt --check --check-format` for all 9 modified locale catalogs (passed; existing header warnings only)
- `python doc/build_i18n.py --all` (passed)
- localized `environments` Sphinx builds for `de`, `es`, `fr`, `it`, `ja`, `ko`, `pt_BR`, `zh_CN`, and `zh_TW` (passed)

Not run: the full CI-style suite; model tests require optional dependencies, model downloads, and/or GPU hardware.


xinference/core/supervisor.py
```diff
@@ -5275,6 +5372,7 @@ async def list_models(self) -> Dict[str, Dict[str, Any]]:
         # Use the timestamp (not the dict) as the cache-validity sentinel
         # so that a genuinely empty model list ({}) is also cached.
         now = time.time()
+        SupervisorActor._expire_worker_model_gpu_memory(self, now)
```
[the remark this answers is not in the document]
m199369309: Fixed in 969b489cc31e12efe866deff9c3bac868c749ac7 by dispatching through `self._expire_worker_model_gpu_memory(now)`.

xinference/core/supervisor.py
```diff
@@ -5293,6 +5391,7 @@ async def list_models(self) -> Dict[str, Dict[str, Any]]:
         # cache (which may have been filled by the preceding sweep).
         async with self._list_models_sweep_lock:
             now = time.time()
+            SupervisorActor._expire_worker_model_gpu_memory(self, now)
```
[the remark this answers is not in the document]
m199369309: Fixed in 969b489cc31e12efe866deff9c3bac868c749ac7; the locked recheck now also calls the method through `self`.

xinference/core/tests/test_list_models.py
```diff
@@ -31,9 +31,17 @@ async def list_models(self) -> Dict[str, Dict[str, Any]]:
...
 
     list_models = SupervisorActor.list_models
+    _invalidate_list_models_debounce_cache = (
+        SupervisorActor._invalidate_list_models_debounce_cache
+    )
+    _clear_worker_model_gpu_memory = SupervisorActor._clear_worker_model_gpu_memory
+    _process_model_gpu_memory_report = SupervisorActor._process_model_gpu_memory_report
+    _is_valid_model_gpu_memory = staticmethod(
+        SupervisorActor._is_valid_model_gpu_memory
+    )
```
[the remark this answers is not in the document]
m199369309: Fixed in 969b489cc31e12efe866deff9c3bac868c749ac7 by borrowing `_expire_worker_model_gpu_memory` in `DummySupervisor`.

xinference/core/supervisor.py
```diff
@@ -5265,6 +5272,96 @@ async def describe_model(self, model_uid: str) -> Dict[str, Any]:
...
+
+        for worker_address in expired_workers:
+            self._worker_model_gpu_memory.pop(worker_address, None)
+            self._worker_model_gpu_memory_update_time.pop(worker_address, None)
+            logger.info(
+                "Expired model GPU memory snapshot for worker %s after %ss",
+                worker_address,
+                XINFERENCE_MODEL_GPU_MEMORY_CACHE_TTL,
+            )
+        self._invalidate_list_models_debounce_cache()
```
qinxuye, line 5316: [P2] Remove stale GPU memory from the model-list fallback path as well

Invalidating the debounce cache does not remove `gpu_memory` already written into the nested specs in `_list_models_cache`: `list_models()` stores each worker result there and subsequently mutates those same spec dictionaries when adding aggregated GPU memory. If a successful listing is followed by snapshot expiry and a failed worker `list_models()` RPC, the fallback returns those dictionaries; because the new aggregation has no GPU entry, the old field is never removed. I reproduced this on the current head: after aging the snapshot past the TTL, `_worker_model_gpu_memory` is empty but `/v1/models` still returns the previous `gpu_memory` value.

Please keep raw worker-cache specs separate from decorated response dictionaries and ensure absence of a current GPU snapshot removes any previous field. Add regression coverage that first populates the list cache, then expires (or explicitly clears) the snapshot and makes the worker RPC fail; the returned model must no longer contain stale GPU memory.

xinference/core/supervisor.py
```diff
@@ -4990,6 +4995,7 @@ async def _terminate_model_replica(
                 ),
             },
         )
+        self._clear_replica_model_gpu_memory(replica_model_uid)
```
qinxuye, line 4998: [P2] Update the replica-termination test fixture for the new cleanup call

`test_terminate_model_replica_updates_active_replica_set` in `xinference/core/tests/test_launch_strategy.py` borrows `_terminate_model_replica` on a `DummySupervisor`, but that fixture has neither `_clear_replica_model_gpu_memory` nor its required GPU-cache state. This new call therefore raises `AttributeError` in the existing test. I reproduced it locally (20 passed, 1 failed in this module), and the Linux 3.10 CI log reports the identical failure.

Please update this fixture with the cleanup method and required state, and run `test_launch_strategy.py` alongside the three modules already validated. Prefer seeding a GPU snapshot and asserting the terminated replica is removed while other replicas remain, so this new termination behavior is covered too.

qinxuye (commented):
Two concrete issues remain: stale GPU memory survives TTL expiry through the model-list fallback cache, and the replica-termination test fixture was not updated for the new cleanup call. Details and regression coverage requests are inline.

xinference/core/supervisor.py
m199369309: Fixed in 0f17907df4f4c430b010d243fcc4ea979f5552d1. `list_models()` now copies each raw worker spec before adding response-only `replica` and `gpu_memory` fields, so `_list_models_cache` remains pristine. Added a regression test covering a successful listing, GPU snapshot expiry, a failed worker RPC, and fallback without stale `gpu_memory`.

xinference/core/supervisor.py
m199369309: Fixed in 0f17907df4f4c430b010d243fcc4ea979f5552d1. The termination test fixture now borrows `_clear_replica_model_gpu_memory`, initializes the required cache state, and verifies that terminating replica 1 removes only its GPU snapshot while preserving replicas 0 and 2. The focused four-module suite passes (169 tests).

qinxuye (approved):
LGTM. Both previous findings are fixed, with 169 focused tests passing on the reviewed pre-rebase head. Verified that all four rebased commits have identical stable patch IDs to the reviewed commits; the new base includes the already reviewed and merged #5518 and #5516. All review threads are resolved.

PR #5521, opened 2026-09-11
FIX: avoid implicit event loop in async client cleanup

qinxuye:
Async client finalizers could raise an unraisable RuntimeError when no event loop was available, including after asyncio.run() returned.

Record the session creation loop and queue best-effort cleanup on it using call_soon_threadsafe. Paused loops execute cleanup when resumed; collection in another thread still targets the owner loop. Missing or closed owner loops are skipped. Callers should continue to explicitly await close() before shutting down their loop.

Validation: 10 cleanup tests passed on Python 3.11 and 3.14, including real aiohttp sessions on paused loops and cross-thread collection with asyncio debug mode enabled. Pre-commit passed.

### DIFF
--- a/xinference/client/restful/async_restful_client.py
+++ b/xinference/client/restful/async_restful_client.py
@@ -73,12 +73,20 @@ async def _release_response(response: aiohttp.ClientResponse):
 def _schedule_session_close(client: Any) -> None:
     if not getattr(client, "session", None):
         return
-    try:
-        loop = asyncio.get_running_loop()
-    except RuntimeError:
-        # Async cleanup cannot run after the owning event loop has stopped.
+    loop = getattr(client, "_session_loop", None)
+    if loop is None or loop.is_closed():
         return
-    loop.create_task(client.close())
+
+    def close_on_owner_loop():
+        loop.create_task(client.close())
+
+    try:
+        # Also queues cleanup while a manually driven loop is paused, and when
+        # the last reference is released by another thread.
+        loop.call_soon_threadsafe(close_on_owner_loop)
+    except RuntimeError:
+        # The owning loop may have closed since the check above.
+        pass
 
 
 class AsyncRESTfulModelHandle:
@@ -92,6 +100,7 @@ class AsyncRESTfulModelHandle:
         self._base_url = base_url
         self.auth_headers = auth_headers
         self.timeout = aiohttp.ClientTimeout(total=1800)
+        self._session_loop = asyncio.get_running_loop()
         self.session = aiohttp.ClientSession(
             connector=aiohttp.TCPConnector(force_close=True)
         )
@@ -1286,6 +1295,7 @@ class AsyncClient:
         self._headers: Dict[str, str] = {}
         self._cluster_authed = False
         self.timeout = aiohttp.ClientTimeout(total=1800)
+        self._session_loop = asyncio.get_running_loop()
         self.session = aiohttp.ClientSession(
             connector=aiohttp.TCPConnector(force_close=True), timeout=self.timeout
         )
--- a/xinference/client/restful/async_restful_client.py
+++ b/xinference/client/restful/async_restful_client.py
@@ -70,6 +70,17 @@ async def _release_response(response: aiohttp.ClientResponse):
     await response.wait_for_close()
 
 
+def _schedule_session_close(client: Any) -> None:
+    if not getattr(client, "session", None):
+        return
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:
+        # Async cleanup cannot run after the owning event loop has stopped.
+        return
+    loop.create_task(client.close())
+
+
 class AsyncRESTfulModelHandle:
     """
     A sync model interface (for RESTful client) which provides type hints that makes it much easier to use xinference
@@ -92,9 +103,7 @@ class AsyncRESTfulModelHandle:
             self.session = None
 
     def __del__(self):
-        if self.session:
-            loop = asyncio.get_event_loop()
-            loop.create_task(self.close())
+        _schedule_session_close(self)
 
 
 class AsyncRESTfulEmbeddingModelHandle(AsyncRESTfulModelHandle):
@@ -1291,9 +1300,7 @@ class AsyncClient:
             self.session = None
 
     def __del__(self):
-        if self.session:
-            loop = asyncio.get_event_loop()
-            loop.create_task(self.close())
+        _schedule_session_close(self)
 
     def _set_token(self, token: Optional[str]):
         if not self._cluster_authed or token is None:
--- a/xinference/client/tests/test_async_client_cleanup.py
+++ b/xinference/client/tests/test_async_client_cleanup.py
@@ -15,7 +15,6 @@
 import asyncio
 import gc
 import threading
-from unittest.mock import MagicMock
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


xinference/client/restful/async_restful_client.py
```diff
@@ -70,6 +70,17 @@ async def _release_response(response: aiohttp.ClientResponse):
     await response.wait_for_close()
 
 
+def _schedule_session_close(client: Any) -> None:
+    if not getattr(client, "session", None):
+        return
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:
+        # Async cleanup cannot run after the owning event loop has stopped.
```
OliverBryant: This introduces a small behavior regression compared to the previous `get_event_loop()` version.

Repro: create a client under a manually driven loop, drop it *between* two `run_until_complete()` calls, then run the loop again:

```python
loop = asyncio.new_event_loop(); asyncio.set_event_loop(loop)
async def mk(): return AsyncClient(...)
c = loop.run_until_complete(mk()); del c; gc.collect()
loop.run_until_complete(asyncio.sleep(0.05))
```

Before: `get_event_loop()` returned the current (non-running) loop, `close()` was scheduled and ran on the next `run_until_complete`, session closed cleanly.
After: `get_running_loop()` raises, we return, and the session is never closed (aiohttp emits `Unclosed client session`).

Since the connector is `force_close=True` there are no live sockets leaked, so I don't consider this blocking. If you want to keep the old coverage (and also avoid depending on whichever loop happens to be running in the collecting thread), one option is to fall back to the session's own loop:

```python
try:
    loop = asyncio.get_running_loop()
except RuntimeError:
    loop = getattr(client.session, "_loop", None)
    if loop is None or loop.is_closed():
        return
loop.create_task(client.close())
```

Trade-off: `_loop` is a private aiohttp attribute. Fine to leave as is if you'd rather not take that dependency.

OliverBryant (commented):
Verified the fix locally on Python 3.12.4 and 3.14.6 with aiohttp 3.14.3. The `Exception ignored in __del__: RuntimeError: There is no current event loop` case (an object escaping `asyncio.run()` and being collected afterwards) is gone on both versions.

One note on scope: the same `RuntimeError` reproduces on 3.12 as well, not only on 3.14, so the title/description could drop the version qualifier.

The change itself looks good and the tests are solid (both would fail if the fix were reverted). Left two inline notes: one about a small behavior regression for manually driven loops, and a wording nit.

xinference/client/restful/async_restful_client.py
```diff
@@ -70,6 +70,17 @@ async def _release_response(response: aiohttp.ClientResponse):
     await response.wait_for_close()
 
 
+def _schedule_session_close(client: Any) -> None:
+    if not getattr(client, "session", None):
+        return
+    try:
+        loop = asyncio.get_running_loop()
```
OliverBryant: Not introduced by this PR, just noting it while we're here: `get_running_loop()` returns the loop running in the *current* thread, which is not necessarily the loop the session belongs to. If GC fires in another thread that has its own running loop, `close()` gets scheduled on the wrong loop. The previous `get_event_loop()` had the same issue. The `session._loop` fallback suggested below would also address this.

xinference/client/restful/async_restful_client.py
```diff
@@ -70,6 +70,17 @@ async def _release_response(response: aiohttp.ClientResponse):
...
+    except RuntimeError:
```
OliverBryant, line 87: Nit: the comment slightly overstates the condition. This branch is hit whenever the collecting thread has no *running* loop, which includes "loop exists but is not currently running" and "GC in a different thread", not only "the owning loop has stopped". Something like `# No running loop in this thread; async cleanup cannot be scheduled.` would be more accurate.

xinference/client/restful/async_restful_client.py
qinxuye: Fixed in 0f4b0030e. Both constructors now record the session creation loop, and finalization queues cleanup on that loop with call_soon_threadsafe. This preserves cleanup between run_until_complete calls without using aiohttp private attributes. Real aiohttp session tests cover paused loops and collection in another thread with its own running loop. All 10 cleanup tests pass on Python 3.11 and 3.14.

xinference/client/restful/async_restful_client.py
qinxuye: Fixed in 0f4b0030e: cleanup always targets the recorded session creation loop via call_soon_threadsafe. The cross-thread regression test enables asyncio debug mode and verifies session.close executes on the owner loop, not the collecting thread loop.

xinference/client/restful/async_restful_client.py
qinxuye: Updated in 0f4b0030e. The implementation now explicitly handles missing or closed owner loops, and the comment describes the possible race with loop closure. Paused owner loops queue cleanup for their next run.

xinference/client/restful/async_restful_client.py
```diff
@@ -81,6 +100,7 @@ def __init__(self, model_uid: str, base_url: str, auth_headers: Dict):
         self._base_url = base_url
         self.auth_headers = auth_headers
         self.timeout = aiohttp.ClientTimeout(total=1800)
+        self._session_loop = asyncio.get_running_loop()
```
OliverBryant, line 103: Non-blocking: with this line, constructing `AsyncClient(...)` / a handle outside a running loop now raises `RuntimeError: no running event loop`.

I checked that this is not a new restriction for current installs: `aiohttp.ClientSession(...)` itself raises the same error outside a running loop since aiohttp 3.10 (verified on 3.10.11 and 3.11.18; 3.9.5 only warns). All in-repo construction sites (`get_model`, the client tests) are already inside coroutines. So the only users affected are those pinned to aiohttp <= 3.9, since `pyproject.toml` leaves `aiohttp` unpinned. Probably acceptable, just flagging it so it's a conscious choice.

xinference/client/tests/test_async_client_cleanup.py
```diff
@@ -0,0 +1,132 @@
...
+import pytest
+
+from ..restful import async_restful_client as client_module
+
+
+@pytest.mark.parametrize(
+    "client_class",
+    [client_module.AsyncRESTfulModelHandle, client_module.AsyncClient],
+)
+def test_destructor_does_not_require_an_event_loop(client_class, monkeypatch):
```
OliverBryant: Nit: after the second commit this test no longer exercises the code path it was written for. `_schedule_session_close` doesn't call `get_running_loop()` anymore, and since `__new__` skips `__init__`, `_session_loop` is missing and the function returns early before touching either patched attribute. It still passes, but it is effectively testing the `_session_loop is None` early return. Fine to keep as a guard against reintroducing `get_event_loop()`, or drop it in favor of `test_destructor_after_owner_loop_closed`, whichever you prefer.

OliverBryant (approved):
Re-verified 0f4b0030e on Python 3.12.4 and 3.14.6 with aiohttp 3.14.3 across the scenarios from the first round:

- collected inside `asyncio.run()`: closed on the owner loop
- collected after `asyncio.run()` returned: no exception, only aiohttp's `Unclosed client session` warning (the loop is already closed, nothing more can be done)
- collected between two `run_until_complete()` calls: closed on the owner loop (this was the regression in the first version, now fixed)
- collected from another thread with its own running loop: closed on the owner loop, not the collecting thread's loop

No scenario is worse than the current `main`, and the three that were broken are fixed. The new real-session test with `set_debug(True)` covers exactly the cases that mattered. LGTM.

Two non-blocking notes inline.

xinference/client/restful/async_restful_client.py
qinxuye: Thanks for checking the version boundary. This is an intentional tradeoff: async clients and model handles should be constructed inside their running event loop, matching aiohttp >=3.10 and the in-repo usage. Users pinned to aiohttp <=3.9 who construct outside a running loop will need to move construction into their coroutine. We will retain explicit owner-loop capture and avoid a legacy implicit-loop fallback.

xinference/client/tests/test_async_client_cleanup.py
qinxuye: Updated in 62ea8d7b6: renamed the test to test_destructor_without_initialized_owner_loop and removed the obsolete get_running_loop mock. The get_event_loop guard remains to prevent implicit loop creation during partial initialization cleanup. All 10 cleanup tests pass on Python 3.14; pre-commit passes.

PR #5526, opened 2026-09-12
feat(audio): add YuE2 music generation

Minamiyama:
## Summary

- Add YuE2-3B text-to-music support with Hugging Face and ModelScope sources.
- Vendor YuE2 v0.1.6 runtime to avoid GitHub installation during model launch.
- Accept and ignore frontend `duration` for YuE2 requests.
- Add engine, registry, runtime, tests, and built-in model docs.

### DIFF
--- a/xinference/model/audio/engine.py
+++ b/xinference/model/audio/engine.py
@@ -33,6 +33,7 @@ from .vllm import VLLMQwen3ASRModel
 from .voxcpm import VoxCPMModel
 from .whisper import WhisperModel
 from .whisper_mlx import WhisperMLXModel
+from .yue2 import YuE2Model
 
 if TYPE_CHECKING:
     from .core import AudioModelFamilyV2
@@ -238,6 +239,20 @@ class PyTorchAceStepAudioModel(AceStepModel, AudioEngineModel):
         return model_family.model_family == "ace_step_1_5"
 
 
+class PyTorchYuE2AudioModel(YuE2Model, AudioEngineModel):
+    required_libs = ("yue2",)
+
+    @classmethod
+    def check_lib(cls):
+        if virtual_env_allows_missing_engine():
+            return True
+        return super().check_lib()
+
+    @classmethod
+    def match(cls, model_family: "AudioModelFamilyV2") -> bool:
+        return has_cuda_device() and model_family.model_family == "yue2"
+
+
 class MLXWhisperAudioModel(WhisperMLXModel, AudioEngineModel):
     required_libs = ("mlx",)
 
@@ -360,6 +375,7 @@ def register_builtin_audio_engines() -> None:
     SUPPORTED_ENGINES["vLLM"] = [VLLMQwen3ASRAudioModel]
     SUPPORTED_ENGINES["PyTorch"] = [
         PyTorchAceStepAudioModel,
+        PyTorchYuE2AudioModel,
         PyTorchF5TTSAudioModel,
         PyTorchKokoroAudioModel,
         PyTorchFunASRAudioModel,
--- a/xinference/model/audio/tests/test_audio_engine.py
+++ b/xinference/model/audio/tests/test_audio_engine.py
@@ -46,8 +46,8 @@ from ..engine import (
     PyTorchKokoroAudioModel,
     PyTorchMeloTTSAudioModel,
     PyTorchQwen3TTSAudioModel,
-    PyTorchYuE2AudioModel,
     PyTorchVoxCPMAudioModel,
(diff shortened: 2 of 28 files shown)


qinxuye (commented):
One reproducible engine-discovery issue remains when virtual environments are explicitly disabled. The existing 45 focused tests pass, but do not cover discovery of the vendored runtime in this mode.

xinference/model/audio/engine.py
```diff
@@ -238,6 +239,20 @@ def match(cls, model_family: "AudioModelFamilyV2") -> bool:
...
 
 
+class PyTorchYuE2AudioModel(YuE2Model, AudioEngineModel):
+    required_libs = ("yue2",)
+
+    @classmethod
+    def check_lib(cls):
+        if virtual_env_allows_missing_engine():
+            return True
+        return super().check_lib()
```
qinxuye, line 250: [P2] Make engine discovery recognize the vendored YuE2 runtime

With `enable_virtual_env=False`, this falls through to `AudioEngineModel.check_lib()`, which uses `find_spec('yue2')`. However, the bundled package lives under `xinference/thirdparty/yue2/src`, and that directory is only added to `sys.path` later by `YuE2Model.load()`. On a fresh process without a separately installed yue2 package, engine discovery therefore reports PyTorch as unavailable even though the runtime is shipped with Xinference; the UI cannot select it to reach load(). Installing the dependencies listed in the model spec does not install a top-level yue2 package either.

I reproduced this through `get_engine_params_by_name('audio', 'YuE2-3B', enable_virtual_env=False)` with the CUDA availability probe mocked true: it returns `{'PyTorch': "Library 'yue2' is not installed"}`. Calling `_ensure_vendored_source_path()` and repeating the same query immediately returns the available PyTorch engine.

Please make the discovery check recognize the vendored runtime before testing availability (while keeping actual external dependency checks appropriate), and add a regression for a fresh process with virtualenv disabled and no separately installed yue2 package. Keep heavy runtime imports lazy.

xinference/model/audio/tests/test_audio_engine.py
```diff
@@ -305,6 +308,75 @@ def test_minimax_music3_without_cuda_is_rejected_before_download():
...
+        sys, "path", [path for path in sys.path if path != _YUE2_VENDOR_ROOT]
+    )
+    monkeypatch.setattr(engine_mod, "has_cuda_device", lambda: True)
+    monkeypatch.setattr(
+        audio_engine_family.importlib.util,
+        "find_spec",
+        find_spec_without_external_yue2,
+    )
+
+    assert get_engine_params_by_name("audio", "YuE2-3B", enable_virtual_env=False) == {
```
qinxuye: [P2] Fix both the discovery fixture and the expected model format

This new test fails in two ways. On a non-CUDA host, the module's autouse `_install()` runs before the test patches `has_cuda_device`, so YuE2 has already been excluded from `AUDIO_ENGINES`; patching the probe at line 368 does not rebuild that registry, and discovery returns `None`. If CUDA is mocked before registration, the test still fails because the current YuE2 spec has no `model_format`: discovery returns `{'PyTorch': [{'model_name': 'YuE2-3B', 'model_format': None}]}`, not the asserted `'pytorch'` value. I reproduced both failures independently (the two audio test files otherwise have 45 passing tests). Please build isolated YuE2 engine registry state after applying the CUDA mock, restore it after the test, and align the expected format with the actual spec. Keep the no-external-yue2 check and verify that discovery does not import the heavy runtime. The production discovery fix itself passes the fresh-process check.

qinxuye (commented):
The original vendored-runtime discovery issue is fixed. The new regression test still fails for two independent reasons below.

xinference/model/audio/engine.py
qinxuye: Verified fixed in 53630e85: discovery now exposes the vendored source before find_spec. A fresh-process check without an externally installed yue2 package successfully discovers PyTorch and leaves yue2 unimported. The separate regression-test failures are covered in a new inline comment.

xinference/model/audio/tests/test_audio_engine.py
qinxuye: Verified fixed in d872bd12: the test rebuilds isolated engine registry state after mocking CUDA, restores it afterward, expects the actual null model format, and checks that yue2 remains unimported. All 46 focused audio tests pass locally.

qinxuye (approved):
LGTM. Both review issues are fixed; all 46 focused audio tests pass. Real model inference was not run locally.

PR #5531, opened 2026-09-12
doc: update README model highlights for v3.4.0

qinxuye:
## Summary

- rotate the New Models highlights to the latest two released versions, v3.3.0 and v3.4.0, keeping eight as a minimum rather than a maximum
- add MiniCPM5-2B, Fish Audio S1-mini/S2-Pro, MonkeyOCR, dots.ocr, and the JoyAI image editing series
- synchronize the root README and all nine localized READMEs with 24 model/model-series entries

## Details

- remove highlights introduced only in v3.2.0
- retain MiniMax-H3 Lightning LoRA because its integration was introduced in v3.3.0, and link only the corresponding PR
- exclude AuK/AuK-Flash because they landed after the v3.4.0 release tag
- do not treat the new Gemma-4 backend or the previously supported MiniCPM5-1B as new models
- change only standalone multilingual README files; no Sphinx source pages or gettext catalogs are affected

## Validation

- verified all ten New Models sections contain exactly 24 entries
- verified the model and PR links match row by row across all ten README files
- verified all 35 unique linked model repositories return HTTP 200
- `git diff --check` and `git diff --cached --check`
- pre-commit ran during commit; no configured hooks apply to Markdown files


OliverBryant (approved):
LGTM

PR #5532, opened 2026-09-12
fix(security): reject local media paths and bound server-side media fetches

bluefish-08:
Chat requests can carry `image_url` / `video_url` / `audio_url` values that the server fetches itself. Today nothing constrains what those values may point at, and nothing bounds how long a fetch may take.

### What a client can currently do

`qwen_omni_utils.fetch_image` (and `qwen_vl_utils`) dispatch on the string prefix:

```python
elif image.startswith("http://") or image.startswith("https://"):
    with requests.get(image, stream=True) as response:   # no timeout
...
elif image.startswith("file://"):
    image_obj = Image.open(image[7:])
else:
    image_obj = Image.open(image)                        # bare path -> local file
```

The `else` branch means any string that is not a recognised URI is opened off the server's filesystem. Xinference's own `_decode_image` / `_decode_image_without_rgb` in `model/llm/utils.py` have the same two problems: `requests.get(_url)` without a timeout, and a `MissingSchema` fallback to `Image.open(_url)`.

I observed all three of these being exercised against a deployment:

- `http://<attacker-host>:28012/ssrf-http` — an out-of-band SSRF probe; the server connected out.
- `/etc/hostname`, `/etc/os-release`, `/root/.xinference/config.json`, `/root/.xinference/` — the errors differ per path (`Is a directory` vs `No such file or directory` vs `cannot identify image file`), which is a working file-existence oracle. A path pointing at a readable *image* would have been read and fed to the model.
- Each probe pinned the model for ~130s (`Leave chat, error: ... elapsed time: 131 s`).

### Why the stall is worse than one slow request

`async_chat` is a coroutine, so `ModelActor._call_wrapper` awaits it directly on the actor's event loop:

```python
if inspect.iscoroutinefunction(fn):
    return await fn(*args, **kwargs)      # on the event loop
return await asyncio.to_thread(fn, *args, **kwargs)
```

`process_vision_info` / `process_audio_info` / `process_mm_info` were called inline inside that coroutine, so their blocking downloads froze the whole event loop for that actor — not just the offending request. In the incident that included `terminate_model`, which logged `Enter terminate_model` and never completed: the model could not even be shut down. Note this is independent of `allow_batch`: vLLM sets `allow_batch = True` so `self._lock is None` and there is no lock contention, yet the actor was still wedged.

### Changes

1. **`xinference/model/media_source.py`** (new) — `validate_media_source()` / `validate_messages_media()`. `data:` URIs and `http(s)://` URLs pass; everything else (a `file://` URI or a bare path) is refused. The point is that the default is deny, rather than falling through to the filesystem.

2. **`ChatModelMixin._transform_messages`** — one validation call. Every engine's multimodal path funnels through it, so this covers vLLM, SGLang and all Transformers multimodal models at once; I checked each of the five `process_vision_info` call sites and they are all downstream of it (`ovis2` reaches it via `build_inputs_from_messages` -> `_generate_chat_data`).

3. **`_decode_image` / `_decode_image_without_rgb`** — validate first, pass `timeout=`, and `raise_for_status()`. `_decode_image` now delegates to `_decode_image_without_rgb` instead of duplicating the body.

4. **vLLM / SGLang `async_chat`** — the media helpers are dispatched through `asyncio.to_thread`, so a slow fetch no longer blocks the actor's event loop.

### Configuration

| Variable | Default | Effect |
| --- | --- | --- |
| `XINFERENCE_MEDIA_ALLOW_LOCAL_PATH` | `false` | Restores reading media off the server filesystem |
| `XINFERENCE_MEDIA_BLOCK_PRIVATE_ADDRESS` | `false` | Also refuse URLs resolving to loopback/private/link-local |
| `XINFERENCE_MEDIA_FETCH_TIMEOUT` | `30` | Seconds allowed for a fetch |

Two deliberate defaults, both arguable:

- Rejecting local paths **is** a behaviour change for anyone passing a server-side path today. I judged that worth it since it is the actual file-read primitive, and `XINFERENCE_MEDIA_ALLOW_LOCAL_PATH=1` restores it. Happy to flip the default if you would rather deprecate it over a release.
- Private-address blocking is **opt-in**, because serving media from an intranet host is a normal deployment and defaulting it on would break those users. That does mean SSRF to internal addresses is still reachable out of the box.

### Limitations

- The private-address check resolves the hostname once and the fetcher resolves again, so DNS rebinding defeats it. Noted in a comment; network policy remains the real control.
- `qwen_omni_utils` / `qwen_vl_utils` are unchanged — this relies on every media path passing through `_transform_messages`. That holds for all current call sites, but a future path calling those helpers directly would bypass it.
- The fetch is now off the event loop but still unpaginated/unbounded in size; a very large remote file is still a resource concern.

### Verification

- 19 new tests. The rejection cases use the exact payloads from the incident (`/etc/hostname`, `/etc/os-release`, `/root/.xinference/config.json`).
- Integration-checked against the real `ChatModelMixin._transform_messages`: the payload raises, a normal `https://` URL still transforms to `{'type': 'image', 'image': ...}`.
- The event-loop regression test fails if either engine goes back to calling the helpers inline (verified by reverting both files).
- `pre-commit run --files` clean on every touched file.

Not run: a live multimodal generation against real weights (no GPU available here), so the `to_thread` change is verified by tests and inspection rather than an end-to-end image request.

### DIFF
--- a/xinference/model/llm/utils.py
+++ b/xinference/model/llm/utils.py
@@ -40,6 +40,7 @@ from typing import (
 import requests
 from PIL import Image
 
+from ...constants import XINFERENCE_MEDIA_FETCH_TIMEOUT
 from ...types import (
     ChatCompletion,
     ChatCompletionChoice,
@@ -57,6 +58,7 @@ from ...types import (
     CompletionUsage,
     ToolCallDelta,
 )
+from ..media_source import validate_media_source, validate_messages_media
 from .core import chat_context_var
 from .reasoning_parser import ReasoningParser
 from .tool_parsers.glm4_tool_parser import Glm4ToolParser
@@ -1166,6 +1168,9 @@ class ChatModelMixin:
         self,
         messages: Union[List[ChatCompletionMessage], List[dict]],
     ):
+        # Every engine's multimodal path funnels through here, so this is the
+        # one place that sees client-supplied media before something fetches it.
+        validate_messages_media(messages)
         transformed_messages = []
         for msg in messages:
             new_content = []
@@ -1423,25 +1428,12 @@ def get_model_version(
 
 
 def _decode_image(_url):
-    if _url.startswith("data:"):
-        logging.info("Parse url by base64 decoder.")
-        # https://platform.openai.com/docs/guides/vision/uploading-base-64-encoded-images
-        # e.g. f"data:image/jpeg;base64,{base64_image}"
-        _type, data = _url.split(";")
-        _, ext = _type.split("/")
-        data = data[len("base64,") :]
-        data = base64.b64decode(data.encode("utf-8"))
-        return Image.open(BytesIO(data)).convert("RGB")
-    else:
-        try:
-            response = requests.get(_url)
-        except requests.exceptions.MissingSchema:
-            return Image.open(_url).convert("RGB")
-        else:
-            return Image.open(BytesIO(response.content)).convert("RGB")
+    image = _decode_image_without_rgb(_url)
+    return image.convert("RGB")
 
 
 def _decode_image_without_rgb(_url):
+    validate_media_source(_url)
     if _url.startswith("data:"):
         logging.info("Parse url by base64 decoder.")
         # https://platform.openai.com/docs/guides/vision/uploading-base-64-encoded-images
@@ -1453,10 +1445,15 @@ def _decode_image_without_rgb(_url):
         return Image.open(BytesIO(data))
     else:
         try:
-            response = requests.get(_url)
+            # Without a timeout a dead host pins the model actor until the OS
+            # gives up, which stalls every other request on that model.
+            response = requests.get(_url, timeout=XINFERENCE_MEDIA_FETCH_TIMEOUT)
         except requests.exceptions.MissingSchema:
+            # validate_media_source already rejected local paths unless they are
+            # explicitly allowed.
             return Image.open(_url)
         else:
+            response.raise_for_status()
             return Image.open(BytesIO(response.content))
 
 
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -2654,7 +2654,10 @@ class VLLMMultiModel(VLLMModel, ChatModelMixin):
                     temp_dir = tempfile.TemporaryDirectory(prefix="xinference-vllm-")
                     self._handle_base64_media(messages, temp_dir.name)
 
-                messages = self._transform_messages(messages)
+                # Validates client-supplied media references, which resolves
+                # hostnames when private-address blocking is on. getaddrinfo
+                # blocks, so keep it off the actor's event loop too.
+                messages = await asyncio.to_thread(self._transform_messages, messages)
 
                 chat_template_kwargs = (
                     self._get_chat_template_kwargs_from_generate_config(
--- a/xinference/model/llm/vllm/core.py
+++ b/xinference/model/llm/vllm/core.py
@@ -2680,15 +2680,24 @@ class VLLMMultiModel(VLLMModel, ChatModelMixin):
                     model_family
                 )
 
+                # These helpers download every referenced image/audio/video
+                # with blocking IO. Run off the event loop: this coroutine is
+                # awaited directly by the model actor, so a slow fetch here
+                # stalls every other request that actor is serving.
                 if "omni" in self.model_family.model_ability:
-                    audios, images, videos, video_kwargs = process_mm_info(
-                        messages, use_audio_in_video=True, return_video_kwargs=True
+                    audios, images, videos, video_kwargs = await asyncio.to_thread(
+                        process_mm_info,
+                        messages,
+                        use_audio_in_video=True,
+                        return_video_kwargs=True,
                     )
                 elif "audio" in self.model_family.model_ability:
-                    audios = process_audio_info(messages, use_audio_in_video=False)
+                    audios = await asyncio.to_thread(
+                        process_audio_info, messages, use_audio_in_video=False
+                    )
                 elif "vision" in self.model_family.model_ability:
-                    images, videos, video_kwargs = process_vision_info(  # type: ignore
-                        messages, return_video_kwargs=True
+                    images, videos, video_kwargs = await asyncio.to_thread(  # type: ignore
+                        process_vision_info, messages, return_video_kwargs=True
                     )
 
                 prompt = self.get_full_context(
--- a/xinference/model/media_source.py
+++ b/xinference/model/media_source.py
@@ -1,0 +1,148 @@
+# Copyright 2022-2026 Xinference Holdings Pte. Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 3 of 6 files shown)


bluefish-08:
Thanks — applied three of the five, pushed in c488f872.

**Blocking DNS on the event loop (both high-priority comments): fixed, and you were right that it mattered.** I had moved the *fetch* off the loop but left `getaddrinfo` on it, which is the same class of bug this PR is about. `_transform_messages` is now dispatched via `asyncio.to_thread` in both vLLM and SGLang `async_chat`. I checked it is safe to run in a thread first: it only rewrites the message list and calls `_normalize_tool_calls`, with no contextvar writes or shared-state mutation (`chat_context_var` is set by `async_chat` itself, not in there). The regression test now covers `self._transform_messages` alongside the media helpers, and I verified it fails when either call goes back inline.

**`parse_env_bool`: applied.** Worth noting it is not only a consistency win — my original `.lower() in ("1", "true", "yes")` silently evaluates an unrecognised value to `False`, so a typo in `XINFERENCE_MEDIA_BLOCK_PRIVATE_ADDRESS` would quietly disable the protection. `parse_env_bool` raises instead, which is the right behaviour for a security flag.

**`urlparse` ValueError: applied.** Wrapped so a malformed URL surfaces as `MediaSourceError` like every other rejection.

**Numeric env var: not applied** — the suggestion is equivalent to the existing code. It already uses a direct `float()` cast with no try/except and no fallback, so it fails fast exactly as recommended; the only difference is passing `30.0` rather than `"30"` as the default to `os.environ.get`, and a str default matches that function's signature. Happy to change it if you prefer the literal form.

xinference/model/llm/vllm/core.py
```diff
@@ -2654,7 +2654,10 @@ async def async_chat(
                     temp_dir = tempfile.TemporaryDirectory(prefix="xinference-vllm-")
                     self._handle_base64_media(messages, temp_dir.name)
 
-                messages = self._transform_messages(messages)
+                # Validates client-supplied media references, which resolves
+                # hostnames when private-address blocking is on. getaddrinfo
+                # blocks, so keep it off the actor's event loop too.
+                messages = await asyncio.to_thread(self._transform_messages, messages)
```
qinxuye, line 2660: [P1] Validate client references before materializing base64 media

Immediately above, `_handle_base64_media` replaces image data URIs with temporary paths and video data URIs with `file://` URIs. `_transform_messages` now rejects those internally generated references with the default `XINFERENCE_MEDIA_ALLOW_LOCAL_PATH=false`, so ordinary base64 image/video requests fail before preprocessing. Reproduced with the existing `TestVLLMBase64Media::test_async_chat_cleans_materialized_video_after_processing` test, which raises `MediaSourceError` here. Validate the original client references before materialization and distinguish trusted request-created files without enabling arbitrary local paths globally. Keep that regression test and add the corresponding image case.

xinference/model/media_source.py
```diff
@@ -0,0 +1,152 @@
...
+    *,
+    allow_local_path: Optional[bool] = None,
+    block_private_address: Optional[bool] = None,
+) -> None:
+    """Raise :class:`MediaSourceError` if ``source`` must not be fetched.
+
+    Non-string sources (already-decoded ``PIL.Image``, numpy arrays, ...) are
+    passed through: they carry no fetch.
+    """
+    if not isinstance(source, str):
```
qinxuye, line 83: [P1] Validate references inside video frame lists

Non-string values are not necessarily decoded media. `video_url: {"url": ["/tmp/private-image.png"]}` passes `CreateChatCompletion` parsing and this validation, and `_transform_messages` forwards the list as `video`. Qwen's video helper explicitly supports lists/tuples of frame references and calls `fetch_image` for each element, including local paths. This leaves the local-image-read primitive reachable with local paths disabled (and also skips private-address checks for URL elements). Recursively validate frame references, or reject reference containers at this boundary; add API-to-transform regression coverage for local paths and private URLs nested in a video list while preserving valid remote/inline frames.

xinference/model/llm/vllm/core.py
```diff
@@ -2680,15 +2683,24 @@ async def async_chat(
...
                     )
                 elif "audio" in self.model_family.model_ability:
-                    audios = process_audio_info(messages, use_audio_in_video=False)
+                    audios = await asyncio.to_thread(
+                        process_audio_info, messages, use_audio_in_video=False
+                    )
                 elif "vision" in self.model_family.model_ability:
-                    images, videos, video_kwargs = process_vision_info(  # type: ignore
-                        messages, return_video_kwargs=True
+                    images, videos, video_kwargs = await asyncio.to_thread(  # type: ignore
```
qinxuye, line 2702: [P2] Apply the timeout to the actual multimodal downloaders

Moving these calls to a thread keeps the event loop responsive but does not bound the fetch. The new timeout is only passed by `_decode_image_without_rgb`; this path instead calls Qwen's own downloader. In qwen-omni-utils 0.0.9, `fetch_image` still invokes `requests.get(image, stream=True)` without a timeout (verified by capturing its request arguments). Thus an unresponsive media host can keep the request and executor thread occupied beyond `XINFERENCE_MEDIA_FETCH_TIMEOUT`; the equivalent SGLang helper path has the same gap. Route remote media through a bounded downloader or otherwise apply timeouts at the actual I/O layer, and test a slow/stalled endpoint through the engine preprocessing path. Merely timing out the await does not stop a blocked thread.

xinference/model/llm/utils.py
```diff
@@ -1453,10 +1445,15 @@ def _decode_image_without_rgb(_url):
         return Image.open(BytesIO(data))
     else:
         try:
-            response = requests.get(_url)
+            # Without a timeout a dead host pins the model actor until the OS
+            # gives up, which stalls every other request on that model.
+            response = requests.get(_url, timeout=XINFERENCE_MEDIA_FETCH_TIMEOUT)
```
qinxuye, line 1450: [P2] Revalidate redirect destinations when private-address blocking is enabled

`validate_media_source` checks only the initial hostname, while `requests.get` follows redirects automatically. A permitted public URL can return `302 Location: http://127.0.0.1/private.png`, and this fetch accesses the private destination even with `XINFERENCE_MEDIA_BLOCK_PRIVATE_ADDRESS=true`. Reproduced using simulated HTTP responses through Requests' real redirect handling: both URLs were fetched, but only the initial public hostname was checked. This does not require DNS rebinding. Disable automatic redirects and validate each destination before fetching it, or enforce the policy in a shared downloader used by the engine helpers too. Add a public-to-private redirect regression test.

qinxuye (commented):
Four concrete issues found at this head. The 19 new tests pass, but the existing vLLM base64-video integration regression fails. Please cover the following together: trusted materialized media, frame-list validation, actual downloader timeouts, and redirect validation.

bluefish-08:
Superseded by #5533.

The approach here was wrong in a way that matters: validating the URL string and returning does not stop SSRF, because the fetch is done afterwards by the vendor reader (`qwen_omni_utils` / `qwen_vl_utils`, LMDeploy's template), which follows redirects. A URL that validates as public can still 302 into the private network, and pre-flight validation never sees that hop.

#5533 takes over the fetch instead — it downloads into a request-scoped directory and rewrites the message to the local copy, so the reader has no URL left to resolve. It also covers `_transform_messages` gaps (`qwen2_audio` overrides it without calling super, `deepseek_vl2` never calls it) by hooking `build_direct_chat_result`, and enumerates media reference keys rather than matching one shape at a time.

PR #5533, opened 2026-09-13
fix(llm): materialize client-supplied media before vendor readers

bluefish-08:
Chat requests carry `image_url` / `video_url` / `audio_url` values that the
server fetches itself. Unrestricted, that is both a local-file-read and an SSRF
primitive, and the readers involved have no timeout, so one request can pin a
model actor until the OS gives up.

## What this changes

`xinference/model/llm/media.py` gates every client-supplied media reference and,
where a vendor reader would otherwise fetch the URL itself, downloads it here
instead.

**Validation** (`validate_messages_media`) rejects local paths and `file://`
URLs, rejects URLs resolving to loopback/private/link-local addresses, and
enforces a scheme allowlist. It is enough wherever the fetch already goes
through `load_media_bytes`.

**Materialization** (`materialize_messages_media`) is the part that actually
closes SSRF on vLLM / SGLang / LMDeploy. `qwen_omni_utils`, `qwen_vl_utils` and
LMDeploy's template do their own `requests.get` with redirects enabled, so a URL
that validated as public can still 302 into the private network. Pre-flight
validation cannot see that hop. Materialization fetches the bytes into a
request-scoped directory and rewrites the message to the local copy, leaving the
reader nothing to resolve.

Fetching reuses `image/utils.py`'s `_open_public_url`, which re-resolves and pins
every redirect hop to the address it validated and refuses https→http downgrade.
Because materialized paths are then re-validated by downstream prompt builders,
`media_workspace()` registers the server's own directories as readable without
opening local reads to clients.

Media reference keys are enumerated rather than matched shape by shape:
`image_url`, `video_url`, `audio_url`, `image`, `video`, `audio`, `url`, `path`,
`base64`. That is a superset of the keys HF's `processing_utils` extracts, and it
matters because `load_image` tries `os.path.isfile()` before decoding base64 —
the key a value arrives under says nothing about how it will be read.

## New environment variables

| Variable | Default | Effect |
| --- | --- | --- |
| `XINFERENCE_MEDIA_ALLOW_LOCAL_PATH` | `false` | Allow `file://` URLs and bare filesystem paths |
| `XINFERENCE_MEDIA_BLOCK_PRIVATE_ADDRESS` | `true` | Refuse URLs resolving to loopback/private/link-local addresses |
| `XINFERENCE_MEDIA_FETCH_TIMEOUT` | `20` | Seconds per remote URL; a request's media as a whole is capped at 3x this |
| `XINFERENCE_MEDIA_MAX_BYTES` | `67108864` | Maximum size of one fetched media file |

Documented in `doc/source/getting_started/environments.rst` with a zh_CN
translation.

## Behaviour changes worth a release note

- `XINFERENCE_MEDIA_ALLOW_LOCAL_PATH` defaults to `false`: an `image_url` of
  `/data/img.png` now errors. Send a `data:` URI or an HTTP(S) URL, or set the
  variable when every API caller is already trusted with the server's filesystem.
- `XINFERENCE_MEDIA_BLOCK_PRIVATE_ADDRESS` defaults to `true`: requests pointing
  at an intranet object store or a Kubernetes Service DNS name now fail. Set it
  to `false` when the deployment is not publicly exposed.

## Known limits

1. **minicpm-v keeps a redirect SSRF.** Those models hand messages straight to
   `processor.apply_chat_template`, and HF's `load_image` fetches with
   `follow_redirects=True`. This PR closes the LFI and direct-internal-address
   cases there, but not the 302 hop. Materialization covers vLLM, SGLang and
   LMDeploy only.
2. **The HF-side readers are still untimed.** They run under `asyncio.to_thread`,
   i.e. the event loop's shared default executor (`min(32, cpu + 4)` threads).
   Enough concurrent slow responses still exhaust it and stall other `to_thread`
   callers, `Model.stop` and `__pre_destroy__` included, so `terminate_model` can
   still hang. The bar moved from one request to the pool size.
3. **DNS cannot be budgeted.** `socket.getaddrinfo` takes no timeout, so one
   fetch is bounded by `max(XINFERENCE_MEDIA_FETCH_TIMEOUT, resolver timeout)`.
4. **`url` / `path` / `base64` are validated regardless of the part's `type`,**
   for the `load_image` reason above. A non-media content part carrying those keys
   would be rejected; no shape currently produced in this repository is affected.
5. **LMDeploy fetches each image twice** — `get_specific_prompt` downloads and
   discards, `_collect_pil_images` downloads again. Pre-existing inefficiency,
   left alone here.
6. **data URIs are decoded twice** on the vLLM path: once for validation, once by
   `_handle_base64_media`.
7. **SGLang's prompt text sees local temp paths** rather than the original URL, a
   side effect of materialization. Qwen templates render vision placeholders and
   are unaffected; a template that rendered the URL into the prompt would change.

## Testing

- `xinference/model/llm/tests/test_media.py`: 50 passed. Covers redirect-into-
  private-network blocked end to end, the header phase bounded against a
  byte-at-a-time server, `Content-Encoding: gzip` decoded rather than handed back
  compressed, per-URL and per-request budgets in both directions, every media
  reference key, and a regression guard asserting no threads exist at import
  time (a module-level thread pool on this path deadlocks vLLM's `EngineCore`).
- `test_sdapi_compat.py` passes; `test_core_chat_model.py` 126 passed / 6 skipped
  / 7 failed, the failures being pre-existing `vllm`-not-installed errors at
  `vllm/core.py:1436` that reproduce on `main`.
- `pre-commit` clean; `msgfmt --check --check-format` clean on the modified
  catalog.
- Deployed to a 5x L20 production host and exercised for two hours with four
  models running (embedding, rerank, audio, and a vLLM multimodal chat model).
  No regression on the inference path.

Supersedes #5532, which validated URLs without taking over the fetch and so left
the redirect bypass open.


xinference/model/llm/media.py
```diff
@@ -0,0 +1,356 @@
...
+    """
+    import tempfile
+    from pathlib import Path
+
+    # Per-part budget, capped for the request as a whole: a single shared budget
+    # starves a legitimate many-image request, while per-part budgets alone
+    # multiply by part count and each one holds a shared executor thread.
+    request_deadline = time.monotonic() + _fetch_timeout() * REQUEST_TIMEOUT_FACTOR
+    for container, key, kind in list(_iter_media_slots(messages)):
+        raw = container[key]
```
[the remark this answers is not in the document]
bluefish-08: Good catch — fixed, though in a different place. Rather than guarding the two call sites separately, the walk is now funnelled through a `_media_slots()` helper that owns the `RecursionError` translation, so `validate_messages_media` and `materialize_messages_media` can't drift apart again.

Added `test_deep_nesting_is_a_client_error` covering both paths with 3000 levels of nesting. Verified it fails when the guard is short-circuited.

xinference/model/image/utils.py
```diff
@@ -301,28 +306,82 @@ def verify_url(url):
...
+            # blocked in recv().
+            sock = getattr(conn, "sock", None)
+            for action in (
+                lambda: sock.shutdown(__import__("socket").SHUT_RDWR),
+                conn.close,
+            ):
+                try:
+                    action()
+                except Exception:
+                    pass
```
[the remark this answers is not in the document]
bluefish-08: Agreed — the old form leaned on the exception handler to absorb a `None`. Now checks `sock is not None` explicitly and drops the `__import__("socket")` inline call for a plain local import.

qinxuye:
Some tests failed, pls take a look.

bluefish-08:
Fixed. The failures were all in the new test file: `json.loads` blew the stack on 3.10/3.11 before reaching the code under test, `base64` only rejects excess padding on 3.13+, and on Windows `local_path()` turned `file:///C:/...` into `/C:/...` so materialized media failed its own validation. Also dropped a bogus timeout tolerance and closed a silent-truncation path where the watchdog's socket shutdown reads as a clean EOF. (The ubuntu-3.13 job was an unrelated port-bind flake.)

xinference/model/llm/media.py
```diff
@@ -0,0 +1,376 @@
...
+def validate_messages_media(messages: Optional[List[Any]]) -> None:
+    """Validate every media url in OpenAI-formatted (or transformed) messages."""
+    for container, key, _ in _media_slots(messages):
+        validate_media_url(container[key])
+
+
+def _read_bounded(reader: Any, deadline: float, limit: int) -> bytes:
+    # read1 returns as soon as any data arrives, so a dribbling server cannot
+    # stay under the per-socket timeout forever without hitting the deadline.
+    read = getattr(reader, "read1", None) or reader.read
```
qinxuye, line 286: [P2] Keep the total deadline effective when read1 is unavailable

With urllib3 2.0.7 (not excluded by the current dependencies), HTTPResponse has no read1, so this falls back to read(64 * 1024). A slow-drip response keeps that call blocked while resetting socket-level timeouts, preventing the deadline checks below from running. The watchdog does not rescue an HTTP/1.0/connection-closing response: the connection's sock is already None while the response still owns its readable socket, so pool.abort() cannot interrupt it.

Reproduced by running `python -m pytest -q xinference/model/llm/tests/test_media.py::test_fetch_media_enforces_total_deadline` with urllib3 2.0.7: the configured 1-second deadline took 40.6 seconds, failing the existing <10-second assertion. A shorter instrumented HTTP/1.0 probe with a 0.2-second budget took 2.57 seconds and confirmed `_active_conn.sock is None` when the watchdog fired. A continuing response can therefore occupy a shared executor thread beyond both the per-URL and request budgets.

Please make the fallback incremental/deadline-aware and ensure cancellation targets the socket actually backing the response, or explicitly enforce a dependency version that guarantees the required read behavior. Add coverage for a no-read1 reader plus a slow HTTP/1.0 or Connection: close response, checking that the fetch actually stops near the deadline rather than raising only after EOF.

qinxuye (commented):
One remaining timeout issue reproduced at this head; the base64 integration regression is fixed.

xinference/model/llm/media.py
bluefish-08: Confirmed and fixed — thanks, this was worse than the urllib3 2.0.7 case.

Reproduced with a dribbling server under a 1s budget: HTTP/1.1 chunked stopped at 1.00s, HTTP/1.0 `Connection: close` ran 25.01s. Instrumenting `socket.shutdown` showed it was never called at all, so `pool.abort()` had never reached the body's socket on any response shape. HTTP/1.1 only looked correct because `read1` returns incrementally and `_read_bounded` re-checks the deadline each round; the watchdog was dead weight. Without `read1` there is nothing left to bound the read.

Rather than dig the socket back out of the response, `_Tracked` now wraps `conn.connect` and keeps its own reference to `conn.sock` — that survives the response taking the connection over, works the same on 1.26 / 2.0.x / 2.x (neither `HTTPResponse.read1` nor `.shutdown()` exists before 2.1), and the SSLSocket case shuts down at the TCP layer. Net effect is fewer lines than before, no private-attribute chains, and no dependency pin needed.

Regression test `test_fetch_media_enforces_deadline_without_read1`: `read1` patched to None against the existing `/slow` handler, which is HTTP/1.0 by default. 41.5s without the fix (matching your 40.6s), 1.0s with it. 102 passed / 6 skipped across test_media.py and test_sdapi_compat.py.

Separately, and outside this PR: `decode_base64_to_image` and `download_civitai_model` call `_open_public_url` without a deadline, so no watchdog is armed there at all — the sdapi one takes a client-supplied URL. Worth its own issue.

qinxuye: Verified fixed at dd94ba6c in the same urllib3 2.0.7 environment that reproduced the failure. Both the original total-deadline test and the new no-read1 regression now pass. The watchdog retains the socket independently of conn.sock. All 71 selected media and vLLM integration tests passed; resolving.

qinxuye (approved):
LGTM. Verified the watchdog fix in urllib3 2.0.7, including the previously failing total-deadline test and the new no-read1 regression. All 71 selected media and vLLM integration tests passed locally. The reported timeout issue is resolved; the documented out-of-scope reader limitations remain.

PR #5538, opened 2026-09-15
feat(llm): add Spark-X2.5 model support

Minamiyama:
## Summary
- Add Spark-X2.5 and Spark-X2.5-Base builtin model specs.
- Add Hugging Face `main` and ModelScope `master` sources.
- Add Transformers, vLLM, SGLang, and llama.cpp compatibility.
- Pin Spark-specific vLLM and SGLang runtime dependencies.
- Add generated builtin-model documentation and family regression coverage.


xinference/model/llm/transformers/spark_x2_5.py
```diff
@@ -0,0 +1,116 @@
...
+            self._get_chat_template_kwargs_from_generate_config(
+                generate_config, self.reasoning_parser
+            )
+            or {}
+        )
+        chat_context_var.set(chat_template_kwargs)
+        full_context_kwargs = chat_template_kwargs.copy()
+        if tools:
+            full_context_kwargs["tools"] = tools
+        return self.get_full_context(
```
qinxuye, line 100: [P2] Normalize tool-call arguments before rendering Spark history

A standard OpenAI tool round trip supplies `assistant.tool_calls[].function.arguments` as a JSON string (e.g. `'{"city":"Beijing"}'`). The official Spark template explicitly requires a mapping and raises `TemplateError: tool_call.function.arguments must be a dictionary; normalize JSON strings before apply_chat_template`. I reproduced this with the official template in this Transformers adapter and the prompt-building paths used by vLLM and SGLang; changing the arguments to a dict makes the same history render successfully. `get_full_context` does not normalize here: Transformers/vLLM pass `chat_template=None`, and SGLang's explicit Spark template does not match `_chat_template_needs_dict_arguments` either. Please normalize valid JSON-object arguments without mutating the caller's history across all three adapters, and add regression coverage for user -> assistant tool call (JSON-string arguments) -> tool result -> next assistant prompt using the real template structure. The current test only exercises a first-turn tools declaration, so it misses this failure.

qinxuye (commented):
One functional issue remains in tool-call continuation across the Spark chat adapters.

xinference/model/llm/tests/test_utils.py
```diff
@@ -2832,6 +2832,78 @@ def test_qwen3_family_get_full_context_handles_string_arguments():
...
+    )
+
+    # SGLang passes the downloaded chat_template.jinja text directly.
+    sglang_prompt = mixin.get_full_context(messages, spark_template)
+
+    class Tokenizer:
+        chat_template = spark_template
+
+        def apply_chat_template(self, rendered_messages, **kwargs):
+            return mixin._build_from_raw_template(
```
qinxuye, line 2893: [P2] Avoid passing add_generation_prompt twice in the test tokenizer

`get_full_context()` passes `add_generation_prompt=True` to this fake tokenizer, which forwards it through `**kwargs` to `_build_from_raw_template()`. That helper already calls `compiled_template.render(messages=messages, add_generation_prompt=True, **kwargs)`, so the new regression test raises `TypeError: ... got multiple values for keyword argument 'add_generation_prompt'`. This is a test-double failure, not a remaining Spark runtime failure: the official template with a real Hugging Face tokenizer works on the updated production paths. Please remove the duplicate option in the fake tokenizer or use a real lightweight tokenizer, then rerun this regression and the related tests. Local run before the formatting-only follow-up: 111 passed, 1 failed, 2 skipped. The latest commit only adjusts formatting and leaves this failing call unchanged.

qinxuye (commented):
The tool-call continuation fix works with the official template across the three prompt paths. One failure remains in the new regression test.

xinference/model/llm/transformers/spark_x2_5.py
qinxuye: Verified the production fix with the official template: Transformers, vLLM shared prompting, and SGLang shared prompting now render JSON-string tool arguments successfully without modifying the input history. Resolving this issue; the new regression test has a separate test-double failure covered in a new inline comment.

xinference/model/llm/tests/test_utils.py
qinxuye: Verified fixed in 2722f59f: the test tokenizer removes add_generation_prompt before calling the raw-template helper. The focused suite now passes: 112 passed, 2 skipped.

xinference/model/llm/vllm/patches/spark_x2_5_plugin.py
```diff
@@ -0,0 +1,60 @@
...
+    with open(target_file, "r") as f:
+        content = f.read()
+    if "[xinference-patch] vLLM 0.29" in content:
+        return False
+    if _ORIGINAL not in content:
+        logger.debug("Spark-X2.5 plugin has no compatible patch target: %s", target_file)
+        return False
+
+    with open(target_file, "w") as f:
+        f.write(content.replace(_ORIGINAL, _PATCHED))
```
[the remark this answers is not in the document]
qinxuye: Verified: both text-file opens explicitly specify encoding=utf-8 in the current head.

xinference/model/llm/sglang/core.py
```diff
@@ -1003,6 +1005,18 @@ async def async_chat(
...
+        ):
+            chat_template_path = os.path.join(self.model_path, "chat_template.jinja")
+            try:
+                with open(chat_template_path, encoding="utf-8") as template_file:
+                    chat_template = template_file.read()
+            except OSError as exc:
+                raise ValueError(
+                    "Spark-X2.5 SGLang requires chat_template.jinja from the "
+                    "official model snapshot"
+                ) from exc
```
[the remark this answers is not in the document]
qinxuye: Reviewed as an optional optimization, not a correctness issue. Template compilation is already cached by the shared helper; no material performance regression has been demonstrated for reading this small local template. Closing as non-blocking, without claiming the proposed instance cache was implemented.

qinxuye (approved):
LGTM. Tool-call continuation and its regression test are fixed. Focused validation: 112 passed, 2 skipped; current CI is green. Real model inference was not run locally.

PR #5545, opened 2026-09-16
feat(api): add opt-in model request logging

m199369309:
## Summary

- add opt-in JSONL logging for model inference requests, including request lifecycle metadata and bounded request-body capture
- add stable `X-Request-ID` correlation across REST responses and Supervisor model lookup logs without replacing model operation IDs used for cancellation/progress
- document the feature and add gettext catalog updates for all maintained locales

## Security and compatibility

- disabled by default with an explicit warning that request values may contain sensitive data
- does not persist request bodies for HTTP 401, 403, or 413 responses
- limits body capture to 16 MiB by default and avoids pre-reading requests whose size is unknown or above the limit
- records multipart field values and filenames, but never uploaded file bytes
- writes through a dedicated rotating logger using restrictive `0600` file permissions
- performs serialization/file I/O off the event loop; logging failures do not change model responses
- preserves `request-id` over `x-request-id`, validates caller IDs, and retains the existing `xinf-<uuid>` generated correlation-ID format
- preserves Anthropic's existing generated `req_...` protocol response IDs

## Configuration

- `XINFERENCE_MODEL_REQUEST_LOG_ENABLED` (default: `false`)
- `XINFERENCE_MODEL_REQUEST_LOG_FILE` (default: `model_request.log`)
- `XINFERENCE_MODEL_REQUEST_LOG_BODY_MAX_BYTES` (default: `16777216`)
- `XINFERENCE_MODEL_REQUEST_LOG_RETENTION_DAYS` (default: `7`)
- `XINFERENCE_MODEL_REQUEST_LOG_MAX_BYTES` (default: `1073741824`)
- `XINFERENCE_MODEL_REQUEST_LOG_BACKUP_COUNT` (default: `7`)

## Validation

- `152 passed` across focused API/core tests covering request logging, request correlation, Anthropic messages, audio, SD API, token-router dispatch, and world API paths
- pre-commit passed for all modified files: Black, EOF/trailing whitespace, Ruff, isort, mypy, and codespell
- `msgfmt --check --check-format` passed for every modified locale catalog
- `python doc/build_i18n.py --all` completed for all locales

A strict whole-document Sphinx build was also attempted, but remains blocked by pre-existing unrelated documentation errors in `doc/source/models/custom.rst` and `doc/source/models/virtualenv.rst`.


xinference/api/model_request_logging.py
```diff
@@ -0,0 +1,601 @@
...
+                if isinstance(exc, asyncio.CancelledError):
+                    status_code = 499
+                elif not isinstance(status_code, int):
+                    status_code = 500
+
+                # Authentication/authorization failures must never persist the body.
+                if XINFERENCE_MODEL_REQUEST_LOG_ENABLED and status_code not in (
+                    401,
+                    403,
+                ):
```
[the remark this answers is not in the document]
m199369309: Fixed in f6088e4f. The exception and response paths now share `_BODY_LOG_EXCLUDED_STATUS_CODES`, including 413, and a regression test verifies that a raised 413 does not persist the request body.

xinference/api/model_request_logging.py
```diff
@@ -0,0 +1,601 @@
...
+        return body.decode("utf-8", errors="replace")
+
+
+def _body_size(request: Request) -> Optional[int]:
+    content_length = request.headers.get("content-length")
+    if content_length is not None:
+        try:
+            return int(content_length)
+        except ValueError:
+            return None
```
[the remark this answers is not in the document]
m199369309: Fixed in f6088e4f. Negative `Content-Length` values are now treated as an unknown size, with a regression test covering the behavior.

xinference/api/model_request_logging.py
```diff
@@ -0,0 +1,601 @@
...
+
+    with _LOGGER_LOCK:
+        if _LOGGER_CONFIGURED:
+            return request_logger
+        handler: Optional[logging.Handler] = None
+        try:
+            filename = XINFERENCE_MODEL_REQUEST_LOG_FILE
+            if not os.path.isabs(filename):
+                filename = os.path.join(XINFERENCE_LOG_DIR, filename)
+            handler = _ModelRequestRotatingFileHandler(
```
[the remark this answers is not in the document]
m199369309: This is already handled by the inherited `SafeTimedRotatingFileHandler.__init__`, which calls `os.makedirs(os.path.dirname(filename), exist_ok=True)` before opening the file (`xinference/deploy/utils.py`). I added a regression test in f6088e4f that constructs `_ModelRequestRotatingFileHandler` with a missing nested parent directory and verifies successful creation, so no duplicate production `os.makedirs` call is needed here.

xinference/api/model_request_logging.py
```diff
@@ -0,0 +1,604 @@
...
+            ):
+                model_uid, stream = await self._log_started(request, request_id)
+
+            actual_model_uid = (
+                getattr(request.state, "_audit_model_uid", "") or model_uid
+            )
+            actual_model_type = str(
+                getattr(request.state, "_audit_model_type", "") or model_type
+            ).lower()
+            if isinstance(response, StreamingResponse):
```
qinxuye: [P2] Track EventSourceResponse lifecycles rather than logging them as completed immediately

The production completion/chat/Anthropic and several other streaming endpoints return `sse_starlette.sse.EventSourceResponse`, which inherits `Response`, not `StreamingResponse`. They therefore take the non-streaming branch here. Reproduced with this route class and a real EventSourceResponse: before the generator's first yield, the log already contains `model_request_finished` with `stream=false`, `success=true`, and sub-millisecond elapsed time; a subsequent generator exception produces no failure event. The correlation context is also absent inside that generator. The current streaming tests use only StreamingResponse and miss the actual response type.

Please include production SSE responses in lifecycle tracking, preserve their wire format, and defer the terminal log until streaming actually completes/fails/disconnects. Also account for existing endpoint generators that catch model exceptions and emit SSE error payloads (or catch cancellation and return), otherwise merely broadening isinstance still labels these cases successful. Add real EventSourceResponse regressions covering normal completion, a raised exception, an emitted SSE error, and client disconnect, checking the terminal event, stream flag, elapsed time, and correlation context.

qinxuye (commented):
One functional gap remains in production SSE request lifecycle logging.

xinference/api/model_request_logging.py
m199369309: Fixed in 76bb6957 (with the stream outcome support added in cdaeb5449).

The route now treats real `EventSourceResponse` instances as streaming responses, chains any existing `client_close_handler_callable` to classify actual client disconnects, and preserves SSE wire encoding by wrapping the original item iterator before `EventSourceResponse` performs encoding. Stream generators explicitly report swallowed model/upstream/protocol failures through a request-scoped first-terminal-state-wins reporter. Correlation context is installed around each source `__anext__()`/`aclose()` call instead of keeping a ContextVar token alive across yields.

Added real `EventSourceResponse` regressions for normal completion (including correlation ID, deferred terminal logging, elapsed time, and exact wire bytes), raised exceptions, emitted SSE error events, and ASGI client disconnect while preserving an existing close handler.

Validation:
- 76 focused API tests passed
- pre-commit passed for all PR files
- all modified gettext catalogs passed `msgfmt --check --check-format`
- `python doc/build_i18n.py --all` passed

xinference/api/model_request_logging.py
```diff
@@ -0,0 +1,669 @@
...
+                        type(outward_error).__name__
+                        if outward_error
+                        else outcome.state.value
+                    ),
+                    "message": outcome.error_message
+                    or (str(outward_error) if outward_error else outcome.state.value),
+                }
+            context_token = _set_current_model_request_id(request_id)
+            try:
+                await cls._log_response(
```
qinxuye: [P2] Protect terminal logging from the stream cancellation scope

On client disconnect, EventSourceResponse cancels its AnyIO task group, but this final log write still runs inside that cancelled scope. `_write_event_async()` awaits `asyncio.to_thread()`, so the terminal write can be cancelled before the executor runs it. The current Linux and Windows CI failures in `test_event_source_client_disconnect_is_logged_and_chains_handler` show zero terminal events. I also reproduced actual loss (not just a premature assertion) with sse-starlette 3.4.11 by occupying a single-worker default executor after the started event: after disconnect, releasing the worker and waiting still leaves only `model_request_started`. Please protect finalization/logging from cancellation and ensure cleanup cannot bypass the terminal event. Add a deterministic regression with a busy executor that verifies exactly one client_disconnected terminal event after cleanup, without relying on thread scheduling timing.

xinference/api/model_request_logging.py
```diff
@@ -0,0 +1,669 @@
...
+                getattr(request.state, "_audit_model_uid", "") or model_uid
+            )
+            actual_model_type = str(
+                getattr(request.state, "_audit_model_type", "") or model_type
+            ).lower()
+            if isinstance(response, (StreamingResponse, EventSourceResponse)):
+                if XINFERENCE_MODEL_REQUEST_LOG_ENABLED:
+                    reporter = get_stream_outcome_reporter(request)
+                    if isinstance(response, EventSourceResponse):
+                        previous_close_handler = response.client_close_handler_callable
```
qinxuye: [P2] Handle supported sse-starlette versions without the close-handler hook

`pyproject.toml` still allows `sse_starlette>=1.6.5`, but `client_close_handler_callable` is absent in, for example, 2.0.0. With that allowed version and request logging enabled, a normal EventSourceResponse now raises AttributeError here before the stream starts; the new normal-completion SSE regression reproduces it. Please either implement lifecycle/disconnect handling compatible with the supported versions or declare a minimum version that provides the required hook. Merely assigning a missing attribute would not make older implementations invoke it. Cover normal SSE completion and disconnect at the declared minimum supported version.

qinxuye (commented):
The SSE response detection and iterator correlation-context fixes are confirmed. Two remaining issues are reproduced below: cancellation can drop the terminal log, and the new close-handler access breaks supported older sse-starlette installations.

xinference/api/model_request_logging.py
m199369309: Fixed in cec6e609. Stream cleanup and terminal response logging now run inside `anyio.CancelScope(shield=True)`, cleanup exceptions cannot bypass the terminal event, and an existing outward cancellation/error remains the propagated exception. Added a deterministic single-worker busy-executor regression that verifies the ASGI request waits for exactly one `client_disconnected` terminal event, plus cleanup-error coverage.

xinference/api/model_request_logging.py
m199369309: Fixed in cec6e609 without raising the dependency floor. Close-handler support is detected from the EventSourceResponse constructor; newer releases chain `client_close_handler_callable`, while supported older releases wrap their actual `listen_for_disconnect` (1.6.5-2.1) or `_listen_for_disconnect` (2.2) receive path, so merely adding an unused attribute is avoided. Normal completion and disconnect were tested against sse-starlette 1.6.5, 2.2.1, and the current 3.4.6 environment.

qinxuye (approved):
LGTM. Verified the cancellation-safe terminal logging and supported older SSE lifecycle handling fixes at cec6e609. Both follow-up threads are resolved. Focused regressions passed; the full CI matrix is still running.

PR #5549, opened 2026-09-17
fix(vllm): abort requests on client disconnect

### LINKED ISSUE #5548: BUG: stop vLLM inference when requests are aborted or disconnected
## Problem

Streaming vLLM requests can continue consuming GPU resources after the client disconnects. The current async-generator `aclose()` propagation is not deterministic across the REST, Xoscar, and vLLM boundaries. In addition, OpenAI-compatible `request_id` values are left inside generation config instead of being propagated to the model actor, so the existing abort endpoint returns `NO_OP` for vLLM.

## Proposed change

- Propagate one request ID from the REST API through the model actor into vLLM.
- Track active vLLM request IDs and call the underlying engine abort API.
- Explicitly abort SSE requests when the client disconnects.
- Keep normal completion abort-free and preserve `DONE`, `NO_OP`, and `NOT_FOUND` outcomes.
- Cover explicit abort, disconnect cleanup, normal completion, unsupported engines, duplicate aborts, and concurrent requests.

## Validation expectations

- Focused unit tests for REST request-ID propagation and vLLM lifecycle cleanup.
- Real GPU validation showing inference stops after disconnect and explicit abort.

amumu96:
## Summary

- propagate a stable request ID from OpenAI completion and chat endpoints into the model actor and vLLM engine
- track active vLLM requests and invoke the engine abort API on explicit abort or early stream closure
- abort SSE work deterministically when clients disconnect while leaving normally completed requests untouched
- preserve `DONE`, `NO_OP`, and `NOT_FOUND` abort outcomes and prevent duplicate engine abort calls

## Release note

- vLLM requests now stop consuming GPU resources when explicitly aborted or when a streaming client disconnects.

Closes #5548

## Validation

- `python -m pytest -q xinference/api/tests/test_llm_request_abort.py xinference/model/llm/vllm/tests/test_abort.py xinference/model/llm/vllm/tests/test_utils.py` (14 passed)
- `python -m ruff check xinference/api/tests/test_llm_request_abort.py xinference/model/llm/vllm/tests/test_abort.py xinference/api/restful_api.py xinference/model/llm/vllm/core.py`
- `python -m compileall -q` on the four changed files
- RTX 3090 Ti / vLLM 0.28.1 real-GPU validation: disconnect and explicit abort both stopped inference; GPU utilization dropped from 100% to 0%; explicit and concurrent aborts returned `DONE`; normal completion did not abort
- Broader local selection: 51 passed and 7 existing unrelated failures (vLLM is not installed locally, leaving `VLLM_VERSION=None`, plus existing tool-call expectation mismatches)


amumu96:
## Validation results

### Focused regression tests

```text
$ python -m pytest -q \
  xinference/api/tests/test_llm_request_abort.py \
  xinference/model/llm/vllm/tests/test_abort.py \
  xinference/model/llm/vllm/tests/test_utils.py
..............                                                           [100%]
14 passed, 1 warning in 0.06s
```

Covered scenarios:

- REST completion/chat request ID propagation
- SSE disconnect invokes explicit model abort
- early stream close aborts the underlying vLLM request
- normal completion does not abort
- unsupported engine returns `NO_OP`
- missing/inactive request returns `NOT_FOUND`
- duplicate abort only invokes the engine once
- concurrent requests are tracked and cleaned independently

Additional checks:

```text
ruff:       passed
compileall: passed
git diff --check: passed
```

### Real GPU validation

Environment: RTX 3090 Ti, vLLM `0.28.1rc1.dev388+g8a728663c`, Qwen3-0.6B.

| Scenario | Result |
| --- | --- |
| Streaming client disconnect | Server logged `Disconnected from client` followed by `Aborted vLLM request`; GPU utilization dropped from 100% to 0% in about one second after disconnect handling |
| Explicit abort endpoint | Returned `{"msg":"DONE"}` and the engine logged the matching request ID as aborted |
| Concurrent requests | Two active requests were aborted independently; both calls returned `DONE` |
| Normal completion | Completed normally without an abort log |
| Request ID propagation | The caller-provided ID appeared unchanged in REST, ModelActor, and vLLM engine logs |

### Broader local selection

A broader test selection completed with **51 passed and 7 existing unrelated failures**. The failures reproduce outside this change and are caused by tests comparing `VLLM_VERSION=None` when vLLM is not installed locally, plus existing tool-call output expectation mismatches.

xinference/model/llm/vllm/core.py
```diff
@@ -2148,6 +2153,55 @@ async def stream_results() -> AsyncGenerator[CompletionChunk, None]:
...
+            finally:
+                try:
+                    if not completed:
+                        await self.abort_request(request_id)
+                except Exception:
+                    logger.exception(
+                        "Failed to abort interrupted vLLM request: %s", request_id
+                    )
+                finally:
+                    self._active_request_ids.discard(request_id)
```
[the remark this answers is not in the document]
amumu96: Fixed in aa892b71. The wrapper now resolves `aiter(results_generator)` once before iteration, keeps that concrete iterator, and on early termination aborts the engine request before calling `aclose()` on the same iterator. This avoids creating and closing a different iterator for AsyncIterable-only implementations. Added a regression test that verifies `aiter()` is called once and the concrete iterator is closed; focused suite now passes 15 tests.

amumu96:
CI failure fixed in `8062774e`.

Root cause: REST request-ID propagation was applied to every physical chat backend. vLLM needs `request_id` as an actor keyword, but Transformers/PyTorch synchronous `chat()` does not accept that keyword, causing `PytorchChatModel.chat() got an unexpected keyword argument request_id` across the CPU and GPU matrices.

Fix:

- pass the actor-level `request_id` keyword only for vLLM backends
- preserve the pre-existing non-vLLM behavior by keeping caller-provided IDs in generation config
- add a regression test for non-vLLM chat dispatch

Validation:

```text
56 passed, 1 warning
ruff: passed
compileall: passed
git diff --check: passed
```

The previously failing `test_physical_chat_keeps_existing_path` now passes locally.

xinference/api/restful_api.py
```diff
@@ -1757,6 +1766,13 @@ async def stream_results():
                     logger.info(
                         f"Disconnected from client (via refresh/close) {request.client} during generate."
                     )
+                    try:
+                        await asyncio.shield(model.abort_request(request_id))
```
qinxuye: [P2] Abort disconnects that occur while SSE is sending a yielded chunk

This abort only runs when CancelledError enters stream_results itself. If the generator is suspended at `yield item` while EventSourceResponse is awaiting ASGI send (e.g. a slow client/backpressure), a disconnect cancels the send task instead, so this handler never executes. Reproduced with a real EventSourceResponse: block send after the first body chunk, return http.disconnect from receive, and await the response; abort_request is called zero times. Explicitly calling response.body_iterator.aclose() afterwards still produces zero abort calls because that injects GeneratorExit, not CancelledError. The finally block only decrements serve count and does not abort or close the remote model iterator, so this path can leave vLLM work running. Please handle disconnect independently of generator advancement, with cancellation-safe cleanup and early-close handling for both completion and chat, while preserving the no-abort behavior for normal completion. Add a real ASGI regression that disconnects during send, as well as during the next-item wait; the current test only raises CancelledError inside the iterator.

xinference/api/tests/test_llm_request_abort.py
```diff
@@ -0,0 +1,157 @@
...
+import pytest
+
+from .. import restful_api
+
+
+class _Request:
+    client = "test-client"
+
+    def __init__(self, body):
+        self._body = body
```
qinxuye, line 28: [P2] Give the request fixture the state required by the current API

This mock has no `state`, but the streaming-outcome code merged through #5545 uses request.state in get_stream_outcome_reporter(). The current merge CI already fails test_completion_disconnect_aborts_propagated_request_id with `AttributeError: '_Request' object has no attribute 'state'` (Python 3.11 job: https://github.com/xorbitsai/inference/actions/runs/35300883265/job/105463638067). The focused tests pass on this branch's older base but not when integrated with main. Please use a real Starlette Request or supply its state on the fixture, then rerun the new tests against the current main integration.

qinxuye (commented):
Two remaining issues: disconnects during ASGI send bypass the abort path, and the new request mock fails against the current main integration used by CI.

amumu96:
Fixed the latest cross-platform CI failure in `67620ecf`.

The abort test uses a lightweight fake Starlette request. After merging the latest `main`, the streaming outcome reporter introduced by #5545 accesses `request.state`, so the fixture failed before reaching the abort assertions. The fake request now initializes `state` with `SimpleNamespace()`.

Focused validation after merging current `origin/main`:
- `63 passed` across abort, token-router, streaming-outcome, vLLM abort/utils, and core model tests
- Ruff passed
- compileall passed
- `git diff --check` passed

Note: a broader local run also hit two pre-existing `test_model_request_logging.py` failures caused by the local Starlette/sse-starlette ExceptionGroup behavior; those are outside this PR and were not present in the shared CI failure signature.

xinference/api/restful_api.py
```diff
@@ -1777,6 +1786,13 @@ async def stream_results():
...
```
qinxuye: [P2] Send-phase disconnect still bypasses abort on the latest head

Reproduced again on 67620ecf for BOTH completion and chat: let the model yield one chunk, block the ASGI send callable on that body message, and return http.disconnect from receive. After EventSourceResponse exits, model.abort_request.await_count is 0; explicitly awaiting response.body_iterator.aclose() afterwards still leaves it at 0. The generator is suspended at yield while cancellation hits the response's send operation, so the CancelledError handler here never runs; closing the generator instead raises GeneratorExit. The newly integrated observe_stream/outcome reporting does not itself abort the remote model request.

Please handle actual disconnect independently of where the generator is suspended, and ensure early response closure performs cancellation-safe, idempotent abort plus remote-iterator cleanup. Apply this to both completion and chat, retaining zero aborts for normal completion. Add real ASGI regressions for disconnect during send and during the next-item wait, plus explicit early close, asserting the correct request ID, exactly one engine abort, and serve-count cleanup. The existing synthetic CancelledError test does not cover this path.

qinxuye (commented):
Re-reviewed 67620ecf. The request.state fixture fix is confirmed and that thread is resolved; all 16 focused tests pass. The send-phase disconnect issue remains reproducible for both completion and chat, as detailed below.

xinference/api/restful_api.py
amumu96: Fixed in `0bc7b393`.

The SSE response now installs a disconnect handler that runs independently of generator advancement. It uses a shared idempotent controller so the disconnect callback and generator `finally` converge on the same engine abort task. On disconnect it aborts first, then closes the response body iterator when possible; the generator `finally` explicitly closes the observed model iterator and decrements serve count. If disconnect happens while `__anext__` is running, the callback aborts and the EventSourceResponse task-group cancellation drives the same cleanup path.

Added real EventSourceResponse ASGI regressions for both completion and chat covering:
- disconnect while ASGI `send` is blocked on a yielded chunk
- disconnect while waiting for the next model item
- explicit early `body_iterator.aclose()`
- normal completion with zero aborts

Each interrupted case asserts the propagated request ID, exactly one `abort_request` call, source iterator closure, and exactly one serve-count decrement. The existing model-request-logging close handler remains chained.

Validation: `73 passed`, Ruff passed, compileall passed, and `git diff --check` passed.

xinference/api/restful_api.py
amumu96: Addressed in `0bc7b393`; implementation and real ASGI regression details are in the latest follow-up thread: https://github.com/xorbitsai/inference/pull/5549#discussion_r4052212239

xinference/api/restful_api.py
```diff
@@ -1787,6 +1929,17 @@ async def stream_results():
                     yield dict(data=json.dumps({"error": str(ex)}))
                     return
                 finally:
+                    if not completed:
+                        await abort_controller.abort()
+                    if observed_iterator is not None:
+                        try:
+                            await _close_async_iterator(observed_iterator)
```
qinxuye: [P2] Protect the entire disconnect cleanup from cancellation

The send-phase abort is fixed, but cleanup is still cancellable when disconnect occurs during the next-item wait. EventSourceResponse cancels its AnyIO task group after the disconnect callback returns; the generator's finally then runs inside that cancelled scope. Shielding only the abort task does not protect iterator closure or decrease_serve_count().

Reproduced on this head for BOTH completion and chat using the new real-ASGI test helper: make decrease_serve_count's mock side effect await asyncio.sleep(0) before recording completion. Disconnect during the next-item wait calls abort exactly once, but the decrement never completes. Adding the same await to the source generator's finally also leaves source cleanup incomplete. Send-phase cases pass. The current immediately-returning AsyncMock and synchronous closed.set() hide these cancellation checkpoints; the real actor RPC can suspend, leaving a serving slot unreleased and eventually causing false request-limit errors.

Please make the full cleanup cancellation-safe and ensure serve-count release still runs if iterator closure fails, for both endpoints. Add regressions with genuinely suspending iterator cleanup and decrement operations, asserting completion (not just await_count), exactly one abort/decrement, and no abort on normal completion.

amumu96: Fixed in `48d6bd50`.

The disconnect handler now protects the full cleanup path with an AnyIO shield and, crucially, cancels and joins the SSE stream task **before** returning to EventSourceResponse. This lets the source iterator unwind while the enclosing task-group scope is still active, instead of first entering its finalizer after that scope has already been cancelled. The generator `finally` then runs a shielded shared cleanup routine that:

1. awaits the idempotent abort task for interrupted streams;
2. explicitly closes the observed/model iterator;
3. releases serve count in `finally`, even if iterator close raises or is cancelled.

The real-ASGI completion/chat tests now make both source finalization and `decrease_serve_count()` cross genuine cancellation checkpoints (`await asyncio.sleep(0)`) and assert completion events, not only await counts. Added completion/chat regressions where iterator close raises after suspending and verified serve-count release still completes. Normal completion continues to assert zero aborts.

Validation: `75 passed`, Ruff passed, compileall passed, and `git diff --check` passed.

qinxuye (approved):
LGTM. Verified disconnect cleanup and serve-count release across asynchronous cancellation checkpoints for completion and chat, including older SSE compatibility. All review threads are resolved.

PR #5550, opened 2026-09-17
feat(model):support MinerU2.5 model

llyycchhee:
## Summary

Add MinerU2.5 support for document parsing with transformers and vLLM backends.

- Register MinerU2.5 as an image model with the `docanalyze` ability, supporting only transformers and vLLM.
- Add the `POST /v1/images/docanalyze` endpoint for PDF and image uploads.
- Add a document parsing UI with structured JSON results, file-type validation, and empty-file checks.
- Add localized capability labels, descriptions, and UI messages in English, Simplified Chinese, Traditional Chinese, Japanese, and Korean.
- Generate the MinerU2.5 model documentation, index entry, and gettext catalogs.

## Validation

- Document upload and localization tests: 5 passed.
- Frontend lint, TypeScript checks, and production build passed.
- MinerU2.5 documentation builds passed for English and all nine maintained documentation locales.
- Diff whitespace checks passed.

The existing capability task panel regression suite could not run because `jsdom` was unavailable in the validation environment. New gettext catalogs are prepared for translation.


xinference/api/restful_api.py
```diff
@@ -2800,6 +2800,60 @@ async def create_ocr(
...
+    async def create_doc_analyze(
+        self,
+        model: str = Form(...),
+        file: UploadFile = File(media_type="application/octet-stream"),
+        kwargs: Optional[str] = Form(None),
+    ) -> Response:
+        if not file.filename or file.size == 0:
+            raise HTTPException(status_code=400, detail="File can't be empty")
+        model_uid = model
+        model_ref = await require_model(
```
qinxuye, line 2816: [P1] Enforce model-level authorization before resolving the model

The route's `Security(..., scopes=["models:read"])` checks the general scope, but does not replace `_check_model_access`. Unlike create_ocr and the other image inference handlers, this handler never performs that model-specific check. Reproduced using a real AdvancedAuthService and an API key restricted to `allowed-model`: validate_model_access(key, 'mineru', 'image') returns False, yet POST /v1/images/docanalyze with model=mineru returns HTTP 200 and invokes the model. Please accept Request and call `_check_model_access(request, model_uid, "image")` before model lookup/inference, following the existing image handlers. Add an authenticated route regression proving a restricted key gets 403 with no model call, while an authorized key succeeds.

xinference/model/image/docanalyze/mineru25.py
```diff
@@ -0,0 +1,552 @@
...
+
+    @property
+    def model_ability(self):
+        return self._abilities
+
+    def load(self):
+        from ....thirdparty.mineru.backend.vlm.custom_logits_processors import (
+            enable_custom_logits_processors,
+        )
+        from ....thirdparty.mineru_vl_utils import MinerUClient, MinerULogitsProcessor
```
qinxuye: [P1] Import the vLLM logits processor only for the vLLM backend

This unconditional import resolves MinerULogitsProcessor through the vendored lazy loader, whose module imports `vllm.config.VllmConfig`. Consequently, selecting backend='transformers' in an environment without vLLM raises `ModuleNotFoundError: No module named 'vllm'` here, before Transformers loads anything. I reproduced this by calling Mineru2_5Model(..., backend='transformers').load(). The model's transformers virtualenv deliberately does not install vLLM, so the advertised independent backend cannot start. Move the vLLM-only import into its backend branch and add a transformers startup regression with vLLM absent; do not mock the entire mineru_vl_utils package, since that hides this import dependency.

xinference/core/model.py
```diff
@@ -1306,6 +1306,28 @@ async def ocr(
             )
         raise AttributeError(f"Model {self._model.model_spec} is not for ocr.")
 
+    @log_async(
```
qinxuye: [P2] Apply the existing request-limit decorator to document analysis

Unlike the neighboring inference methods, docanalyze has only log_async and bypasses `@request_limit`. The configured request limit and serve-count accounting therefore do not apply to document parsing. Reproduced with _request_limits=1 and _serve_count=1: ModelActor.docanalyze still calls _call_wrapper_json and returns successfully instead of rejecting the request at capacity. Please apply the standard request-limit wrapper and add regressions for rejection at capacity and serve-count restoration after success/error, so this new capability honors the same launch-time limit as other inference methods.

qinxuye (commented):
Three confirmed integration issues below: model-level authorization is bypassed, the transformers backend requires vLLM unexpectedly, and docanalyze bypasses the configured request limit. These findings concern the Xinference integration, not the vendored implementation.

xinference/api/restful_api.py
llyycchhee: Thanks for catching this. Added the Request parameter and model-level authorization check before model lookup, following the existing image handlers. Authenticated route regressions use a real AdvancedAuthService to verify that a restricted API key receives 403 without model lookup or inference, while an authorized key succeeds.

xinference/model/image/docanalyze/mineru25.py
llyycchhee: Fixed. The vLLM-only logits processor and configuration helper are now imported only in the vLLM backend branch. Added a transformers startup regression that blocks vLLM imports while retaining the real vendored package and lazy loader; it successfully initializes the transformers client.

xinference/core/model.py
llyycchhee: Fixed by applying the existing @request_limit decorator to docanalyze. Added regressions verifying rejection at capacity without inference and serve-count restoration after success, errors, and cancellation.

qinxuye (approved):
LGTM. Verified the authorization, Transformers-only startup, and request-limit fixes; all 56 focused tests pass.

PR #5551, opened 2026-09-17
feat(audio): support FishAudio-S2-Pro with MLX

qinxuye:
Adds an MLX engine for FishAudio-S2-Pro on Apple Silicon, using the 8-bit checkpoint from both Hugging Face and ModelScope. PyTorch remains the default engine, and the MLX checkpoint uses a separate cache directory.

The adapter decodes reference audio at the model sample rate, maps reference text and token-limit arguments, and joins generated audio segments. The shared MLX TTS argument handling also accepts reference_text as an alias for prompt_text, including for Qwen3-TTS. Includes engine/catalog regression coverage, regenerated model documentation, and gettext entries with English fallback.

Known limitation: Fish MLX retains upstream's default max_tokens=1024, matching the existing PyTorch Fish token limit. Plain text without speaker markers is not automatically split by upstream and long speech may be silently truncated at this budget. Callers can increase max_tokens (or max_new_tokens), for example to 4096, or split long inputs; increasing the budget is not a guarantee of complete output for arbitrary lengths.

Validation: 57 focused audio tests, including catalog filtering on simulated Linux, Windows, Intel macOS, and Apple Silicon; pre-commit; gettext validation and compilation; focused documentation builds for all nine maintained locales. Full-weight inference has not been run.


xinference/model/audio/mlx_audio.py
```diff
@@ -490,6 +492,27 @@ def _build_generation_kwargs(
...
+            )
+            if reference_audio_path:
+                from mlx_audio.utils import load_audio
+
+                sample_rate = int(getattr(self._model, "sample_rate", 44100))
+                generate_kwargs["ref_audio"] = load_audio(
+                    reference_audio_path,
+                    sample_rate=sample_rate,
+                )
+                generate_kwargs["ref_text"] = prompt_text
```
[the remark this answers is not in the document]
qinxuye: Not adopting this validation: the pinned mlx-audio 0.4.6 implementation explicitly accepts missing reference text and normalizes it with `ref_text or ""` in `_prepare_reference_prompt`: https://github.com/Blaizzy/mlx-audio/blob/v0.4.6/mlx_audio/tts/models/fish_qwen3_omni/fish_speech.py#L502-L523 . Our existing PyTorch Fish adapter also permits omitted reference text. Requiring it here would reject calls supported by the upstream API; reference-text quality considerations do not establish a mandatory parameter.

OliverBryant (commented):
Reviewed the adapter against the pinned upstream (`mlx-audio` 0.4.6: `fish_speech.py`, `utils.py`, `config.py`, `prompt.py`, `fish_s1_dac.py`). The integration holds up well and I found **no blockers**. Notes on what I verified, plus three observations (one worth acting on).

**Verified correct**

- Argument mapping matches the upstream signature `generate(text, voice, ref_audio, ref_text, instruct, max_tokens, speed, ...)` exactly, with `**kwargs` absorbing extras.
- `load_audio(path, sample_rate=...)` returns an `mx.array`, and `_prepare_reference_prompt` consumes it via `.ndim` / `mx.mean`. Decoding the reference to an array rather than forwarding a path is therefore required, not optional — the adapter gets this right.
- `Model.sample_rate` is a real property, and the `44100` fallback matches `ModelConfig.sample_rate`'s upstream default.
- Catalog and dispatch: MLX specs are filtered out on non-Apple-Silicon hosts (covered by the new test), `FishAudio-S1-mini` keeps the legacy path, and the PyTorch entry stays first in the JSON so PyTorch remains the default engine.
- Pins are real: HF revision `c8d4481b3f7cbfe64d855c8b7cda7739502fc3ff` resolves, the ModelScope repo exists, and `mlx-audio==0.4.6` publishes a `tts` extra. The virtualenv block is structurally identical to the Qwen3-TTS / VoxCPM entries.
- Test strength: I mutated the new Fish test five ways (removing the `reference_text` fallback, the `max_tokens` mapping and the `instruct` forwarding; hardcoding a wrong sample rate; switching `pop` to `get`). All five were caught.

**On the reference-text suggestion**

Declining it was the right call, and the stated reasoning checks out: upstream normalizes explicitly with `prompt_texts.append(ref_text or "")` in `_prepare_reference_prompt`, and the PyTorch adapter already treats reference text as optional. No change needed.

See the inline comments for the remaining points.

xinference/model/audio/mlx_audio.py
```diff
@@ -490,6 +492,27 @@ def _build_generation_kwargs(
...
+            if instruct:
+                generate_kwargs["instruct"] = instruct
+            max_new_tokens = generate_kwargs.pop("max_new_tokens", None)
+            if max_new_tokens is not None:
+                generate_kwargs.setdefault("max_tokens", max_new_tokens)
```
OliverBryant, line 515: Long inputs are silently truncated here, because this branch has no default token budget while the sibling `qwen3_tts` branch deliberately sets one:

```python
# mlx-audio defaults to 1200 codec tokens, which is only about nine
# seconds of Qwen3-TTS audio and can truncate ordinary paragraphs.
generate_kwargs.setdefault("max_tokens", 4096)
```

The FishAudio branch only forwards `max_tokens` when the caller supplies `max_new_tokens`, so an ordinary request falls back to upstream's `generate(..., max_tokens=1024)`.

Working out what that bounds, from the pinned codec config: `encoder_rates=[2, 4, 8, 8]` gives `hop_length=512`, `frame_length = hop_length * 4 = 2048`, so at 44.1 kHz one semantic frame is ~46.4 ms, i.e. **~21.5 frames/sec → ~47.6 s of audio at 1024 tokens**.

The per-call budget is `min(max_new_tokens, max(32, text_tokens * 12))`, so the cap starts binding once the text passes ~85 tokens (~64 words) and stops growing with the input after that. Crucially, `_split_generation_text` only splits when the text contains `<|speaker:N|>` markers — plain TTS text is sent as a single chunk, so chunking does not rescue long input:

| input | speech needed | ceiling | result |
| --- | --- | --- | --- |
| 100 words | ~40 s | 47.6 s | ok |
| 200 words | ~80 s | 47.6 s | truncated |
| 400 words | ~160 s | 47.6 s | truncated |

The truncation is completely silent: `_generate_codes_for_batch` just exits its loop at the budget and the partial codes decode normally — only a zero-token result raises. The caller receives audio cut off mid-sentence with no error or warning.

I am not calling this a blocker, since the PyTorch Fish adapter also defaults to `max_new_tokens=1024`, making this consistent across both engines rather than a regression introduced here. But given that the branch directly above this one exists specifically to guard against this failure mode, the omission reads more like an oversight than a decision. Suggest either `generate_kwargs.setdefault("max_tokens", 4096)` (~190 s), or recording it as a known limitation in the PR description.

xinference/model/audio/mlx_audio.py
```diff
@@ -453,6 +453,8 @@ def _build_generation_kwargs(
         generate_kwargs = dict(kwargs)
         prompt_speech = generate_kwargs.pop("prompt_speech", None)
         prompt_text = generate_kwargs.pop("prompt_text", None)
+        reference_text = generate_kwargs.pop("reference_text", None)
+        prompt_text = prompt_text or reference_text
```
OliverBryant, line 457: Minor, and I would not change the code: this fallback sits in the shared prologue rather than inside the FishAudio branch, so it also changes `qwen3_tts` behavior.

I confirmed by running it — a `qwen3_tts` call passing `prompt_speech` plus only `reference_text` (no `prompt_text`) raised `ValueError: prompt_text is required when prompt_speech is provided` before this change and now succeeds.

That is a widening rather than a break, and it aligns MLX with the PyTorch adapter's `kwargs.get("prompt_text", kwargs.get("reference_text", ""))`, so it arguably fixes an inconsistency. Worth a line in the PR description, since the change currently reads as Fish-only.

xinference/model/audio/engine.py
```diff
@@ -152,6 +154,25 @@ def is_model_family_supported(cls, model_family: "AudioModelFamilyV2") -> bool:
         return model_family.model_name == "Kokoro-82M"
 
 
+class PyTorchFishAudioModel(FishSpeechModel, AudioEngineModel):
```
OliverBryant, line 157: Not a blocker — unreachable today — but flagging it as a latent inconsistency now that FishAudio is dual-engine.

In `core.py`'s legacy dispatch, every other dual-engine family carries an MLX check (`whisper`, `funasr`, `MeloTTS`, `qwen3_asr`, `qwen3_tts`, `VoxCPM`), for example:

```python
elif model_spec.model_family == "qwen3_tts":
    if (model_spec.engine or "").lower() == "mlx":
        model = MLXAudioTTSModel(...)
    else:
        model = Qwen3TTSModel(...)
```

The FishAudio branch still constructs the PyTorch adapter unconditionally:

```python
elif model_spec.model_family == "FishAudio":
    model = FishSpeechModel(model_uid, model_path, model_spec, **kwargs)
```

I checked reachability: `MLXAudioTTSEngineModel.match` requires Darwin + arm and `engine == "MLX"`, and since `FishAudio-S2-Pro` is now in `AUDIO_ENGINES` the engine-aware dispatch sets `audio_cls` first, so an MLX spec never reaches the legacy branch. It only triggers under unsupported conditions, hence not a blocker — but it would load the MLX 8-bit checkpoint with the PyTorch adapter if the dispatch order ever changed.

OliverBryant (approved):

xinference/model/audio/mlx_audio.py
qinxuye: Confirmed against pinned mlx-audio 0.4.6. Taking the suggested documentation option: the PR description now records the 1024-token default, possible silent truncation for long plain text, and the max_tokens/max_new_tokens override or input splitting. Keeping the upstream default in this integration; increasing it alone would still leave a finite cap.

xinference/model/audio/mlx_audio.py
qinxuye: Agreed. The PR description now explicitly notes that the shared reference_text alias also widens Qwen3-TTS argument handling. No code change needed.

xinference/model/audio/engine.py
qinxuye: Not changing the legacy fallback in this PR: FishAudio-S2-Pro uses registered engine-aware dispatch, so its MLX checkpoint cannot reach this branch under the supported configuration. Existing dispatch tests cover default PyTorch and explicit MLX selection. A change to dispatch order should revisit this fallback with corresponding coverage.

PR #5552, opened 2026-09-18
perf: modify model_ability to suit frontend pages

llyycchhee:
## Summary

Add model-specific prefixes to embedding and rerank sub-abilities so the frontend can distinguish them from general model abilities and adapt its pages accordingly.

- Embedding: `vision`, `video`, and `audio` become `embed_vision`, `embed_video`, and `embed_audio`.
- Rerank: `vision`, `video`, and `audio` become `rerank_vision`, `rerank_video`, and `rerank_audio`.
- Add the six sub-abilities to the frontend ability mappings and translations while preserving existing general abilities.

This change updates model ability metadata. It does not change API endpoints, request schemas, or inference behavior.

## Validation

- Updated embedding and rerank ability tests: 2 passed.
- Verified parsing and description output for all six sub-abilities.
- Frontend TypeScript, Prettier, and diff checks passed.


xinference/model/embedding/core.py
```diff
@@ -95,7 +95,7 @@ class EmbeddingModelFamilyV2(BaseModel, ModelInstanceInfoMixin):
     max_tokens: int
     language: List[str]
     # Extra accepted input modalities. Text is implicit for every embedding model.
-    model_ability: List[Literal["vision", "video", "audio"]] = Field(
+    model_ability: List[Literal["embed_vision", "embed_video", "embed_audio"]] = Field(
```
qinxuye, line 105: [P2] Normalize legacy modality values before validating persisted registrations

Replacing the accepted literals outright also changes CustomEmbeddingModelFamilyV2, which inherits this schema; the same applies to CustomRerankModelFamilyV2 and the changed rerank schema. Previously valid persisted V2 registrations containing model_ability=['vision'] (or video/audio) now raise ValidationError. Reproduced for both custom family classes using an otherwise valid normalized model spec. During startup, register_custom_model() catches that error, warns, and skips the registration, so affected custom models disappear after an upgrade/restart. Updating the bundled JSON files does not migrate user registrations under the model directory. Please normalize the old values to embed_* / rerank_* before literal validation while continuing to emit the new names. Add regressions for loading legacy embedding and rerank registrations, as well as accepting the new values and rejecting genuinely unknown abilities.

llyycchhee: Thanks for catching this. The literal replacement breaks loading existing custom model registrations that contain legacy modality values.

I'll add pre-validation normalization in both model family schemas to map legacy values to the corresponding embed_* / rerank_* names. This will preserve compatibility with persisted registrations while continuing to emit the new names and reject unknown abilities.

I'll also add regression tests for loading legacy embedding and rerank registrations, accepting the new values, and rejecting unknown abilities.

qinxuye (approved):
LGTM. Verified legacy embedding/rerank ability normalization, persisted registration loading, new-value acceptance, and unknown-value rejection. All 16 focused tests passed and the review thread is resolved. Full CI is still running.

PR #5553, opened 2026-09-18
feat(audio): support Irodori TTS Small with MLX

qinxuye:
## Summary
- Add Irodori-TTS-v4.1-Small MLX support using mlx-audio 0.5.0 and the mlx-community 8-bit checkpoint, with both Hugging Face and ModelScope sources and a separate cache.
- Preserve PyTorch quantization variants and support voice descriptions, reference audio, and seeds in the MLX adapter.
- Filter audio quantization choices and cache indicators by engine, resetting incompatible selections when switching engines.
- Make voice-design instructions optional in the generic speech form; keep music descriptions required.
- Generate model documentation and update all nine translations.

## Validation
- 69 focused Python audio tests passed.
- 3 audio quantization tests and 9 capability/speech panel tests passed.
- ESLint, TypeScript, formatting, and commit hooks passed.
- Translation catalogs validated and compiled; targeted documentation builds passed for all nine locales during implementation.
- User verified real audio generation with and without voice instructions.

## Notes
- Japanese text input only; MLX uses a fixed 8-bit checkpoint. The default quantization selector value does not indicate unquantized weights.
- No recommendation-branch changes are included.


doc/source/gen_docs.py
```diff
@@ -539,7 +539,27 @@ def main():
...
+                    + '\n'.join(f"* ``{engine}``" for engine in engines)
+                    + '\n\nAvailable quantizations by engine\n'
+                    '^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n\n'
+                    + '\n'.join(
+                        f"* ``{engine}``: " + ', '.join(
+                            f"``{spec['quantization'] or 'none'}``"
+                            for spec in engine_specs if spec['engine'] == engine
+                        ) for engine in engines
+                    )
+                )
```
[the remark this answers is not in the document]
qinxuye: Not changing this: engine_specs is constructed immediately above with explicit engine, quantization and model_id keys, so direct indexing cannot raise the reported KeyError even for legacy entries. None is retained intentionally for legacy models without a named engine. No current mixed-engine quantized group contains a null engine, and documentation generation and both remote documentation builds pass. Defaulting all legacy entries to PyTorch would introduce an unsupported engine assumption.

frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -346,6 +339,7 @@ export default function LaunchDialog({
   }, [
     cacheIndex.quantizations,
     model?.modelSpecs,
+    modelEngineValue,
     modelFormatValue,
     modelSizeInBillionsKey,
     modelType,
```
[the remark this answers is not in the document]
qinxuye: Not applicable: isCachedSpec is a module-level named import from ../utils, not a function or callback created inside the component. Its identity does not depend on renders, so it is not a reactive dependency. ESLint and TypeScript checks pass; no dependency change is needed.

xinference/model/audio/tests/test_irodori_tts.py
```diff
@@ -61,15 +61,37 @@ def synthesize(self, request):
...
+        ("linux", "x86_64", False),
+        ("win32", "AMD64", False),
+        ("darwin", "i386", False),
+        ("darwin", "arm", True),
+    ],
+)
+def test_irodori_catalog_registers_quantization_variants(
+    monkeypatch, system, processor, supports_mlx
+):
+    from .. import platform, sys
```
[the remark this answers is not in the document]
qinxuye: Cleaned up in 60c3ac198: import platform directly at module scope and reuse the existing sys import. The original imports worked because audio/__init__.py imports both modules, but direct standard-library imports are clearer. All 33 focused Irodori and MLX tests and commit hooks pass.

xinference/model/audio/mlx_audio.py
```diff
@@ -513,6 +513,21 @@ def _build_generation_kwargs(
             max_new_tokens = generate_kwargs.pop("max_new_tokens", None)
             if max_new_tokens is not None:
                 generate_kwargs.setdefault("max_tokens", max_new_tokens)
+        elif model_family == "Irodori-TTS":
+            caption = generate_kwargs.pop("caption", None)
+            instruction = generate_kwargs.pop("instruction", None)
+            if caption is None:
+                caption = instruct or instruction or prompt_text
+            generate_kwargs.update(text=input, caption=caption)
```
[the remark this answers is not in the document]
qinxuye: Already handled before family dispatch: _build_generation_kwargs copies kwargs and immediately calls generate_kwargs.pop("instruct", None) while extracting the instruction aliases. Therefore instruct cannot remain in generate_kwargs in the Irodori branch. The existing parameterized test exercises instruct with a strict fake generate signature, including reference-audio variants; all 33 focused tests pass. No duplicate pop is needed.

doc/source/gen_docs.py
```diff
@@ -539,7 +539,27 @@ def main():
 
         for model in sorted_models:
             engine_specs = model['engine_specs']
-            if any(spec['quantization'] is not None for spec in engine_specs):
+            engines = list(dict.fromkeys(spec['engine'] for spec in engine_specs))
+            if len(engines) > 1 and any(spec['quantization'] is not None for spec in engine_specs):
```
OliverBryant, line 543: Re the bot's `KeyError` concern on `spec['engine']` — **this cannot raise**, so no change is needed.

The entries are constructed a few lines above (in the same file, ~line 528):

```python
rendered_model['engine_specs'].append(
    {
        'engine': model.get('engine'),
        'quantization': model.get('quantization'),
        'model_id': model_id,
    }
)
```

The `'engine'` key is always present — `.get()` supplies `None` rather than omitting it. I also replayed this exact branch logic over the real `model_spec.json` (97 specs, 40 of which have no `engine` key): zero `KeyError`s, and no model reaches the new branch with a `None` engine. The pre-existing `elif len(engine_specs) > 1` branch below has been indexing `spec['engine']` the same way already.

There *is* a latent cosmetic issue in the neighbourhood, though it is unreachable today: if a model ever had legacy specs (no `engine`) **plus** a new engine-tagged spec **plus** quantizations, this branch would render literally `- **None (INT8) model ID:** ...` and `* ``None```. I confirmed that by constructing such a spec list. Nothing in the catalog hits it — and this PR actively removes the possibility for Irodori by tagging all six existing specs `"engine": "PyTorch"`. Worth a `or 'PyTorch'` only if you want to harden against a future contributor adding an MLX spec without backfilling the PyTorch ones.

xinference/model/audio/mlx_audio.py
OliverBryant, line 521: Re the bot's suggestion to also pop `"instruct"` here — **`instruct` can never be in `generate_kwargs` at this point**, so the extra pop would always yield `None`.

It is consumed in the shared prologue of this same function (~line 460), before any family branch runs:

```python
instruct = (
    generate_kwargs.pop("instruct", None)
    or generate_kwargs.pop("control_instruction", None)
    or generate_kwargs.pop("instruct_text", None)
)
```

I ran `_build_generation_kwargs` for all four caption sources to confirm the resulting kwargs:

```
input {'instruct': 'calm voice'}  ->  keys=['caption', 'text']  caption='calm voice'  'instruct' present=False
input {'caption': 'c'}            ->  keys=['caption', 'text']  caption='c'           'instruct' present=False
input {'instruction': 'i'}        ->  keys=['caption', 'text']  caption='i'           'instruct' present=False
input {'prompt_text': 'p'}        ->  keys=['caption', 'text']  caption='p'           'instruct' present=False
```

Separately, the premise that this would `TypeError` downstream is also wrong — upstream treats the two as synonyms (`irodori_tts.py:471`):

```python
caption = caption or kwargs.pop("instruct", None)
```

The current `caption = instruct or instruction or prompt_text` chain is correct as written.

xinference/model/audio/tests/test_irodori_tts.py
```diff
@@ -61,15 +61,37 @@ def synthesize(self, request):
...
+
```
OliverBryant, line 77: Re the bot's suggestion to replace this with plain `import platform` / `import sys` — **please don't; it would silently defeat the test.**

The indirection is deliberate. `xinference/model/audio/__init__.py` imports both modules at its top (lines 18–19) and `_need_filter` reads them from that namespace:

```python
def _need_filter(spec: dict):
    if (sys.platform != "darwin" or platform.processor() != "arm") and spec.get(
        "engine", ""
    ).upper() == "MLX":
```

Since `load_model_family_from_json` lives in that same module, `from .. import platform, sys` is what lets `monkeypatch.setattr` rebind the objects the filter actually consults. Importing them locally would patch this test module's own references instead, the monkeypatch would become a no-op, and the `supports_mlx=True` / `False` cases would stop exercising the MLX-visibility filter while still passing. So the current form is not fragile — it is the only form that works, and `ImportError` is not a risk because the parent package genuinely binds those names.

frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
```diff
@@ -298,24 +299,16 @@ export default function LaunchDialog({
...
+        const matchingSpecs = audioSpecs.filter(
           (spec) => toOptionValue(spec.quantization) === quantization
         );
-        const cached = matchingSpecs.some(
-          (spec) =>
-            cacheIndex.quantizations.has(
-              createCacheKey(toOptionValue(spec.model_format), '', quantization)
-            )
-        );
+        const cached = matchingSpecs.some(isCachedSpec);
```
OliverBryant, line 312: Two notes on this line.

**1. The bot's dependency-array finding does not apply.** `isCachedSpec` is a module-scope import from `../utils` (line 44 of this file), not a `useCallback` defined in the component. `react-hooks/exhaustive-deps` does not ask for stable module bindings, and the bot's own premise was conditional ("if `isCachedSpec` is defined inside the component") — that condition is false here. Adding it would be harmless but meaningless.

**2. This refactor quietly fixes a real bug, which may be worth a line in the PR description.** The previous form looked cached-ness up by key:

```js
cacheIndex.quantizations.has(createCacheKey(toOptionValue(spec.model_format), '', quantization))
```

That key omits the engine, so two engines sharing a `model_format` would alias. I compared old vs new behaviour on a spec pair where PyTorch `none` is **not** cached and MLX `none` **is**:

```
-- distinct model_format (pytorch / mlx) --
engine=PyTorch quant=none  OLD=false NEW=false  same
engine=MLX     quant=none  OLD=true  NEW=true   same

-- same model_format across engines --
engine=PyTorch quant=none  OLD=true  NEW=false  *** differs (new is correct) ***
engine=MLX     quant=none  OLD=true  NEW=true   same
```

Irodori uses distinct formats so nothing regresses today, but the new version is strictly more correct. Both paths ultimately read `spec.cache_status` on the same objects (`cacheIndex` is itself built gated on `isCachedSpec` at line 227), so this is equivalent-or-better, not a semantic loss.

OliverBryant (commented):
Reviewed against the pinned upstream (`mlx-audio` 0.5.0 `irodori_tts.py`) and traced the engine-scoped quantization change end to end through `worker.py` → `normalizeModels` → the launch dialog. **No blockers.**

**Verified correct**

- Argument mapping matches upstream `generate(text, voice, ref_audio, caption, stream, **kwargs)`; `rng_seed` / `seconds` / `num_steps` reach `generate_latents` through the `**kwargs` passthrough. `ref_audio` accepts a `str` path here (unlike FishAudio, which needs a decoded array), so forwarding the temp-file path is right.
- `voice` really is unused upstream, so the warning is accurate rather than defensive.
- `apply_mlx_audio_seed` already declared `-> Optional[int]`, so capturing its return value needed no signature change.
- Pins are real: HF repo and revision `c81956b675ce21422c305ef683fb2ac59d7faf6a` both resolve, the ModelScope repo exists, and `mlx-audio==0.5.0` is published.
- **Engine-scoped quantizations work end to end.** `worker.py`'s audio branch emits `model_engine`, `model_format` and `quantization` (defaulted to `"none"`) per family, `normalizeModels` passes `model_specs` through untouched, and with the real catalog data `audioSpecsForEngine` yields 6 quantizations for PyTorch and `none` only for MLX — matching the regenerated docs.
- Adding `"engine": "PyTorch"` to the six existing specs is what keeps `gen_docs.py`'s new multi-engine branch from rendering a literal `None` engine, so that part is load-bearing rather than cosmetic.
- Test strength: I mutated the Irodori MLX adapter eight ways (breaking each link of the caption chain, dropping `rng_seed`/`ref_audio` forwarding, ignoring `reference_speech`, reverting the seed capture) and the 4×3 parametrization caught every one. I also mutated `audioSpecsForEngine` four ways (case-sensitive compare, dropping the empty-engine early return, widening the legacy fallback, removing it) — all four caught.

**One improvement worth calling out**

Switching the audio cache indicator from `cacheIndex.quantizations.has(createCacheKey(...))` to `matchingSpecs.some(isCachedSpec)` is not just a simplification, it fixes a latent bug. The old key was `(model_format, '', quantization)`, so when two engines shared a `model_format` a cached spec on one engine made the *other* engine's same-named quantization show as cached. Irodori uses distinct `pytorch` / `mlx` formats so it does not trigger today, but the new form is strictly more correct.

**On the four bot findings — none hold up.** Details inline; summarising here because the pattern matters:

1. `gen_docs.py` `KeyError` — impossible. Each entry is built as `{'engine': model.get('engine'), ...}`, so the key is always present.
2. `isCachedSpec` missing from the `useMemo` deps — it is a module-scope import from `../utils`, not a component-local function, so `exhaustive-deps` does not require it.
3. `from .. import platform, sys` is fragile — it is deliberate and load-bearing, and the suggested fix would silently break the test.
4. `instruct` leaking into `generate()` and raising `TypeError` — it is already popped in the shared prologue, and upstream aliases `instruct` to `caption` anyway.

I ran each of these rather than reasoning about them; the reproductions are in the inline comments.

doc/source/gen_docs.py
qinxuye: Thanks for verifying. Keeping the current behavior: this catalog has no mixed-engine quantized group with a null engine, and the six Irodori PyTorch variants are explicitly tagged. We should not assume every legacy audio model uses a named PyTorch engine just to cover a hypothetical future catalog entry.

xinference/model/audio/mlx_audio.py
qinxuye: Thanks for checking upstream and the argument mapping. Agreed; no additional pop is needed. Keeping the existing implementation and caption-source regression coverage.

xinference/model/audio/tests/test_irodori_tts.py
qinxuye: The direct imports in current head 60c3ac198 do not break this test. Python returns the same module objects: audio.platform is platform and audio.sys is sys are both True. monkeypatch.setattr(sys, "platform", ...) mutates an attribute on that shared module object; it does not rebind the local import. I verified the actual audio._need_filter with these direct imports: darwin/arm returns False for an MLX spec, then linux returns True. All 11 Irodori tests pass, including the four platform cases. Patching audio.platform itself would be a different operation, but this test only patches module attributes. Keeping the direct imports.

frontend/src/components/pages/launch-model/launch-dialog/launch-dialog.tsx
qinxuye: Thanks for verifying both points. The PR summary already calls out filtering audio quantization choices and cache indicators by engine. The engine-scoped matchingSpecs.some(isCachedSpec) behavior is intentional; no dependency-array change is needed.

OliverBryant (approved):
LGTM — approving on `60c3ac198`.

**Correction on my part first:** my earlier claim that switching to direct `import platform` / `import sys` would "silently defeat the test" was wrong, and your reply is right. `import platform` in two modules returns the same object out of `sys.modules`, so `monkeypatch.setattr(platform, "processor", ...)` mutates an attribute on the object `audio/__init__.py` also holds a reference to. I verified this rather than take it on faith:

```
pkg.sys is sys           -> True
pkg.platform is platform -> True
darwin/arm -> need_filter(MLX) = False   (MLX visible)
linux/x86  -> need_filter(MLX) = True    (MLX hidden)
```

The distinction I had in mind only applies to `setattr(audio_module, "platform", ...)`, i.e. rebinding the package's own name, which this test never does. The new imports are equivalent and clearer.

I re-mutated the platform filter against the four parametrized cases on the current head to confirm the assertions still discriminate: dropping the `arm` check, never filtering, and always filtering MLX are each caught. (One mutation — dropping the `darwin` check alone — survives, but that is pre-existing `_need_filter` coverage and needs a non-Darwin host reporting `processor() == "arm"` to matter. Out of scope here; not worth holding this PR for.)

**Verified nothing else moved:** `7f597fc...60c3ac1` touches only `test_irodori_tts.py` (1+/2-), so `mlx_audio.py`, `gen_docs.py`, `audio-quantization.ts` and `form-panels.tsx` are byte-identical to what I already reviewed — upstream argument mapping, the pinned HF revision and ModelScope sources, and the end-to-end engine-scoped quantization behaviour (PyTorch → 6 quantizations, MLX → `none`) all still hold.

Agreed on the three declines: the `gen_docs.py` `None`-engine case stays hypothetical and defaulting legacy entries to PyTorch would assert an engine those models never declared; `isCachedSpec` is a module-scope import and not a reactive dependency; and the duplicate `instruct` pop is dead code given the shared prologue plus upstream's own `caption or kwargs.pop("instruct")` alias.

No blockers. CI is green on lint, both docs builds and the completed build jobs, with the macOS/Windows matrix still running — worth a glance before merge.

PR #5554, opened 2026-09-19
fix(worker): run supervisor reporting RPCs on actor loop

m199369309:
## What does this PR do?

Runs the worker's Supervisor heartbeat and status-reporting RPC lifecycle on
the worker actor loop.

The periodic reporting task still uses its dedicated isolation loop for
resource and GPU-status collection. Only the Supervisor-facing operations are
dispatched to the actor loop:

- Supervisor ActorRef lookup and initialization
- `receive_heartbeat`
- `report_worker_status`
- RPC timeout and cancellation handling
- generation-conditional reference cleanup
- worker re-registration and replica replay

## Why?

The periodic status task runs in a separate thread and event loop. Although
Supervisor ActorRef initialization was already dispatched to the worker actor
loop, the ActorRef method call and `xo.wait_for()` still ran on the reporting
isolation loop. This could create and reuse a separate xoscar client/listener
for periodic Supervisor RPCs.

In an affected long-running worker, Supervisor requests were handled and their
responses reached the worker TCP receive queue, but the reporting connection
stopped consuming responses. The corresponding heartbeat and status-reporting
futures then timed out repeatedly.

This change keeps the complete Supervisor RPC lifecycle on the actor loop while
leaving potentially blocking resource collection on the isolation loop.

## Implementation notes

- `_run_on_actor_loop` accepts a coroutine factory so RPC-related async objects
  are created only after switching to the actor loop.
- `_call_supervisor` centralizes ActorRef lookup, RPC execution, timeout, and
  generation-safe cleanup on that loop.
- Cancellation of the isolation-loop caller also cancels the actor-loop future,
  preventing registration locks or pending tasks from being left behind during
  worker shutdown.
- Existing heartbeat cadence, full-status cadence, reconnection behavior, and
  public interfaces are unchanged.

## Verification

- `pytest -q xinference/core/tests/test_worker.py`
  - 134 passed
- `pre-commit run --files xinference/core/worker.py xinference/core/tests/test_worker.py`
  - all hooks passed
- Manual real-xoscar TCP cross-loop stress check
  - 500 heartbeat RPCs passed
  - 500 status-reporting RPCs passed
  - no timeout, reference reset, or pending-task leak

Not run: the full CI-style test suite; many model tests require optional
dependencies, model downloads, GPU hardware, or platform-specific runtimes.


qinxuye (approved):
LGTM. Reviewed actor-loop dispatch, cancellation propagation, and generation-safe supervisor reference cleanup. All 134 worker tests pass locally. The Windows 3.14 CI error occurs during pool setup with a socket bind PermissionError (WinError 10013), before the test logic runs.

PR #5555, opened 2026-09-19
feat(audio): support Breeze TTS 2 with MLX

qinxuye:
## Summary
- Register Breeze-TTS-2 with selectable PyTorch and MLX engines while preserving the existing CUDA runtime.
- Add the fixed mlx-community 8-bit checkpoint from Hugging Face and ModelScope, with an isolated MLX cache and mlx-audio[tts]==0.5.1.
- Map voice instructions, reference audio/transcripts, speaker tags, guidance scale, and token limits to the MLX API. Validate incomplete cloning inputs and clean temporary audio files on errors.
- Regenerate model documentation and add translations for all nine maintained locales.

## Validation
- 108 focused audio tests passed after rebasing onto latest main, covering parameter mapping, platform filtering, engine/source selection, and existing Irodori behavior.
- All commit hooks and git diff --check passed.
- All nine translation catalogs passed msgfmt validation and compilation; targeted HTML builds passed in all nine locales.
- Verified checkpoint and bundled codec/tokenizer files on both hubs.

## Limitations
- Real-weight audio generation has not been tested locally.
- MLX output is non-streaming through the existing shared adapter; PyTorch streaming is unchanged.
- The MLX variant uses fixed 8-bit weights; no additional quantization variants are registered.
- Existing Breeze research/non-commercial license restrictions are preserved.


xinference/model/audio/mlx_audio.py
```diff
@@ -513,6 +513,39 @@ def _build_generation_kwargs(
...
+            instruction = generate_kwargs.pop("instruction", None)
+            instruction = instruct or instruction
+            ref_audio = prompt_audio_path or reference_audio_path
+            if ref_audio and not (prompt_text and prompt_text.strip()):
+                raise ValueError(
+                    "Breeze-TTS-2 requires the reference audio transcript in "
+                    "`prompt_text` when reference audio is provided."
+                )
+            if not ref_audio:
+                instruction = instruction or prompt_text
```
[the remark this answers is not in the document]
qinxuye: Fixed in 622e98ea2. The Breeze MLX branch now converts instruction and reference text to strings and strips whitespace before validation, matching the existing PyTorch normalization. Added regression cases for numeric inputs, surrounding whitespace, blank instructions using the default, and blank reference transcripts being rejected. All 116 focused audio tests and commit hooks pass.

xinference/model/audio/mlx_audio.py
```diff
@@ -513,6 +513,43 @@ def _build_generation_kwargs(
...
+                cfg_scale = guidance_scale if guidance_scale is not None else 1.0
+            max_new_tokens = generate_kwargs.pop("max_new_tokens", None)
+            if max_new_tokens is not None:
+                generate_kwargs.setdefault("max_tokens", max_new_tokens)
+            speaker = generate_kwargs.pop("speaker", None)
+            generate_kwargs.update(
+                text=input,
+                voice=speaker
+                or (voice if voice not in self._OPENAI_VOICES else None)
+                or "S0",
```
OliverBryant, line 549: Minor: this membership test is case-sensitive, while the sibling VoxCPM branch lower-cases before the same check (`elif voice and voice.lower() not in self._OPENAI_VOICES:`, ~line 576 of this file). `_OPENAI_VOICES` holds lowercase names only, so a capitalized OpenAI voice slips through and becomes a Breeze speaker tag instead of falling back to `S0`.

I ran the branch to confirm:

```
voice=alloy    -> generate voice='S0'
voice=Alloy    -> generate voice='Alloy'     <-- becomes speaker tag [Alloy]
voice=ALLOY    -> generate voice='ALLOY'
voice=echo     -> generate voice='S0'
voice=Echo     -> generate voice='Echo'
voice=S1       -> generate voice='S1'        (correct: real speaker tag)
voice=''       -> generate voice='S0'
```

Reachability: `restful_api.py` passes `body.voice` straight through to `model.speech()` with no normalization or enum validation, so a client sending `"Alloy"` arrives here verbatim. The blast radius is small — OpenAI's documented voice names are lowercase, so this needs a client that capitalizes them — and the result is a nonsense speaker tag rather than a crash. Not a blocker.

`voice.lower() not in self._OPENAI_VOICES` would match the VoxCPM branch and close it. Worth adding a case to the parametrized test if you do change it; none of the new Breeze tests currently exercise a capitalized voice.

OliverBryant (commented):
Reviewed against the pinned upstream (`mlx-audio` 0.5.1 `breeze_tts/breeze_tts.py`). **No blockers** — one small inconsistency inline that is worth a one-word fix, plus a recurring pattern note.

**Verified correct**

- Argument mapping matches upstream `generate(text, voice, instruct, ref_audio, ref_text, cfg_scale, max_tokens, ...)` exactly, with `**_` absorbing extras. `ref_audio` accepts `str | Path | mx.array`, so forwarding the temp path is right.
- `"S0"` is upstream's own default speaker (`_speaker()` returns `S0` for an empty voice and wraps it in brackets), so the fallback agrees rather than inventing a tag.
- Importing `_DEFAULT_INSTRUCTION` / `_validate_cfg_scale` from `breeze_tts` is safe here: that module has no import-time torch dependency (torch is under `TYPE_CHECKING` only), so the MLX venv does not pull it in. Reusing the PyTorch validator also keeps the two engines' `cfg_scale` semantics identical rather than duplicating the rule.
- Precedence behaves as documented — I ran each combination: `cfg_scale` beats `guidance_scale` (2.0 vs 4.0), falls back to `1.0`; `max_tokens` beats `max_new_tokens` (100 vs 200) and is omitted entirely when neither is given, so upstream's own default applies.
- Pins are real: HF repo and revision `c6e4a2ff6ab9afba68b7853de802273ffe23fb49` both resolve, ModelScope exists, `mlx-audio==0.5.1` is published.
- Test strength: I mutated the branch nine ways — dropping the transcript validation, either `.strip()`, the default instruction, the `cfg_scale` validation, the instruction→`prompt_text` fallback, inverting the `cfg_scale`/`guidance_scale` precedence, and sending `ref_text` without `ref_audio`. All nine are caught. The last one is caught only by the `instruction_key="prompt_text"` × `reference_key=None` cell of the parametrized matrix asserting `ref_text is None` — nice that the cross-product covers it rather than it being a happy accident of a single case.
- The temp-file cleanup on the validation error is genuinely exercised (`test_mlx_breeze_rejects_missing_transcript_and_cleans_temp_audio` asserts the saved path is gone after the raise), which is the part most easily gotten wrong when validation moves before `generate`.

**Recurring pattern, not for this PR**

Breeze is now the third dual-engine audio family (after FishAudio and Irodori-TTS) whose `core.py` legacy-dispatch branch constructs the PyTorch adapter unconditionally, while `whisper` / `funasr` / `MeloTTS` / `qwen3_asr` / `qwen3_tts` / `VoxCPM` all carry an `engine == "mlx"` check. It remains unreachable for the same reason you gave on #5551 — registered engine-aware dispatch sets `audio_cls` first — so I am not asking for a change. But at three occurrences it is drifting from "one exception" toward "the new default shape", and whoever eventually reorders dispatch will have three of these to find rather than one. Might be worth a follow-up issue so it is written down somewhere other than a merged PR thread.

OliverBryant (approved):

xinference/model/audio/mlx_audio.py
qinxuye: Fixed in b10ea10d9. OpenAI voice aliases are now matched case-insensitively, while real speaker tags and explicit speaker overrides retain their original values. Added seven regression cases including Alloy, ALLOY, Echo, empty/None voices and S1, with speaker override assertions. All 123 focused audio tests and commit hooks pass.

PR #5556, opened 2026-09-19
ENH(llm): extend Ling-3.0-flash model support

Minamiyama:
## Summary

Extend existing inclusionAI/Ling-3.0-flash support with official dialogue checkpoints and quantized variants from Hugging Face and ModelScope.

- Register official dialogue model variants and quantized checkpoints.
- Extend Transformers support for `BailingMoeV3ForCausalLM`.
- Add Ling-3.0-flash compatibility to vLLM and SGLang.
- Support offline FP4 and Int4 checkpoints with SGLang 0.5.19.
- Extend SGLang startup timeout for the 124B checkpoint.
- Update generated built-in model documentation.


xinference/model/llm/llm_family.json
```diff
@@ -1256,7 +1256,8 @@
     "virtualenv": {
       "packages": [
         "#transformers_dependencies# ; #engine# == \"Transformers\"",
-        "#sglang_dependencies# ; #engine# == \"sglang\"",
+        "numpy<2.3 ; #engine# == \"sglang\"",
```
qinxuye: [P2] Restore the SGLang dependency placeholder for HuatuoGPT

This change is in HuatuoGPT-o1-Qwen2.5, not Ling. Replacing #sglang_dependencies# with only numpy/pandas removes the engine and its supporting dependencies from this existing model's virtualenv. Using the same package expansion/filtering helpers as worker setup yields only ['numpy<2.3', 'pandas<3']; there are no SGLang model-format defaults to restore the missing engine. On a host without SGLang already installed, launching this model with SGLang therefore fails at import instead of provisioning the engine. Please restore the original placeholder here and keep Ling-specific dependency changes scoped to Ling. Add a regression asserting that HuatuoGPT's resolved SGLang packages still include the engine.

xinference/model/llm/sglang/core.py
```diff
@@ -283,9 +288,15 @@ def _load():
                 model_path=self.model_path,
                 tokenizer_path=self.model_path,
                 port=sgl_port,
+                launch_timeout=self._get_launch_timeout(),
```
qinxuye: [P2] Preserve explicit launch_timeout without passing it twice

The model config retains user-supplied launch_timeout, so adding an explicit keyword alongside **self._model_config makes an existing custom timeout fail before Runtime is called. Reproduced through SGLANGChatModel.load() with {'launch_timeout': 1200}: both Ling-3.0-flash and HuatuoGPT-o1-Qwen2.5 raise TypeError: got multiple values for keyword argument 'launch_timeout'. The distributed branch has the same duplication. Please apply the family-specific timeout as a default in the config (or pop the override before forwarding), preserving explicit user values and passing the keyword exactly once. Cover default and overridden timeouts for Ling and another family, including the distributed construction path.

qinxuye (approved):
LGTM. Verified restoration of the HuatuoGPT SGLang dependency placeholder and preservation of explicit launch_timeout overrides in both single-worker and distributed paths. All nine new regression cases pass; all review threads are resolved.

PR #5558, opened 2026-09-20
feat(audio): support MiniMax Music3 with MLX

qinxuye:
## Summary
- Add a dedicated mlx-audio 0.5.0 music adapter for MiniMax-Music3 on Apple Silicon, preserving the lyrics and music-description API and 44.1 kHz stereo output.
- Register the fixed 8-bit checkpoint on Hugging Face and ModelScope with an isolated cache and an explicit 8-bit quantization label.
- Select the matching MLX weights when MLX is the default engine; preserve the existing NVIDIA CUDA/diffusers path. When quantization is omitted, automatically select a sole available quantization; reject explicit unsupported values.
- Support non-streaming WAV, MP3, FLAC and OGG output, validate generation parameters and stereo results, and run loading/inference on a dedicated MLX thread.
- Regenerate model documentation and add all nine locale catalogs.

## Validation
- 178 focused audio tests passed, covering engine/source/default selection, quantization selection, platform filtering, stereo sample preservation, output formats, parameter validation and MLX thread affinity.
- All commit hooks and git diff --check passed.
- All nine catalogs passed msgfmt validation and compilation; targeted documentation builds passed in all nine languages.
- Verified the pinned Hugging Face revision and ModelScope weight/tokenizer files.

## Limitations
- Real-weight audio generation has not been verified locally; adapter tests use a fake generator.
- Only the fixed 8-bit MLX checkpoint is registered, approximately 14 GB of weights; runtime memory requirements are higher.
- No streaming or voice cloning. Lyrics and a music description are required; use [instrumental] for instrumental output.
- Model weights remain subject to the MiniMax-Music3 Community License.


xinference/model/audio/core.py
```diff
@@ -279,7 +285,7 @@ def create_audio_model_instance(
 
     enable_virtual_env = kwargs.pop("enable_virtual_env", None)
     model_name, model_engine = resolve_audio_model_name_and_engine(
-        model_name, model_engine
+        model_name, model_engine, use_default_engine=model_name == "MiniMax-Music3"
     )
```
OliverBryant, line 289: Not a blocker, and I understand why it is here — but flagging a hardcoded model name in otherwise model-agnostic dispatch code.

I traced why it is load-bearing: without `use_default_engine=True`, a launch with no `--model-engine` leaves `model_engine=None`, so `match_audio` never engine-filters, `selected_quantization` stays `"none"`, and the MLX-only `8-bit` spec cannot be selected. With it, `resolve_audio_model_name_and_engine` picks the first registered engine, which then drives both spec selection and the new single-quantization auto-select. So this and the `match_audio` change are two halves of one mechanism.

I also checked the default-engine ordering, since `SUPPORTED_ENGINES` iterates in insertion order (`transformers, vLLM, PyTorch, MLX, diffusers`) and MLX is registered *before* diffusers — which would matter if both could ever register for this model:

```
host=cuda         registered=['diffusers']  default -> diffusers
host=apple        registered=['MLX']        default -> MLX
host=plain-linux  registered=[]             default -> None  (hits the improved error)
```

They are mutually exclusive by host (`MLXMiniMaxMusic3AudioModel` requires Darwin+arm, `DiffusersMiniMaxMusic3AudioModel` requires CUDA), so ordering never actually decides anything and the docs' `--model-engine diffusers` example stays right on a CUDA box. No issue today.

The maintainability concern is just that the next model needing this behaviour will either grow this into `model_name in {...}` or silently not get it. Since the real precondition is "this model's engines are host-exclusive, so the default engine must be resolved before quantization selection", a predicate derived from the spec — or defaulting `use_default_engine` to True for any model whose registered engines all declare a quantization — would express it without naming the model. Entirely fine to leave as-is for one model; worth a follow-up if a second one appears.

OliverBryant (commented):
Reviewed against the pinned upstream (`mlx-audio` 0.5.0 `music/minimax_music3.py` + `music/utils.py`) and the pinned checkpoint's own `config.json`. **No blockers** — this is the cleanest of the four MLX audio PRs so far. One inline note about a hardcoded model name, which is a maintainability observation rather than a defect.

**Verified correct**

- `from mlx_audio.music import load` is valid in 0.5.0 (`music/__init__.py` re-exports `load`/`load_model`), `load(model_path, **kwargs)` matches the call, and `MODEL_REMAPPING = {"minimax_music3": "minimax_music3"}` confirms the model type is supported.
- `generate(text, *, lyrics, duration, steps, seed, **_)` matches the adapter exactly — `lyrics` is keyword-only upstream and the adapter passes everything by keyword.
- **Stereo orientation is right, and the comment is accurate.** Upstream computes `waveform = audio[0].transpose(1, 0)` and reports `samples = waveform.shape[0]`, so it really does yield sample-major `(N, 2)`. I traced the `.T` → `_normalize_audio` round trip at N = 1, 2, 3 and 100 and channels stay correct in every case, including the ambiguous `(2, 2)` one where the array is symmetric and the transpose is its own inverse. The `shape[1] != 2` guard also rejects mis-oriented input before it can reach that branch.
- The pinned checkpoint corroborates all three hard-coded expectations: `sample_rate = 44100` (so the strict rate check passes rather than being a latent failure), `model_type = minimax_music3`, and `quantization.bits = 8` matching the `"8-bit"` label.
- Pins are real: HF repo and revision `10aa4ca578d04c6f5256c1bc22fc8405a09602b5` resolve, ModelScope exists. `mlx-audio==0.5.0` without the `[tts]` extra is the right choice for a `mlx_audio.music` consumer.
- `duration` bounds are strictly narrower than upstream's (adapter `[0.04, 360]` vs upstream `(0, 360]`), i.e. erring toward rejection, and 0.04 s is ~1 frame at the config's 25 fps. Reusing the parent's `_validate_speech_request` keeps diffusers and MLX behaviourally identical instead of duplicating the rules.
- `steps` is popped before validation so it escapes the parent's "unsupported parameter" sweep and gets its own explicit 1–30 check — deliberate, and it matches upstream's own bound.
- Exceptions propagate correctly through the MLX thread: `MLXModelThreadMixin` uses `submit(...).result()`, which re-raises in the caller, so the validation errors genuinely surface. (My first harness wrapped the call in a bare `Thread` and swallowed them — the code was fine, my scaffolding wasn't.)
- Unlike the TTS families, MiniMax has **no** `core.py` legacy-dispatch branch at all — it raises early when the model is absent from `AUDIO_ENGINES` — so it is structurally immune to the missing-MLX-check pattern I flagged on #5555. Worth noting as the better shape.
- Test strength: I mutated the adapter nine ways (dropping the stereo/ndim/finite checks individually, the sample-rate check, the empty-output check, the `steps` range and its bool guard, omitting the `.T`, and concatenating on the wrong axis). All nine are caught. The 15 invalid-request and 5 invalid-output cases plus the two-sample stereo test cover the edges I would have written myself.

**On the quantization auto-select** — this was the change I scrutinised hardest, since it touches generic `match_audio` logic. It is correctly scoped: it fires only when `quantization is None` **and** the (already engine-filtered) family list has exactly one distinct quantization. I replayed old vs new selection across every model in the catalog and every engine scope, and exactly **one** case differs — the intended one:

```
MiniMax-Music3  engine=MLX   OLD=ValueError  NEW=['8-bit']
total differing cases: 1
```

No existing model changes behaviour. Explicit rejection still works and is case-insensitive (`8-BIT` resolves, `INT4` and a `none` request against MLX both raise).

xinference/model/audio/core.py
```diff
@@ -279,7 +285,7 @@ def create_audio_model_instance(
 
     enable_virtual_env = kwargs.pop("enable_virtual_env", None)
     model_name, model_engine = resolve_audio_model_name_and_engine(
-        model_name, model_engine
+        model_name, model_engine, use_default_engine=model_name == "MiniMax-Music3"
```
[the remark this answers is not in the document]
qinxuye: Fixed in ede84c305. Default engine resolution now runs before spec matching for all audio models, removing the MiniMax-specific condition. Added a regression test with Breeze whose first spec is PyTorch but only available engine is MLX, asserting matching MLX weights, hub and isolated cache for both Hugging Face and ModelScope. All 180 focused audio tests and commit hooks pass.

xinference/model/audio/core.py
qinxuye: Addressed in ede84c305, pushed while this review was arriving. Removed the model-name predicate and resolve the default engine before selecting weights for all audio models. The factory already selected that same default later, so this aligns weight selection with dispatch without changing engine priority. Added a non-MiniMax regression (Breeze, MLX-only available engine, both hubs) and all 180 focused audio tests pass.

OliverBryant (approved):
LGTM — approving on `ede84c305`. No blockers.

The head moved after my review (`9ef3315` → `ede84c305`), so I re-reviewed the delta rather than approving what I had already read. **Your generalization is strictly better than what I suggested**, and it fixes more than I flagged.

I suggested deriving a predicate from the spec; you replaced the hardcoded name with `use_default_engine=True` for every audio model. That is a wider blast radius, so I checked it properly:

- **Legacy no-engine models are untouched.** 33 catalog models declare no `engine` at all (including `Irodori-TTS-v4.1-Anime` with six quantizations). `match_audio`'s `if engine_families:` guard skips the filter when the resolved engine matches no spec, so their resolution is byte-identical. I replayed old vs new for the whole catalog to confirm.
- **Models absent from `AUDIO_ENGINES` are untouched.** `if available_engines and model_engine is None:` short-circuits on the falsy `None`, so `model_engine` stays `None`, the `or model_engine is not None` clause stays `False`, and dispatch is unchanged. The "does not support engine selection" warning is still only reachable via an explicit `--model-engine`, as before.
- **For models in `AUDIO_ENGINES`, the engine value is identical.** The dispatch block already did `model_engine = next(iter(AUDIO_ENGINES[name]))`; you moved that same resolution earlier so `match_audio` can see it. Same engine, same `audio_cls` — the only change is that the weights now match the engine that will actually run.

That last point is a **pre-existing bug this closes**, beyond my original note. On an MLX-only host with no `--model-engine`:

```
OLD: match_audio(engine=None) -> BreezeBlue/Breeze-TTS-2            (engine='PyTorch')
     dispatch -> MLXAudioTTSEngineModel        <-- MLX class + PyTorch weights
NEW: match_audio(engine='MLX') -> mlx-community/Breeze-TTS-2-mlx-8bit (engine='MLX')
     dispatch -> MLXAudioTTSEngineModel        <-- consistent
```

So it affected every dual-engine audio family, not just MiniMax-Music3. `test_default_engine_selects_matching_weights` pins it down well — reverting to `use_default_engine=False` fails both the `model_id` and `cache_name` assertions (the `model_engine` one alone would still pass, since the dispatch block sets it afterwards, so it is good that the weight assertions are there).

Everything from my first pass still holds: the upstream `generate` mapping, sample-major stereo handling verified across the `(2, 2)` ambiguity, the checkpoint's `sample_rate=44100` / `bits=8` corroborating the hard-coded expectations, the narrower-than-upstream `duration` bounds, exception propagation through `MLXModelThreadMixin`, and nine adapter mutations all caught.

One process note: CI had only just started on this head when I approved — `changes` passed, `lint` and both docs builds were still running, and the test matrix had not reported. Worth confirming it goes green before merge; I am approving the code, not a green board.

PR #5564, opened 2026-09-21
fix(logging): normalize model request IDs and event types

m199369309:
## Summary

- generate standards-compliant UUID request IDs without the historical `xinf-` prefix
- keep caller-provided valid request IDs unchanged
- emit `event_type` alongside the legacy `event` field for model-request lifecycle records
- update request logging tests for generated IDs, rejected caller IDs, auth failures, streaming responses, and event compatibility

## Backward compatibility

- the legacy `event` field remains present
- existing stored `xinf-<uuid>` IDs remain queryable through the compatibility logic added to #5560
- caller-supplied request IDs are not rewritten

## Validation

- `pytest -q xinference/api/tests/test_model_request_logging.py` — 35 passed
- focused pre-commit checks — passed
- `git diff --check` — passed


qinxuye (approved):
LGTM.

PR #5565, opened 2026-09-21
fix(logging): keep uvicorn access records out of application logs

m199369309:
## Summary

- add a `DropAccessLogFilter` that excludes all `uvicorn.access` records from Xinference application logs
- attach the filter only to the `uvicorn.access` logger
- keep `uvicorn.error` and other application/runtime error records unaffected
- cover successful and failed HTTP status codes plus the logging configuration wiring

## Rationale

HTTP request outcomes are already represented by the audit pipeline. Keeping Uvicorn access lines out of `xinference.log` avoids duplicate request records and prevents log-center correlation timelines from being polluted by access-log entries.

## Validation

- `pytest -q xinference/tests/test_logging_filters.py` — 20 passed
- focused pre-commit checks — passed
- `git diff --check` — passed


qinxuye (approved):
LGTM.

PR #5566, opened 2026-09-21
fix: correct embedding benchmark concurrency and throughput

qinxuye:
Embedding benchmarks can over-dispatch requests, stop timing while workers are still running, and report failed requests as throughput. Opening a new HTTP session for every request also adds connection setup overhead.

Reuse one session with a connection limit matching the requested concurrency, claim each selected input once before sending, wait for all workers, and report successful/failed counts with success-only QPS. Clean up pending workers on cancellation. Add local HTTP regression tests and document the measurement behavior.

Validation:
- `python -m pytest -q benchmark/tests/test_benchmark_embedding.py`: 8 passed, covering exact request counts, full completion, connection reuse, concurrency up to 128, failures, and cancellation.
- Black, isort, Ruff, CLI help, and `git diff --check` passed. Repository pre-commit hooks exclude `benchmark/`, so formatting/lint checks were run directly.
- RTX 3090 Ti / bge-m3: 36,000 measured requests succeeded across sentence_transformers and vLLM, one and two replicas, using the corrected runner and fixed synthetic Chinese inputs. Server logs confirmed batching and balanced two-replica routing. This validates the benchmark path; it is not a server throughput improvement claim.


OliverBryant (approved):
LGTM — approving. **No blockers, no majors.**

I reproduced all three bugs against the old code rather than taking the description on trust. Replaying the previous `worker` + `asyncio.wait(FIRST_COMPLETED)` loop:

```
n=100 conc=16  at timing-stop: dispatched=115 completed=112  workers still running=3
      after cancel: total dispatched=115 (target 100), unique=70
```

So all three claims hold, and the third was the worst of them: 115 sent for 100 requested, and only **70 unique** — the random per-worker start index meant inputs were simultaneously duplicated and skipped. Combined with `args.num_query / benchmark_time`, the reported QPS had a numerator that was neither what was sent nor what succeeded, over a denominator that stopped while three workers were still in flight.

**The new claim logic is correct.** `while self.left > 0` → `index = len - left` → `left -= 1` has no `await` between the check and the claim, so single-threaded asyncio cannot interleave it — which is exactly what the comment asserts. Verified by instrumenting dispatch at concurrency 1/3/10/16/128/… against 1/7/10/100/128 inputs: exactly-once and complete in every combination, `left == 0` at the end.

**Tests.** I ran them (`8 passed`, matching your validation) and then mutation-tested the fix with 11 mutations. Every bug this PR fixes is pinned:

```
decrement AFTER await (over-dispatch)      CAUGHT (4 failed)
FIRST_COMPLETED (early timing stop)        CAUGHT (4 failed)
default connector limit (100 cap)          CAUGHT (1 failed)
throughput counts failures                 CAUGHT (1 failed)
drop ValueError (invalid JSON)             CAUGHT (1 failed)
treat non-200 as success                   CAUGHT (1 failed)
leak _session on exit                      CAUGHT (5 failed)
drop input validation                      CAUGHT (2 failed)
```

These are real integration tests against a live local aiohttp server, not mocks, and three assertions are doing more work than they look: `peak == concurrency` would fail on an unintended 100-connection pool cap, `Counter(seen[5:])` pins exactly-once rather than just the count, and `transports` pins session reuse. The `first_wave` gate to hold the first measured wave until every worker holds a connection is a neat way to make the concurrency assertion deterministic.

Three mutations survived; I checked each and none is a defect:

- `min(concurrency, len(inputs))` → `concurrency`: dispatch is byte-identical with or without it (surplus workers see `left == 0` and exit immediately). Verified at n=3/conc=10 and n=1/conc=128. Purely defensive, so the mutation is a semantic no-op.
- Dropping the `finally` cancel in `_run`: `await asyncio.gather(*tasks)` already cancels its awaited tasks when the outer coroutine is cancelled, so the explicit cleanup is belt-and-braces.
- Moving `warm_up()` inside the timed window: the only genuine measurement question of the three, so I measured it — 10 requests × 20 ms at concurrency 1 gives `benchmark_time = 0.217s`, not the ~0.30s it would be if the 5 warm-up requests were included. **Warm-up is correctly excluded**; only the assertion is absent. Minor test gap, not worth blocking.

**Two details I liked.** Moving `load_dataset` into `main()` is what makes the module importable without `datasets` installed, which is what lets these tests exist at all. And the README's "`--num-query` caps the number of dataset rows used; it does not repeat a smaller dataset" is a real behavior change worth disclosing — `input_requests[: args.num_query]` is a plain slice now, whereas the old `index % len(...)` cycling made it *look* like `num_query` was always honoured. Documenting that rather than quietly preserving the old padding is the right call, since padding is what produced the duplicate-input measurements in the first place.

Switching `time.time()` → `time.perf_counter()` for interval timing is also correct, and reusing the existing `RequestOutput` brings this runner in line with the other benchmarks instead of keeping a bare-latency-list shape.

PR #5569, opened 2026-09-21
feat(UI): enhance running model capability experience

maoyuehui:
1. Improve multilingual content for the running model details page and all capability panels.
2. Add the Segmented component.
3. Extend Rerank support for image, video, and audio sub-capabilities.
4. Extend Embedding support for image, video, and audio sub-capabilities.


frontend/src/components/pages/running-model-detail/panels/capability-task-panel.tsx
```diff
@@ -466,10 +469,10 @@ const CapabilityTaskPanel = forwardRef<CapabilityTaskPanelMethod, CapabilityTask
         >
           <div className="flex shrink-0 items-center justify-between border-b bg-card/80 p-4">
             <div className="flex items-center gap-3">
-              <h3 className="text-base font-semibold">Results</h3>
+              <h3 className="text-base font-semibold">{t('runningModels.detail.results')}</h3>
               {latencyMs !== undefined && !loading && (
                 <span className="text-xs font-medium text-muted-foreground">
-                  Latency {formatLatency(latencyMs)}
+                  {t('runningModels.detail.latency')} {formatLatency(latencyMs)}
```
qinxuye, line 475: [P2] Update the latency regression test alongside this localization change

Running `npm run test:capability-task-panel` now fails `latency is shown only for a successful completion and cleared by Reset`: the test's i18n mock returns the key unchanged, so the rendered text contains `runningModels.detail.latency 1 ms`, while the assertion still expects `/Latency/`. Please update the mock or the localized-label assertions, retaining the checks that latency appears only after success and disappears after Reset, and rerun this test suite.

qinxuye (approved):
LGTM.

PR #5571, opened 2026-09-22
feat(image): add Qwen-Image-2.1 support

Minamiyama:
Adds Qwen-Image-2.1 as a built-in image model.

The model is registered in the per-model image catalog and supports:

- Text-to-image and image-to-image generation
- Hugging Face (`main`) and ModelScope (`master`) sources
- Multiple reference images
- RGBA reference preservation through the OpenAI-compatible image edit endpoint

The Diffusers dependency is installed from the Hugging Face Git repository because the current PyPI release does not include `QwenImage21Pipeline`.


qinxuye (approved):
LGTM.

PR #5576, opened 2026-09-22
feat: recommend model sizes using offline memory metadata

qinxuye:
## Summary

- Recommend the largest eligible LLM size estimated to fit the worker's available memory, while preserving explicit user constraints. Keep compatibility-only selection for non-LLM models and unverified multi-GPU configurations.
- Bundle validated memory metadata across 164 LLM catalog files, including MLX, PyTorch and GGUF sources. Support per-quantization repository metadata and record source URLs and hashes.
- Use bundled metadata without downloads during recommendation. Fix GQA/head-dimension KV-cache estimates, normalize MLX quantization names, and reject unsupported GGUF quantizations rather than returning zero weight memory.
- Add maintenance collectors for configuration JSON and bounded GGUF metadata headers, schema validation, regression tests, and updated launch documentation with all nine existing translations.

## Selection and limitations

- Estimates assume one sequence, 2048 tokens and FP16 KV cache, with an 80% available-memory budget. CPU/Apple Silicon use available system memory; automatic GPU placement uses the lowest free memory among eligible devices, without pooling GPU memory or changing placement.
- Known oversized candidates are excluded. Missing estimates retain a conservative smallest-size fallback. No resources are reserved and launch success is not guaranteed.
- The full catalog scan covered 5,533 source/quantization combinations: 2,980 have collected dense metadata, 2,167 retain metadata tagged as unsupported by the dense estimator, and 386 could not be collected reliably due to inaccessible repositories or missing configuration fields. Unsupported architectures are not silently treated as dense models; exotic quantizations can also remain unverified.
- Existing estimator callers retain their download fallback; the recommendation path explicitly disables downloads. Collection is an explicit maintenance operation, not a launch-time action.

## Validation

- 216 focused tests passed across collection, metadata, memory estimation, recommendation and catalog schemas.
- All applicable pre-commit hooks passed, including catalog schema validation, Black, Ruff, isort, mypy and codespell; `git diff --check` passed.
- Qwen3 MLX regression verifies all six dense sizes on Hugging Face and ModelScope and selects 14B/4bit with a simulated 20 GiB available-memory budget instead of falling back to 0.6B.
- Compared 373 generated model documentation pages with and without metadata: identical output.
- Updated launch documentation translations were compiled and rendered during implementation.
- No real GPU/model inference benchmark was run; estimates remain heuristics, not measured capacity guarantees.


OliverBryant (changes requested):
There is one blocking correctness issue:

**[P1] Do not assume every GPU backend reports free memory in bytes** (`xinference/core/model_recommendation.py:134`)

`get_gpu_info()` does not expose a uniform byte contract. NVML returns bytes, but the torch and ROCm fallbacks use `free = 1` as an unknown sentinel, while the Ascend parser reports MiB values from `npu-smi`. Dividing every value by `1024**2` turns those values into approximately zero MiB. When catalog estimates are available, the recommendation is then treated as verified and every candidate is filtered out, returning `no_recommendation` instead of preserving the conservative smallest-size fallback. This breaks recommendation on supported non-NVIDIA GPU paths and on NVIDIA systems using the torch fallback.

Please normalize units at the device backend boundary (or expose an explicit unit), and treat unknown/sentinel readings as unavailable before applying the memory budget. Please also add a regression test covering at least one sentinel or MiB-reporting backend.

qinxuye:
Fixed the backend memory-unit issue in 878aaca52. Device backends now expose an optional normalized free_memory_mib measurement: NVML/MetaX convert bytes to MiB, while Ascend retains its MiB reading. Recommendation no longer interprets legacy free values, so torch/ROCm sentinel readings and backends without a normalized measurement remain unknown and preserve the conservative fallback. Existing free fields are unchanged for other callers. Ascend composite keys are mapped to scheduling indices, including with ASCEND_RT_VISIBLE_DEVICES filtering. Added regression tests exercising both real sentinel producers and mocked npu-smi parsing with/without visibility filtering. All 221 focused tests and commit hooks passed. No real accelerator hardware was exercised. @OliverBryant please take another look.

OliverBryant (approved):
The blocking memory-unit issue from my previous review is resolved in 878aaca. The recommendation snapshot now consumes only explicitly normalized `free_memory_mib` values, unknown torch/ROCm sentinel readings correctly retain the conservative fallback, and Ascend MiB values are no longer converted twice. The added regression coverage exercises both paths. I found no remaining blocking issue. LGTM.

